PR #2360, opened 2022-11-14
Avoid add citations in incorrect cases

quevon24:
After @flooie started to import the missing volumes from Federal Appendix reporter (F. App'x) we noticed that some citations were added to incorrect cases, these are two examples:

https://www.courtlistener.com/opinion/622150/united-states-v-truman-scott/ correct citation is 463 F. App'x 227

https://www.courtlistener.com/opinion/622149/united-states-v-frank-esquivel/ correct citation is 463 F. App'x 226

We have many functions to check if two cases are the same, and after digging in the process of Harvard imported, I realized that the problem might be when we try to check if there is an overlap in the case names which causes to return the incorrect case as has previously been imported, and this case to add the incorrect citation.

This is a way to reproduce the problem:

First, we remove the incorrect citation from (I did this in my local env) https://www.courtlistener.com/opinion/622149/united-states-v-frank-esquivel/

Then we try to import this case: https://ia801003.us.archive.org/21/items/law.free.cap.f-appx.463/225.4117247.json which is the one with the correct citation: 463 F. App'x 225

The script searches for possible cases, and because we deleted the incorrect citation the first check fails, then we try again to search for possible cases again, but this time we use the court and the date filed as filter params and then we get some matches, “United States v. Frank Esquivel” is in the matches.

Then we start to perform some comparisons to see if the case to import is the same as the ones of the results we got previously. All the checks says that the case is the same.

But I believe the problem is in one of the checks, specifically in this function case_names_dont_overlap (https://github.com/freelawproject/courtlistener/blob/main/cl/corpus_importer/management/commands/harvard_opinions.py#L754)

We need that this function returns a True value to say that there is no overlapping between the case names we are comparing. And having a True value will let us say that the cases are not the same, break the condition and avoid adding the citation to incorrect case. You can see more about this here: (https://github.com/freelawproject/courtlistener/blob/main/cl/corpus_importer/management/commands/harvard_opinions.py#L854)

Here is an example of what is causing the failure, we compare these two case names to check if there is an overlap:

“United States v. Frank Esquivel” and “United States v. Vinson UNITED STATES of America, Plaintiff-Appellee, v. Wayne VINSON, Defendant-Appellant”

As you can see, there is an overlap only with “united” and “states” (some words are removed due the use of harmonize function to juriscraper and other processes), the problem here is that in this specific case “united states” doesn’t have the enough relevance for us to say that both cases are the same, what difference these two cases is the other party. 

One way to solve this is to remove that kind of strings in the check what can cause this unexpected behavior.

This PR removes the string “united states” in that specific check to avoid an overlap with case names that include that string.

But why this happened?

It happened that two consequential cases in the same volume had the same decision date, court, and very similar names: “United States v. Vinson (463 F. App'x 225)”, “United States v. Frank Esquivel (463 F. App'x 226)” and “United States v. Scott (463 F. App'x 227)”

This caused to have many potential possible cases and similar names, and this caused an overlap in the names, making the script to return the case to add the citation instead to jump to verify next case.

Another way to fix this would be also adjusting or improving the other functions that makes the checks, for example cosine_similarity_too_different() which checks if the case names are the same using a different approach, we could apply something similar of what i suggest above.

Currently i'm checking how valid would be to modify the other functions that performs the comparison between case, for example the function that compare the length of content between cases (length_too_different function) and the function that checks the similarity between the content of the cases (content_too_different function)


quevon24:
I've been doing some testing with compare_documents() function https://github.com/freelawproject/courtlistener/blob/43ed1fd10cad9aef6af89b74f80edba1c1632e52/cl/corpus_importer/management/commands/harvard_opinions.py#L998 

This function seems to work very well but i decided to try other algorithms to see if we can have some improvements.

I reviewed some algorithms like: Hamming comparison, the drawback with this is that the text to be compared must be of the same size.

The next one i checked was jaro-winkler's similarity algorithm, this returns a value in the range of [0,1] where 1 means the strings are equal and 0 means no similarity,  this algorithm seems to give good results, but not close as those given by the current algorithm used(substring match).

And finally, I tried Levenshtein's distance algorithm, which returns the minimum number of character substitutions, insertions, deletions required to convert one string into the other. With that number of operations we can calculate the similarity between the two strings with a value in the range of [0,1]. This algorithm also gives good results, but still a bit far from the algorithm actually used.

For example, comparing the opinion from https://ia801003.us.archive.org/21/items/law.free.cap.f-appx.463/225.4117247.json which citation is added incorrectly to https://ia801003.us.archive.org/21/items/law.free.cap.f-appx.463/226.4117753.json, comparing both opinions we have these results:

Substring match: 95%
Jaro-Winkler: 90.88 %
Levenshtein: 82.59 %

So it's probably not worth changing this. I'm still trying to understand some of the conditions defined in content_too_different function https://github.com/freelawproject/courtlistener/blob/43ed1fd10cad9aef6af89b74f80edba1c1632e52/cl/corpus_importer/management/commands/harvard_opinions.py#L717 to see if we can do something there to improve the functionality.

With the check implemented in length_too_different function i think it's probably not worth changing this, is a simple check that can be useful, but not very reliable because we can't tell much about two opinions just from their length, that why we require the other checks.

So far the solution i proposed seems to work, i'm trying to figure out how to write better tests for this using the factories that @flooie created.

Any suggestion or comment is always welcome 👍

flooie:
Are you sure that we should be focusing on the compare document for difference in content?

flooie:
It seems to me that the real problem here is not in comparing the text of the opinion because the opinions appear to be templatized.  it's that 

1) UNITED STATES of America, Plaintiff—Appellee, v. Wayne VINSON, Defendant—Appellant
and 
2) UNITED STATES of America, Plaintiff—Appellee, v. Frank ESQUIVEL, Defendant—Appellant

should've been able to effectively identify these as different opinions based on the name alone.  

and maybe also the docket number but I do need to look what was in the system.

quevon24:
> Are you sure that we should be focusing on the compare document for difference in content?

I decided to check the compare documents fuction because the opinion from https://www.courtlistener.com/opinion/622150/united-states-v-truman-scott/ differs a bit from the opinion in https://ia801003.us.archive.org/21/items/law.free.cap.f-appx.463/226.4117753.json 

This was one option that worth to try, i but so far the current algorithm works great. 

I think the problem is related to the comparison between case names, specifically with the ones that has a templatized name or name abbreviation like: "United States v. XXXXX", "XXXXX v. United States" or "UNITED STATES of America, Plaintiff—Appellee, v. XXXXX, Defendant—Appellant"

That is the main reason i proposed the solution i commited in this PR.

quevon24:
Thanks to @mlissner comments I think we can use this https://github.com/freelawproject/juriscraper/blob/main/juriscraper/lib/string_utils.py#L550, it is already used in the harvard importer to create the short case names but only when we add a new case, maybe this can be used in some specific conditions when we compare the case names

cl/corpus_importer/management/commands/harvard_opinions.py
```diff
@@ -982,10 +982,20 @@ def winnow_case_name(case_name: str) -> Set:
...
     # Remove all non alphanumeric characters
     case_name = harmonize(case_name)
     case_title = re.sub(r"[^a-z0-9 ]", " ", case_name.lower())
 
+    # Remove strings that can cause an unnecessary overlap
+    case_title = false_positive_strings_regex.sub("", case_title)
+
     # Remove one letter words, initials etc.
     case_title = re.sub(r"\b[^ ]\b", "", case_title)
     cleaned_set = set(case_title.split())
```
flooie, line 1001: This all looks good to me.

cl/corpus_importer/tests.py
```diff
@@ -791,3 +811,20 @@ def test_no_volume_citation(self):
         )
         self.read_json_func.return_value = case_law
         self.assertSuccessfulParse(1)
+
+    def test_failing_name(self):
+        """
+        Check if there is an overlap between two similar case names.
+        """
```
flooie, line 818: Probably could use a more descriptive name.  

Test case name winnowing comparison ... or 

test_removal United States... something something

quevon24: for the function name or the description, or both?

flooie: both.

quevon24: What do you think about this

```
    def test_case_name_winnowing_comparison(self):
        """
        Test removing "United States" from case names and check if there is an
        overlap between two case names.
        """
        case_name_full = "UNITED STATES of America, Plaintiff-Appellee, " \
                         "v. Wayne VINSON, Defendant-Appellant "
        case_name_abbreviation = "United States v. Vinson"
        harvard_case = f"{case_name_full} {case_name_abbreviation}"

        case_name_cl = "United States v. Frank Esquivel"
        overlap = winnow_case_name(case_name_cl) & winnow_case_name(
            harvard_case)
        self.assertEqual(len(overlap), 0)
```

flooie (commented):
Thanks @quevon24

PR #2368, opened 2022-11-17
Add Territory and Tribal Courts to model 

flooie:
This PR replaces a previous PR that adds territory and tribal courts to CL

It also removed and separately pushed a PR for fixes to tests 
that were not related to this very specific fix


flooie:
@mlissner Please this!

PR #2376, opened 2022-11-23
Improve log message for harvard importer

quevon24:
Some case names are reduced to blank in the server but working correctly in local environment, I improved the log message to identify those case names.

I added one test with the case name failing in the server but working correctly in local environment.

### DIFF
--- a/cl/corpus_importer/management/commands/harvard_opinions.py
+++ b/cl/corpus_importer/management/commands/harvard_opinions.py
@@ -1021,7 +1021,6 @@ def winnow_case_name(case_name: str) -> Set:
     if not case_title:
         # Log case name if the process reduce it to blank
         logger.warning(f"Case name: \"{case_name}\" reduced to blank.")
-        print(f"Case name: \"{case_name}\" reduced to blank.")
 
     # Convert case name to set of words
     cleaned_set = set(case_title.split())
--- a/cl/corpus_importer/management/commands/harvard_opinions.py
+++ b/cl/corpus_importer/management/commands/harvard_opinions.py
@@ -1018,7 +1018,7 @@ def winnow_case_name(case_name: str) -> Set:
     # Remove one-letter words, initials etc.
     case_title = re.sub(r"\b[^ ]\b", "", case_title)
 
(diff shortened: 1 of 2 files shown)


quevon24:
I did not take into account that there could be cases without a name in courtlistener, and that is what caused the problem, this PR will only register if any case name is reduced to nothing, but only if the name is not empty from the beginning.

cl/corpus_importer/management/commands/harvard_opinions.py
```diff
@@ -1000,27 +1000,27 @@ def winnow_case_name(case_name: str) -> Set:
...
     # Remove strings that can cause an unnecessary overlap
     case_title = false_positive_strings_regex.sub("", case_title)
 
     # Remove one-letter words, initials etc.
     case_title = re.sub(r"\b[^ ]\b", "", case_title)
 
-    if not case_title:
+    if not case_title and case_name:
         # Log case name if the process reduce it to blank
-        logger.warning(f"Case name: {case_name} reduced to blank.")
```
flooie, line 1023: I think this maybe shouldn't be warning that clogs up the logs.

cl/corpus_importer/management/commands/harvard_opinions.py
```diff
@@ -1000,27 +1000,27 @@ def winnow_case_name(case_name: str) -> Set:
...
+        logger.warning(f'Case name: "{case_name}" reduced to blank.')
```
flooie, line 1023: I'm not really sure I understand what is going on here.  Why would be also include case_name and instead of prepared_case_name and why would we do both?

quevon24: We log the original case name, not the one reduced to blank, the condition ensures that the name entered is not empty and it was reduced to blank. To avoid additional warnings in the logs, with this, the message will be logged unless is really necessary. I ran the function against all case names from case.law metadata and none throwed a warning. So we could also remove the message.

mlissner:
Looks like this needs some resolution too, no?

PR #2379, opened 2022-11-25
feat(opinion-page): Add vlex sponsored by

### LINKED ISSUE #2378: Need to identify v|Lex blurb timeframe
We need to identify a way to identify when a v|Lex sponsored blurb should leave the page.  

Currently we can identify when a blurb should go on to an opinion more or less, but the blurb is not indefinite and we have no mechanism right now to decide when that time has lapsed.

flooie:
Fixes #2378 

Add vlex logo to assets
Add css for asset
Check if opinion qualifies as sponsored
Add sponsor to bottom or sidebar


flooie:
On Safari Mobile 
-----------------



On MacOS Desktop
--------------------
Screenshot 2022-11-25 at 2 48 57 PM

cl/assets/static-global/css/override.css
```diff
@@ -1399,3 +1399,7 @@ textarea {
   max-width: 100%;
   height: auto;
 }
+
+#vlex {
+    background: black;
+}
```
mlissner, line 1405: If this is the best logo we've got for them, we should just add a black background in the png file instead of carrying extra CSS around forever.

cl/opinion_page/views.py
```diff
@@ -521,6 +522,14 @@ def view_opinion(request: HttpRequest, pk: int, _: str) -> HttpResponse:
         cluster,
     ).prefetch_related("representative",)[:3]
 
+    # Identify opinions updated/added in partnership with v|lex for 3 years
+    sponsored = False
+    if (
+        cluster.date_created.date() > datetime.datetime(2022, 6, 1).date()
+        and cluster.filepath_json_harvard
```
mlissner, line 529: This seems pretty good. Do you need the `.date()` casts? Why not just compare the datetime objects and have a little slop in the comparison?

cl/opinion_page/templates/opinion.html
```diff
@@ -212,6 +211,20 @@ <h3><span>Visualizations</span></h3>
...
+
+        {# Sponsored by #}
+        {% if sponsored %}
+            <div id="sponsored-by" class="sidebar-section">
+                <h3>
+                  <span>Sponsored By</span>
+                </h3>
+                <a href="https://www.vlex.com" rel="sponsored">
+                    <img id="vlex" src="{% static "img/vlex-white.png" %}" class="img-responsive v-offset-below-1" width="150" height="75" >
+                </a>
```
mlissner, line 223: Don't forget an alt text here and let's do a `target="_blank"` too.

mlissner (changes requested):
A couple little comments. 

Aesthetically, did you try putting the logo below the "This opinion added..." blurb? It feels a bit jagged to me?

cl/assets/static-global/css/override.css
flooie: I left it black because I wasn't sure if we wanted something more subtle so we could debate the color or if they had a different image which it looks like they do so we can drop this

cl/opinion_page/templates/opinion.html
flooie: ah yes.

cl/opinion_page/views.py
flooie: `can't compare offset-naive and offset-aware datetimes`

I thought it was easier then adding timezones

flooie:
Screenshot 2022-11-28 at 7 18 59 PM

I like the text change, but I think the logo feels like it's missing something now

mlissner:
Yeah, the logo has a spacing problem now. Can you bring it closer to the paragraph above? It looks too far away?

flooie:
@mlissner I think the spacing issue is fixed

flooie:
Maybe we just drop the blurb - and it says sponsored by?.  The blurb language is from the contract.  fwiw

mlissner:
Can I get another screenshot? The blurb language was negotiated, so if we left it out, I'm sure they'd want it back.

flooie:
Screenshot 2022-12-01 at 6 59 24 PM
Screenshot 2022-12-01 at 6 58 50 PM

mlissner:
Still needs alt text, but it's good otherwise. Feel free to merge once that's in place. 

Thank you!

flooie:
excuse me while I look up alt text

mlissner (approved):

PR #2383, opened 2022-11-30
Show filepath_json_harvard field in clusters API

quevon24:
This small change allows clusters API to show filepath_json_harvard field, which was originally excluded


mlissner (approved):
Seems OK to me, at a glance. I leave it to @flooie to make the final call.

flooie:
looks good to me

PR #2393, opened 2022-12-05
Fix reporter mismatch warning for harvard opinions importer

quevon24:
In the logs we found many warning messages like this: "WARNING Reporter mismatch for cluster: 3200805 on cite: 2016 Ohio 2813"

This problem is caused due Citation model restrictions (reporter type is required to create the citation) , we were using a get_or_create() to get existent citation or add it as a new citation.  This happened because if a reporter type is updated in reporters-db and there are already citations from that reporter in courtlistener causes the reporter type to mismatch and then tries to get the object, the problem is that if the reporter type changed, the get_or_create fails, then tries to create the object and raises the IntegrityError exception.

These changes improve the process to create citations in harvard importer, this proccess allows to update the reporter type if this changed.


flooie:
I'm not sure I'm convinced this is the correct approach here.  Have you looked at the reporter that was causing this issue.  We cant just change the reporter type, we need to make a determination, a better determination on the type of reporter.  Here Ohio can be both neutral and state.

quevon24:
I checked the reporter type and it hasn't changed, so the problem isn't there.

I inspected all the related functions to get the reporter type integer value in CL and everything seems right.

I proceeded to check the API and the opinion i mention (2016 Ohio 2813) was created in 2016 [https://www.courtlistener.com/api/rest/v3/clusters/3200805/](https://www.courtlistener.com/api/rest/v3/clusters/3200805/) , so i guess that this was loaded with other thing different than the harvard importer.

The i saw the opinion source is from the court website, so i guess the opinion was loaded using the scrapper, but if it was loaded with the scrapper, taking a look to git history, the Citation model was added in 2018, so maybe the citation was added later?

My question is, how the citation was added? or how was this opinion loaded without the harvard importer?

The changes i suggested are with the goal to not modify the model unless is really needed (get_or_create breaks if reporter type doesn't match and reporter_type is required field), and to handle the case where the reporter type for some reason changed (this shouldn't happen be very often) but this way, we can update the reporter type in case of being needed.

flooie:
I think this is the wrong approach.  I think this is something that needs to be addressed in reporters-db.

flooie:
This should be addressed in reporters-db and not here.  The citation is actually correct - on the system and it should register these errors as neutral.  The citation regex format on reporters is too narrow for our real world data.

Issue #1452, opened 2020-11-09
Nebraska Scraping Fail


flooie:
[307 Neb 1 over and over](https://www.courtlistener.com/c/Neb./307/1/)

I stumbled onto a duplicating opinion bug that is repeatedly allowing cases to be collected over and over and over again.  The case above has been scraped and added to our system 21 times as of 11/9/20 at 3:22 PM EST.

flooie:
This may just a Nebraska issue but a but in the REST API is curtailing a deeper dive into the problem.  

It appears that in 2017 we started capturing the same opinion multiple times, and it is now becoming a larger issue for anyone that wants to research Nebraska case law. 

```
Volume 294 has 71 opinions.
Volume 300 has 1719 opinions.
Volume 307 the current volume already has 460 opinions.
```

Additionally, we have listed in volumes, Volume 925- which is not real.  The volume should have been added to Vol 295. 

It is unclear how or why this is happening.  This needs further investigation

flooie:
This is still quite a problem.  And should be addressed because its causing issues in the opinion import.

quevon24:
The problem is caused because the sha generated from content mismatch causing the creation of duplicates.

The sha doesn’t match because the stamp on the document changed

For example, in https://www.courtlistener.com/opinion/9303423/bleich-v-bleich/ we have:
![image](https://user-images.githubusercontent.com/6474994/206011675-e23f339a-93b6-488c-b321-c127fc64106a.png)

And in the duplicate: https://www.courtlistener.com/opinion/9303466/bleich-v-bleich/
![image](https://user-images.githubusercontent.com/6474994/206011697-bebae90f-cf04-4631-929e-303b5c50092d.png)

quevon24:
The problem started in volume 292: https://www.courtlistener.com/c/neb/292/

This volumes are missing: 281, 282, 283, 284, 287, 288, 289, 290, 291

And the volumes, starting from 280, are incomplete.

PR #2404, opened 2022-12-07
Fix for nebraska scraping

quevon24:
As i mentioned in issue #1452 the problem with nebraska opinion files is that they are updating them causing the timestamp to be updated in pdf, causing the generated sha to not match the one stored in CL opinions.

The solution I propose is really simple, they are updating the pdf files but the download url is still the same, for example, this opinion https://www.courtlistener.com/opinion/6905720/schaaf-v-schaaf/ is duplicated 26 times, the first timestamp is from 07/22/2022 09:05 AM CDT and the last is from 10/14/2022 09:07 AM CDT but the download url is always the same: https://www.nebraska.gov/apps-courts-epub/public/viewOpinion?docId=N00008736PUB

What makes this fix is that when the scrapper for Nebraska is run, it will use the download_url as the lookup field instead of sha field which can easily change even if they just update the timestamp.

This is really simple, but does the job, and in case of having a similar situation to the one with nebraska, we just need to add the court_str to the exception list.


flooie:
I'd like to get @mlissner thoughts on this - but I think we shouldn't modify the cl scrape opinion for a one off court - if we can resolve this by tightening up the citation formatting in reporters-db

flooie:
@quevon24, I actually on second thought, wonder if this should be a default check just in general and not specific to Nebraska.

mlissner:
> I wonder if this should be a default check just in general and not specific to Nebraska.

The balance here is between opinions that are legitimately being corrected by the court and opinions that are  getting new sha1 hashes every time you download them. 

If we add a check like this by default, we'll stop getting legitimate updates from the court, which is not so great. I think I like @quevon24's fix for that reason. We maintain our ability to get updated opinions, except in dumb jurisdictions where they generate PDFs. 

A bigger fix, I suppose, would be to diff the contents after each addition and make some sort of clever guess about whether they fixed something or just tweaked a timestamp, but that is hard because our text extraction is asynchronous. 

And the biggest fix, of course, is to do better things in the UI when we identify duplicated opinions. It wouldn't be such a big deal to have these duplicates if the UI coalesced them together in all relevant places. Then, people could see that there were "29 versions of this opinion" or whatever, and then look at the others. Still, this bug would be annoying, but we'd push analysis of it to users, who can make their own decisions about it. We'll get that working next year, I think. I don't think we'll have much choice pretty soon.

In the meantime, I think the simple, surgical solution just for NEB is fine?

flooie:
I think it would be perfectly reasonable if this is a problem in the future to remove the annotation stamp - I think it would be easy enough to do - and maybe something we want to get good at - if we eventually want to clean up the footers and headers that destroy our HTML, regardless this is good.  

@quevon24 I would of course love a test but its small enough of a change I think.

PR #2422, opened 2022-12-20
Set throttle rate for user

quevon24:
Set throttle rate for me to perform tests, clone cases in local env(requires many requests to clone one case), etc.


mlissner:
This is an insane amount of queries, but OK. If you notice the site getting slow, maybe back off. :)

PR #2433, opened 2022-12-30
feat(maine): Add portal for Maine SJC uploads

flooie:
Rename TennWorkersForm to CourtUploadForm
Make publish.html work between courts
Add Maine Jud. Seal to use as image

Update portal form to work with Maine SJC
Improve doc strings


flooie:
PR adds support for Maine SJC Reporter of Decision to upload directly to Courtlistener.  

Also includes a bump to the permissions structure to allow for individualized court permissions, which have been updated in Courtlistener already.  

Please see a demo of the page and fields. 

Screenshot 2022-12-29 at 5 26 49 PM

cl/opinion_page/forms.py
```diff
@@ -212,15 +282,42 @@ def __init__(self, *args, **kwargs) -> None:
...
+                ]
+            )
+
+        elif self.pk == "me":
+            self.fields["cite_reporter"].widget = forms.Select(
+                choices=[("ME", "ME")],
+                attrs={"class": "form-control"},
+            )
+            self.drop_fields(["lead_author", "second_judge", "third_judge"])
         else:
```
mlissner, line 301: It's worth being explicit in this `else` statement since things are getting complex. I'd suggest making it:

```
elif self.pk == "xyz'
    do_stuff
else:
    Raise_an_error or throw a 500 or something with an "unknown court" error.
```

cl/opinion_page/templates/publish.html
```diff
@@ -37,7 +37,11 @@
     <div class="row">
       <div class="col-xs-1"></div>
       <div class="col-xs-2">
-        <img src="{% static "img/tn-court.jpg" %}" class="img-responsive" width="359" height="359" >
+        {% if form.initial.court.id == "me" %}
+          <img src="https://seals.free.law/v2/128/me.png" class="img-responsive" height=128 width="128" />
+        {% else %}
```
mlissner, line 42: Being explicit in this else statement seems good too.

cl/opinion_page/views.py
```diff
@@ -97,19 +99,29 @@ def court_homepage(request: HttpRequest, pk: str) -> HttpResponse:
     return TemplateResponse(request, "court.html", render_dict)
 
 
-@group_required("tenn_work_uploaders")
+@group_required(
+    "tenn_work_uploaders", "tennworkcompcl", "tennworkcompapp", "me"
```
mlissner, line 103:
```suggestion
    "tenn_work_uploaders", "uploaders_tennworkcompcl", "uploaders_tennworkcompapp", "uploaders_me"
```

cl/opinion_page/views.py
```diff
@@ -97,19 +99,29 @@ def court_homepage(request: HttpRequest, pk: str) -> HttpResponse:
...
     :param request: A GET or POST request for the page
     :param pk: The CL Court ID for each court
     """
+    if pk not in ["tennworkcompcl", "tennworkcompapp", "me"]:
+        raise Http404(
+            "Court pages only implemented for Tennessee Worker Comp Courts and Maine SJC."
+        )
+    # Validate the user has permission
+    if not request.user.is_staff and not request.user.is_superuser:
+        if not [g.name for g in request.user.groups.all() if pk == g.name]:  # type: ignore
```
mlissner, line 117: I think you can tweak this to use an `exists()` check instead, which should be faster. Check this out:

https://stackoverflow.com/questions/4789021/in-django-how-do-i-check-if-a-user-is-in-a-certain-group

mlissner (changes requested):
Nice. A couple things:

1. We don't need the seal in here anymore, right?

3. We gotta get tests passing.

Otherwise, good stuff. Let's make sure we get their homepage up and running next.

cl/opinion_page/forms.py
```diff
@@ -96,10 +97,9 @@ class DocketEntryFilterForm(forms.Form):
     )
 
 
-class TennWorkersForm(forms.Form):
+class CourtUploadForm(forms.Form):
 
     court_str = forms.CharField(required=True, widget=forms.HiddenInput())
-
     case_title = forms.CharField(
         label="Case Title",
```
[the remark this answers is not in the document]
flooie: do you mean for the label and not the variable name

cl/opinion_page/templates/publish.html
flooie: Great.

cl/opinion_page/forms.py
flooie: √

PR #2434, opened 2022-12-30
Track changes for models in favorites app

quevon24:
This PR contains the needed changes to track the all models in favorites app.

To revert an object, you need to open your django shell, get the object, and then get the history from that object and select the one that you want to revert.

For example, we want to revert user tag object to the last change saved:

```
# Get the object
user_tag = UserTag.objects.get(pk=1)
# Get the object history
event_objects = user_tag.event.order_by("-pgh_id")
# Revert object, revert() returns the updated/created object
user_tag = event_objects[0].revert()
```

The  related_name can easily be changed in the model decorator.

**New feature:**
Now you can revert objects from admin, all you need to do is register the model in admin.py with EventHistoryAdmin imported from cl.lib.pghistory_admin instead of usual admin.ModelAdmin

For each revert object you can see a diff from the previous change.


cl/favorites/tests.py
```diff
@@ -333,6 +334,33 @@ def test_user_can_change_favorites(self) -> None:
...
+            kwargs={"username": params["username"], "tag_name": tag_name},
+        )
+        request = RequestFactory().get(path)
+
+        # Increment view counter by 1
+        increment_view_count(test_tag, request)
+        self.assertEqual(test_tag.view_count, 1)
+
+        # Revert object to last change, and check view counter
+        test_tag.history.order_by("-pgh_id")[0]
```
mlissner, line 361: Huh? Does looking at the history table automatically revert from it? Where did the revert happen??

cl/favorites/models.py
```diff
@@ -125,6 +138,10 @@ class Meta:
         index_together = (("user", "name"),)
 
 
+@pghistory.track(
+    pghistory.Snapshot(),
+    obj_field=pghistory.ObjForeignKey(related_name="history"),
```
mlissner, line 143: If we don't do this, the name is `event`, right? I think I prefer `history` too, but I'm not sure it's worth the cruft and the need to remember to break the default whenever we do this. I also like keeping with defaults because it makes understanding online guides easier.

If it's just event vs history and event is the default, I'd say we go with that. If there's something more here, let me know.

cl/favorites/models.py
```diff
@@ -83,6 +92,10 @@ class Meta:
         unique_together = (("docket", "tag"),)
 
 
+@pghistory.track(
+    pghistory.Snapshot(),
+    obj_field=pghistory.ObjForeignKey(related_name="history"),
+)
 class UserTag(AbstractDateTimeModel):
```
mlissner, line 99: We should ignore view_counts, to save ourselves an insert for every page load.

cl/favorites/models.py
```diff
@@ -67,6 +72,10 @@ class Meta:
         )
 
 
+@pghistory.track(
+    pghistory.Snapshot(),
+    obj_field=pghistory.ObjForeignKey(related_name="history"),
+)
```
mlissner, line 78: Since this is a through table without any extra fields, I think we don't need to track it.

mlissner (changes requested):
This is super cool and looks like an amazing solution. 

A couple comments so far, but one other thing we need is the SQL migration. Can you generate that too? I think you just need to do the sqlmigrate management command once you've got the migration file.

Looks like a nice start.

cl/favorites/models.py
quevon24: We can exclude the field, the problem is that we won't be able to revert the object, according to the documentation if we exclude fields and try to revert an object it will throw an RuntimeError exception: https://django-pghistory.readthedocs.io/en/2.5.1/reversion.html#reverting-objects

cl/favorites/tests.py
quevon24: My mistake, i didn't write the whole line in the test

cl/favorites/models.py
quevon24: Yes, the default related name is "event", so i'll leave them with the default value

quevon24:
> This is super cool and looks like an amazing solution.
> 
> A couple comments so far, but one other thing we need is the SQL migration. Can you generate that too? I think you just need to do the sqlmigrate management command once you've got the migration file.
> 
> Looks like a nice start.

Yes you can get the sql with "sqlmigrate" management command, you get the usual sql from migrations but also the sql for triggers.

cl/favorites/models.py
mlissner: That's OK if we can't revert easily. I think that's preferred vs having the history table get really big or incur a performance penalty when a page becomes viral.

cl/favorites/tests.py
```diff
@@ -333,6 +334,37 @@ def test_user_can_change_favorites(self) -> None:
...
+        )
+        request = RequestFactory().get(path)
+
+        # Increment view counter by 1, event object created
+        increment_view_count(test_tag, request)
+        self.assertEqual(test_tag.view_count, 1)
+
+        # Revert object to previous change, we use the second-to-last because
+        # the latest event always contains the current version of the model
+        test_tag = test_tag.event.order_by("-pgh_id")[1].revert()
```
mlissner, line 363: Maybe I'm wrong, but this approach doesn't seem totally trustworthy to me. From my eyes, you're running the revert command on the `event` model manager, which is returning a model instance through some magic. I'm not _completely_ convinced that the model instance isn't coming from the `event` table. 

I think I'd rather see something simpler, like this:

```suggestion
        # Check that view count set to 0 again
        test_tag = test_tag.event.order_by("-pgh_id")[1].revert()
        self.assertEqual(test_tag.view_count, 0)
  
        # Grab the original item from the DB, just to be sure
        test_tag.refresh_from_db()
        self.assertEqual(test_tag.view_count, 0)
```

mlissner (changes requested):
Sorry, we need one more round of tweaks, but they should be easy ones. 

A couple comments below. Also:

1. We actually check the `sql` file into the repo. Poke around in a couple of the `migrations` folders, and you'll see examples. This just means to do `manage.py sqlmigrate > filename.sql` and check that into the PR.

2. Please see the DEVELOPMENT.md file for commit message formatting. No need to re-do your messages here unless that's something that's easy for you (I'm bad at it!), but for future PRs, your commits should follow that style, please.

Thank you!

cl/favorites/tests.py
quevon24: Inspecting the revert() property inside the library I saw that it performs an update_or_create using the pk and the values from event object as default and then returns the only object that is supposed to be updated.
(https://github.com/Opus10/django-pghistory/blob/001bbfeb5ee7931d6e9b72df965a9ae69d644502/pghistory/models.py#L175)

I was thinking to use a get to get the updated object, but i prefer your suggestion

quevon24:
> Sorry, we need one more round of tweaks, but they should be easy ones.
> 
> A couple comments below. Also:
> 
> 1. We actually check the `sql` file into the repo. Poke around in a couple of the `migrations` folders, and you'll see examples. This just means to do `manage.py sqlmigrate > filename.sql` and check that into the PR.
> 2. Please see the DEVELOPMENT.md file for commit message formatting. No need to re-do your messages here unless that's something that's easy for you (I'm bad at it!), but for future PRs, your commits should follow that style, please.
> 
> Thank you!

Yes, i just reread DEVELOPMENT.md again, i'll be more careful with my commits, i just updated my pc and had a few issues that caused me to commit wrong code many times 🤦‍♂️, but now this should be fixed

quevon24:
> Sorry, we need one more round of tweaks, but they should be easy ones.
> 
> A couple comments below. Also:
> 
> 1. We actually check the `sql` file into the repo. Poke around in a couple of the `migrations` folders, and you'll see examples. This just means to do `manage.py sqlmigrate > filename.sql` and check that into the PR.
> 2. Please see the DEVELOPMENT.md file for commit message formatting. No need to re-do your messages here unless that's something that's easy for you (I'm bad at it!), but for future PRs, your commits should follow that style, please.
> 
> Thank you!

Yes, i just reread DEVELOPMENT.md again, i'll be more careful with my commits, i just updated my pc and had a few issues that caused me issues many times 🤦‍♂️, but now this should be fixed

cl/favorites/models.py
mlissner: Did we exclude view count somewhere that I didn't see?

mlissner (changes requested):
OK, a few things here:

1. Your SQL migration file is screwy because of Windows. When I download it and open it in my editor, it looks like this:

    image

    And Github won't display it in the code review page because it thinks it's a binary file. (Strangely, it does show up on pages like this one though: https://github.com/freelawproject/courtlistener/blob/e42f6f5e8088b709c26580b707ff3f248d2cdb6c/cl/favorites/migrations/0002_auto_20230103_0920.sql)

    Anyhow, my suspicion is that Windows is doing something dumb with the BOM, and looking at the first few bytes of the file seems to bear that out:
    
    ```
    In [3]: with open('/tmp/0002_auto_20230103_0920.sql', 'rb') as f:
       ...:     print(f.read()[0:10])
    b'\xff\xfeB\x00E\x00G\x00I\x00'
    ```
    
    Those bytes [seem to correspond](https://stackoverflow.com/questions/65841649/detect-byte-order-mark-bom-in-python) with the BOM for `UTF-32 LE`, though even that has the `B` and `E` from the sql `BEGIN` statement mixed in. I'm sure _you_ didn't do anything to make this so weird, but Windows certainly did. I suggest you find a solution somehow or perhaps now is a good time to get Ubuntu working.

2. I'm sorry to say that the admin work you did is more than we need, and I think it falls into YAGNI. It looks like creative, good code, but as we discussed, can you please pull it out? It'd rather not have to maintain this code forever, and I'd much prefer if it were part of pghistory (though who knows how hard that is, and whether it's worth our time).

3. Your test is in a selenium test, which it shouldn't be. It should be in the simplest, fastest test possible.

Beyond the above, I've got a few other line  by line comments, and I'll do a full review once the above are fixed.

Sorry to delay the response on this one, and thanks for putting it together!

cl/favorites/models.py
quevon24: I missed the last message, i just pushed this change.

quevon24:
> OK, a few things here:
> 
> 1. Your SQL migration file is screwy because of Windows. When I download it and open it in my editor, it looks like this:
>    image
>    And Github won't display it in the code review page because it thinks it's a binary file. (Strangely, it does show up on pages like this one though: https://github.com/freelawproject/courtlistener/blob/e42f6f5e8088b709c26580b707ff3f248d2cdb6c/cl/favorites/migrations/0002_auto_20230103_0920.sql)
>    Anyhow, my suspicion is that Windows is doing something dumb with the BOM, and looking at the first few bytes of the file seems to bear that out:
>    ```
>    In [3]: with open('/tmp/0002_auto_20230103_0920.sql', 'rb') as f:
>       ...:     print(f.read()[0:10])
>    b'\xff\xfeB\x00E\x00G\x00I\x00'
>    ```
>    
>    
>        
>          
>        
>    
>          
>        
>    
>        
>      
>    Those bytes [seem to correspond](https://stackoverflow.com/questions/65841649/detect-byte-order-mark-bom-in-python) with the BOM for `UTF-32 LE`, though even that has the `B` and `E` from the sql `BEGIN` statement mixed in. I'm sure _you_ didn't do anything to make this so weird, but Windows certainly did. I suggest you find a solution somehow or perhaps now is a good time to get Ubuntu working.
> 2. I'm sorry to say that the admin work you did is more than we need, and I think it falls into YAGNI. It looks like creative, good code, but as we discussed, can you please pull it out? It'd rather not have to maintain this code forever, and I'd much prefer if it were part of pghistory (though who knows how hard that is, and whether it's worth our time).
> 3. Your test is in a selenium test, which it shouldn't be. It should be in the simplest, fastest test possible.
> 
> Beyond the above, I've got a few other line by line comments, and I'll do a full review once the above are fixed.
> 
> Sorry to delay the response on this one, and thanks for putting it together!

Thanks for the comments, i'm on it.

mlissner:
@quevon24 I'm sorry this PR is generating so much trouble, but can you review it one more time before I dig in? I think there are a couple more clean up tasks to do before it's ready for my full review.

quevon24:
> @quevon24 I'm sorry this PR is generating so much trouble, but can you review it one more time before I dig in? I think there are a couple more clean up tasks to do before it's ready for my full review.

Done 👍

mlissner:
Well, I totally screwed this up while trying to rebase it and fix the conflicts in it (a poetry upgrade meant that our old poetry.lock files are no longer valid and the new ones conflict with the old). 

I still don't know how I screw this up every now and then, but I sure do and it sure is upsetting. I'm going to try to squash and force push your changes only so that this branch gets fixed. I won't be able to do that for a little bit. I still really hate git sometimes.

mlissner:
> I still don't know how I screw this up every now and then,

Yes, I think I do. It's like this. `main` got merged into this branch several times over the past few weeks, and now I just tried to rebase this branch onto main. So, in effect I'm rebasing main into main. Gross. No wonder it's a mess.

Now to fix it.

mlissner:
OK, so to fix this I cherry picked all the changes from the old branch and put them in a new one that was branched from the latest code in main. That did two things. It got us the code from #2466, which renames one of the models, and it got us a fresh start with zero commits and just raw changes. 

From there, I regenerated the poetry.lock file using the newest version of poetry (which we had to upgrade to today, frustratingly), and then I deleted and regenerated the migration files.

At that point the code was ready, so I committed it to a new local branch. I was also able to do this in four tidy commits such that if you remove them one by one, each former commit still works. Check out the order of them and let me know if this doesn't make sense. In theory, this is how you want to have your commits, but it's rarely possible unless you have a situation like this where you totally screw up your code. 

Once the code was committed to my new branch, I rebased the branch for this PR onto main one more time and pushed that (it was a TOTAL mess at this point, but at least it branched from the same place as my new branch). And then, finally, I force pushed my new branch into this one on Github, which nuked all the old commits in this branch, replacing them with the four I made.

A few other notes:

1. Please check this carefully, Kevin. I did my best not to drop any of your changes, but I'm not sure I didn't screw this up somehow. Are all your changes here?

2. I noticed that your migration file had a bad name, so I changed it to something better. The ones generated by django tend to suck, we usually give them more meaningful names.

3. I reformatted the SQL file using PyCharm. It's almost readable now!

I think that's it and this should be good. I'm off to bed. Hopefully tests will pass and you'll say it's good, if so, I'll merge tomorrow and the rest of the models should be easier.

mlissner (approved):

PR #2435, opened 2022-12-30
Add media file directory to dockerignore

quevon24:
I added the 'media' directory to the dockerignore file as this was causing the docker image to take longer to build locally, this happened because if you download files using the management commands to download files from IA this can increase the directory size a lot making the context size bigger.


mlissner:
Thanks.

PR #2436, opened 2022-12-30
Management commands to clone opinions and dockets in local environment

quevon24:
This are two useful management commands for development purposes, these commands allows you to clone opinion clusters or only dockets in case of need.

For example:

```
docker-compose -f docker/courtlistener/docker-compose.yml exec cl-django python manage.py clone_from_cl --type OpinionCluster --id 9355884
docker-compose -f docker/courtlistener/docker-compose.yml exec cl-django python manage.py clone_from_cl --type Docket --id 5377675
docker-compose -f docker/courtlistener/docker-compose.yml exec cl-django python manage.py clone_from_cl --type Person --id 16207
docker-compose -f docker/courtlistener/docker-compose.yml exec cl-django python manage.py clone_from_cl --type Court --id usnmcmilrev
```

These commands only work if you have the env variable DEVELOPMENT set to True.


mlissner:
I haven't looked at the code yet, but any reason you can't put these into one management command? 

Just notionally:

manage.py clone_from_cl --obj-type search.Opinion --id 123

manage.py clone_from_cl --object-type search.Docket --id 123

I guess it'd be nice if it used object types similar to those in other management commands. I think the cl_update_search command has some examples (they might even be in an `enum` or a class that you could re-use?

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+
+class Command(VerboseCommand):
+    help = (
+        "A helper function clone a data from courtlistener.com into local "
+        "environment "
+    )
+
+    def __init__(self, *args, **kwargs):
+        super(Command, self).__init__(*args, **kwargs)
+        self.verbosity = None
```
mlissner, line 52: I don't think verbosity is used, is it?

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+        self.options = []
```
mlissner, line 53: I don't think `self.options` is used either, right?

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+            "--type",
+            type=str,
+            choices=VALID_TYPES,
+            help="Object type to clone. Current choices are %s"
+            % ", ".join(VALID_TYPES),
+        )
+
+        parser.add_argument(
+            "--id",
+            help="Object id, " "object id to clone",
```
mlissner, line 67: Funky here.

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def handle(self, *args, **options):
+        super(Command, self).handle(*args, **options)
+        self.verbosity = int(options.get("verbosity", 1))
+        self.type = options.get("type")
+        self.id = options.get("id")
+
+        if not self.id:
+            self.stdout.write("Object id required!")
+            sys.exit(1)
```
mlissner, line 78: I think we can use argparse to do this properly in the `add_argument` line. Maybe via the `nargs` parameter, or maybe via the `required` parameter? I forget exactly, but that's the better place to do this kind of check.

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+                self.clone_docket()
+            elif self.type == "Person":
+                self.clone_person()
+            elif self.type == "Court":
+                self.clone_court()
+            else:
+                self.stdout.write("Invalid type!")
+
+        else:
+            self.stdout.write("Command not enabled for production environment")
```
mlissner, line 93: A couple things here:
 - Better to do an early abort `if not DEVELOPMENT` than to do an if/else with nested if statements.
 - This is a good time to use `match`.
 - I tweaked this to use the full model name (as mentioned earlier)
 - This moves the methods on the class to functions on the module by removing `self`. More on this in a sec.

```suggestion
        if not DEVELOPMENT:
            self.stdout.write("Command not enabled for production environment")
        
        match self.type:
            case "search.Opinion":
                clone_opinion()
            case "search.Docket":
                clone_docket()
            case "person_db.Person":
                clone_person()
            case "search.Court":
                clone_court()
            case _:
                self.stdout.write("Invalid type!")
```

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+
+    def clone_opinion(self) -> None:
```
mlissner, line 95: All of these methods are probably better off being functions on the module. The advantage is they're easier to test and easier to import into other systems later. For example, if you're working in the Python shell, it's fairly easy to do:

```
from cl.scrapers.management.commands.clone_from_cl import clone_court
clone_court('cand')
```

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--type",
+            type=str,
+            choices=VALID_TYPES,
+            help="Object type to clone. Current choices are %s"
+            % ", ".join(VALID_TYPES),
+        )
+
+        parser.add_argument(
```
mlissner, line 65: Could ID take multiple values so that you can clone tens or hundreds of items with something like:

```
clone_from_id --type foo --ids abc def foo bar
```

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+
+        try:
+            obj = OpinionCluster.objects.get(pk=self.id)
+            print(f"OpinionCluster with id: {self.id} already in local env.")
+            return
+        except OpinionCluster.DoesNotExist:
+
+            s = requests.session()
+            s.headers = {
+                "Authorization": f"Token {os.environ.get('CL_API_TOKEN', '')}"
```
mlissner, line 108: I guess it makes sense to get this from an env, but I wonder if it's just easier to do it as a command line arg?

If you think the env is better, that's fine, but let's provide an example of passing it as part of the command. Like, I think this works, but I'm not sure:

CL_API_TOKEN='asdf' manage.py blah

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+            }
+            cluster_endpoint = f"https://www.courtlistener.com/api/rest/v3/clusters/{self.id}/"
```
mlissner, line 110: We defined cluster_endpoint up above. Shouldn't we use that?

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+            results = s.get(cluster_endpoint).json()
```
mlissner, line 111: Always have a timeout with python requests. [Always :)](https://github.com/psf/requests/issues/3070).

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+
+            cluster_datum = results
```
mlissner, line 113: Why do we need the results variable if we throw it away?

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+    def clone_opinion(self) -> None:
+        """Download opinion cluster data from courtlistener.com and add it to
+        local environment
+        """
+
+        try:
+            obj = OpinionCluster.objects.get(pk=self.id)
+            print(f"OpinionCluster with id: {self.id} already in local env.")
+            return
+        except OpinionCluster.DoesNotExist:
```
mlissner, line 104: We shouldn't have so much stuff in the `except` clause. Usually it should only be a few lines. I think the good news is you can just put a `pass` statement under the exception and then outdent the rest.

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+            del docket_datum["absolute_url"]
+            del docket_datum["audio_files"]
+            del docket_datum["tags"]
+            del docket_datum["panel"]
+            del cluster_datum["docket"]
+            del cluster_datum["citations"]
+            del cluster_datum["sub_opinions"]
+            del cluster_datum["absolute_url"]
+            del cluster_datum["panel"]
+            del cluster_datum["non_participating_judges"]
```
mlissner, line 134: Can we do this in a more tidy way? Maybe have a list of all the attributes we need to delete then do them in a loop, or have cleanup functions for each object type? Having ten `del` statements in a row is hard to read.

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+            with transaction.atomic():
+                # Get or create required court
+                try:
+                    ct = Court.objects.get_or_create(**court_datum)
+                except:
```
mlissner, line 139: We shouldn't have an empty except statement. This should list the exceptions it's supposed to catch.

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+                    ct = Court.objects.filter(pk=court_datum["id"])
+                docket_datum["court"] = ct[0]
```
mlissner, line 141: This only works if we hit the exception and do the `filter`, right? Oh, wait, no, it works because `get_or_create` returns a two-tuple and `filter` returns a list. That's a little too magic. If you want to do that you should do:

```
try:
    ct, _ = ...get_or_create()
except SomeException:
    ct = ...filter()[0]
```

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+
+                # Create docket
+                Docket.objects.create(**docket_datum)
+
+                for cite_data in citation_data:
+                    cite_data["cluster_id"] = cluster_datum["id"]
```
mlissner, line 147: This should go below with the citation related stuff, right?

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+                for op in opinion_data:
+                    op_data = s.get(op).json()
+                    # delete fields with fk or m2m relations or unneeded fields
+                    del op_data["opinions_cited"]
+                    del op_data["cluster"]
+                    del op_data["absolute_url"]
+                    del op_data["resource_uri"]
+                    del op_data["author"]
+                    del op_data["joined_by"]
+                    op_data["cluster_id"] = cluster_datum["id"]
```
mlissner, line 167: Shouldn't this go above with the similar code?

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+
+                    # Create opinion
+                    op = Opinion.objects.create(**op_data)
+
+                    # Add opinion to search engine
+                    add_items_to_solr.delay([op.id], "search.Opinion")
+
+                print(
+                    f"http://localhost:8000/opinion/{cluster_datum['id']}/{docket_datum['slug']}/"
```
mlissner, line 176: Why not use `reverse` here? We should never write out URLs.

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+                )
+
+    def clone_docket(self) -> None:
+        """Download docket data from courtlistener.com and add it to local
+        environment
+        """
```
mlissner, line 182: Could you refactor this and the other methods so they are called from the clone_opinion method, instead of repeating code?

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+    @staticmethod
+    def get_court_from_url(court_url: str) -> Court | None:
+        """Get court data from courtlistener api url"""
+        if not court_url:
+            return None
+        s = requests.session()
+        s.headers = {
+            "Authorization": f"Token {os.environ.get('CL_API_TOKEN', '')}"
+        }
+        court_data = s.get(court_url).json()
```
mlissner, line 259: This is repeated a lot. Seems like it should be pulled out, right? DRY.

mlissner (changes requested):
OK, LOTS of comments here, but I think this will be a good one to drive home a few good practices. Let me know what you think, and thanks for putting together this useful tool.

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,298 @@
...
+
+from cl.lib.command_utils import VerboseCommand
+from cl.people_db.models import Person
+from cl.search.models import Citation, Court, Docket, Opinion, OpinionCluster
+from cl.search.tasks import add_items_to_solr
+
+env = environ.FileAwareEnv()
+DEVELOPMENT = env.bool("DEVELOPMENT", default=False)
+
+VALID_TYPES = ("Opinion", "Docket", "Person", "Court")
```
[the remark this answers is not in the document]
quevon24: I think this was done in cl_update_index this way because solr required the correct model to index the item and query the id to correct object class, in this specific situation i think we don't need to that exactly the same because we are working with django, so we can do the imports directly instead of using get_model()

cl/scrapers/management/commands/clone_from_cl.py
quevon24: It works this way because an opinion cluster can have many opinions, and we need to delete those keys from each opinion

cl/scrapers/management/commands/clone_from_cl.py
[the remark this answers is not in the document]
mlissner: Yeah, the solr one uses this in a bit of a clever way. Here it's nice to do just to be consistent across scripts so that when any script needs a model as an input, you can be sure it needs something like `search.Xyz`

cl/scrapers/management/commands/clone_from_cl.py
mlissner: Sure, but I'd still expect to see all the `del` work getting done first, not some here, some there, particularly because you're in a transaction block now. Ideally when you're in the transaction you shouldn't be doing very much beyond saving things to the DB. If this change means you need a for-loop up above and a second one here, that's OK.

quevon24:
the changes you requested are ready @mlissner

mlissner:
There are still some unresolved comments. Can you please take another look and get those resolved?

quevon24:
> There are still some unresolved comments. Can you please take another look and get those resolved?

Yes, I think there were many comments and that's why some were hidden and that's why I didn't see them🤦, i'm on it 👍

quevon24:
All changes should be ready @mlissner

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,406 @@
...
+            prepared_opinion_data.append(op_data)
+
+        with transaction.atomic():
+            # Create opinion cluster
+            opinion_cluster = model.objects.create(**cluster_datum)
+
+            for cite_data in citation_data:
+                # Create citations
+                cite_data["cluster_id"] = opinion_cluster.pk
+                Citation.objects.create(**cite_data)
```
mlissner, line 135: You could do this with `bulk_create` for a performance boost, but not worth re-doing now. Something to think about in the future though.

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,406 @@
...
+    :param session: a Requests session
+    :param docket_ids: a list of docket ids
+    :param object_type: Docket app name with model name
+    :return: list of docket objects
+    """
+
+    dockets = []
+
+    for docket_id in docket_ids:
+        print(f">> Cloning docket id: {docket_id}")
```
mlissner, line 174: I used to do a lot of these kinds of things with `>` in my log lines or fancy indentation, but it makes logs harder to parse automatically. I'd suggest simplifying to just:

```suggestion
        print(f"Cloning docket id: {docket_id}")
```

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,406 @@
...
+                clone_court(
+                    session, [docket_data["appeal_from"].split("/")[7]]
+                )[0]
+                if docket_data["appeal_from"]
+                else None
+            )
+
+            docket_data["assigned_to"] = (
+                clone_person(
+                    session, [docket_data["assigned_to"].split("/")[7]]
```
mlissner, line 224: A helper function for this would probably be good:

```
.split("/")[7]]
```

Perhaps:

```
def get_id_from_url(api_url: str) -> str:
    """Get the PK from an API url"""
    return api_url.split("/")[-1]]
```

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,406 @@
...
+
+            dockets.append(docket)
+            print(
+                ">> View cloned docket here:",
+                reverse(
+                    "view_docket",
+                    args=[docket_data["id"], docket_data["slug"]],
+                ),
+            )
+
```
mlissner, line 240: No solr for dockets?

mlissner (changes requested):
This is about 1,000× better, thank you @quevon24! I think it's good now, but I made a few more comments that I think will make it great. 

Thanks for sticking with this one. It's small, but it's covering a lot of ground in terms of our norms and standards around here that I think will help with many of your future PRs.

quevon24:
> This is about 1,000× better, thank you @quevon24! I think it's good now, but I made a few more comments that I think will make it great.
> 
> Thanks for sticking with this one. It's small, but it's covering a lot of ground in terms of our norms and standards around here that I think will help with many of your future PRs.

I appreciate all your comments, I know this will help me improve my skills and code better 😊

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -0,0 +1,406 @@
...
+    :param cluster_ids: a list of opinion cluster ids
+    :param object_type: OpinionCluster app name with model name
+    :return: list of opinion cluster objects
+    """
+
+    opinion_clusters = []
+
+    for cluster_id in cluster_ids:
+        print(f">> Cloning opinion cluster id: {cluster_id}")
+        model = apps.get_model(object_type)
```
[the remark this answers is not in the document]
quevon24: It's not really necessary, but i did it like this so that all clone functions are consistent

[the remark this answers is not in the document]
mlissner: I don't see anywhere that you *couldn't* remove this though, right? You can have them consistent _and_ have none of them pass the arg?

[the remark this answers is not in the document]
quevon24: I did it this way to accept the object type arg as an input and use it to get the model, the argument could be removed from all of them but then i'll have to import all of the models manually and the object type input would be only a string to select which function to call. 

The other way is to remove only the object_type argument only in clone_opinion_cluster function, and let the others to  use the default object type value in case of being called from clone_opinion_cluster function or use the object type input arg in case you call the other functions individually(clone only dockets, courts or people) 

what do you think?

quevon24:
Almost all requested changes are complete @mlissner I only left you a comment on one of the requested changes

mlissner:
Two of our CI tools broke, so I fixed them in another branch that I just merged and then rebased this on. 

This is set to auto-merge as of now. I'll check back and make sure it gets into the system. 

Thank you for taking the initiative to write and land these tools.

mlissner (approved):

PR #2439, opened 2023-01-01
Feat: first pass at db model for record of citation by RECAP document

ss108:
Here is my first take at the DB model for keeping track of citations from RECAP documents (such as briefs) to court opinions. 

The most difficult part of this portion of the full feature is naming this model/table.


johnhawkinson:
Is it worth considering citation **to** other RECAP documents that are not opinions (e.g. motions, briefs)?

ss108:
Well, I guess I will figure out the migration issue once you guys approve the contents of the model.

@johnhawkinson IDK if you're asking me or Mike only, but the basics of my thoughts on that are that it depends on whether you want to prioritize surfacing/exposing as much info as possible versus want to improve the user experience.

johnhawkinson:
I'm just suggesting that we are often stuck with DB models for decades, so it is wise to consider future work when defining the model. I offer no opinion as to who should respond.

ss108:
Good point. 

Seems like a citation to a docket entry would be a different model though.

mlissner:
> Is it worth considering citation to other RECAP documents that are not opinions (e.g. motions, briefs)?

No, it's not really worth it for now, beyond realizing that such a feature would just require a new table. New tables are easy. Hard tables are the ones that add or remove columns from big tables. Thanks for raising this though.

mlissner (changes requested):
Yeah, this is about right. I think you pretty much nailed the model name. I think I'd go with `OpinionsCitedByRECAPDocument` though, if you don't mind tweaking that in the class name and the verbose_name_plural, that'd be great. 

The one other thing we need here is migration SQL. See [here](https://github.com/freelawproject/courtlistener/blob/main/DEVELOPING.md#special-notes-for-special-types-of-code) for a few (brief) details. If you think more would be helpful, feel free to add that to this PR. (Our docs can always be improved and you're the perfect person to do so.)

Otherwise, this looks good to me, thanks.

mlissner:
Oh, and I guess @ss108, if you want to really dial things in, we do have a commit message template you could apply. I don't usually bug volunteers about it, but since you've been around awhile, I'm guessing you'd actually care. See item three here: 

https://github.com/freelawproject/courtlistener/blob/main/DEVELOPING.md#guidelines-for-contributions

Thank you again!

ss108:
@mlissner do you want me to put the generated SQL anywhere, or does it not matter, given that we know we aren't going to run into known issues?

mlissner:
Usually we just put it next to the migration file with an sql extension and the same file name, and that's it.

ss108:
@mlissner I think this is a good cutoff point to merge. What do you think?

cl/search/models.py
```diff
@@ -2907,6 +2907,28 @@ class Meta:
...
+        RECAPDocument, related_name="cited_opinions", on_delete=models.CASCADE
+    )
+    cited_opinion = models.ForeignKey(
+        Opinion, related_name="citing_documents", on_delete=models.CASCADE
+    )
+    depth = models.IntegerField(
+        help_text="The number of times the cited opinion was cited "
+        "in the citing document",
+        default=1,
+        db_index=True,
```
mlissner, line 2921: Sorry, I just caught one more thing. This [should be done in the `indexes` meta attribute](https://github.com/django/django/pull/12716). It's easier to adjust and manage when it's there. Probably I should teach semgrep to throw an alert when db_index is used...someday. 

Anyhow, mind making that last tweak, and then we'll be good to go?

mlissner (changes requested):
Sorry for the slowness reviewing. I didn't realize it was ready for me. I found one other small thing that could be better, if you don't mind. Things should get smoother once this is merged!

cl/search/models.py
mlissner: And in case it's not obvious, after you do this, the SQL migration and the migration file both probably need regeneration.

ss108:
@mlissner is this right?

I read the PR you linked to and the threads linked therein. Seems like the issue is that using `db_index=True` actually creates 2 indices, thus taking up extra disk space? And additionally, is harder to update and manage going forward, for whatever reason? Do I have that about right?

mlissner:
Yeah, I think you have that right. I think the bottom line is that Django tried to be helpful when you set db_index=True, but over the years Django supported more kinds of indexes, so now it's best to do them using the more featureful meta parameter. In my experience the ones made with the meta parameter are much easier to reconfigure and deal with than the traditional ones.

cl/search/models.py
```diff
@@ -2907,6 +2907,29 @@ class Meta:
...
+        RECAPDocument, related_name="cited_opinions", on_delete=models.CASCADE
+    )
+    cited_opinion = models.ForeignKey(
+        Opinion, related_name="citing_documents", on_delete=models.CASCADE
+    )
+    depth = models.IntegerField(
+        help_text="The number of times the cited opinion was cited "
+        "in the citing document",
+        default=1,
+        # db_index=True,
```
mlissner, line 2921:
```suggestion
```

mlissner (approved):
This looks great. I'll get it deployed tomorrow.

ss108:
Thanks. I have started on the actual functionality.

mlissner:
Our pleasure, thank you.

PR #2445, opened 2023-01-04
Tweak tagging docs

mlissner:
This just makes the docs a bit nicer for tagging and favorites, plus adds a couple little tweaks that have been in my change history.


albertisfu (approved):
These changes seems good to me.
Merging... thanks!

PR #2450, opened 2023-01-06
Possible log injection.

openrefactory:
In file: views.py, the moosend_webhook method stores user-provided data into logs without validating the data. This allows an attacker to corrupt the log file structure. The body of the HTTP request should be sanitized before it can be added to logs.


mlissner:
Fixed via #2452.

PR #2455, opened 2023-01-09
Document tip for selenium tests in windows and wsl

quevon24:
If you develop on Windows, you will probably run into some issues when trying to run a test that requires selenium.
I've added a solution in the doc for that issue so anyone can run the tests.


mlissner:
Nice one. Thanks.

PR #2457, opened 2023-01-10
fix(docker-compose): Switch to firefox

flooie:
We use firefox on the server.
We use firefox in juriscraper.
We should be consistent and use firefox here.


flooie:
This messes up tests and we dont need to do this.  

We just need to switch over our production server.

PR #2460, opened 2023-01-11
feat(harvard): Update Harvard importer

flooie:
Update the harvard importer

Step one in the merger code 

Add and log the filepath json for the
associated opinions in the database


mlissner:
Seems fine to me!

PR #2469, opened 2023-01-13
Normalize judge values to opinion and cluster objects from harvard import

quevon24:
With this new management command we will be able to complete author field in the opinion using author_str value if the judge already exists in CL, otherwise last name will be logged.

The same for panel in opinion clusters, using the judges string field we will try to find all the judges and add them to panel. Only those found will be added.

For non_participating_judges i think we can't add the Person objects to it because OpinionCluster model doesn't have a string field for non participating judges. Probably we will need to add an string field for this.

Execute the command as a regular management command:

`docker-compose -f docker/courtlistener/docker-compose.yml exec cl-django python manage.py normalize_judges_opinions
`


cl/corpus_importer/management/commands/normalize_judges_opinions.py
```diff
@@ -0,0 +1,79 @@
...
+        self.normalize_panel_in_opinioncluster()
+
+    def normalize_authors_in_opinions(self):
+        """Normalize author_str in opinions"""
+
+        # Get opinions that have no author object and have author_str
+        opinions_with_author_str = Opinion.objects.exclude(
+            author__isnull=False
+        ).exclude(author_str__exact="")
+        for opinion in opinions_with_author_str:
```
mlissner, line 28: You'll want to do this with an iterator so that you don't pull millions of rows into memory: https://docs.djangoproject.com/en/dev/ref/models/querysets/#iterator

cl/corpus_importer/management/commands/normalize_judges_opinions.py
```diff
@@ -0,0 +1,79 @@
...
+
+        self.normalize_panel_in_opinioncluster()
+
+    def normalize_authors_in_opinions(self):
+        """Normalize author_str in opinions"""
+
+        # Get opinions that have no author object and have author_str
+        opinions_with_author_str = Opinion.objects.exclude(
+            author__isnull=False
+        ).exclude(author_str__exact="")
```
mlissner, line 27:
```suggestion
        ).exclude(author_str="")
```

cl/corpus_importer/management/commands/normalize_judges_opinions.py
```diff
@@ -0,0 +1,79 @@
...
+        for opinion in opinions_with_author_str:
+            date_filed = opinion.cluster.docket.date_filed
+            court_id = opinion.cluster.docket.court.pk
```
mlissner, line 30: This will avoid a query by getting the ID from the docket table:

```suggestion
            court_id = opinion.cluster.docket.court_id
```

cl/corpus_importer/management/commands/normalize_judges_opinions.py
```diff
@@ -0,0 +1,79 @@
...
+
+    def normalize_panel_in_opinioncluster(self):
+        """Normalize panel in opinion cluster object"""
+
+        # Get opinion cluster that have no panel objects and have judges_str
+        opinion_clusters_with_judges_str = OpinionCluster.objects.exclude(
+            panel__isnull=False
+        ).exclude(judges__exact="")
+
+        for opinion_cluster in opinion_clusters_with_judges_str:
```
mlissner, line 55: You'll want an iterator here too.

cl/corpus_importer/management/commands/normalize_judges_opinions.py
```diff
@@ -0,0 +1,79 @@
...
+        "populate panel in opinion clusters using judges field. "
+    )
+
+    def handle(self, *args, **options):
+
+        self.normalize_authors_in_opinions()
+
+        self.normalize_panel_in_opinioncluster()
+
+    def normalize_authors_in_opinions(self):
```
mlissner, line 21: You should move this and your other method to be functions, like in the command to load from the API. If you do that, your tests get easier.

cl/corpus_importer/tests.py
```diff
@@ -872,3 +883,99 @@ def test_case_names_with_abbreviations(self):
...
+                    author=None,
+                ),
+            ),
+        )
+
+        # Check that the opinion doesn't have an author
+        self.assertEqual(cluster[0].sub_opinions.all().first().author, None)
+
+        # Run command to normalize authors in opinions
+        Command().normalize_authors_in_opinions()
```
mlissner, line 947: This will be easier if your methods are elevated to functions.

cl/corpus_importer/tests.py
```diff
@@ -872,3 +883,99 @@ def test_case_names_with_abbreviations(self):
...
+
+        # Reload field values from the database.
+        cluster[0].refresh_from_db()
+
+        #  Check that the opinion now have an author
+        self.assertEqual(
+            cluster[0].sub_opinions.all().first().author, self.judge
+        )
```
mlissner, line 955: Nice.

mlissner (changes requested):
Nice. Looks pretty good to me. Just a couple little things.

quevon24:
the changes you requested are ready @mlissner

cl/corpus_importer/management/commands/normalize_judges_opinions.py
```diff
@@ -0,0 +1,88 @@
...
+            logger.info(f"Author updated in opinion id: {opinion.pk}")
+
+
+def normalize_panel_in_opinioncluster():
+    """Normalize panel in opinion cluster object"""
+
+    # Get opinion cluster that have no panel objects and have judges_str
+    opinion_clusters_with_judges_str = OpinionCluster.objects.exclude(
+        panel__isnull=False
+    ).exclude(judges__exact="")
```
mlissner, line 45: I don't think there's almost any reason to ever use `__exact`. Let's remove it here too:

```suggestion
    ).exclude(judges="")
```

mlissner (approved):
One last thing is that you have another instance of using `__exact`. I don't think there are many reasons to use it, but it doesn't harm anything so I'm going to merge. Thanks for sticking with this. 

Let me know if you need help running it, but I imagine that'll be @flooie's domain.

mlissner:
Oh, sorry, one last bit of feedback. One of your commit messages is:

```
refactor(cl.corpus_importer): Refactor normalize_judges_opinions management command, use iterators, use id column instead of .pk, convert class methods to functions and refactor tests to use functions
```

Commit messages should be much shorter than this. I think it's 72 chars on the first line and up to 80 on the remaining ones. If you're using PyCharm, I'm pretty sure it'll give you warnings about this.

quevon24:
> Oh, sorry, one last bit of feedback. One of your commit messages is:
> 
> ```
> refactor(cl.corpus_importer): Refactor normalize_judges_opinions management command, use iterators, use id column instead of .pk, convert class methods to functions and refactor tests to use functions
> ```
> 
> Commit messages should be much shorter than this. I think it's 72 chars on the first line and up to 80 on the remaining ones. If you're using PyCharm, I'm pretty sure it'll give you warnings about this.

Thanks for the comment, i had the option unchecked, i guess that's why i didn't get a warning

Issue #2493, opened 2023-01-25
Add court_id to the docket API


mlissner:
We need this for the Big Cases bot. The court URL is there, but the ID is not, and it's annoying.

PR #2495, opened 2023-01-25
feat(docket): Add court_id to the docket serializer

ERosendo:
This PR adds the court_id to the `DocketSerializer` and resolves https://github.com/freelawproject/courtlistener/issues/2493.

### DIFF
--- a/cl/search/api_serializers.py
+++ b/cl/search/api_serializers.py
@@ -49,6 +49,7 @@ class DocketSerializer(DynamicFieldsMixin, HyperlinkedModelSerializerWithId):
         view_name="court-detail",
         queryset=Court.objects.exclude(jurisdiction=Court.TESTING_COURT),
     )
+    court_id = serializers.CharField(read_only=True)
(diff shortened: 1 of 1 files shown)


cl/search/api_serializers.py
```diff
@@ -49,6 +49,7 @@ class DocketSerializer(DynamicFieldsMixin, HyperlinkedModelSerializerWithId):
         view_name="court-detail",
         queryset=Court.objects.exclude(jurisdiction=Court.TESTING_COURT),
     )
+    court_id = serializers.CharField(read_only=True)
```
mlissner, line 52: Did you try using the ReadOnlyField? It's what HyperlinkedModelSerializerWithId uses?

ERosendo: No, I didn't try to use that field, and Yes! the HyperlinkedModelSerializerWithId uses it. We can use the ReadOnlyField too because It simply returns the value of the field without modification.

mlissner:
Automerge set, thanks.

Issue #2489, opened 2023-01-24
Integrate pghistory in more apps and models


quevon24:
This is the list of models that were discussed to integrate pghistory. Let me know if i need to add other models. 

From search app:
- [x] Docket
- [x] OpinionCluster
- [x] Opinion
- [x] Court
- [x] Citation
- [x] OriginatingCourtInformation
- [x] DocketEntry
- [x] RECAPDocument
- [x] BankruptcyInformation
- [x] Claim
- [x] ClaimHistory
- [x] Tag

From people_db app:

- [x] Person
- [x] School
- [x] Position
- [x] RetentionEvent
- [x] Education
- [x] Race
- [x] PoliticalAffiliation
- [x] Source
- [x] ABARating

All models in apps:

- [x] Disclosures app
- [x] Donate app

Models from the others apps:

- [x] UserProfile
- [x] Alert
- [x] DocketAlert
- [x] Webhook
- [x] Audio

Core models:

- [x] Group
- [x] Permissions
- [x] User

PR #2497, opened 2023-01-26
Integration of pghistory in many models

quevon24:
The pghistory implementation is ready. You can see the tracked models in issue #2489 

I added some proxy models to track m2m relationships.

Comments are always welcome 😊


quevon24:
mypy seems to have problems detecting "through" model definition causing this error: `Name "Audio.panel.through" is not defined`, that is the reason why it is ignored in AudioPanel proxy model

cl/search/models.py
```diff
@@ -179,6 +180,7 @@ class Meta:
         verbose_name_plural = "Originating Court Information"
 
 
+@pghistory.track(pghistory.Snapshot())
```
mlissner, line 183: We need to ignore the `view_count` here. Can you also check if it's anywhere else that I can't remember? I don't think so, but it'll be annoying to deal with later.

cl/search/models.py
```diff
@@ -933,8 +935,29 @@ def reprocess_recap_content(self, do_original_xml: bool = False) -> None:
...
+    obj_field=None,
+)
+class DocketPanel(Docket.panel.through):
+    """A model class to track docket panel m2m relation"""
+
+    class Meta:
+        proxy = True
+
 
+class DocketEntry(AbstractDateTimeModel):
```
mlissner, line 960: We should track this, right?

cl/search/models.py
```diff
@@ -2907,6 +2971,17 @@ class Meta:
...
 
+@pghistory.track(
+    pghistory.Snapshot(),
+    obj_field=None,
+)
+class OpinionOpinionsCited(OpinionsCited):
+    """A model class to track opinion opinions cited m2m relation"""
+
+    class Meta:
+        proxy = True
```
mlissner, line 2982: I don't think we need this one, actually. The opinions cited from one to another are generated by eyecite, and if we need them we can always regenerate them.

mlissner (changes requested):
This looks great. A couple other comments that didn't fit in elsewhere:

1. A few models didn't get tracking that should probably have it: ABARating, OriginatingCourtInformation, DocketEntry, RECAPDocument, BankruptcyInformation, Claim, ClaimHistory, and Tag. Could you add these too, please? Sorry if I'm contradicting what I said before.

2. What about the tables that are part of Django? Have you thought about / looked at those? Obviously, there's the `User` table, but looking at the tables in prod, I see some others:

```
auth_group
 auth_group_permissions
 auth_permission
 auth_user
 auth_user_groups
 auth_user_user_permissions
```

I think, ideally, we'd want all of these too if that's not too hard?

cl/api/models.py
```diff
@@ -20,6 +21,7 @@ class WebhookEventType(models.IntegerChoices):
 )
 
 
+@pghistory.track(pghistory.Snapshot(), model_name="WebhookHistoryEvent")
```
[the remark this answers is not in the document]
quevon24: This is because we already have a model named "WebhookEvent", so i need to manually define the event model name

cl/people_db/models.py
```diff
@@ -1405,6 +1425,18 @@ def __str__(self) -> str:
...
+@pghistory.track(
+    pghistory.Snapshot(),
+    obj_field=None,
+)
+class Parties(PartyType):
+    """A model class to track parties m2m relation in docket model in
+    search app"""
+
+    class Meta:
+        proxy = True
```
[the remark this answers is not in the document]
quevon24: then I'll remove it

cl/search/models.py
quevon24: I'll remove it

cl/search/models.py
quevon24: yes, no problem, i'll exclude that field

quevon24:
> This looks great. A couple other comments that didn't fit in elsewhere:
> 
> 1. A few models didn't get tracking that should probably have it: ABARating, OriginatingCourtInformation, DocketEntry, RECAPDocument, BankruptcyInformation, Claim, ClaimHistory, and Tag. Could you add these too, please? Sorry if I'm contradicting what I said before.
> 2. What about the tables that are part of Django? Have you thought about / looked at those? Obviously, there's the `User` table, but looking at the tables in prod, I see some others:
> 
> ```
> auth_group
>  auth_group_permissions
>  auth_permission
>  auth_user
>  auth_user_groups
>  auth_user_user_permissions
> ```
> 
> I think, ideally, we'd want all of these too if that's not too hard?

I'm going to add the track to those models.

I think we can track auth tables using a proxy model, i'll check the documentation again

quevon24:
the changes are complete @mlissner

mlissner:
Looks great. Here we go!

mlissner (approved):

mlissner:
OK, this is deployed. It wasn't *super* smooth, I'm sorry to say:

1. The migration SQL for `search` was based on migration 0001 instead of 0011. It was generated using a shell script, so it must have had a bug. I'll push the corrected migration in a moment, since I had to generate it so I could apply it.

2. There are around 6,000 lines of SQL in this migration of which 2,000 need to be applied to the replica, because we don't want the triggers applied there (only the tables). Normally, I apply everything in the SQL files on the replica, but not in this case (or else we'd get doubled items in the event tables (maybe), which would be very bad.)

    Anyhow, to do this correctly, I made a new migration file that copy/pasted all the important parts of the SQL files for the replica. It wound up being 2,000 lines, but it applied cleanly.

4. After the new tables were created on the replica, we had a bunch of changes piled up in the publisher, so I had to fix that with:

    ```sql
    alter subscription from_prod_to_replica refresh publication;
    ```

Anyway, all seems well now, but it was a bit rockier than I'd have preferred.

PR #2501, opened 2023-01-28
Display judges outside opinion tabs

quevon24:
Display judges outside opinion tabs

image

image

image

image


flooie:
Can you add some screenshots to the PR @quevon24, which a few examples?

flooie:
Does author still stay inside the tabs if there are multiple opinions?

quevon24:
> Can you add some screenshots to the PR @quevon24, which a few examples?

Done @flooie

quevon24:
> Does author still stay inside the tabs if there are multiple opinions?

yes, i just added another screenshot at the end

flooie:
@quevon24 thank you.  this is helpful.  

One thing that jumps out - is - we list judges and authors together even when it's just one judge.  It's redundant I think.

quevon24:
> @quevon24 thank you. this is helpful.
> 
> One thing that jumps out - is - we list judges and authors together even when it's just one judge. It's redundant I think.

@flooie I have a question about this, will the judge and the author always be the same if there is only one opinion? or the author and the judges can differ?

flooie:
Judges is any judge involved in the group of decisions.  Author is the author of an opinion.  

If there is only one judge - then no  they should be the same.  but not if there are multiple judges involved.

quevon24:
@flooie in this cluster: https://www.courtlistener.com/api/rest/v3/clusters/9303920/ the judges and the author of the opinion is the same https://www.courtlistener.com/api/rest/v3/opinions/9299008/

so i guess this is an error, right?

flooie:
@quevon24 I wouldnt say its an error - its correct, I think it looks weird.  I was wondering what you thought.

mlissner:
And we should get this merged too, please. This isn't on anybody's project board. It should probably be on both of yours so that it doesn't get forgotten.

flooie:
@quevon24 this looks good to me.

flooie:
There is something about the `Western Indemnity Co. v. Pillsbury, 151 P. 398 (Cal. 1915)` that I couldn't quite verbalize.  but now that ive looked closer it seems like an issue.  We have judges here, and we have authors.  But clearly if we have multiple authors, we should be able to use that field instead of the judges field?  right?  clearly all those authors are participants in the cluster.  @quevon24 

Perhaps this isn't an issue with the PR and should be addressed as a cleanup issue?

quevon24:
@flooie i added the author in the opinion tab as you suggested 
image

I left the author inside the opinion because sometimes we have a person object, that way we can still go to its details page

quevon24:
@flooie I removed the "opinion" text from display options in the model to display the opinion tab names correctly

I also saw that the admin options for opinions shows the text as: "010combined opinion", i changed it to use the display name

image

cl/search/models.py
```diff
@@ -2749,12 +2749,12 @@ class Opinion(AbstractDateTimeModel):
     ON_THE_MERITS = "080onthemerits"
     ON_MOTION_TO_STRIKE = "090onmotiontostrike"
     OPINION_TYPES = (
-        (COMBINED, "Combined Opinion"),
-        (UNANIMOUS, "Unanimous Opinion"),
-        (LEAD, "Lead Opinion"),
-        (PLURALITY, "Plurality Opinion"),
-        (CONCURRENCE, "Concurrence Opinion"),
-        (CONCUR_IN_PART, "In Part Opinion"),
```
flooie, line 2757: I dont think we should mess with this.

flooie: I think we should do any changes to the HTML only.

flooie:
@quevon24 can you make sure to update the branch to - just validate the tests in the future and also - can you add some screenshots for the cases you showed above to see what it will look like now?

quevon24:
@flooie I updated the screenshots, i used cut template tag to remove "Opinion" in the tabs

cl/opinion_page/templates/opinion.html
```diff
@@ -403,7 +404,7 @@ <h4>Summary of Opinion</h4>
                             <a href="#{{ sub_opinion.type }}{{ forloop.counter }}"
                                aria-controls="{{ sub_opinion.type }}{{ forloop.counter }}"
                                role="tab"
-                               data-toggle="tab">{{ sub_opinion.get_type_display }}</a>
+                               data-toggle="tab">{{ sub_opinion.get_type_display|cut:"Opinion" }} {% if sub_opinion.author_str or sub_opinion.author %}<span class="text-capitalize">({% if sub_opinion.author %}{{ sub_opinion.author.name_last }}{% else %}{{ sub_opinion.author_str }}{% endif %})</span>{% endif %}</a>
```
flooie, line 407: Can you unwind this one change and then lets merge this

quevon24: done, I'll merge it when the checks are done

PR #2546, opened 2023-02-21
Tweak: Add mastodon bots to the open graph detector 

mlissner:
This is a a small PR to make it so that we detect the Mastodon crawler on CourtListener. By doing this, we should start getting open graph images on certain pages on CourtListener when Mastodon crawls. Their crawler is described here: 

https://github.com/monperrus/crawler-user-agents/blob/master/crawler-user-agents.json#L3037-L3040

This also adds the financial disclosures to the bulk data docs, since I noticed it was missing.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -75,24 +75,25 @@ jobs:
           --follow-imports=skip \
           --exclude 'migrations/*' \
           cl/api/api_permissions.py \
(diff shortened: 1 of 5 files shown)


mlissner:
Just a note to future passers-by, I had to also tweak this in the Lambda Edge function.

mlissner:
And deploy that to version of the function to CloudFront. :)

PR #2556, opened 2023-02-23
add select-all class to docket # on docket page per comment on PR #2545

DasWordNerd:
This PR adds the new `select-all` CSS class merged in #2545 to one additional instance of the `docket.docket_number` prop in the docket entries view of the recap archive. I missed this one in #2545.

I also searched for `docket_number`, `docket_Number`, `docketnumber`, and `docketNumber` to see if there were any other instances of this prop that I might have missed the first time through but I found none where adding the new class would be desirable.


mlissner:
Perfect, thank you.

mlissner:
Just tried this out live. It's great, thank you!

PR #2573, opened 2023-03-03
Increase timeout to handle large file uploads

ttys0dev:
It appears the recap extension can fail to upload larger documents/files on connections with slower upload speeds due to a 3 minute connection timeout.

### DIFF
--- a/docker/django/Dockerfile
+++ b/docker/django/Dockerfile
@@ -122,7 +122,7 @@ CMD gunicorn cl_wsgi:application \
     # Reset each worker once in a while
     --max-requests 10000 \
     --max-requests-jitter 100 \
-    --timeout 180 \
+    --timeout 1800 \
     --bind 0.0.0.0:8000
 
 #freelawproject/courtlistener:latest-scrape-rss


mlissner:
Hi, thanks for submitting this. I'm not sure that gunicorn is the only problem though. I think there are timeouts configured elsewhere. In any case though, I think three minutes is as high as we want to go. In other areas of the service, we want requests to eventually time out. I may be wrong about this, but I think if you can't upload a file in three minutes, you've got bigger problems usually.

ttys0dev:
> I think if you can't upload a file in three minutes, you've got bigger problems usually.

Seems the only problem is the timeout from what I'm estimating.

Lets take a ~50MB upload as an example:

To complete this upload within the timeout period of 3 minutes one needs to average ~2.22Mbits/s in upload speed.

[DLS broadband packages](https://www.att.com/support/article/u-verse-high-speed-internet/KM1010095/) in the US have an upload speed as slow as ~0.256Mbits/s, this works out to an upload time of roughly 26 minutes for this example 50MB file.

mlissner:
That speed is pretty abysmal and that file is pretty huge. The download speed that comes with that is so slow that I don't think you'd actually get much use out of such a connection. 

I'd be more curious about the average speed in the U.S., say, but it's going to take a lot to convince me this is an issue, since we've been running with this timeout for a while with no complaints (until now).

ttys0dev:
> That speed is pretty abysmal and that file is pretty huge.

I'm not saying it's a common case but it's a decent way to calculate a timeout setting like this(which is not a value based on the average/typical scenario but rather should be derived from worst case scenario numbers).

An upload speed of 2.22Mbits/s(minimum required for a ~50MB file to complete within the 3 minute timeout) while fairly common these days is far from guaranteed, especially on slower cellular based connections or connections that may be oversubscribed.

> The download speed that comes with that is so slow that I don't think you'd actually get much use out of such a connection.

I'm not saying it's great but a six megabit DSL connection(which is enough to watch HD netflix) is also offered which only has a max upload speed of ~0.512Mbits/s, so you'd still be looking at a 15 minute timeout for a ~50MB file even with that.

> I'd be more curious about the average speed in the U.S.

That's pretty much irrelevant for calculating the timeout value, timeouts should be high enough to handle worst case realistic scenarios(ie largest expected file with slowest expected connection).

> since we've been running with this timeout for a while with no complaints (until now)

The upload timeout failure is effectively silent unless you look at the extension network/console logs so that's not really all that surprising. It also tends to be rather transient depending on network conditions as well.

mlissner:
My concern here is that lengthening this leaves us open to DOS attacks where people tie up our connections endlessly. Killing connections after some period of time provides some modicum of protection against that, so that's why I'm not eager to make this change. 

I do think we'd be hearing more about it if it were an issue though. It's not a silent failure, because your upload would fail and you'd wonder why you can't upload big files. We had a problem in the extension uploading big files a few years ago (the method we used in the extension crashed or something, IIRC). We heard about it once a month or so.

I guess I'd be curious to get some research on what normal timeouts look like. It's not something I've looked at closely.

ttys0dev:
> My concern here is that lengthening this leaves us open to DOS attacks where people tie up our connections endlessly.

Is that an actual issue that you're seeing happen in practice? Mitigating potential DOS attacks prematurely when there are side effects for legitimate connections like this is probably not a good idea.

> Killing connections after some period of time provides some modicum of protection against that, so that's why I'm not eager to make this change.

I mean, it's not going to be a particularly effective DOS attack mitigation technique by itself, sure there's some attacks it may mitigate but there's better ways to go about that which don't impact legitimate connections. If long running workers are an issue you could just stick nginx in front with [proxy_request_buffering](http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_request_buffering) enabled or something along those lines so that you sidestep the issue entirely by buffering the transfers using a reverse proxy before sending them to the application server.

> It's not a silent failure, because your upload would fail and you'd wonder why you can't upload big files.

Sure, you might notice it there, but it's also somewhat transient due to network conditions varying in a lot of cases so you might not.

> I guess I'd be curious to get some research on what normal timeouts look like.

I mean, the timeout happens precisely 3 minutes after the upload begins in network requests, probably easiest way to test is to use the [throttling feature](https://developer.chrome.com/docs/devtools/settings/throttling/) in devtools to simulate a slow upload connection.

mlissner:
> I mean, the timeout happens precisely 3 minutes after the upload begins in network requests, probably easiest way to test is to use the [throttling feature](https://developer.chrome.com/docs/devtools/settings/throttling/) in devtools to simulate a slow upload connection.

Sorry, I meant that I'm curious what other websites use for their timeout values, if there's any kind of norm that's established here.

ttys0dev:
> Sorry, I meant that I'm curious what other websites use for their timeout values, if there's any kind of norm that's established here.

Hmm, looks like the issue is actually due to the worker type being used by gunicorn, switching to `gthread` workers like in #2585 supposedly is the right way to handle large transfers without increasing the worker timeout as only `sync` workers have an effective request timeout equal to the worker timeout.

mlissner:
Is there documentation on why gthread would be better?

ttys0dev:
> Is there documentation on why gthread would be better?

Well other than the [timeout docs](https://docs.gunicorn.org/en/stable/settings.html#timeout) there's some additional info [here](https://docs.gunicorn.org/en/stable/design.html#asyncio-workers) on `gthread` workers. There's also docs here on [choosing worker type](https://docs.gunicorn.org/en/stable/design.html#choosing-a-worker-type).

Issue #1162, opened 2020-01-30
Merge or stub all known cases into CL


flooie:
Need to add method to merge or stub all known cases into CL

nemobis:
This would be great, especially if there is at least basic information such as the date and nature of suit. (I'm interested in new and old copyright cases which have not been archived yet.)

mlissner:
We're working on it.

quevon24:
After inspecting the models, this is the minimum data for each model required to create a stub case:

Docket: court and source
Cluster: docket and date_filed
Opinion: cluster and type
Citation: cluster, volume, reporter, page and type

Also i ran some tests on the lexis and westlaw datasets and found that we don't always have a case name, a court, date filed or even some cases don't have citations. I was thinking that we could add a new opinion type to easily identify stub cases

I found a problem with Citation model, why volume field only accept integers but if you parse citation with eyecite it allows numbers and characters in volume? 

I found some citations in lexis data like: 71-A A.F.T.R.2d (RIA) 3088 or 88-1 U.S. Tax Cas. (CCH) P9373, i think this field in citation model should be a text field instead of a positive integer field, i just need to analyze if this change won't affect any other functionality in the system

quevon24:
> Yes, we clearly need to update the citation object, but thats a big update. I think you mentioned we need to discuss stubbing your PR so here goes.
> 
> I think I am learning towards creating a new table to handle stubbed cases. I think that makes a lot of sense and would let us keep the model relatively safe. That will also for the most part protect our users for displaying a stubbed case that @mlissner suggested.
> 
> I think we need to dive deeper in citations to make sure that we can safely change the volume field because but I think we may also want to think bigger about citations in general. Perhaps the easiest thing would be to allow a citation to have a text field for atypical citation formats.

Yes, definitely citations is a complex thing, if we allow the volume to be a text field that might break other things, like happened with page number in this issue: #2474 

I think the new table is the best way to handle the stubbed cases and to avoid showing them to the end user, probably one table for the case data and another table to store the stub case citations, that way we could implement a management command  to try to check if we can add that information to an existent case

PR #2582, opened 2023-03-09
Westlaw citations merger

quevon24:
The citations merger for westlaw data is ready, we don't have much  information in the dataset so this was the best way i could do it without messing with very complex algorithms to guess which case each citation belongs to. I have left comments in the code so you know what it does.

This is the way to run the command, i added a csv sample.
`docker-compose -f docker/courtlistener/docker-compose.yml exec cl-django python manage.py westlaw_citation_merger --csv /opt/courtlistener/cl/citations/management/commands/westlaw_citation_merger.csv  `

`docker-compose -f docker/courtlistener/docker-compose.yml exec cl-django python manage.py westlaw_citation_merger --csv /opt/courtlistener/cl/citations/management/commands/westlaw_citation_merger.csv --debug`


mlissner:
@flooie why don't you do first review on this, and I'll do a second pass, once you think it's ready.

cl/citations/management/commands/westlaw_citation_merger.py
```diff
@@ -0,0 +1,354 @@
...
+        #     # Add date filed to search params if exist
+        #     prep_date_filed = prepare_date(date_filed)
+        #     if prep_date_filed:
+        #         additional_search_params[
+        #             "date_filed"
+        #         ] = prep_date_filed.strftime("%Y-%m-%d")
+        #         # We have an exact date
+        #         additional_search_params[
+        #             "date_filed_is_approximate"] = False
+
```
flooie, line 157: this has a TODO - but I think it was on the right track.  you can use dates or locations to find more things.

cl/citations/management/commands/westlaw_citation_merger.py
```diff
@@ -0,0 +1,354 @@
...
+    except IntegrityError:
+        logger.warning(
+            f"Reporter mismatch for cluster: {cluster_id} on cite: {citation}"
+        )
+
+
+def prepare_date(date_str: str) -> Optional[datetime]:
+    """Convert dates like 'February 28, 2011' or '2011-02-28' to date object
+    :param date_str: date string
+    :return: date object or None
```
flooie, line 87: don't we normally have spacing here.

cl/citations/management/commands/westlaw_citation_merger.py
```diff
@@ -0,0 +1,354 @@
...
+        additional_search_params = {}
+
+        if court:
+            # Add court to search params if exists
+            if court.endswith("."):
+                # Remove dot at end, court-db fails to find court with
+                # dot at end, e.g. "Supreme Court of Pennsylvania." fails,
+                # but "Supreme Court of Pennsylvania" doesn't
+                court = court[:-1]
+
```
flooie, line 135: why not just strip `.` from court instead of checking ?

cl/citations/management/commands/westlaw_citation_merger.py
```diff
@@ -0,0 +1,354 @@
...
+                if (
+                    citation_cluster_results.first().pk
+                    != parallel_citation_cluster_results.first().pk
+                ):
+                    logger.warning(
+                        f'Possible duplicated cases for citations: "{citation}" with clusters: {citation_cluster_results.first().pk} and {parallel_citation_cluster_results.first().pk}'
+                    )
+                else:
+                    logger.warning(
+                        f'Cluster {citation_cluster_results.first().pk} already have both citations "{citation}" and "{parallel_citation}"'
```
flooie, line 251: I think this shouldn't be a warning ... more like info ...

cl/citations/management/commands/westlaw_citation_merger.py
```diff
@@ -0,0 +1,354 @@
...
+            elif results_count > 1 and parallel_results_count == 0:
+                # We got more than one result using the citation, we can
+                # try to find the case name using csv data and try to
+                # match it with the correct case
+
+                # Check if case name from csv is exactly the same in
+                # case_names list and get index, this is just a workaround
+                # to use as much of westlaw data as possible, we could try
+                # to implement something more complex
+                indexes_found = find_indexes(list(results_names), case_name)
```
flooie, line 262: I think we should be doing this anyway.  there are too many duplicate case citations to not do a little checking.

cl/citations/management/commands/westlaw_citation_merger.py
quevon24: you are right, it's easier that way

cl/citations/management/commands/westlaw_citation_merger.py
```diff
@@ -0,0 +1,354 @@
...
+                        f'parallel citation: "{parallel_citation}"'
+                    )
+
+            elif results_count == 0 and parallel_results_count == 0:
+                # No match for both citations
+                logger.warning(
+                    f'No results for case: "{case_name}" with '
+                    f'citation: "{citation}" and parallel citation: '
+                    f'"{parallel_citation}"'
+                )
```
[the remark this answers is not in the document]
quevon24: yes, i would leave that as last resort, because in the tests i did, i almost got the same results count filtering by court, with date i found many times that we have a different dates than the ones from the csv, i'll give it a try and let you know when is ready

cl/citations/management/commands/westlaw_citation_merger.csv
```diff
@@ -0,0 +1,36 @@
+"Title","Court Line","Filed Date","Citation","Parallel Cite","Docket Num","Volume"
+"Grooms v. United States","United States Court of Federal Claims.","January 07, 2013","113 Fed.Cl. 651","2013 WL 6919668","11-743C","113"
```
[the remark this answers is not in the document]
mlissner: On top of not committing them, I regret to point out, in case it's not obvious, that simply removing it and doing another commit won't fix the problem. You'll need to re do this branch and force push it to remove the commit that contained the CSV. If you don't, the commit adding it and the one removing it will both be added when we merge.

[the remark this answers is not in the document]
flooie: Regrettable indeed.

cl/citations/management/commands/westlaw_citation_merger.py
```diff
@@ -0,0 +1,354 @@
...
+        try:
+            date_obj = datetime.strptime(date_str, date_format)
+            return date_obj
+        except ValueError:
+            logger.warning(f"Invalid date string: {date_str}")
+            continue
+    return None
+
+
+def prepare_citation(citation: str) -> Union[List[FullCaseCitation], List]:
```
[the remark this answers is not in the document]
mlissner: OK, so in that case, I think you'd want to do something like:

`list[FullCaseCitation] | list[WHATEVER_GOES_HERE]`

I just wanted to make it clear that:

 - `Union` is now `|`
 - `List` is now `list`
 - `Optional` is now `None`

quevon24:
@flooie The changes are ready to review

cl/citations/management/commands/westlaw_citation_merger.csv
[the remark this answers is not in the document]
flooie: @quevon24 what do you think?

flooie:
@quevon24 this doesn't seem ready to review.  We still need to remove teh CSV files that were committed - ie create a new branch.  

I also dont see any code that would be used for stubbing out citations without opinions.  https://github.com/freelawproject/courtlistener/issues/1162

PR #2585, opened 2023-03-11
Switch to gthread workers to prevent upload timeouts with sync workers

ttys0dev:
Using gthread workers should prevent uploads from timing out at 3 minutes.

See docs:
https://docs.gunicorn.org/en/stable/settings.html#timeout

### DIFF
--- a/docker/django/Dockerfile
+++ b/docker/django/Dockerfile
@@ -117,6 +117,8 @@ CMD gunicorn cl_wsgi:application \
     --group www-data \
     # Set high number of workers. Docs recommend 2-4× core count`
     --workers ${NUM_WORKERS:-48} \
+    --worker-class gthread \
+    --threads 10 \
     # Allow longer queries to solr.
     --limit-request-line 6000 \
     # Reset each worker once in a while


ttys0dev:
> This one is a month old and we haven't gotten more than one person asking for it in all this time.

Can confirm that the problem is still present, although in practice it's hit fairly rarely. The reproduction steps are trivial so I don't see why more bug reports would be helpful here to confirm an already confirmed reproducable bug.

> I also haven't been shown how this change fixes the problem or what the pros/cons of it are.

Either way sync workers are generally a bad idea when dealing with file uploads like this unless you put a reverse caching proxy like nginx in front(I assume there is not one right?). I think in general it's a bad idea to not have a caching reverse proxy in front. See details https://dev.to/lsena/gunicorn-worker-types-how-to-choose-the-right-one-4n2c.

> I think it's defacto closed already since it's stalled for a month, but if others are also feeling the pain of this issue, please chime in.

Not sure why it stalled, are the reproduction steps not working? The bug is really obvious once you look at the timings on file uploads in the extension debugger and see that they always fail at 3 minutes exactly.

Issue #2531, opened 2023-02-07
Make pghistory work on first changes to models


mlissner:
I did an experiment with @flooie today. We grabbed a docket that had never been changed, and we changed it to see if we could revert it. It turns out that because django-pghistory tracks the NEW data, not the OLD, and writes events on INSERT, not DELETE, we can't revert the change. Here's the demo:

```txt
# Start by checking the event table for rows (there are none)
courtlistener=> select * from search_docketevent where id = 4761309;
(0 rows)

# Then, we update the row via CL's admin interface and we have an event with the NEW values
courtlistener=> select * from search_docketevent where id = 4761309;
 pgh_id  |        pgh_created_at         | pgh_label |   id    |         date_created         |         date_modified         | source | appeal_from_str |  assigned_to_str  | referred_to_str | panel_str |    date_last_index     | date_cert_granted | date_cert_denied | date_argued | date_reargued | date_reargument_denied | date_filed | date_terminated | date_last_filing | case_name_short |                         case_name                         |                          case_name_full                          |                           slug                           | docket_number | docket_number_core | pacer_case_id |         cause         |     nature_of_suit      | jury_demand | jurisdiction_type | appellate_fee_status | appellate_case_type_information | mdl_status | filepath_local | filepath_ia |                                      filepath_ia_json                                      | ia_upload_failure_count | ia_needs_upload |  ia_date_first_change  | date_blocked | blocked | appeal_from_id | assigned_to_id | court_id | idb_data_id | originating_court_information_id |            pgh_context_id            | pgh_obj_id | referred_to_id ---------+-------------------------------+-----------+---------+------------------------------+-------------------------------+--------+-----------------+-------------------+-----------------+-----------+------------------------+-------------------+------------------+-------------+---------------+------------------------+------------+-----------------+------------------+----------------+-----------------------------------------------------------+------------------------------------------------------------------+----------------------------------------------------------+---------------+--------------------+---------------+-----------------------+-------------------------+-------------+-------------------+----------------------+---------------------------------+------------+----------------+-------------+--------------------------------------------------------------------------------------------+-------------------------+-----------------+------------------------+--------------+---------+----------------+----------------+----------+-------------+----------------------------------+--------------------------------------+------------+----------------
[...]

PR #2593, opened 2023-03-16
2531 make pghistory work on first changes to models

quevon24:
This PR contains the changes requested in #2531. This contains a few changes:

- A custom model tracker to store OLD value instead of NEW when making a change to an object, the custom tracker will also allow us to avoid duplicates when only updating the auto_now fields(save without making any changes).
- Replace Snapshot tracker with CustomSnapshot tracker
- Django migrations and sql files


quevon24:
@mlissner The changes are complete, @albertisfu helped me to fix the tests related to #2591

cl/lib/pghistory.py
```diff
@@ -0,0 +1,79 @@
+import pgtrigger
+from pghistory import trigger
+from pghistory.core import DatabaseTracker, _get_name_from_label
+
+
+class CustomSnapshot(DatabaseTracker):
```
mlissner, line 6: Perhaps:

```suggestion
class AfterUpdateOrDeleteSnapshot(DatabaseTracker):
```

Calling something "custom" doesn't tell the next person much about why it exists. :)

mlissner (changes requested):
This is good stuff, Kevin. Two thoughts:

1. We should try to get this upstream instead of managing it ourselves. To that end, I updated the OLD vs. NEW bug in pghistory to ask if this would be a welcome feature, and I created a new issue to ask if `ignore_auto_now` would be a welcome feature: https://github.com/Opus10/django-pghistory/issues/83

    If we can get this upstreamed, we'll have the benefit of other people looking at your code, and we won't have to maintain it ourselves forever. Plus: Others might find it useful! 

2. The `Snapshot` class saves events when things are INSERTed or UPDATEd, right? But we want ours to do UPDATE or DELETE. I don't think I saw the events for delete though, right? We should add a test for this too.

I think that's it. Maybe let's fix number 2 now and we can proceed to merge this regardless of what happens with number 1. If we get your code into pghistory, we can eventually swap out our implementation for the one in the package.

Thank you!

cl/lib/pghistory.py
quevon24: @mlissner Yes, the Snapshot class works for inserts and updates, we just talked about the behavior when we do an update, we don't have triggers for delete action, i can add them, i think we should also store the old value when we do a deletion, right?

mlissner: Yep, exactly.

quevon24:
@mlissner, all changes are complete, i added the "ignore_field_changes" argument to exclude fields from row comparison, but not from copy

quevon24:
@mlissner the changes are ready with the new version of pghistory, 

i added by default the option to exclude all auto_now fields in a custom snapshot class instead of using the new barebone tracker "BeforeUpdateOrDelete" to avoid those extra records in the db caused by auto_now fields, this is ok, or should we use the barebone tracker?

cl/people_db/models.py
```diff
@@ -60,7 +61,7 @@
 
 
 @pghistory.track(
-    pghistory.Snapshot(),
+    AfterUpdateOrDeleteSnapshot(),
```
mlissner, line 64: Can you make all of these into one line, instead of three?

cl/search/models.py
```diff
@@ -81,7 +82,7 @@ class PRECEDENTIAL_STATUS:
 
 
 @pghistory.track(
-    pghistory.Snapshot(),
+    AfterUpdateOrDeleteSnapshot(),
```
mlissner, line 85: More triple-line issues here. I don't usually complain, but there are so many of them! :)

mlissner (changes requested):
This looks great. Just a couple little comments, and I think we'll be ready to merge this monster. Thank you.

mlissner:
Oh, sorry, and it looks like poetry is unhappy in the merge. Revert the file and add the dependency fresh?

quevon24:
@mlissner changes are ready

mlissner (approved):

mlissner:
Fantastic. Here we go, thank you!

PR #2603, opened 2023-03-29
Light code refactoring of code updating citation clusters for opinions

ss108:
@mlissner if this looks good, I can squash the commits and make the new commit message conform to the guidelines. 

I think the next PR will involve some migrations because of changes to the `RECAPDocument` model. That model will need to reference the new model created as part of this feature and previously merged, `OpinionsCitedByRECAPDocument`


cl/citations/match_citations.py
```diff
@@ -293,3 +293,6 @@ def do_resolve_citations(
         resolve_shortcase_citation=resolve_shortcase_citation,
         resolve_supra_citation=resolve_supra_citation,
     )
+
+
+# def get_resolved_citations()
```
mlissner, line 298: Can you please remove this?

cl/citations/tasks.py
```diff
@@ -88,120 +102,123 @@ def find_citations_and_parentheticals_for_opinion_by_pks(
...
-                        Parenthetical(
-                            describing_opinion_id=opinion.pk,
-                            described_opinion_id=_opinion.pk,
-                            text=clean,
-                            score=parenthetical_score(clean, opinion.cluster),
-                        )
+
+def store_opinion_citations_and_update_parentheticals(
+    opinion: Opinion, index: bool
+):
```
mlissner, line 115: A return type would be nice here, as well as some docstrings.

cl/citations/tasks.py
```diff
@@ -88,120 +102,123 @@ def find_citations_and_parentheticals_for_opinion_by_pks(
...
+        return
+
+    # try:
+    citation_resolutions: Dict[
+        MatchedResourceType, List[SupportedCitationType]
+    ] = do_resolve_citations(citations, opinion)
+    # except ResponseNotReady as e:
+    #     # Threading problem in httplib, which is used in the Solr query.
+    #     raise self.retry(exc=e, countdown=2)
+
```
mlissner, line 130: We want to keep this retry functionality in place, right? So we have a couple options:

1. Use the try/except here and return a special value in the except clause that allows the calling function to do a retry (this feels ugly).
2. Let it crash here and catch it in a try/except up in the calling function that can do a retry up there.
3. Probably something else, but those are the options that come to mind. Either way, we should make sure we maintain this try/except clause somehow.

cl/citations/tasks.py
```diff
@@ -88,120 +102,123 @@ def find_citations_and_parentheticals_for_opinion_by_pks(
...
+    #     if _opinion.pk not in currently_cited_opinions:
+    #         opinion_ids_to_update.add(_opinion.pk)
+
+    opinion_ids_to_update = set(
+        [
+            o.pk
+            for o in citation_resolutions.keys()
+            if o.pk not in currently_cited_opinions
+        ]
+    )
```
mlissner, line 154: This is a good tweak, but remove the old commented out code, please.

cl/citations/tasks.py
```diff
@@ -88,120 +102,123 @@ def find_citations_and_parentheticals_for_opinion_by_pks(
...
-                )
-
             # Save all the changes to the citing opinion (send to solr later)
-            opinion.save(index=False)
+        opinion.save(index=False)
 
-    # If a Solr update was requested, do a single one at the end with all the
-    # pks of the passed opinions
-    if index:
-        add_items_to_solr.delay(opinion_pks, "search.Opinion")
```
mlissner, line 207: I'm not deep into this area of the code, but it's weird that this is getting removed. I'm guessing it's just duplicative of the solr command above, but are we entirely sure? Is it worth looking at the commit that added this and seeing if there's a clue about why this is here?

mlissner: Looks like that happened here: https://github.com/freelawproject/courtlistener/commit/c5eefc5dfb80de73de9923bf9272c2eeb7502b10#diff-4d550ad2f1f29ad62f61c3b21017f6c7908f5c6fd737f221b6dcd0fc22a0b351R175-R181

mlissner (changes requested):
This looks pretty good, but needs a bit more work in a few places. 

Two substantive things are:

1. You removed a lot of comments during your refactor. Why? Should we put these back?
2. You removed a call to solr that seems to change the functionality. Can you make sure that it's definitely redundant?

cl/citations/tasks.py
ss108: Sorry, this was something I made a mental note I needed to uncomment and have in place, but it slipped my mind.

Just pushed an update. Option 2 is the path of least resistance. The `self` whose `retry` method is called is the Celery task, and that doesn't exist within the context of the new separate storage function.

cl/citations/tasks.py
ss108: I believe I have preserved this particular call: https://github.com/ss108/courtlistener/blob/docket-citations-fn/cl/citations/tasks.py#L113

mlissner (changes requested):
This looks good, but we're still missing some of the comments in the old version. Can you put those back in, and I think we'll be good to go?

cl/citations/tasks.py
mlissner: Yep, my bad. Thanks.

ss108:
1. Added back the comments, sorry about that oversight.

> You removed a call to solr that seems to change the functionality. Can you make sure that it's definitely redundant?

That's referring to the Solr update thing we already resolved, right?

mlissner:
Cool. Now comes the fun part. Thank you!

PR #2614, opened 2023-04-02
Add a default sort setting for Dockets

troglodite2:
The default for Dockets is to sort the DocketEntries in ascending order. This adds a boolean to the User Profile, updates forms and templates to allow the user to modify and see the setting.

The Docket view was updated to test for the new setting.


troglodite2:
image

No other user visible changes.

cl/opinion_page/views.py
```diff
@@ -264,7 +265,14 @@ def view_docket(request: HttpRequest, pk: int, slug: str) -> HttpResponse:
...
     form = DocketEntryFilterForm(request.GET)
     if form.is_valid():
+        user: UserProfile.user = request.user
+        desc = user.is_authenticated and user.profile.sort_desc
         cd = form.cleaned_data
+        if desc and not cd.get("order_by"):
+            cd["order_by"] = DocketEntryFilterForm.DESCENDING
+            form = DocketEntryFilterForm(cd)
+            form.is_valid()
+            cd = form.cleaned_data
```
mlissner, line 275: I think this is a bit hacky, because we're creating and validating the form two times. There are a couple ways to work around this. 

1. Set the ordering before making the form.
2. Set the ordering within the form using a `clean_x` function.

I think number one is a so-so approach, but it suffers because request.GET is immutable, so to do it, you have to make a copy of request.GET and then fiddle with that.

I think the second option is a bit better, and the trick for it, I believe, is to pass the request as a second parameter to a custom `__init__` method on the `DocketEntryFilterForm`. This pattern is described here: 

https://stackoverflow.com/a/1057640/64911

If you do that, you'll have the `request` object in the form, and you can do a `clean_order_by` function that takes it into account and returns a nice value. This approach seems nicer to me too than the others, because it moves this fiddly logic out of the view and into the form, where it's hidden away.

I get the sense that you're trying to bang this out, and probably don't want to refactor this, but if you don't mind attempting this approach (assuming you agree with it generally), that'd be really great.

cl/users/migrations/0011_add_sort_desc.sql
mlissner, line 1: Nice work reading the docs and making us an SQL file. I appreciate that!

cl/users/models.py
```diff
@@ -138,6 +138,10 @@ class UserProfile(models.Model):
         "subscribe to the case.",
         default=True,
     )
+    sort_desc = models.BooleanField(
```
mlissner, line 141: This isn't something that I feel really strongly, but I *think* this should probably mention dockets in the name, since CL is an ever-growing platform and `sort_desc` could refer to lots of things. Maybe a better name could be `desc_dockets` or `descending_dockets`, the latter would avoid the ambiguity of being interpreted as "description of dockets" or something like that?

cl/users/templates/profile/settings.html
```diff
@@ -170,6 +171,19 @@
...
+		    {{ profile_form.sort_desc }}
+		    Sort dockets in descending order</label>
+		{% if profile_form.sort_desc.errors %}
+		    <p class="help-block">
+			{% for error in profile_form.sort_desc.errors %}
+			    {{ error|escape }}
+			{% endfor %}
+		    </p>
+		{% endif %}
+	    </div>
```
mlissner, line 186: Could you clean up the indentation here a bit, please, so it's more like (or better than) the others (which I acknowledge are kind of weird)?

mlissner (changes requested):
This looks good! I'm sorry I wound up making more comments than I had hoped to, but most are relatively small if you're familiar with Django.

Let me know if I'm asking too much and we could certainly run with what you've done so far!

cl/opinion_page/views.py
troglodite2: I like to program in what I call "railroad siding" style.  Whenever there is a logical track to follow and something is outside of that track that creates a "siding" you want to rejoin the main line as soon as you can so no other logic needs to change. 

I like the idea of just creating or updating `clean_order_by` to handle the process.  I remember it being painful to get `request` into the clean part but it can be done.  Nice catch

cl/users/migrations/0011_add_sort_desc.sql
troglodite2: You are welcome.  The docs need to include how to run `mypy` as well as `pre-commit`

cl/users/models.py
troglodite2: `docket_default_order_desc` ?

cl/users/migrations/0011_add_sort_desc.sql
mlissner: Sorry. We're really not good at docs overall. It's probably our weakest point.

cl/users/models.py
mlissner: Lengthy, but unambiguous. Sure.

cl/users/templates/profile/settings.html
troglodite2: I'm sorry, I'm looking at the block immediately above `wants_newsletter` and it looks to my blind eyes to be formatted exactly like I formatted the `sort_desc` block.  I don't know what you are seeing.

mlissner: Maybe your editor is lying to you? 

image

cl/users/migrations/0011_add_sort_desc.sql
troglodite2: I've updated the documents but it needs more and will be in a different PR.  Resolving this conversation.

mlissner:
OK, only one thing left, which is to make it so it only has one migration. They tend to pile up and make tests and dev slow, so we always make sure there's just one in each PR. 

I was going to fix this for you because I've already put you through a lot, but I don't have push permission to your fork. Since I can't do it for you, the process is simple:

1. Undo your two migrations with:

        ./manage.py migration users 0010

2. Delete the four migration files.

3. Make one new migration with:

        ./manage.py makemigrations --name add_docket_order_preference

4. Make the sql with:

        ./manage.py sqlmigrate users 0011 > add_docket_order_preference.sql

I think that'll be close? Sorry to put you through the ringer. I'm pretty sure this is the last thing and the front end change will be easier/smaller.

mlissner (approved):

mlissner:
OK, auto-merge is enabled. Once tests pass, this will merge automatically. Once the merge is complete, we have automated deployment that'll kick in. It will fail because there's a database migration, which we always handle by hand (big data and replicated databases make it best to do these by hand), so I'll do that. Once the migration is complete, I'll re-run the automated deployment, and this will go live.

Thanks again for this nice enhancement. Once we get it on the docket sheet, people are going to be jazzed.

troglodite2:
This PR has an issue that is not resolved.  Most people will not notice.

On the Docket list the ascend/descending buttons do not reflect the default when the default is set to descending.

This is because with the new method, the cleaned data for 'order_by' is not what is passed into the template when from is passed in.  Instead `None` is passed in which is value of the submitted data.

mlissner:
> This is because with the new method, the cleaned data for 'order_by' is not what is passed into the template when from is passed in. Instead None is passed in which is value of the submitted data.

Oh shoot. Want to include that in the next PR when do you the front end part?

PR #2618, opened 2023-04-04
Lexis citations merger

quevon24:
The lexis citations merger is ready for review, all the coed have comments to understand what it does each part.

How to run the command:
`docker-compose -f docker/courtlistener/docker-compose.yml exec cl-django python manage.py lexis_citation_merger --csv /opt/courtlistener/cl/citations/management/commands/lexis_citation_merger.csv ` 

`docker-compose -f docker/courtlistener/docker-compose.yml exec cl-django python manage.py lexis_citation_merger --csv /opt/courtlistener/cl/citations/management/commands/lexis_citation_merger.csv --debug`

I left one TODO in the code because there is a case when we have only one valid citation from, but for that i need to add a slightly different process to find the correct case


flooie:
@quevon24 this should not contain the csv files - similarly to the west law data

PR #2622, opened 2023-04-04
Westlaw citation merger

quevon24:
The citations merger for westlaw data is ready, we don't have much  information in the dataset so this was the best way i could do it without messing with very complex algorithms to guess which case each citation belongs to. I have left comments in the code so you know what it does.

This is the way to run the command, i added a csv sample.
`docker-compose -f docker/courtlistener/docker-compose.yml exec cl-django python manage.py westlaw_citation_merger --csv /opt/courtlistener/cl/citations/management/commands/westlaw_citation_merger.csv  `

`docker-compose -f docker/courtlistener/docker-compose.yml exec cl-django python manage.py westlaw_citation_merger --csv /opt/courtlistener/cl/citations/management/commands/westlaw_citation_merger.csv --debug`

### DIFF
--- a/cl/citations/management/commands/westlaw_citation_merger.py
+++ b/cl/citations/management/commands/westlaw_citation_merger.py
@@ -1,0 +1,350 @@
+import itertools
+import re
+from datetime import date, datetime
+from typing import List, Union
+
+import numpy as np
+import pandas as pd
+from courts_db import find_court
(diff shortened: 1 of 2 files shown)


quevon24:
@flooie I finished all the changes, stub case creation is included, i refactored  the code to make it work better.

Issue #2619, opened 2023-04-04
Ascending/Descending buttons don't reflect default value


troglodite2:
If the Docket Default Sort order is set to descending the Docket view shows the ascending button as active.

PR #2630, opened 2023-04-07
2619 docket search order buttons

troglodite2:
Fixes: #2619


mlissner:
Simple enough. Thanks.

Issue #1434, opened 2020-11-03
Analyze order of docket sheet to optimize user workflows


mlissner:
Right now, the most recent filings are at the bottom of the docket so it reads in chronological order. I didn't give this a lot of consideration when I implemented it originally, but I just heard from a user that wants it switched. I think they're probably right that we should do so. 

I think the counter argument is that folks like to read the initial complaint. To the extent that's true, maybe we should do both: Most recent first, and pin the complaint at the top somehow. 

Thoughts?

mlissner:
Twitter poll here: https://twitter.com/FreeLawProject/status/1323720475982262279

brianwc:
I even get annoyed when docket alerts reverse-chron the docket to show me
the latest filing first. I ALWAYS want the docket to appear in regular
chron and would prefer a user-setting that controls that.

johnhawkinson:
I don't think that this is a good proposal, and I think the poll is misguided, because it asks people what they want, rather than evaluates their workflows to determine what they want (i.e. they are "threadbare recitals of a [solution], supported by mere conclusory statements." 
_Ashcroft v. Iqbal_, 556 U.S. 662 (May 18, 2009) (misquoting to characterize the notice pleading standard as a poll worthiness standard…yet it fits.)

The problem we really want to solve is that readers want to see what's new, and they don't get that when they land on the first page of a paginated docket sheet. We can't ignore that, and hopefully it will help guide us to a solution.

This is not "easy pickins" — it's actually a really hard design problem. How do our peers solve this problem, and what are the reasonable web design pattersn more broadly?

People want access to the case initiating documents, and people want access to the latest documents. I think those are the most common cases, and then there are people who want stuff in between, who are less common. And there are people who want to Cmd/Ctrl+F and search the page for a string, and those people are sad.

There's a strong argument for keeping it the way it is, because you have created a reliance interest.
Furthermore, you're not designing a website from the ground up, you are mimicing PACER/CMECF, and should take strong direction from their design choices, even when they are not great (but not, perhaps, when they are abysmal—which this is not).
And of course there are good use cases for it.

A whole deeper debate is whether there should be a knob. I would tend to think not, because that distracts us from finding a good solution.

As you proposed, providing access to the case-initiating document as well as the latest documents is an intriguing suggestion.
As is my indepenently-generated three-quarters -joking suggestion of giving the first 50 and the last 50, instead of the first or last 100.
[...]

PR #2631, opened 2023-04-07
Fixes: #1434 add accordion to docket

troglodite2:
When a user looks at a docket they will sometimes want a quick context. That context might be to see the last few entries of the docket or it could be the initial filing, which is often the complaint.

The quick look accordion presents a view into the first entry and the latest entries on the docket.  The latest entries are the last 4 plus any docket entries filed on the same date `date_filed` as the 4th entry.

This means that if somebody filed 3 entries yesterday and 3 entries today they would see the first entry and the last 6 entries.

If they filed 3 entries today and 1 entry yesterday and 2 entries the day before they would only see today's and yesterday's entries.

Further work:
* Provide filters for quick look such as this week, current week and last week, last 30 days and so forth
* Store the current setting of the accordion in the session so when the user goes to the next page the accordion is in the same state.state

### DIFF
--- a/cl/opinion_page/views.py
+++ b/cl/opinion_page/views.py
@@ -264,6 +264,24 @@ def view_docket(request: HttpRequest, pk: int, slug: str) -> HttpResponse:
     sort_order_asc = True
 
     de_list = docket.docket_entries.all().prefetch_related("recap_documents")
+    de_as_list = list(de_list)
+    first_entry = de_as_list[:1]
+    n = min(len(de_as_list) - 1, 4)
+    if n < 4:
+        last_entries = de_as_list[-n:]
+    else:
+        last_entries = de_as_list[-n:]
+        current = last_entries[0]
+        n = n - 1
+        while (
+            n > 0
+            and current.date_filed
+            and de_as_list[n].date_filed
+            and de_as_list[n].date_filed == current.date_filed
+        ):
+            current = de_as_list[n]
+            n = n - 1
+            last_entries.insert(0, current)
     form = DocketEntryFilterForm(request.GET, request=request)
     if form.is_valid():
         cd = form.cleaned_data
@@ -295,6 +313,8 @@ def view_docket(request: HttpRequest, pk: int, slug: str) -> HttpResponse:
         {
             "parties": docket.parties.exists(),  # Needed to show/hide parties tab.
             "docket_entries": docket_entries,
+            "first_entry": first_entry,
+            "last_entries": last_entries,
             "sort_order_asc": sort_order_asc,
             "form": form,
             "get_string": make_get_string(request),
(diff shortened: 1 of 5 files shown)


troglodite2:
@mlissner there are a lot of spacing changes in this PR.  This is because I re-indented `de-list.html` when I pulled out the div.  This should be a one time pain.

mlissner:
Can you include some screenshots, please?

troglodite2:
image

image

troglodite2:
Looking at that, I think there is a look and feel fix to somehow get the second entry to iterate the background differently.

cl/opinion_page/views.py
```diff
@@ -264,6 +264,24 @@ def view_docket(request: HttpRequest, pk: int, slug: str) -> HttpResponse:
...
+        n = n - 1
+        while (
+            n > 0
+            and current.date_filed
+            and de_as_list[n].date_filed
+            and de_as_list[n].date_filed == current.date_filed
+        ):
+            current = de_as_list[n]
+            n = n - 1
+            last_entries.insert(0, current)
```
mlissner, line 284: I think I'd approach this a bit differently, with two goals in mind. **First,** unfortunately, some dockets have tens of thousands of entries, so pulling all of those into a list is...not great for the poor database or for our application tier's memory. It's important we never do that (we've made that mistake before). 

**Second**, we shouldn't show the quick look if we already have those same documents on the page. A lot of dockets are really short (if they're new, say), so it's silly to show a quick look when you can already see the whole thing. It's also unnecessary to show the initial complaint when you're on page one of a DESC docket. It'd be half an inch away already. 

So with that in mind, I think what we need here is probably three queries:

 - The current one to get the usual stuff, and then paginate it. If there's more than one page, then:
     - If the initial complaint isn't in the selected page (i.e., it's not page 1):
         - A second query to show the initial complaint. This should probably just query `entry_number=1`, I think? 
     - If we're not on the last page:
         - A third query for the last few items.
 - Else, if there's only one page, no quick look needed; just do the one query.

I think if we don't have the initial complaint, it makes sense to show an error in the quick look saying as much. If we don't have the last few items, we won't know it. 

That sound OK?

troglodite2: I disagree regarding "we shouldn't show if..."  The quick look is there for those that want to open it and look.  If they can see everything they want, they wont open it.

If they don't see what they want, they will open it and it will be in a standard view.

The goal is to not have to scroll to see what you are interested in.

mlissner: Having looked at a lot of dockets, scrolling isn't usually the issue so much as pagination. I don't want to repeat content in the quick look that's already there on the page.

troglodite2: I disagree with your opinion.

I'll code up the changes per your suggestions and push up the commits shortly.

troglodite2:
The logic for this is a bit worse than might be expected.  We have to run all of the

mlissner:
I don't feel like we have consensus on this one even though we talked about it a lot. I'm going to close it for now since I know I won't have time to get back to it for a while, but it was still a very useful process for moving the conversation forward.

Hopefully when we've had some time and distance from the issue, we can take another swing at it and get it across the finish line. 

Thanks for pushing on it. I'm just trying to be realistic as I clear out my backlog. I hope you understand.

Issue #2637, opened 2023-04-11
Add Sort ordering to docket-alerts page


troglodite2:
It would be nice to have the ability to order RECAP Docket Alerts

mlissner:
Nice! I think that's the right idea, but I'd suggest using the Angle Up/Down icons from Font Awesome: 

https://fontawesome.com/icons/angle-up?f=classic&s=solid

Also:
 - Maybe don't make the headers blue? 
 - Maybe use the `gray` class from our CSS on the icons?

Thanks. Nice little enhancement. We need about a hundred of these.

troglodite2:
![image](https://user-images.githubusercontent.com/82225842/231289005-4f2ec257-7203-49c4-8830-a55d8471f850.png)

mlissner:
The header should still be black like it is now. Just the icons get the gray?

troglodite2:
The headers, "Case Name ^" is an anchor. I put `class="gray"` on the <a> element.  I think I would need to put a <span> around italics to get it to be gray and the rest of the anchor black.  The cheat uses a template tag to figure out which direction the caret should be.

```
<tr>
              <th><a class="gray" href="[?order_by=name](view-source:http://127.0.0.1:18000/profile/docket-alerts/?order_by=name)">Case Name&nbsp;<i class="fa fa-angle-up"></i></a></th>
              <th><a class="gray" href="[?order_by=court](view-source:http://127.0.0.1:18000/profile/docket-alerts/?order_by=court)">Court&nbsp;<i class="fa fa-angle-down"></i></a></th>
              <th colspan="2"><a class="gray" href="[order_by=hit](view-source:http://127.0.0.1:18000/profile/docket-alerts/order_by=hit)"> Last&nbsp;Hit&nbsp;<i class="fa fa-angle-up"></i></a></th>
          </tr>
```
Let me know what class(es) you want for the `<a>` and `<i>` elements and I'll put them in there.
I couldn't find a class for "black" easily.

mlissner:
Try:

```
class="no-underline black-link"
```
With the gray on the icon?

troglodite2:
![image](https://user-images.githubusercontent.com/82225842/231293350-c3f55a6c-255d-4c11-baea-5ba80e266eb5.png)

mlissner:
Nice! Thanks!

PR #2639, opened 2023-04-11
Fixes: #2637 Add Sorting Option to Lists

troglodite2:
Makes the headers of docket-alerts clickable to toggle sort order by the different columns

This only adds the sorting capabilities to docket alerts.  The code can be used with other lists.  All that needs to be done is to add the tags from extras.

### DIFF
--- a/cl/custom_filters/templatetags/extras.py
+++ b/cl/custom_filters/templatetags/extras.py
@@ -5,6 +5,7 @@ from django.core.exceptions import ValidationError
 from django.template import Context
 from django.utils.formats import date_format
 from django.utils.html import format_html
+from django.utils.http import urlencode
 from django.utils.safestring import SafeString, mark_safe
 
 register = template.Library()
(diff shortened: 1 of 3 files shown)


troglodite2:
This PR is not ready for merging. Please do initial review while I fix sort by hit.

troglodite2:
@mlissner I do not know why flynt is failing.  pre-commit passes in my docker instance.

mlissner:
I think we can ignore flynt today. Github is having all kinds of failures for some reason. I'm hopeful it'll be better tomorrow. 

When this is ready, please request a review, and I'll come take a look. Screenshots of the final version are also always appreciated, if they wind up different from the ones in the issue.

troglodite2:
The images are the same as on the issue.  I can't request a review because a required workflow failed.
@mlissner

cl/users/views.py
```diff
@@ -88,9 +88,25 @@ def view_search_alerts(request: HttpRequest) -> HttpResponse:
...
+        direction = ""
+    name_map = {
+        "name": "docket__case_name",
+        "court": "docket__court__short_name",
+        "hit": "date_last_hit",
+    }
+    if order_by not in name_map.keys():
+        order_by = "date_created"
+    else:
+        order_by = direction + name_map[order_by]
```
mlissner, line 105: How about:
```suggestion
    order_by = direction + name_map.get(order_by, "date_created")
```

cl/custom_filters/templatetags/extras.py
```diff
@@ -120,3 +121,30 @@ def get(mapping, key):
...
+        else:  # order_by for different column
+            dict_[field] = value
+    else:  # No order_by
+        dict_[field] = value
+    return urlencode(sorted(dict_.items()))
+
+
+@register.simple_tag
+def sort_caret(request, value) -> SafeString:
+    current = request.GET.get("order_by", "*UP*")
```
mlissner, line 145: Sorry if I'm missing it, but how is `*UP*` used?

mlissner (changes requested):
Neat solution. A couple little things, but looks like it should be a nice tweak.

cl/custom_filters/templatetags/extras.py
troglodite2: It is not `None` which makes the following code simpler.  The text was chosen because nobody can name a field with an asterisk and the UP tells the developer that it is going to default to the up direction.

mlissner: Cool. I think I'd have used a comment, b/c this made me go hunting for how that value was used, but all good. Merging in a sec.

mlissner (approved):
Thanks. This is nice. One last thing, since you're doing a lot of PRs. I didn't ask for a tweak on it here, but our indent style is two spaces for HTML. It's in our editorconfig, if your editor supports that. We used to do four spaces, so lots of files still are that way too, but we're slowly moving to only two, in theory, at least.

Thanks again. Merging!

cl/custom_filters/templatetags/extras.py
troglodite2: Sorry.  It is a standard pattern from way back when.

PR #2641, opened 2023-04-12
Allow for directory wide settings for the one true Editor

troglodite2:
Ignore directory/sub-directory settings files for emacs. This will allow for developers to use emacs with CL specific settings, such as indenting rules in templates.


mlissner:
No idea what that file is (I don't use emacs), but sure, looks good, thanks.

PR #2643, opened 2023-04-12
1368 - New Pagination Include

troglodite2:
Creates a new pagination include that will optionally create a first, prev, clickable numbers, next, last page navigation row.

To use: {% include "includes/new_pagination.html" with page_obj=<page>
only %}

The "only" is there for security and everything functions without it.

Options:
 hide_first - hide the first button
 hide_numbers - replace the clickable numbers with Page n of x
 hide_last - hide the last button


mlissner:
I haven't looked too closely, but a few quick questions:

1. You say:

    > The "only" is there for security and everything functions without it.

    Sorry, what is the risk this is mitigating? I'm not sure I follow. I couldn't even see how it did...anything? :)

2. Sadly, somehow, you've got a merge conflict. That's surprisingly bad luck.

3. Why keep the old pagination around? Why not just blow it away if this is subsuming it, and then we don't get stuck with two pagination footers forever?

troglodite2:
@mlissner the "only" keyword is Django.  It tells the template to ONLY pass the context variables listed on the include.  Habits from the old time.  If the included routine doesn't need it, don't give it access.  In this case it makes no difference if only is used.  

Yes, because I did the two modifications that both required changes to extras.py it caused a merge conflict.  I tried to fix it on my end but it didn't work.  Just except both sides.

There is magic in the old pagination code for webhooks.  I've not looked into exactly why it exists but it does.  Until I understand what webhooks needs differently it wasn't safe to pull the old code.

Maybe what is needed is code in pagination.html that spits an error message if it isn't a webhook.

mlissner:
> the "only" keyword is Django.

Woah. Great. Never learned that one before. I'll take a look at the webhooks pagination magic when I get a sec and try to form an opinion accordingly.

mlissner:
Sorry, I don't understand. Why is there a merge conflict? That should be fixed before I can merge this one.

cl/assets/templates/includes/new_pagination.html
```diff
@@ -0,0 +1,83 @@
...
+          {% if not hide_numbers %}
+            {% for page in page_obj|page_prev_range %}
+              <a class="btn btn-default" href="?{{ get_string }}page={{ page }}">{{page}}</a>
+            {% endfor %}
+            <span class="btn btn-default disabled"> {{ page_obj.number }}</span>
+            {% for page in page_obj|page_next_range %}
+              <a class="btn btn-default" href="?{{ get_string }}page={{ page }}">{{page}}</a>
+            {% endfor %}
+          {% else %}
+            Page {{ page_obj.number }} of {{ page_obj.paginator.num_pages }}
```
mlissner, line 47: The old one had this:

```suggestion
            <span class="hidden-xs">Page </span> {{ page_obj.number|intcomma }} of {{ page_obj.paginator.num_pages|intcomma }}
```

Seems like the intcomma and hiding the word "Page" are still good ideas?

cl/assets/templates/includes/new_pagination.html
```diff
@@ -0,0 +1,83 @@
...
+			        <span class="hidden-xs hidden-md">Prev.</span>
+            </span>
+          {% endif %}
+        </div>
+        <div class="col-xs-8 text-center">
+          {% if not hide_numbers %}
+            {% for page in page_obj|page_prev_range %}
+              <a class="btn btn-default" href="?{{ get_string }}page={{ page }}">{{page}}</a>
+            {% endfor %}
+            <span class="btn btn-default disabled"> {{ page_obj.number }}</span>
```
mlissner, line 42: We should add `intcomma` everywhere that makes sense. E.g.:
```suggestion
            <span class="btn btn-default disabled"> {{ page_obj.number|intcomma }}</span>
```

cl/assets/templates/includes/new_pagination.html
```diff
@@ -0,0 +1,83 @@
...
+            <span class="btn btn-default" >
+              <i class="fa fa-caret-left no-underline" ></i>
+			        <span class="hidden-xs hidden-md">Prev.</span>
+            </span>
+          {% endif %}
+        </div>
+        <div class="col-xs-8 text-center">
+          {% if not hide_numbers %}
+            {% for page in page_obj|page_prev_range %}
+              <a class="btn btn-default" href="?{{ get_string }}page={{ page }}">{{page}}</a>
```
mlissner, line 40:
```suggestion
              <a class="btn btn-default" href="?{{ get_string }}page={{ page }}">{{ page|intcomma }}</a>
```

cl/assets/templates/includes/new_pagination.html
```diff
@@ -0,0 +1,83 @@
...
+            {% endfor %}
+            <span class="btn btn-default disabled"> {{ page_obj.number }}</span>
+            {% for page in page_obj|page_next_range %}
+              <a class="btn btn-default" href="?{{ get_string }}page={{ page }}">{{page}}</a>
```
mlissner, line 44:
```suggestion
              <a class="btn btn-default" href="?{{ get_string }}page={{ page }}">{{ page|intcomma }}</a>
```

cl/assets/templates/includes/new_pagination.html
```diff
@@ -0,0 +1,83 @@
...
+              </a>
+            {% endif %}
+          {% else %}
+            <span class="btn btn-default" >
+			        <span class="hidden-xs hidden-md">Next</span>
+              <i class="fa fa-caret-right no-underline" ></i>
+            </span>
+            {% if not hide_last %}
+              <span class="btn btn-default" >
+			          <span class="hidden-xs hidden-md">Last</span>
```
mlissner, line 73: You have `hidden-md` on a bunch of things. How come? Do you have screenshots at each breakpoint? That'd definitely help, if you can provide them.

cl/search/templates/search.html
```diff
@@ -274,7 +274,7 @@ <h2 id="result-count" class="bottom">
             {% include "includes/no_results.html" %}
         {% endif %}
 
-        {% include "includes/pagination.html" %}
+        {% include "includes/new_pagination.html" with page_obj=results only%}
```
mlissner, line 277: We cannot allow users to go to the last page of search results (it's a performance nightmare). Let's be sure to disable that.

mlissner (changes requested):
OK, a handful of comments here. It looks pretty good, but a few things would make it better:

1. Some screenshots would be very helpful to see how the pagination looks at different breakpoints. 

3. Let's call this pagination.html and rename the old pagination as htmx_pagination. Then we can take out the `if webhooks` parts of it, and keep it around (but simplified). 

Thank you!

cl/assets/templates/includes/new_pagination.html
troglodite2: I considered this, but disagree.  Consider the following `1,234 1,235 1,236` Those are easy to read individually, but for passing information I think `1234 1235 1236` is likely a better option.

cl/assets/templates/includes/new_pagination.html
troglodite2: I was attempting to copy what the other template had. 

image
image
image

I am going to say that the different breakpoints are not right yet.

cl/search/templates/search.html
troglodite2: You have said this before, but I don't understand how the last page is any more expensive than any other page.  Regardless of which page you go to you are going to have to apply the query filter, WHERE in the SELECT statement, after that you are going to have to sort the results.  Finally, you do the magic where you see which rows you want returned.

`ASC LIMIT 5 OFFSET 15` is what the ORM generates for `queryset[15:20]`

Given how it works, why do you think it is a performance nightmare?

cl/assets/templates/includes/new_pagination.html
mlissner: Yeah, I think I'd suggest dropping features as screens get narrower. Start by dropping the words "First," "prev", etc, then drop the page buttons, finally, if you need to, drop the First/Last buttons, but in no case should we wind up on two lines like this.

cl/assets/templates/includes/new_pagination.html
mlissner: Want to do screenshots of both? That'd help me see it, but I think commas help, generally, at least when the numbers are far enough away from each other (and have borders and boxes) that they don't run together.

mlissner:
A few replies above. Once we get the screenshots looking good for various breakpoints, I'll do another code review.

cl/assets/templates/includes/new_pagination.html
troglodite2: The Bootstrap 3 break points are at 768, 922, and 1200.  Even at the 767 everything still looks good.  It isn't until we get to about 1/2 of the "xs" breakpoint does it start to stack the first/prev and next/last buttons.

We can design for the extra small to be 767, or we can design for extra small meaning some smaller number.  Once I know what the design parameters are, then I can move forward.

mlissner: The iPhone 12/13 mini is 375px and fairly new Samsung phones are 360, so I think it should look good until at least 360.

troglodite2: 360:
image
767:
image
768 and 800:
image
992 and 1199:
image
1500:
image
Just some samples.

troglodite2:
Per review, we are going to remove the number buttons and leave just the first and last buttons.  The issue is making the page look good at all levels of pagination.  Specifically, when the page number gets larger.

troglodite2:
Mini 360:
image
xs 767:
image
small 800:
image
sm 991:
image
md 1199:
image
lg 1500:
image
Desktop:
image

mlissner:
Looks good, but I think you made the text smaller?

troglodite2:
I didn't make the text smaller, but it doesn't look right with the same size as the button text.  I've increased the size to "medium"
image

mlissner:
In the live pagination it has a class of `large`: 

image

(that's from the dom [here](https://www.courtlistener.com/?type=r&q=unicorn&order_by=score+desc))

troglodite2:
image
image

mlissner:
Much better.

cl/assets/templates/includes/pagination.html
```diff
@@ -1,61 +1,73 @@
...
-<div class="well v-offset-above-3 hidden-print">
-  <div class="row">
-    <div class="col-xs-2 col-sm-3">
-      {% if results.has_previous %}
-        <div class="text-left">
-          {% if webhooks %}
-            <button id="webhooks-paginator"
-               hx-get="{% url 'webhook_events-list' format="html" %}?debug={{debug}}&page={{ results.previous_page_number }}&webhook__event_type={{ type_filter }}&event_status={{ status_filter }}"
-               hx-target="#webhook_table_body"
+{% load extras %}
```
mlissner, line 2: Do we still need extras here?

cl/assets/templates/includes/pagination.html
```diff
@@ -1,61 +1,73 @@
...
+<div class="col-xs-12" >
```
mlissner, line 3: Is there a reason you wrapped this in a col-xs-12? It wasn't nested this way before, except on the docket and it usually causes weirdness unless you have col's inside row's properly.

cl/assets/templates/includes/pagination.html
```diff
@@ -1,61 +1,73 @@
...
+              <span class="hidden-xs btn btn-default" >
+                <i class="fa fa-caret-left no-underline" ></i>
+                <i class="fa fa-caret-left no-underline" ></i>
+			          <span class="">First</span>
+              </span>
+            {% endif %}
+            <span class="btn btn-default" >
+              <i class="fa fa-caret-left no-underline" ></i>
+			        <span class="hidden-xs">Prev.</span>
+            </span>
```
mlissner, line 34: If there's not a prev or next, let's not have a button at all instead of having a fake button. That'll keep with current design.

cl/assets/templates/includes/pagination.html
```diff
@@ -1,61 +1,73 @@
...
+			        <span class="hidden-xs">Next</span>
+              <i class="fa fa-caret-right no-underline" ></i>
+            </span>
+            {% if not hide_last %}
+              <span class="hidden-xs btn btn-default" >
+			          <span class="">Last</span>
+                <i class="fa fa-caret-right no-underline" ></i>
+                <i class="fa fa-caret-right no-underline" ></i>
+              </span>
+            {% endif %}
```
mlissner, line 67: So I think this can go too, to keep with current design.

cl/assets/templates/includes/htmx_pagination.html
```diff
@@ -0,0 +1,61 @@
+{% load humanize %}
+
+{% if results.has_previous or results.has_next %}
+<div class="well v-offset-above-3 hidden-print">
+  <div class="row">
+    <div class="col-xs-2 col-sm-3">
+      {% if results.has_previous %}
+        <div class="text-left">
+          {% if webhooks %}
```
mlissner, line 9: Correct me if I'm wrong, but I think this is only used by webhooks now, right? If so, can we simplify this by removing these if/then statements?

mlissner (changes requested):
OK, I finally had time to review this carefully. It's largely there. Just a couple of tweaks, please and we'll finally get this merged!

cl/assets/templates/includes/htmx_pagination.html
troglodite2: Clean the htmx_pagination.html file so that only webhook code remains, removing conditionals?

mlissner: Yep, exactly. It's only being used by webhooks, so why do we need the `if webhooks` junk and the unused `else`'s, right?

cl/assets/templates/includes/pagination.html
mlissner: You marked this resolved, but it's still here?

troglodite2: The "col-xs-12" is still there because it is what was there to begin with.  When the new pagination.html was created, I pulled the formatting from the old version.  The old version had a `class="col-xs-12 col-sm-3"`  In testing the responsive layouts, the col-sm-3 was deleted, leaving just the col-xs-12.

In particular, it adds whitespace above the pagination header.

I marked it as resolved because the resolution is to keep it there.

mlissner: Where do you see the col-xs-12 in the old one? The version in `main` begins:

```
{% load humanize %}

{% if results.has_previous or results.has_next %}
<div class="well v-offset-above-3 hidden-print">
  <div class="row">
    <div class="col-xs-2 col-sm-3">
      {% if results.has_previous %}
        <div class="text-left">
```

troglodite2:
```
{% load humanize %}

{% if results.has_previous or results.has_next %}
<div class="well v-offset-above-3 hidden-print">
  <div class="row">
    *** <div class="col-xs-2 col-sm-3"> ***
      {% if results.has_previous %}
        <div class="text-left">
```
During testing, the "col-sm-3" was found to be redundant and was removed.

troglodite2: Wait, re looking.

troglodite2:
```
git diff 1368-update-pagination-links -- docket.html
diff --git a/cl/opinion_page/templates/docket.html b/cl/opinion_page/templates/docket.html
index a0e460a9b..012a42dfe 100644
--- a/cl/opinion_page/templates/docket.html
+++ b/cl/opinion_page/templates/docket.html
@@ -56,5 +56,9 @@
 {% endif %}
 
 
-{% include "includes/pagination.html" with page_obj=docket_entries %}
+{% if docket_entries.paginator.num_pages > 1 %}
+    <div class="col-xs-12">
+    {% include "includes/pagination.html" with results=docket_entries %}
+    </div>
+{% endif %}
 {% endblock %}
```

mlissner (approved):

PR #2647, opened 2023-04-13
Fix DataError exception with citation redirector

quevon24:
This PR fixes issue #2474 

To solve the problem of sorting when we have page numbers with letters is to use natural sorting to be able to sort number strings correctly, i added the natsort library that allow us to identify numbers anywhere in a string and sorts them naturally (just like we humans do).

That way, we don't need to cast the page number to an integer.


quevon24:
@mlissner The PR is ready for review

cl/opinion_page/views.py
```diff
@@ -827,17 +828,39 @@ def citation_handler(
...
+        # Find the position of the item that we added
+        citation_item_position = sort_possible_matches.index(citation_item)
+
+        if citation_item_position > 0:
+            # if the position is greater than 0, then the previous item in
+            # the list is the closest citation, we get the id of the
+            # previous item, and we get the object
+            possible_match = OpinionCluster.objects.filter(
+                id=sort_possible_matches[citation_item_position - 1][0]
+            ).first()
```
mlissner, line 862: This can be a `get()` instead of a `filter()`, right?

cl/opinion_page/views.py
```diff
@@ -827,17 +828,39 @@ def citation_handler(
     if cluster_count == 0 and page.isdigit():
         # Do a second pass for the closest opinion and check if we have
         # a page cite that matches -- if it does give the requested opinion
-        possible_match = (
+
+        possible_match = None
+
+        # Create a list of the closest opinion clusters id and page to the
+        # input citation
+        closest_opinion_clusters = list(
```
mlissner, line 836: It took me a bit of time to remember what this code is doing, so I think it just needs some better comments. Github doesn't let me make suggestions on unchanged lines of code, but can we tweak this area of the code to be:

```
    if cluster_count == 0 and page.isdigit():
        # We didn't get an exact match on the volume/reporter/page. Perhaps it's a pincite.
        # Try to find the citation immediately *before* this one in the same book. To do so,
        # Get all the opinions from the book, sort them by page number (in Python, b/c pages
        # can have letters), then find the citation just before the requested one.

        possible_match = None

        # Get all the citations in the book
        closest_opinion_clusters = list(
```

mlissner (changes requested):
Sorry, this one has some merge conflicts, and I made a few small comments, but otherwise, I think it's ready to go. Thank you!

cl/opinion_page/views.py
flooie: can I just ask, what are we going to do with citations with page numbers that have letters.

mlissner: What do you mean? They should sort correctly and more or less work. Can you be more concrete?

flooie: well. we have an issue inside the lexis and west law citation import - for formats that dont fit.  117A XYZ 13A just to make up something.  

It's probably a problem for down the road but we are probably going to need to update this as soon as we update the citation model.  Presuming we update the citation model.

mlissner: Sorry, I still don't understand. The page is a TextField in the model. What's the problem? What do you mean they don't fit?

flooie: this won't be working soon `if cluster_count == 0 and page.isdigit():` because we need to move away from page being a digit.

mlissner: OK, now I understand your concern. Yeah, I think that `isdigit()` was put in to fix a bug a while back. @quevon24 could you check in the git blame and see if you can learn why we check for digits here? I suspect the fixes in this PR make the `isdigit` check unnecessary going forward.

quevon24: I think you added isdigit() because the citation handler probably failed with a "wrong" page number (with a letter) and it worked, but now that we have page numbers with letters, we don't need that condition.

That part can be removed now because with natural sort we don't have to worry about whether page numbers contain letters or not.

I made the requested changes

mlissner (approved):

Issue #2652, opened 2023-04-14
It would be nice to be able to copy/paste citations directly from the page


troglodite2:
When citing documents the different pieces are often on the page but not organized in a way that makes it easy to just cut and paste a citation.

With the metadata on dockets, recap documents and others, it would be nice if there was just one place to look to see a citation that can be used by normals.

PR #2653, opened 2023-04-14
2652 add citation field

troglodite2:
2652 Add a citation field to dockets and RECAP documents
    
    Add includes/citation.html which takes an object and a format string and
    creates a metadata field.  This includes pure javascript for copying the
    citation to the clipboard and for copying the URL to the clipboard.
    
    Added a citations template tag.  This takes in an object and a format
    string.  If there is no format string then it uses a default format
    string.  The code pulls information out of the object and then formats
    it in reasonable ways.
    
    If certain data is missing the tag will clean up the format before using
    it.
    
    The format string is of the form:
            {thing1} {thing2} {thing3}
    This is a standard python formatting string.
    
    There are currently a few things that the tag knows about, more can be
    added.
    
    Future work: Create a UserProfile value that allows users to select and
    potentially create their own citation formatting strings.
    
    WARNING: The code takes the formatting string as is and returns the
    results as a safe string.  At this time it is not safe to hand user
    created formatting strings directly to the citation tag.
Fixes: #2652


cl/custom_filters/templatetags/extras.py
```diff
@@ -148,3 +151,63 @@ def sort_caret(request, value) -> SafeString:
...
+
+@register.simple_tag
+def citation(obj, fmt=None) -> SafeString:
+    if isinstance(obj, Docket):
+        docket = obj
+        date_of_interest = (
+            docket.date_terminated
+            if docket.date_terminated
+            else docket.date_last_filing
+        )
```
mlissner, line 164: Why are we using terminated here?

cl/opinion_page/templates/docket_tabs.html
```diff
@@ -184,7 +184,7 @@ <h2>{{ docket.court }}</h2>
         <span class="meta-data-value">{{ docket.date_argued }}</span>
       </p>
     {% endif %}
-
+    {% include "includes/citation.html" with obj=docket fmt=cite_format %}
```
mlissner, line 187: Where is `cite_format` coming from or being used?

cl/custom_filters/templatetags/extras.py
```diff
@@ -148,3 +151,63 @@ def sort_caret(request, value) -> SafeString:
...
+        ecf = ""
+    elif isinstance(obj, DocketEntry):
+        docket = obj.docket
+        date_of_interest = obj.date_filed
+        ecf = obj.entry_number
+    else:
+        return mark_safe("Unknown object type")
```
mlissner, line 171: I'd sort of rather if this totally crashed so this doesn't get accidentally used somewhere.

cl/custom_filters/templatetags/extras.py
```diff
@@ -148,3 +151,63 @@ def sort_caret(request, value) -> SafeString:
...
+            name_short=docket.case_name_short,
+            case_bb=f"No. {docket.docket_number}",
+            reporter="",
+            court_bb=docket.court.citation_string,
+            court_short=docket.court.short_name,
+            court_full=docket.court.full_name,
+            date_bb=date_of_interest,
+            ecf_bb=f"ECF {ecf}",
+        )
+    )
```
mlissner, line 213: This feels overly complicated and hard to understand. 

Could we accomplish the same with something like:

```
cite = []
if docket.case_name:
    cite.append(f"docket.case_name,")
if next_thing:
    cite.append(next_thing)
cite_str = " ".join(cite)
```

I think I'd also just like to see different `if/else` blocks for each format it can do. 

So at the top, it should do:

```
if it's a docket:
    # Format should be XYZ
    ....make the docket citation...
elif it's a document:
    # Format should be ABC
    ...make the document citation...
elif it's something else:
    CRASH
```

Could you try to refactor it that way?

cl/custom_filters/templatetags/extras.py
```diff
@@ -148,3 +151,63 @@ def sort_caret(request, value) -> SafeString:
...
+    if not ecf:
+        # remove all references to ecf
+        fmt = re.sub(r"\{ecf.*?\}\s*", "", fmt)
+    fmt = re.sub(r"\(\s*?\)", "", fmt)  # remove empty ()
+    fmt = re.sub(
+        r"\s\s+", " ", fmt
+    )  # change multiple spaces to a single space
+    fmt = re.sub(r",\s?,+", ",", fmt)  # change double comma to a single comma
+    fmt = re.sub(r"\s+$", "", fmt)  # remove trailing whitespace
+    fmt = re.sub(r",+$", "", fmt)  # remove trailing commas
```
mlissner, line 198: What are all these cleanup things for? Do we know we need them?

mlissner (changes requested):
A few comments. Not sure I like your approach, sorry. I think it could be simpler to understand, if, perhaps a bit longer.

cl/assets/templates/includes/citation.html
```diff
@@ -0,0 +1,7 @@
+{% load extras %}
+<p class="bottm" >
+  <span class="meta-data-header" >Citation:</span>
+  <span id="citation" href="" class="meta-data-value">{% citation obj fmt %}</span>
+  <i class="fa fa-copy" onClick="navigator.clipboard.writeText(document.getElementById('citation').textContent);" title="Copy citation"></i>
+  URL<i class="fa fa-copy" onClick="navigator.clipboard.writeText(window.location.href)" title="Copy URL"></i>
```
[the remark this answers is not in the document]
troglodite2: I don't think leaving it there is an actual issue.  Is this a look thing?

cl/opinion_page/templates/docket_tabs.html
troglodite2: Moved to cl.settings.django as DEFAULT_CITE_FORMAT

cl/custom_filters/templatetags/extras.py
troglodite2: Yes, they are needed.  The methodology is to create a format screen with all the potential parts that could be used.  Then we evaluate if those parts are actually going to be used.  If they are not used, then we want to remove those parts along with standard connecting characters.  This can leave excess spaces, commas and other things.

cl/assets/templates/includes/citation.html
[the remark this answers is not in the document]
mlissner: Yes. It's clutter without much utility.

cl/opinion_page/templates/docket_tabs.html
mlissner: We definitely don't need this as a global setting, since it's not something that I expect will be used in many places. It'd be much better to put it at the top of this file or even just in this function until it's proved that it's needed somewhere else.

cl/custom_filters/templatetags/extras.py
mlissner: Let's not over-complicate it. When/if we want to build that feature, we'll think of the best way to do so. For now, we dictate the format, so we can put it in the code.

cl/custom_filters/templatetags/extras.py
mlissner: OK, so sounds like we can remove/reduce this if we simplify as discussed above, right?

cl/custom_filters/templatetags/extras.py
troglodite2: I disagree that this is overly complicated.  I will refactor to code the format into the python.

mlissner (changes requested):
Home stretch. A few more comments and we'll have it.

Thank you!

cl/custom_filters/templatetags/extras.py
```diff
@@ -148,3 +151,36 @@ def sort_caret(request, value) -> SafeString:
...
+    if date_of_interest:
+        date_of_interest = date_of_interest.strftime("%b %d, %Y")
+    result = f"{docket.case_name}, {docket.docket_number}, ("
+    result = result + docket.court.citation_string
+    if date_of_interest:
+        result = f"{result} {date_of_interest}"
+    result = f"{result})"
+    if ecf:
+        result = f"{result} ECF No. {ecf}"
+    return result
```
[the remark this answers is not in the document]
troglodite2: The joke is that RegEx's are write only. I actually broke up the original regex's to be easier to understand.

cl/assets/templates/includes/citation.html
```diff
@@ -0,0 +1,6 @@
+{% load extras %}
+<p class="bottom" >
+  <span class="meta-data-header" >Citation:</span>
+  <span id="citation" href="" class="meta-data-value">{% citation obj fmt %}</span>
+  <i class="fa fa-copy" onClick="copy_element_text('citation');" title="Copy citation"></i>
```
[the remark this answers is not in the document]
troglodite2: So this will select but not copy.  Ok.  I think that looks nicer but is not an UX that I expect.

cl/assets/static-global/js/base.js
mlissner, line 1: We don't need this anymore, right?

cl/custom_filters/templatetags/extras.py
```diff
@@ -1,4 +1,5 @@
 import random
+import re
```
mlissner, line 2: Since I'm doing one more look through here, we don't need this import anymore either, right?

cl/opinion_page/templates/docket_tabs.html
mlissner, line 187: These don't need `fmt` anymore.

mlissner (changes requested):
Back at you with a few more things. I think you should have caught these.

cl/custom_filters/templatetags/extras.py
troglodite2: Looks like you did the remove

mlissner: I don't know what you're looking at, but it's still there.

cl/assets/static-global/js/base.js
troglodite2: I was trying to figure out what wasn't needed anymore.  You are correct, we can pull the function from base.js.  I did that.
Unfortunately what I saw was a request to pull "import re" which I had already done.

mlissner:
32 messages later, and we got this one. Great. Merging.

mlissner (approved):

PR #2654, opened 2023-04-14
feat(opinions): Add logic to redirect requests to the attachment page.

ERosendo:
This PR checks if the document of a docket entry got converted into an attachment before raising the `ResponseNotFound` exception. The new logic redirects users to the **attachment page** for cases like the one described [here](https://github.com/freelawproject/bigcases2/issues/193#issuecomment-1504157836).

This PR fixes https://github.com/freelawproject/bigcases2/issues/193.


cl/opinion_page/views.py
```diff
@@ -471,6 +471,34 @@ def view_recap_document(
                     redirect_to_pacer_modal = True
 
     except IndexError:
+        # Avoid checking if a document got converted when the user tries
+        # to load an attachment
```
mlissner, line 475: We need a better comment here since this is so weird. Something like:

> Unable to find the docket entry the normal way. In appellate courts, this can be because the main document was converted to an attachment, leaving no main document behind. See:
> 
>     https://github.com/freelawproject/courtlistener/pull/2413
> 
> When this happens, try redirecting to the first attachment for the entry, if it exists.

cl/opinion_page/views.py
```diff
@@ -471,6 +471,34 @@ def view_recap_document(
...
+                kwargs={
+                    "docket_id": docket_id,
+                    "doc_num": doc_num,
+                    "att_num": 1,
+                    "slug": slug,
+                },
+            )
+            if request.GET.urlencode():
+                attachment_page += f"?{request.GET.urlencode()}"
+            return HttpResponseRedirect(attachment_page)
```
mlissner, line 500: You'll need the code above starting on line 459 here too, right? Maybe factor that out so it can go here as well?

mlissner (changes requested):
A couple tweaks, please! Thank you.

cl/opinion_page/fixtures/appellate_attachment_doc.json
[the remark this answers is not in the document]
ERosendo: sure!

cl/opinion_page/views.py
ERosendo: You're right. This comment is a lot better.

mlissner (approved):
Looks good, thanks.

Issue #2648, opened 2023-04-13
Docket Alert Sort order on last hit puts "Never" as newest


troglodite2:
The recently added code did not account for never correctly.  It puts the "Never" as more recent than any given date.
This can be fixed by putting using the correct `.desc(nulls_last=True)` or `.desc(nulls_first=True)` on the queryset.

mlissner:
Nice one. Did you do a PR for this too?

troglodite2:
Not yet

PR #2663, opened 2023-04-20
Put "never" as the oldest item in a docket alert list

troglodite2:
NULLs are sorted in magic ways by RDBMS.  Often you get the correct answer.  When sorting by date it is more difficult, does a NULL date mean that it is so new that it doesn't have a date or so old that it doesn't have a date?

Django added a modifier to the `desc()` and `asc()` methods of an `F` filter.

We updated the code to use an `F()` filter so we can explicitly say where we want NULLS to sort in the Docket Alert List.

Fixes: #2648


mlissner:
Sure, simple enough. Thanks.

PR #2621, opened 2023-04-04
Lexis citations merger 

quevon24:
The code have comments to understand what it does each part.

How to run the command:
`docker-compose -f docker/courtlistener/docker-compose.yml exec cl-django python manage.py lexis_citation_merger --csv /opt/courtlistener/cl/citations/management/commands/lexis_citation_merger.csv `

or

`docker-compose -f docker/courtlistener/docker-compose.yml exec cl-django python manage.py lexis_citation_merger --csv /opt/courtlistener/cl/citations/management/commands/lexis_citation_merger.csv --debug`

PR #2664, opened 2023-04-20
Model for stub cases

quevon24:
This is the model to store the stub cases, it's very simple, all fields are optional because we don't always have the same information (like the westlaw and lexis datasets)

This also contains a very simple tests for object creation, i added the citations string in a specific format because i'll be preprocessing the citations in westlaw and lexis citation merger to use a semicolon to split the citations.

These changes need to be merged in order to update #2621 and #2622


quevon24:
@mlissner What do you think is the best name for the model?

@flooie has already given me several suggestions for the model's name

1. StubOpinion
2. OpinionStub 
3. OpinionMetadata
4. StubOpinion
5. PartialOpinion

mlissner:
I don't feel like "Stub" is the best word for it, but I can't think of much better. I think my mind has been corrupted by calling it a stub for so many years. I guess if I had to do it, I'd go with `OpinionStub`.

quevon24:
I think that at the moment we no longer need this new model since all the processing of the citation datasets was done locally, we already know which citations to add to which cluster and we already have the command to do it.

Should we close this PR? @mlissner

mlissner:
Yeah, I think that's fine. Let's close it, and we can re-open if we need to. Thanks for revisiting this.

Issue #2662, opened 2023-04-20
manage.py make_dev_data aborts with duplicate key


troglodite2:
By default, certain objects make other objects. In this case, make_dev_data --make-objects 104 creates docket entries.  We want lots of docket entries.  Each docket entry creates a fake court.

The CourtFactory creates a fake position via Faker.  Since this is a random number, there is a chance of a collision.  The position field is unique in the database.  When the random numbers happen to match up, the command fails.

mlissner:
I've run into this too. The good news is if you just re-run things, it'll probably be fine, but it'd be nice to fix it, for sure.

troglodite2:
I'm looking for the "right way" to fix this.  It will be a bit, I just didn't want to lose it.

troglodite2:
It looks like there are two possible ways to move forward with this.  The first is to not use the `Subfactory()` factory and instead use the `Iterator()` factory.  `Iterator()` will take a list and select from that list in order.  So we go from
```
class DocketParentMixin(DjangoModelFactory):
    docket = SubFactory(
        "cl.search.factories.DocketFactory",
        # Set the case names on the docket to the ones on this object
        # if it has them. Else generate the case name values.
        case_name=LazyAttribute(
            lambda self: getattr(
                self.factory_parent,
                "case_name",
                str(Faker("case_name")),
            )
        ),
```
To: 
```
class DocketReuseParentMixin(DjangoModelFactory):
    docket = Iterator(Docket.objects.all())
```
The other option, which I've not explored yet, is to somehow pass in a docket id.  Then we can use the `django_get_or_create` metadata so that we only create a docket when it doesn't already exist and all the DocketEntries being created are attached to the named docket.

PR #2665, opened 2023-04-21
2662 make dev data aborts

troglodite2:
When adding DocketEntries the Court Factory will sometime abort

The Court Factory randomly creates a `position`.  There is a small
chance of a collision with existing courts.  Not a serious issue when
only adding a "few" courts.

When adding DocketEntry's it is not uncommon to add thousands of
DocketEntry's.  This increases the chances of a collision.

This adds two new factories for creating DocketEntry's.  The first will
create and the new objects to existing Docket's.  The second allows for
creating DocketEntry's for a particular docket.

Fixes: #2662


mlissner:
@albertisfu can you take a look at this and see if it solves the problem in #2662 in a good way? My first glance is that it feels more complicated than we need and that maybe all we need to do is UUID's or something *less* random for the Court IDs when we create them. Mind taking a look?

troglodite2:
In looking at the request for review, it becomes obvious to me that I didn't "fix" the issue, but instead provided a workaround.

There are three different use cases at play here:
To make Docket Entries that create a new Docket for each entry.  This is the current object 104 method
To make Docket Entries that are attached to a particular Docket.  This is something I needed in order to test pagination.
To make Docket Entries that are attached to pre-existing Dockets.  This could have replaced 104.

albertisfu (commented):
@mlissner Yeah, regarding the original issue of courts colliding due to the same position value, as you mentioned, it might be easily solved using a kind of `UUID` so that the chance of collisions is too low.

For instance we could use a new Faker provider like:

```
def random_unique_number(self) -> int:
        """Generate a random number with a very small chance of repetition.

        :return: A big random number with small chance of repetition.
        """
        return secrets.randbits(64)
```
An using it for the `Court` `position`:
`position = Faker("random_unique_number")`

So the risk of collisions is practically zero.

About the new @troglodite2 `107 DocketEntryForDocketFactory` also seems quite useful to me when you need to add multiple entries for a specific docket.

`108: DocketEntryReuseParentsFactory`, I think this might not be needed if we apply the `uuid `approach to solve the original issue.


On the other hand this is another thing I found while reviewing this, but it is not related to this PR. However, it seems to be an issue.

When creating `DocketEntryWithParentsFactory`, the related docket that is created gets a bad name like `factory.faker.Faker object at 0xffff7071bf40`. I think the problem is that the child object, in this case DocketEntry, doesn't generate a `case_name`. Generating an object where the case name is generated in the child object like `AudioWithParentsFactory` works well.

Screenshot 2023-04-24 at 13 50 41

troglodite2:
I dislike using "big random" to avoid a collision.  Maybe the better option is to do a try/except for the actual creation of the Court object  and if it fails on a not unique, try again with a different position?

troglodite2:
Now that Alberto has pointed out the issue with case_names I'll go look to see if I can fix it in this PR.

As for the UUID, the issue is that what is conflicting is the "position" field.  From the help field: A dewey-decimal-style numeral indicating a hierarchical ordering of jurisdictions

The CourtFactory creates this with a `Faker("pyfloat", positive=True, right_digits=4, left_digits=3)`

To use UUID for position would require a change in scheme, and we will lose the hierarchical aspect. 

The try would happen in DocketFactory

mlissner:
Oh, right. I was thinking it's the `id` field. Am I crazy, or does this mean it's a one-in-a-million bug: `right_digits=4, left_digits=3`?

troglodite2:
If it were truly random, yeah, 1 in a million.  It is not truly random as
we get hits way more often than 1/1,000,000

mlissner:
OK, then I'm back to this:

> If the try/except were focused in a single place (like in the faker or the factory), that seems fine. If it would have to be done whenever we create courts, that'd be a bummer.

Otherwise, Alberto's solution to make this really, really unlikely seems fine.

albertisfu:
Yeah, I forgot to mention that in the example I put it uses [secrets](https://docs.python.org/3/library/secrets.html)  
`import secrets`  so the randomness is more reliable.

troglodite2:
This is going to fail the selenium tests until we get the other PR merged.

cl/search/factories.py
```diff
@@ -45,6 +47,23 @@ class Meta:
...
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
+        count = 1
+        while True:
+            try:
+                obj = model_class(*args, **kwargs)
+                obj.save()
+                return obj
+            except IntegrityError as exp:
+                print(f"Unexpected {exp=}, {type(exp)=}")
```
albertisfu, line 59: Yeah, this method solves the issue I would just recommend using `logger.info() ` instead of `print()` for logging the error.

albertisfu (commented):
@mlissner the new @troglodite2 approach solves the issue using a `try/except` when the collision happens, seems good to me, just a small suggestion about logging.

Also, the docket `case_name` for `DocketEntryWithParentsFactory` is solved.

mlissner:
Cool. Set for auto-merge. Thanks for rebasing.

Issue #2683, opened 2023-04-25
Actual citations are not saved to bulk data


mlissner:
A user has reported that we're not exporting the search_citations table. They're right; we should.

PR #2684, opened 2023-04-25
fix(bulk): Add citations export

mlissner:
Fixes: #2683

I'll need to deploy this to the cron job in k8s when it's ready.


albertisfu (commented):
This seems good to me!

I've rebased it and set it auto-merge.

Issue #2670, opened 2023-04-24
Feature request: When looking at docket pages it would be nice to have a next/prev button at the top


troglodite2:
Currently, the filter header looks like this:
![image](https://user-images.githubusercontent.com/82225842/234108123-19bd0e9c-4ffb-4d5b-8ff6-cf60ea6bc83d.png)

Adding a next/previous in the header should be doable.

mlissner:
Want to try a mockup/prototype with the smallest of buttons you can cram into the far left and right?

troglodite2:
![image](https://user-images.githubusercontent.com/82225842/234383914-6a3ed719-1e09-42c0-8bba-0c6d1a6c5ef3.png)

This is the full size mock up.  If this is what you are looking for, then I'll go a head and code the responsive versions as well.

mlissner:
That's not bad. I was actually thinking of something simpler. Could you mock up just the arrows themselves on the far left and far right, maybe with a mouseover that says Previous/next? That's what's in my mind's eye, and I think it'd be pretty elegant.

troglodite2:
I heard that you are looking for something different.  I want to present a couple of different options:
Option 1:
![image](https://user-images.githubusercontent.com/82225842/234657673-faf331d4-9872-4270-9aaf-3b843e1874cd.png)
Prev. at the front, next at the end of the line.  The ascend/descend buttons changed so that only the clickable version shows.

Option 2: 
The same layout but now only the next/previous arrows showing.  I've also changed the order to show the button changing
![image](https://user-images.githubusercontent.com/82225842/234658346-d8bf9733-7875-4268-b6b7-806d5a90ea3c.png)

Option 3:
Previous/next arrows at the end of the bar, still with the ascend/descend showing only one button.
![image](https://user-images.githubusercontent.com/82225842/234659404-f7a073fa-446b-4d5b-8512-928878194368.png)
We can tweak the placement of the arrows in there as well.

Option 4:
Previous/Next plus arrows at the end of the filters.
![image](https://user-images.githubusercontent.com/82225842/234659695-e74ce35e-6007-42ac-85d7-a131f0260457.png)

Option 5-8: 
The same as above, but going back to showing the redundant sort order button.

If there is something else you would like to suggest, let me know.

mlissner:
[...]

PR #2687, opened 2023-04-27
Add previous/next navigation buttons to the filter bar.

troglodite2:
Currently dockets have upto 200 entries on them.  We have navigation buttons at the very bottom that allow navigation.  Many people use scrolling to move through a page and a 200 item docket can be time consuming.

We add Previous and Next buttons to the filter bar to allow people to navigate more easily when at the top of the page.

During testing it became apparent that the Bootstrap XS version was pretty ugly, per CL design specifications.  Code was added to override.css to make things look better.

When testing at a screen width of 360px it again looked horrible. Having things look right at XS width and at 360 wasn't possible.

To allow for another breakpoint a series of classes were added to override.css that provide cl-col-xxs- and cl-col-xs- as well as some auxilary classes.  These allow for that extra breakpoint.

the cl-col-xs- are required because the col-xs- classes are always "on" and we needed a method to override them.

We also added some specific cl-pull classes to allow different alignment based on screen size.

Fixes: #2670


mlissner:
What happened to the boostrap-sm that you had here: https://github.com/freelawproject/courtlistener/issues/2670#issuecomment-1524120128

troglodite2:
I can try to get back to it.  I think going from small to large, I kept the wrong look.

troglodite2:
Per discussions, we are designing this for standard bootstrap break points.  The 360px is what it is.
360px:
image
Bootstrap xs: 767px:
image
Bootstrap sm: 991px
image
Bootstrap md: 1199px
image
Bootstrap lg and desktop: 1500px
image

Once these images are approved, I'll push up the next commit.

mlissner:
These look good, but the screenshots aren't the most difficult widths. 767, 991, and 1199 are the best cases for xs, sm, and md sizes. The harder question is one pixel bigger: The smallest size before the breakpoint kicks in. Can you share 768, 992, and 1200? 

My general feedback though is we should try to stick to two lines max, except on xs, and should drop pagination on the smaller sizes if we need to, to make that happen.

troglodite2:
768px:
image
992px:
image
1200px:
image

mlissner:
Cool. Those look great. Do what you can for xs, and let's ship.

cl/assets/static-global/css/override.css
```diff
@@ -1428,3 +1436,21 @@ textarea {
...
+  margin-bottom: 5px;
+  display: inline-block;
+  vertical-align: middle;
+}
+@media (max-width: 768px) {
+  .cl-xs-pull-left {
+    float: left !important;
+  }
+  .cl-xs-pull-right {
+    float: right !important;
```
mlissner, line 1454: I don't think we need to namespace these, and it seems like the convention in BS4 and BS5 is to call them `pull-right-xs` and and `pull-left-xs`.

mlissner (changes requested):
A couple little CSS things, but we're on the homestretch.

mlissner (approved):

Issue #2689, opened 2023-04-28
Docket Alert Sort order still sorts date_last_hit incorrectly


troglodite2:
Clicking on the sort buttons for alerts does something, but it is not consistent.  In addition, "Never" (NULL/None) shows up at the wrong end of the list.

PR #2690, opened 2023-04-28
2689 fix alert sort ordering

troglodite2:
Fix the logic for Docket Alert Sort Order
    
It was wrong, it is better now.
    
Fixes: #2689


mlissner:
This has a bunch of changes from another PR. Please rebase or cleanup somehow and let me know when it's ready by requesting a review.

PR #2700, opened 2023-04-29
Add {{ get_string }} to docket page navigation within filter bar

troglodite2:
The docket page navigation buttons within the filter bar loses any parameters that are on the current page.  That means that it is possible to loose sort order and other parameters.


mlissner:
Thanks for catching this!

Issue #2730, opened 2023-05-05
Dockerfile has misplaced ARG statement


troglodite2:
Build arguments exist only within a build sequence.  Terms might be incorrect here.

This means that the ARG BUILD_ENV=prod only exists until the next FROM statement.  Because of this, the build argument is not available further than the 2nd line of the Dockerfile.

PR #2736, opened 2023-05-05
Fix misplaced ARG commands in DockerFile

troglodite2:
Build arguments only exist until the next FROM command.

Fixes: #2730

### DIFF
--- a/docker/django/Dockerfile
+++ b/docker/django/Dockerfile
@@ -1,6 +1,6 @@
-ARG BUILD_ENV=prod
-
 FROM python:3.10-slim as build-base
+
+ARG BUILD_ENV=prod
 
 RUN apt-get update --option "Acquire::Retries=3" --quiet=2 && \
     apt-get install \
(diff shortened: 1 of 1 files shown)


mlissner:
Can you explain what problem this fixes and how it possibly worked (it did, right?) since according to what you're saying, line 77 shouldn't have worked?

troglodite2:
I added a RUN echo to the Dockerfile
image

Here are the results
image

Here I've added ARG in front of the RUN and FROM lines
image
Here are the results:
image

I am not a docker expert.  I discovered this because I was attempting to use BUILD_ENV on a RUN statement.  You can see that what happens in that case is that the variable is empty.

The question is why was it working on that particular FROM line.  My best guess is that build arguments are always available on certain lines, but it is only the ARG command that brings them into the build environment.

Another way to fix this might be to remove the ARG BUILD_ENV completely, but I think that would just be confusing.  It is self documenting.

mlissner:
Weirdly, your test is failing in a way that indicates your changes made this worse?

troglodite2:
In the comment above, I was working on `main`, not my fixes. I am showing you what the current condition is.

I inserted the RUN above line 77 (only change) to show the value of BUILD_ENV.  I then added an ARG line above that to show that inserting the ARG resolves the issue.

mlissner:
OK, but your test is failing with:

> ERROR: failed to solve: failed to parse stage name "build-": invalid reference format

mlissner:
Fixed via #3130

Issue #2735, opened 2023-05-05
Revamp Citation Lookup


troglodite2:
Two issues, one is that in the Opinions drop down the text is "look up" and on the actual page it is "lookup tool"
![image](https://user-images.githubusercontent.com/82225842/236576368-afc309c8-a039-41b2-9c44-273b98353749.png)
![image](https://user-images.githubusercontent.com/82225842/236576332-d63e3eb1-3ed1-4206-8b0b-592a8121cef1.png)

The current Citation Lookup tool requires the user to parse a citation and input volume, reporter, and page.  We have the tools to do for the user using eyecite.

Change from using 3 input fields on the Citation Lookup home page to just one field.
![image](https://user-images.githubusercontent.com/82225842/236576597-29f536d7-6ccb-4bfc-96dc-182b1f19a6ee.png)
Becomes:
![image](https://user-images.githubusercontent.com/82225842/236576794-9d17ccb1-7305-47c1-8ef7-161ca229fb54.png)

![image](https://user-images.githubusercontent.com/82225842/236576879-6c036b79-2115-4a3b-89d4-cb82a9f103d0.png)

360px:
![image](https://user-images.githubusercontent.com/82225842/236577017-bb2a227f-ab75-465e-997a-05bf22ad0818.png)

767px:
![image](https://user-images.githubusercontent.com/82225842/236577102-04a5c504-cace-44b6-b001-ecef23ea5c5a.png)

768px:
![image](https://user-images.githubusercontent.com/82225842/236577179-33cc784f-4872-4668-bb0c-066a2200739a.png)

mlissner:
Looks like a nice improvement. I think bootstrap makes it pretty easy to do a combined lookup button. Want to try something like on the homepage? 

![image](https://user-images.githubusercontent.com/236970/236581635-d82532c2-543a-4a6b-b2f6-27dace52ee3f.png)

It'd be nice if it weren't full width as well, kind of like the homepage, since full width just looks sort of out of control.

troglodite2:
Desktop:
![image](https://user-images.githubusercontent.com/82225842/236646635-3c053af6-89c4-4abe-88af-51298493687d.png)

Mini:
![image](https://user-images.githubusercontent.com/82225842/236646664-4a5fe150-ee92-4651-9c38-be9416412606.png)

xs:
![image](https://user-images.githubusercontent.com/82225842/236646676-8b23b4e5-6018-4f0e-b950-0e9f05211b56.png)

[...]

PR #2737, opened 2023-05-07
2735 citation lookup revamp

troglodite2:
The citation lookup tool was designed to easily convert reporter,
volume, page into a citation URL which then redirects to the correct
item, if found.

We have eyecite, which can parse a text and extract a citation.  The
code was revamped to provide a single field which is capable of taking
a citation.

EyeCite then parses that and turns it into a reporter, volume, page and
redirects to the correct item, if found.

This appears to be much cleaner, it allows for cut-and-paste options,
and in general looks like it is a better user experience.

All the `/c/` lookup URLs still work the same way, and the Citation
Lookup Tool home page still has a list of reporters that can be browsed.

Fixes: #2735

### DIFF
--- a/cl/opinion_page/forms.py
+++ b/cl/opinion_page/forms.py
@@ -41,7 +41,7 @@ class CitationRedirectorForm(forms.Form):
         widget=forms.TextInput(
             attrs={
                 "class": "form-control input-lg",
-                "placeholder": "Full Citation",
+                "placeholder": "Paste any text containing a citation",
             }
         ),
         required=True,
--- a/cl/opinion_page/forms.py
+++ b/cl/opinion_page/forms.py
@@ -33,13 +33,15 @@ class CitationRedirectorForm(forms.Form):
         widget=forms.TextInput(
             attrs={"class": "form-control input-lg", "placeholder": "Volume"}
         ),
-        required=True,
+        required=False,
     )
+    # We change the place holder to allow people to continue to use the rest
+    # of the redirect API with no modifications.
     reporter = forms.CharField(
         widget=forms.TextInput(
             attrs={
                 "class": "form-control input-lg",
-                "placeholder": "Reporter",
+                "placeholder": "Full Citation",
             }
         ),
         required=True,
@@ -48,7 +50,7 @@ class CitationRedirectorForm(forms.Form):
         widget=forms.TextInput(
             attrs={"class": "form-control input-lg", "placeholder": "Page"}
         ),
-        required=True,
+        required=False,
     )
 
 
--- a/cl/opinion_page/views.py
+++ b/cl/opinion_page/views.py
@@ -4,6 +4,7 @@ from itertools import groupby
 from typing import Dict, Tuple, Union
 from urllib.parse import urlencode
 
+import eyecite
 import natsort
 from django.contrib import messages
 from django.contrib.auth.models import AnonymousUser, User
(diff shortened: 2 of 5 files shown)


cl/opinion_page/forms.py
```diff
@@ -33,13 +33,15 @@ class CitationRedirectorForm(forms.Form):
         widget=forms.TextInput(
             attrs={"class": "form-control input-lg", "placeholder": "Volume"}
         ),
-        required=True,
+        required=False,
     )
+    # We change the place holder to allow people to continue to use the rest
+    # of the redirect API with no modifications.
```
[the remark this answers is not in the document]
troglodite2: I tried that first.

One option, make a new form that just had full_citation in it.  Then make decisions based on `is_valid()` as to which form to use.  In addition, it adds some complication to the `reverse()` functions, making sure that the correct URL pattern is picked.

Another option is to add a full_citation field to the current form.  This would require all the fields to be `required=False`.  This gets confusing and might have other UI interactions.

Having tried three different methods of accomplishing this particular input processing, this produced the cleanest code.

I use a style of programming which I call "Railroad siding".  In this style, there is always a main line of progression.  Anything that moves us away from that main line moves us onto a siding.  The goal is to do as little modification as possible and then rejoin the main line.

Thus, this code checks to see if there is a full citation.  If there is a full citation, then it sets the reporter, volume, and page and rejoins the main line.  Nothing after that one check and set has major changes.

I would suggest that we keep my method this time.  If you don't like it, we'll decide on which method to use, and I'll code that.

cl/opinion_page/views.py
```diff
@@ -1087,8 +1111,13 @@ def citation_homepage(request: HttpRequest) -> HttpResponse:
         if form.is_valid():
             # Redirect to the page as a GET instead of a POST
             cd = form.cleaned_data
+            kwargs = {"reporter": cd["reporter"]}
+            if cd["volume"]:
+                kwargs["volume"] = cd["volume"]
+            if cd["page"]:
+                kwargs["page"] = cd["page"]
```
mlissner, line 1118: Do you need this? I'm just imagining here, but if you just pass through `cd` like before, aren't volume and page None, and doesn't that work fine?

cl/opinion_page/forms.py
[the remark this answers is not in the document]
mlissner: OK.

cl/opinion_page/forms.py
```diff
@@ -33,13 +33,15 @@ class CitationRedirectorForm(forms.Form):
...
     reporter = forms.CharField(
         widget=forms.TextInput(
             attrs={
                 "class": "form-control input-lg",
-                "placeholder": "Reporter",
+                "placeholder": "Full Citation",
```
mlissner, line 44: I think we can do better here. How about:

```suggestion
                "placeholder": "Paste any text containing a citation…",
```

mlissner (changes requested):
A few small comments. Thanks for including a test!

cl/opinion_page/views.py
troglodite2: I originally tried it with just passing in 'cd' but that caused the reverse() function to blow up because it didn't find a match in urls.py that it liked.

mlissner: Got it, thanks.

mlissner (approved):

PR #2741, opened 2023-05-08
Improve opinion model

flooie:
This PR adds three new fields to the opinion cluster. 

1) HeadMatter Footnotes

We add a field for headmatter footnotes - a field we take from harvard for all of the footnotes grouped before the opinion.  

Unfortunately these footnotes can not be reversed back to their original content in some cases, or are floating or duplicates or mistakes in the books. 

2) Book Images

This can take some unexpected but important context from the books.  This can take the form of a written note - for example (Judge X Recused) in handwriting.  Or dates of the opinion.  There is not slot for these otherwise. 

https://cite.case.law/pdf/8500715/Jordan%20v.%20Touro%20Infirmary%20&%20Hebrew%20Benevolent%20Ass'n,%207%20Pelt.%20199%20(1922).pdf

3) Arguments

Arguments are the legal cases used or the arguments pushed during oral arguments and/or summarized by the court about the positions taken in writing before appellate hearings.  These relate to attorneys, and will need to include the attorney information to retain the context of who is presenting an opinion.


mlissner:
Can you tune up your migrations, please? You'll want to rebase first too.

Issue #2745, opened 2023-05-08
Upgrade replication server to postgres 14


mlissner:
This might not be too hard, but it needs to happen in order to restore our healthcheck endpoint. 

Right now our health check endpoint is not checking the health of postgresql because Django 4.2 was crashing whenever we did. 

The thing that makes me most nervous about this is that we have a lot of replicas that this fans out to, so if those break or otherwise fail, it'll be really sad. 

Probably the trick is to stop replicating to this server, upgrade it, and then restore replication. That'll be a pain, but it probably won't be too terrible.

mlissner:
OK, so it looks like for many databases, this would be really easy, but for ours, since we have logical replication set up, it's going to be bad. The docs say:

> An upgrade can't occur if the instance has any logical replication slots. Before upgrading, make sure that you know the purpose of any logical replication slots that are in use, and confirm that it's okay to delete them. 
> If the logical replication slots are still being used, you shouldn't delete them, and you can't proceed with the upgrade. 

I don't know why that would be. I'd expect that if I paused a replication connection, did the upgrade, and then resumed it, that would work fine. 

If we really need to stop all replication, that's going to be hard since this server sends replication to a number of client servers. 

There's an outside chance that I'm confusing replication *slots* with replication *subscriptions*. Maybe when you pause a subscription, it removes the slot? I'm not sure, but this looks pretty gloomy.

---

Making matters worse, an email from AWS that I've been trying to pay more attention to says:

> RDS will automatically upgrade your PostgreSQL 10 databases to PostgreSQL 14 during a scheduled maintenance window between April 17, 2023 00:00:01 UTC and July 18, 2023 00:00:01 UTC. After July 18, 2023 00:00:01 AM UTC, any PostgreSQL 10 databases that remain will be automatically upgraded to version 14 regardless of your scheduled maintenance window

So I guess there's a chance that AWS could upgrade this DB at any moment, and if it does, that that will break all our client replication systems. Great.

mlissner:
I'm doing a lot more research on this today and I'm feeling a bit more positive about it. First, this blog post gets close to the problem that we have and goes into some detail about how they worked around it: 

https://elephanttamer.net/?p=58

[...]

Issue #1109, opened 2019-12-31
DB Migration and Replication Post-Mortem (2019-12-27 to 2019-12-30)


mlissner:
This issue notes the various causes, solutions, and issues that occurred during the database migration begun on 2019-12-27 and mostly completed on 2019-12-30. This migration, while successful, encountered numerous problems, which had to be solved on the fly. 

# Architecture and goals analysis

The CourtListener replication system uses the PostgreSQL system to replicate data from our servers to those of our clients. The current architecture has five components:

## The servers

 - The `master` server. This is our core database that serves all SQL requests and which serves as the ground truth. This is an on-premise server that is incredibly powerful.

 - The `old-master` server. This is our previous master server/database. It subscribes to the `master` publisher, and then publishes changes to our `AWS-server`. This server should be removed from the architecture, but we have not figured out how to do so (more below). This is an on-premise server with medium horsepower.

 - The `AWS-server``. This server is smaller and is only used to publish changes to our customer servers. It serves no SQL and uses AWS RDS.

 - Our `client-servers`. These servers are in AWS or Google Cloud and are controlled by our clients. They are usually about the same size and power as our `AWS-server`, to which they subscribe, but clients can scale them up and down as needed for their use case.

The data and schema of these servers is all kept in sync at all times.

## The `proxy`

Because AWS RDS does not provide static externally-accessible IP addresses on RDS instances, we need a `proxy` that does (see #949). This server sits between our AWS server and our clients, so that their connections to replication can use a domain name instead of an IP address. Failure to do this causes headaches when the IP of our `AWS-server` changes, which it does from time to time (like when you scale it).

This `proxy` is only used by half of our clients because not all have been upgraded to use it at this time. In general, there's no need to think about the proxy server — it just provides a domain name and doesn't do much else.

## Data flow

When a data changes comes in, it flows thus:

    master --> old master --> our AWS --> (proxy server sometimes) --> client servers

# The plan

There were two goals of this plan:

1. Add several new fields to the `search_opinion` and `search_opinioncluster` tables. These would add support for the harvard database.

2. Tweak three text fields so that they are no longer nullable, per Django's convention. (Text fields in Django are never NULL. If you lack a value, they're always simply blank, `''`.)

To achieve this, the process is generally to use Django's built in migration system to migrate the `master` server, then use the SQL it generates to manually migrate all the other servers (`old-master`, `AWS-server`, and `client-servers`). 

# What went wrong?

Lots. 

[...]

Issue #1106, opened 2019-12-28
Stop using Django's default DB migration system


mlissner:
Sigh, this is a sad one, but it looks like Django's DB migration system is just...not good enough. Currently we use it on our master DB and then use SQL on all our replicas, so this ticket is about changing the process to using SQL everywhere. 

Some of the problems:

1. When adding text columns (which per convention are not nullable), Django's migrations use `ALTER` statements that are just...brutally bad. Like, everybody knows they're bad and yet it uses them anyway.

    Here's an example from a recent migration:

        ALTER TABLE "search_opinion" ADD COLUMN "joined_by_str" text DEFAULT '' NOT NULL;
        ALTER TABLE "search_opinion" ALTER COLUMN "joined_by_str" DROP DEFAULT;

    This migration adds a new column, sets the default value to blank, and makes it not nullable. Then in a second `ALTER`, it removes the default value. The result is that you have a column full of blanks. This is a fine outcome in Django-land, but it's terrible from a performance perspective because in Postgres < 11 (the very latest version), to create a column with a default value, you have to copy the entire table. Woo, let's hope we have disk space for that on a large table! Here's what the [postgres docs][docs] say:

    > Adding a column with a DEFAULT clause or changing the type of an existing column will require the entire table and its indexes to be rewritten.

    A better approach, that I believe has the same outcome if performed in a transaction is:

    ```
    ALTER TABLE "search_opinion" ADD COLUMN "joined_by_str" text;
    UPDATE "search_opinion" set "joined_by_str" = '' WHERE "joined_by_str" IS NULL;
    ALTER TABLE "search_opinion" ALTER COLUMN "joined_by_str" SET NOT NULL;
    ```

    I believe that would create a new column (instantly), set all the values to blank (slowly, but without locking or copying the table), update the column to not allow NULLs anymore, as above.

    The [django docs](https://docs.djangoproject.com/en/2.2/topics/migrations/#postgresql) also mention that:

    > The only caveat is that prior to PostgreSQL 11, adding columns with default values causes a full rewrite of the table, for a time proportional to its size. For this reason, it’s recommended you always create new columns with null=True, as this way they will be added immediately.

    So that's an option too, if you remember. I haven't figured out a way to automatically prevent this though. Maybe a git hook?

[...]

PR #2746, opened 2023-05-09
feat(search.models.py): Update opinion cluster model

flooie:
Add Arguments, HeadMatter Footnotes and Book Images


flooie:
This PR adds three new fields to the opinion cluster.

HeadMatter Footnotes
We add a field for headmatter footnotes - a field we take from harvard for all of the footnotes grouped before the opinion.

Unfortunately these footnotes can not be reversed back to their original content in some cases, or are floating or duplicates or mistakes in the books.

Book Images
This can take some unexpected but important context from the books. This can take the form of a written note - for example (Judge X Recused) in handwriting. Or dates of the opinion. There is not slot for these otherwise.

https://cite.case.law/pdf/8500715/Jordan%20v.%20Touro%20Infirmary%20&%20Hebrew%20Benevolent%20Ass'n,%207%20Pelt.%20199%20(1922).pdf

Arguments
Arguments are the legal cases used or the arguments pushed during oral arguments and/or summarized by the court about the positions taken in writing before appellate hearings. These relate to attorneys, and will need to include the attorney information to retain the context of who is presenting an opinion.

flooie:
An example of an image - worth keeping

Screenshot 2023-05-08 at 4 43 36 PM

cl/search/models.py
```diff
@@ -2389,6 +2389,22 @@ class OpinionCluster(AbstractDateTimeModel):
...
+    head_matter_footnotes = models.TextField(
+        help_text="Footnotes that appear before an opinion or at the conclusion "
+        "but that are not part of the opinion",
+        blank=True,
+    )
+    book_images = models.TextField(
+        help_text="Book images containing cluster content "
+        "(for example handwritten recusals) as HTML base64 images",
+        blank=True,
+    )
```
mlissner, line 2407: I'm quite hesitant to put images in the DB. We've never done so before and it's usually an anti-pattern. Any reason not to do our usual S3 approach here? 

Second question: OK, so we have an image associated with the opinion. Is it possible to have more than one? How do we know where to put that image in the opinion?

mlissner (commented):
Well, migrations can be difficult, and I'm afraid this one will be. If you look in MIGRATIONS.md, you'll see that:

> ## New columns With DEFAULT values in Postgresql < v11
> 
> The biggest issue we run into is that creating new columns with DEFAULT values,
> can be fine in regular environments, but cause a crisis in huge tables like the
> ones we have. Django does this kind of migration by default when you create a
> new text column with `blank=True`. That's very bad and until we upgrade to
> Postgresql 11 we will have to contend with this issue.
> 
> Here is some background reading on why this is a problem:
> 
> https://github.com/freelawproject/courtlistener/issues/1106
> 
> https://github.com/freelawproject/courtlistener/issues/1109

If you follow those two links, you'll see they link to further pages. I just read most of them, and honestly, it's a mess. Doing the migrations here in Postgres 10 is just going to be a small disaster, I'm afraid. 

Elsewhere, you'll see that upgrading postgres to version 11 has become a high priority: 

https://github.com/freelawproject/courtlistener/issues/2745

It turns out that postgresql 10 is EOL, so AWS is going to upgrade our database for us literally at any moment, unless we beat their bots to it—And if that happens, it'll probably break lots of things, like replication.

So, given that doing this migration would be really difficult and might be a total disaster, and that we have an urgent need to upgrade to postgresql 11 (at least), I suggest we wait on this PR (somehow), until we've got the upgrade complete. Then, at that point, this migration should be quite straightforward (as much as they ever are, that is).

cl/search/models.py
flooie: yes it is possible to have more than one - but Im not sure we've had two outside the opinion.  I'll double check

mlissner:
The DB is upgraded. Is this ready for a final review and merge?

cl/search/models.py
mlissner: I'm still waiting on some clarity here:

1. Shouldn't these go in S3?
2. Might we have more than one image per opinion?
3. How does the opinion know to insert an image?
4. How does the opinion know *where* to insert an image? 

It sort of feels like maybe images should go in a joined table, but maybe I'm over thinking it?

quevon24: That field has already been removed.

mlissner:
Cool. This goes from nearly impossible (postgres 10) and complicated (images?) to easy and straightforward. Merging and deploying by hand. Here we go.

mlissner:
A few migration notes:

1. This is an additive migration, so I added the columns to our clients, then our replica server, then prod. 

1. Once we added the columns to our database with not-null constraints, replication broke because publishers were still sending data with null values (because the columns didn't exist on the publisher yet). To fix this, I made all the columns nullable, migrated the publisher, then made them not nullable again.

1. One of our clients is still on postgresql 10, meaning that rolling this out was a bit of a pain, but it's not as bad as if our server were still on pg10, because it doesn't affect us at least. I didn't catch this at first, but once I tried to add the column, their DB hung endlessly, until I did things the old way:
    - Add the column as nullable and without a default value.
    - Do an update to change the null value to an empty string.
    - Make the column not null and add a default value.

1. This is still ongoing because the customer on pg10 is also not in our cloud, meaning that I'm still waiting for replication to flush to them. Once their replication catches up, I can do the last step above. 

As ever, this was annoying, but it's the price we pay, so so be it.

PR #2762, opened 2023-05-17
Update NOT_JUDGE_WORDS list to improve judge/author name extraction

quevon24:
I found some strings where were the judge/author name was extracted but also some wrong words, for example:

| Judge/author string  | Extracted names | Incorrect data |
| ------------- | ------------- | ------------- |
| Quillin, P.J., concurs in judgment and opinion except as to footnote 1.  | Except, Quilin | Except |
| Swepston, Justice, concurs, except as to Statute of limitations.  | Except, Limitations, Statute, Swepston | Except, Limitations, Statute |
| Before Justice Garrison sitting alone. | Alone, Garrison | Alone |
| Johnson, C. J., also dissents. | Also, Johnson | Also |
| *440MORRISON, Presiding Judge, concurring on Appellant’s Motion for Rehearing. | Appellant, Morrison, Motion | Appellant, Motion |

To solve this i decided to lean on the nltk library with averaged perceptron tagger which is a pretrained english model tagger, and used it to extract all words except those tagged with "PERSON" tag. This way i narrowed down the possible incorrect words to a list, i had to inspect that list manually because sometimes the last names can also have a different meaning, for example: Stone in string "Before STONE, WOODROUGH and VCGEL, Circuit Judges." depending on the context  it can be a last name, or it can refer to a rock, the problem here is that we are using a traditional algorithm where we don't care about the context, that's why we can't add "Stone" to the list.

These new words in the list help to improve the results, you just keep in mind that this list does not solve typos, these have to be fixed in opinionated if you are importing the information from a harvard json.


mlissner:
Wow, so many words! @flooie, let's get this merged. @quevon24, should we create a new issue to apply this to opinionated?

quevon24:
> Wow, so many words! @flooie, let's get this merged. @quevon24, should we create a new issue to apply this to opinionated?

This applies when we import an opinion from harvard or when we run the harvard merger, one way to run this might be to create a management command to get clusters where author_str or judges are not empty and apply the cleanup process to update the names.

mlissner:
I don't have a sense of that anymore, so I'll duck out of the way and let you and @flooie decide the best thing to do there.

flooie:
@quevon24 can you provide the links to the examples in the table above.

flooie:
So after investigating - there are some more concerning features of this method - specifically its inability to handle names with apostrophes or hyphens.

quevon24:
> @quevon24 do we still need this? 

probably not

PR #2821, opened 2023-06-15
2499 Add Opinion Ordering

flooie:
@mlissner 

This PR - is a small change.  

Add support for [Django Ordered Model](https://github.com/django-ordered-model/django-ordered-model)
Adds ordering for opinions inside opinion cluster.  
Updates all fixtures to include a default value of 1 for order.  

Once we reprocess Columbia - this should actually really shrink the code for the harvard merger and for the HTML template code displaying opinions.


mlissner (commented):
I gave this a quick look, but I won't be able to deploy it until I get to Montreal, so a deep look will have to wait until then. 

A few quick thoughts though:

1. This library looks good, but one thing it says it does is [add default ordering](https://github.com/django-ordered-model/django-ordered-model#custom-ordering-field) to the `Meta` class. That's logical for this library, but I don't really like default ordering because it slows down queries that don't need ordering. For example, this query is rough on the poor DB on a good day:

        Opinion.objects.all()

    But after this change, it becomes:

        Opinion.objects.all().order_by('order')

    Because the `order_by` is always applied by default. Adding ordering when you don't need it, naturally, hurts performance in the DB.

    I think it's much better to have developers explicitly request ordering when they need it, rather than having to remember to disable ordering when they don't.

2. Did you check that the `order` field is properly added to any DRF responses?

flooie:
@mlissner 

Working backwards - Order does show up in DRF - filtering to fields order and id to make it easier to see

Screenshot 2023-06-16 at 11 22 48 AM

flooie:
@mlissner back to you if you can 

I tried a couple things but override the custom ordering with a custom ordering manager was (appears to be) the right move.

mlissner:
@flooie I'm looking over my outstanding tasks and I think this one is back to you. It needs a rebase, conflicts resolved, and tests to pass, and I forget what it needed wrt to tweaks to the actual code that we were discussing when I was in an airport. :)

Mind giving it a careful look and letting me know when it's ready for another review?

quevon24:
@flooie Well, after inspecting django-ordered-model I came to the conclusion that the only way to disable the default ordering is to overwrite some of the code in the library (and that is not an option because with major changes we will need to update the overridden code, I opened an issue in their repo to ask the reason to force the order, maybe we can contribute with that). You could leave the default behavior since it is not common to fetch all the objects in the frontend, the problem would be in the API, since by default it would return the opinions ordered by the ordering field.

I also found a library named django-orderable (https://github.com/incuna/django-orderable) which adds manual sort order to django objects via an abstract base class and admin classes, this library has no methods to move objects, its main goal is to be integrated into the admin and use a drag and drop interface to reorder the objects.

The best alternative is to manually add the ordering field in the opinion model (integerfield or floatfield), to do this, what we need to answer is:

1. We plan to move the objects in a simpler way using methods? Like: move_up, move_down, move_top, move_bottom, etc. Doing this will require to update the position in the previous and/or next objects if we use ints, if we use floats we can generate smaller numbers between where you want to place the object.
2. Or will we manually indicate the position in certain scripts like management commands? This will require to update the positions of previous/next objects directly in the command.
3. We plan to have the migration auto-populate the field? (it could be copy the pk as the position since opinions are generally created in order) 
4. if you use this option, would we implement all the methods within the Opinion model? Or would we create an abstract model that could be reused elsewhere?
5. This ordering only applies to the frontend when we show the cluster and its opinions?

So far these are my comments, any comment is welcome

mlissner:
Yeah, my reaction is that we're over thinking this and the various libraries don't really fit our needs. I think what we need is:

1. A way of storing the ordering information:
    - A table to join to or as a field on the model?
    - Floats or ints?
4. Some methods/functions that allow you to set the ordering for an item. 
5. Some tweaks to the front end to show items in the correct order.

I think that's mostly it. I don't think we need functions for `move_up` or `move_top`, fancy admin integration, or anything like that. I imagine when you guys are doing the import, you'll just need a setter, right? 

Notionally:

```
opinion.order_position = 1
opinion.save()
```

Right? The libraries feel either too big or full of features we don't really want, so let's take from them what we do want, and leave the rest behind. I, for one, will be glad to have one fewer library using memory and having to be upgraded and maintained.

If this all easily works via a mixin, great. If not, we only need it for one model so far, so it's probably fine to not abstract it out, at least not yet. 

What do you think?

quevon24:
> 2\. to



> Yeah, my reaction is that we're over thinking this and the various libraries don't really fit our needs. I think what we need is:
> 
> 1. A way of storing the ordering information:
>    
>    * A table to join to or as a field on the model?
>    * Floats or ints?
> 2. Some methods/functions that allow you to set the ordering for an item.
> 3. Some tweaks to the front end to show items in the correct order.
> 
> I think that's mostly it. I don't think we need functions for `move_up` or `move_top`, fancy admin integration, or anything like that. I imagine when you guys are doing the import, you'll just need a setter, right?
> 
> Notionally:
> 
> ```
> opinion.order_position = 1
> opinion.save()
> ```
> 
> Right? The libraries feel either too big or full of features we don't really want, so let's take from them what we do want, and leave the rest behind. I, for one, will be glad to have one fewer library using memory and having to be upgraded and maintained.
> 
> If this all easily works via a mixin, great. If not, we only need it for one model so far, so it's probably fine to not abstract it out, at least not yet.
> 
> What do you think?

I think the same, I think it would be enough to add a field to the Opinion model and maybe some methods to handle the ordering of the object.  Maybe modify the save method or use a signal in case it is necessary to move the opinions from the same cluster.

I also don't know if it would be a good idea, but I was thinking of using a custom migration to fill in the new field based on the pk of the cluster opinions, for example if we have the opinions [<Opinion 258965>, <Opinion 258966>, < Opinion 258967>] for a specific cluster we can fill the field in order: 1, 2, 3 since generally the opinions from sources such as Columbia or Harvard are created in the order of appearance. This can be done in the migration or with a management command to fill the field.

flooie:
I'll die on the other hill.  I dont see the major harm from just using the default ordering code.  I prefer integers so I can say give me the second opinion - or the first opinion and know I can just also order it 1 2 3 and not worry about finding the right float value to make it work. 

If you say itll be a performance issue down the line or the API - I guess I would cave, but I dont see it.

quevon24:
> I'll die on the other hill. I dont see the major harm from just using the default ordering code. I prefer integers so I can say give me the second opinion - or the first opinion and know I can just also order it 1 2 3 and not worry about finding the right float value to make it work.
> 
> If you say itll be a performance issue down the line or the API - I guess I would cave, but I dont see it.

Usually there were performance hits when you have a very large data set in postgresql and you needed to apply ordering and pagination (I mention it because of the API and the deeper you go in the results), but I just remembered that we already updated to postgresql 15 and in this version there were improvements in this aspect (https://www.percona.com/blog/postgresql-15-new-features-to-be-excited-about/#:~:text=Improved%20sorting%20algorithms,performance%20gains%20%E2%80%A6%20they%20just%20work!) and the pagination in the API is limited to a certain number of pages. 

So it's probably not a big problem to use the library.

@mlissner @flooie which do you think is the best option?

quevon24:
And what do you think is the best approximation to fill the field with the correct value? With a migration or with a management command?

flooie:
I think we need to update all of harvard and Columbia - so a management command - I think - would be needed to parse the respective XMLs and assign an order.

mlissner:
Kevin, if we use the library you suggest (sorry, which one is it?), it'll add default ordering to the Opinion table, right? Would that mean that the API would be ordered by the new ordering table, or does it mean that we just have to make sure the API has ordering (which it does)? 

For the migration, I'd suggest just using what we have now as a first pass, if necessary, then adding the correct ordering data when we re-add this data. Why do it in two passes if we know the ordering and are about to do a huge ingestion that will correct this?

quevon24:
> Kevin, if we use the library you suggest (sorry, which one is it?), it'll add default ordering to the Opinion table, right? Would that mean that the API would be ordered by the new ordering table, or does it mean that we just have to make sure the API has ordering (which it does)?
> 
> For the migration, I'd suggest just using what we have now as a first pass, if necessary, then adding the correct ordering data when we re-add this data. Why do it in two passes if we know the ordering and are about to do a huge ingestion that will correct this?

@mlissner After inspecting it a lot, the library that @flooie suggested [Django Ordered Model](https://github.com/django-ordered-model/django-ordered-model) works good, it requires to add the ordering option to meta class, for example:

    class Meta:
        ordering = ("order",)

This ordering would only affect the queryset when we retrieve objects using for example:

```
Opinion.objects.all()
Opinion.objects.filter(type="010combined")
```

but it won't apply if you execute:
```
Opinion.objects.filter(type="010combined").order_by("id")
Opinion.objects.order_by("date_created")
```

Also, this won't apply to every endpoint in the API because we are using ModelViewSet for Opinions endpoint which extends [GenericAPIView](https://www.django-rest-framework.org/api-guide/generic-views/#genericapiview) and according to the documentation it requires to explicitly define the queryset, so every filter or ordering will be applied to the defined queryset, in this case the queryset is defined as:

```
queryset = (
        Opinion.objects.select_related("cluster", "author")
        .prefetch_related("opinions_cited", "joined_by")
        .order_by("-id")
    )

```
So if you apply a filter to the Opinions endpoint it will still return the results ordered by id unless you pass an order_by to the endpoint.

On the other hand, if you look at the API endpoint for the clusters, in the serializer we use `Opinion.objects.all()` for the queryset, so when you go to /api/rest/v3/clusters/, the sub_opinions will be displayed in order.

And regarding how to fill the order field, we continue to discuss it to see how to integrate it in the best possible way, for now we need to have the ordering field ready.

mlissner:
Sorry, Kevin, it took me all day to get to your comment. Yes, this all sounds great! I don't think we have any more blockers as to strategy. Let's go with our original approach and let me know when this is ready for code review. (There's an expression about having to go around the world to discover how nice home is....I feel like we did this here!)

quevon24:
Thanks @mlissner, the code is ready

mlissner:
The current plan is to hold off on this until Harvard merger is finished, and then supplemented by the missing Columbia data. Once those are in, we'll merge this, then do a pass to backfill it into existing cases.

mlissner:
Kevin, you requested a review here, but according to [my last comment](https://github.com/freelawproject/courtlistener/pull/2821#issuecomment-1660978653), we were planning to wait on this until Colombia and Harvard data are merged, right?

quevon24:
> Kevin, you requested a review here, but according to [my last comment](https://github.com/freelawproject/courtlistener/pull/2821#issuecomment-1660978653), we were planning to wait on this until Colombia and Harvard data are merged, right?

Sorry, it was my mistake, I was reviewing all my pending PRs, merging the PRs with main and correcting conflicts and I think I accidentally hit it

quevon24:
The only detail with the library is that when trying to use the order_with_respect_to option to order the opinions with respect to the cluster (https://github.com/django-ordered-model/django-ordered-model?tab=readme-ov-file #subset-ordering) It seems that this option causes problems with the django-model-utils library, specifically it gives us this error:

`AttributeError: 'Opinion' object has no attribute '_es_o_field_tracker'`

I tried to inspect the library but its implementation is not very descriptive so I would need to review it in detail if we wanted to solve that problem. At first glance it seems that the library does not like that the model has attributes that are not fields (perhaps an option would be to move the order_with_respect_to attribute to the meta class in the django-ordered-model library, I'm just guessing).

Once that option is removed, the tests pass without problems. In my opinion I think that this option is not necessary unless we wanted to order the opinions of a cluster and this can be done directly by applying the order_by where necessary.

Implementing the library in the admin did not present any problem, it is just adding a TabularInline to the cluster and modifying the OpinionCluster admin

image

You can move opinions one by one using the up, down, move to top or bottom arrows (you can't drag the opinions to reorder them). By default, the order field is filled with a 1. The command (#3078) will be in charge of filling this field for Harvard and Columbia, so I don't know if it is useful.

If we don't need the additional features of the library, we could simply add an integer field with a default value, and have the management command fill it with the correct position. This is taking into consideration that there are no splitted opinions from other sources, since it would be necessary to write additional commands for each source.

mlissner:
OK, I just spent some time getting back up to speed on this. I stand by my comment that:

> we're over thinking this and the various libraries don't really fit our needs. I think what we need is:
> 
> 1. A way of storing the ordering information:
>        
>    * A table to join to or as a field on the model?
>    * Floats or ints?
> 
> 2. Some methods/functions that allow you to set the ordering for an item.
> 
> 3. Some tweaks to the front end to show items in the correct order.

I'm just not excited about adding more big libraries to CL that we'll have to worry about breaking, and this one seems to already be causing weird and hard to debug issues.

So, here's what I think we should do:

1. An int field on the opinions objects that's null by default.
2. A unique_together constraint on the `cluster_id` and ordering fields, so that you cannot have two opinions in a cluster with the same ordering value.
     - (This means that if we need to re-order things you create a slot first by moving objects out of the way, then you fill the slot with the new value.)
4. A management command (or two) to populate the correct values in the field.

If we do that, I think we'll have a really simple system that works and we can put this to rest. 

What we lose:
 - No admin drag and drop (but we'd rarely want that, and it's the kind of thing that would break across Django versions)
 - No handy utilities for putting things in places, swapping values, moving up/down, etc. (But it's not like we'll be doing a lot of that.)

We also gain:
 - No default ordering (this is good, because more often than not you forget it's there and it takes DB CPUs you didn't intend).

Thoughts?

flooie:
@mlissner i agree completely that we should scrape the libraries and just use an int

quevon24:
@mlissner I also completely agree with what you mentioned, I'll start working on it

cl/search/models.py
```diff
@@ -3320,6 +3320,14 @@ class Opinion(AbstractDateTimeModel):
             "sha1",
         ]
     )
+    order = models.IntegerField(null=True, blank=True)
```
mlissner, line 3323: I'm just passing through here, but I'm not sure that `order` is the right name for this. it has two problems:

1. It could be interpreted as a noun because courts issue orders, so at a glance you might think it's a boolean that represents whether the opinion row is an order. Hrm.

2. It's quite generic, and intersects in a lot of other things in the DB. For example, this is a bit weird: `Opinion.objects.order_by('order')`.

I _think_  this field is more like `ordering_key` or even `cluster_order_key`. I don't love those either. Other ideas?

quevon24: mmm maybe `position` or `spot`?

mlissner: `spot` feels wrong to me. We have a `position` field on the courts table that deals with sorting, but it also can have two meanings. It can also mean the role that somebody has in an organization.

I think my fave remains `ordering_key`. 

@flooie do you have any thoughts?

mlissner: How about something like `publication_order`?

cl/search/fixtures/test_objects_query_counts.json
```diff
@@ -418,7 +423,8 @@
          "date_created":"2015-08-15T14:10:56.801Z",
          "html_lawbox":"",
          "per_curiam":false,
-         "type":"010combined"
+         "type":"010combined",
+         "ordering_key": 4
```
flooie, line 427: im not sure I understand what is happening here.  

How do we have ordering keys for combined opinions - shouldn't they be null 

and also why are they also alternatively not 1

flooie (commented):
It looks like we aren't accounting for ordering key correctly for combined opinions-  combined opinions dont have an order as they are all of the sub opinions

cl/search/tests/tests.py
```diff
@@ -300,6 +300,80 @@ def test_custom_manager_chained_filter(self) -> None:
...
+            date_filed=date(1978, 3, 10),
+            source="U",
+            precedential_status=PRECEDENTIAL_STATUS.PUBLISHED,
+        )
+
+        # Create three opinions
+        op_1 = OpinionFactory(
+            cluster=cluster,
+            type="Concurrence Opinion",
+            ordering_key=1,
```
flooie, line 325: this should be 2

cl/search/tests/tests.py
```diff
@@ -300,6 +300,80 @@ def test_custom_manager_chained_filter(self) -> None:
...
+        op_2 = OpinionFactory(
+            cluster=cluster,
+            type="Dissent",
+            ordering_key=2,
+        )
+
+        op_3 = OpinionFactory(
+            cluster=cluster,
+            type="Lead Opinion",
+            ordering_key=3,
```
flooie, line 337: this should be 1

cl/search/tests/tests.py
```diff
@@ -300,6 +300,80 @@ def test_custom_manager_chained_filter(self) -> None:
...
+        op_1 = OpinionFactory(
+            cluster=cluster,
+            type="Concurrence Opinion",
+            ordering_key=1,
+        )
+
+        op_2 = OpinionFactory(
+            cluster=cluster,
+            type="Dissent",
+            ordering_key=2,
```
flooie, line 331: this should be 3 most likely.

cl/corpus_importer/management/commands/update_opinions_order.py
```diff
@@ -0,0 +1,415 @@
...
+            type=int,
+            default=0,
+            help="Start id for a range of clusters (inclusive)",
+        )
+        parser.add_argument(
+            "--end-id",
+            type=int,
+            default=0,
+            help="End id for a range of clusters (inclusive)",
+        )
```
flooie, line 379: I think we should revamp this.  I find the start and end id stuff kinda confusing and I dont know why we need it.  

lets switch this to skip-until and --limit 

Otherwise I dont know how you would know what the end id you want to go until is.

flooie:
@mlissner - assuming the test doesnt fail - I think this is ready for your review

cl/search/models.py
```diff
@@ -3338,6 +3347,11 @@ def get_absolute_url(self) -> str:
     def clean(self) -> None:
         if self.type == "":
             raise ValidationError("'type' is a required field.")
+        if self.ordering_key is not None and self.ordering_key != "":
```
mlissner, line 3350: How about:

```suggestion
        if self.ordering_key:
```

Or:

```suggestion
        if isinstance(self.ordering_key, int):
```

Or, I guess, even better would be to combine these with an `and`:

```suggestion
        if isinstance(self.ordering_key, int) and self.ordering_key < 1:
            ...
```

cl/corpus_importer/management/commands/update_opinions_order.py
```diff
@@ -0,0 +1,96 @@
...
+    limit = options.get("limit", None)
+
+    base_filter = (
+        OpinionCluster.objects.exclude(filepath_json_harvard="")
+        .annotate(opinions_count=Count("sub_opinions"))
+        .filter(opinions_count__gt=1)
+    )
+
+    if skip_until:
+        base_filter &= Q(pk__gte=skip_until)
```
mlissner, line 29: Not that it matters, with the refactor I suggest below, but I don't think this would have worked? Can you add a Q to a QuerySet? Hmmm....

cl/corpus_importer/management/commands/update_opinions_order.py
```diff
@@ -0,0 +1,96 @@
...
+
+    harvard_clusters = (
+        OpinionCluster.objects.annotate(opinions_count=Count("sub_opinions"))
+        .filter(base_filter)
```
mlissner, line 33: I don't *think* this will do what you want it to. I think you'll be passing a QuerySet into a filter, which does a subquery. It might work, but I don't think it's efficient. It's worth checking by printing the query. 

Some other things:
 - It's also confusing that you do the annotation twice, here and above. 
 - If we're going to iterate over all the subopinions, we should do a select_related.

Putting this all together, I think you can delete everything above and replace it with something like:

```suggestion
    harvard_clusters = (
        OpinionCluster.objects.exclude(filepath_json_harvard="")
        .select_related("sub_opinions")
        .annotate(opinions_count=Count("sub_opinions"))
        .filter(opinions_count__gt=1)
        .order_by("id")
    )
    if skip_until:
        harvard_clusters = harvard_clusters.filter(pk__gte=skip_until)
```

Would that check all the boxes?

cl/corpus_importer/management/commands/update_opinions_order.py
```diff
@@ -0,0 +1,96 @@
...
+    for cluster in harvard_clusters:
+        logger.info(f"Processing cluster id: {cluster}")
+        sub_opinions = cluster.sub_opinions.exclude(
+            type=Opinion.COMBINED,
+        ).order_by("id")
+        if not sub_opinions:
+            logger.info(
+                f"No sub_opinions left to order for cluster id: {cluster}"
+            )
+            continue
```
mlissner, line 48: Instead of doing this query millions of times, could you do a check inside your for loop below and `continue` when you run into `Opinion.COMBINED`? You'd have to fiddle with the `opinion_order` value, but I think it could work, and save our poor DB a bit.

mlissner (changes requested):
Cool. Simple enough, which I appreciate. A few high-level things. 

First, this needs the same DB tweaks from the wiki that our other PR needed a minute ago.

Second, I'd like to check if we can avoid sending all of these to Elastic and Solr. That'd be a good thing to be able to do.

Third, are we 100% certain that ordering by ID and then adding keys is going to work properly? Is there really no instance when we had to do cleanup of an import or where an import was interrupted? Maybe we handled that via a transaction, but I'd like to be sure, if we can, before we make such a big change.

Finally, like with our other update, we should add a `--delay` parameter here, particularly if we are sending it to Elastic and Solr, so that it doesn't go wild and impact things more than we intend.

Thank you guys! This is close after over a year of revision, tweaks, and changes of direction!

quevon24:
The changes are ready @flooie

flooie:
@quevon24 thanks.  There are some migration file changes and customer migration file still needed for this.  

@mlissner I can think of no instances where the opinions would have been made out of order.  We could though Audit it and look at the metadata tags that identify page structure

quevon24:
@flooie I've already made the adjustments for the migrations

mlissner:
Sorry, what do you mean by this, @flooie: 

> We could though Audit 

You mean we could *do an audit*? Just making sure I'm not missing something.

flooie:
nothing more than we could verify using the metadata of the book scans.  I would need to re-familiarize with the metadata of the tags but I think it's all built into xml/tags thats all we could add that as a second verification as well.  It may not work for all the modifications we made but almost all if not all of those fixed opinions were single opinion fixes

mlissner:
Seems like a good idea if we have any concerns about the id-ordering being correct. Hopefully not entirely hard (he says naively)? Just find the first page in each of the sub-opinions, then sort?

We can still land this without that though, just to get the data modeling in place.

quevon24:
I think the worst case scenario is that the opinions won't be displayed in the correct order although I see it as unlikely since the harvard importer created the data in order of appearance, but as Bill suggests we can use tags within the xml, I think the tag is <page-number>, though I don't remember if everyone has it

flooie:
@quevon24 im not talking about page number - let me find the exact language im talking about and get back to you

mlissner:
> harvard importer created the data in order of appearance

Any chance that was done in a transaction, or do we have other ways of being sure that resuming things wouldn't cause problems? I think you're probably right that it's fine, but I want to be sure, if we can.

flooie:
@mlissner - it would be trivial to double check our order - the fast case data I don't believe is split into separate opinions, but the metadata embedded in each paragraph provides an order.  The general structure is

`<p id="b123-1"` .... `<p id="b123-2` .... `<p id="b124-1` interspersed with `<p id='AuX` tags for authors

We should be able to just validate the order of ID in nearly all cases.

mlissner:
Great, but do we need to?

quevon24:
I think all the opinions imported using the harvard importer should have been created in the correct order, or at least that is what can be inferred from the code (https://github.com/freelawproject/courtlistener/blob/main/cl/corpus_importer/management/commands/harvard_opinions.py#L573)

It uses the find_all() function from BeautifulSoup, that function maintains the order of the tags as they appear in the content. We search for <opinion> tags and tags with data-type="opinion" attribute (were added in opinionated for mislabeled content)

The process of creating the docket, cluster and opinions is done within a transaction block, so it is unlikely the the opinions were not created in the correct order sorting by id ascending (https://github.com/freelawproject/courtlistener/blob/main/cl/corpus_importer/management/commands/harvard_opinions.py#L494)

This applies to all opinions imported from the Harvard source (source=U) and not merged with any other data source like lawbox(Bill told me that only Harvard and Columbia's opinions can be splitted), there are ~4.2M opinions that are only from Harvard where I consider that what I mentioned above applies as long as they have more than one opinion.

mlissner:
>  the docket, cluster and opinions is done within a transaction block

Hell yes! That's great, nicely done, Kevin.

So let's take one more look at this with your findings, and then merge once all fixes are ready. 

Thank you!

flooie:
@quevon24 thanks for diving back into the old code archives.

mlissner (approved):
I didn't evaluate the ingestion script carefully. I'll leave that you you and Bill, but the migrations are good and we want to hand them off to Ramiro, so I'm going to merge them now.

Issue #2828, opened 2023-06-21
Warnings emitted during GitHub Workflow execution


cweider:
There are many warnings from the actions run in the checks of `pull_request`. Virtually all relate to deprecation of one kind or another.

The majority of issues relate to:

* [GitHub Actions: Deprecating save-state and set-output commands](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)
* [GitHub Actions: All Actions will begin running on Node16 instead of Node12](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/)

Both of which are of the variety “we’re turning this off in Summer 2023” – which would be bad. Enforcement of Node16 [became effective on the 14th of June](https://github.blog/changelog/2023-06-13-github-actions-all-actions-will-run-on-node16-instead-of-node12-by-default/) (with no serious effects) and collective inaction has led the `set-state` command disabling to be postponed indefinitely (from 31st of May 2023). Nevertheless, the deprecated command should be replaced – it’s best to address these warnings on a schedule we control.

cweider:
Tally of workflow warnings:
```
 3 Automate build and deploy
 3 CodeQL
 8 Lint
 0 Push new dev images
 1 Semgrep
50 docker build and test
 0 Sentry-release
```

PR #2830, opened 2023-06-21
chore(workflows): upgrade workflow actions

cweider:
When run, the majority of the workflow warnings mentioned in #2828 are:

```
The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```
```
Node.js 12 actions are deprecated. https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/
```

And these can be resolved by upgrading offending actions to more recent versions. After that, the patch adds Dependabot integration so that, prospectively, action upgrades are easier to manage.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -21,5 +21,4 @@ updates:
 - package-ecosystem: "github-actions"
   directory: "/"
   schedule:
-    # Check for updates to GitHub Actions every week
-    interval: "monthly"
+    interval: monthly
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,3 +17,9 @@ updates:
   # Ignore all patch updates.
   - dependency-name: '*'
     update-types: ["version-update:semver-patch"]
+
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    # Check for updates to GitHub Actions every week
+    interval: "monthly"
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -12,7 +12,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         # We must fetch at least the immediate parents so that if this is
         # a pull request then we can checkout the head.
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -32,7 +32,7 @@ jobs:
       id: vars
       run: echo "::set-output name=sha_short::$(git rev-parse --short HEAD)"
     - name: Configure AWS credentials
-      uses: aws-actions/configure-aws-credentials@v1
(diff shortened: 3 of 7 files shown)


cweider:
Special attention is needed for [docker/build-push-action@v4](https://github.com/docker/build-push-action/releases/tag/v4.0.0). It comes with the note: 
> Buildx v0.10 enables support for a minimal [SLSA Provenance](https://slsa.dev/provenance/) attestation, which requires support for [OCI-compliant](https://github.com/opencontainers/image-spec) multi-platform images. This may introduce issues with registry and runtime support (e.g. [Google Cloud Run and AWS Lambda](https://github.com/docker/buildx/issues/1533)). You can optionally disable the default provenance attestation functionality using provenance: false.

This may affect Courtlistener. If so, v4 could still be adopted using `provenance: false`.

.github/dependabot.yml
```diff
@@ -17,3 +17,9 @@ updates:
   # Ignore all patch updates.
   - dependency-name: '*'
     update-types: ["version-update:semver-patch"]
+
+- package-ecosystem: "github-actions"
+  directory: "/"
+  schedule:
+    # Check for updates to GitHub Actions every week
```
mlissner, line 24: Month or week?

mlissner (commented):
This all looks pretty ok. For the docker thing, I guess this means it's going to build multiarchitecture images, and when it does those may or may not work depending on where they are used?

mlissner:
Man, I just tried to learn more about SLSA and province and whether it would affect us, and Crazy Max is just always on a different level from me. I can understand what the purpose of these things is, but I have no idea if the comment will affect me and little idea how to check aside from just chancing it.

.github/dependabot.yml
cweider: Lemme just strike the whole comment – effect of the configuration is evident enough.

…pushing a `--fixup` commit – TBD if/how Github works with this and `rebase --autosquash`.

cweider:
Yea, “YMMV, the only way to find out is to try” seems to be the situation. Surprisingly, change was made in the minor release, [v3.3.0](https://github.com/docker/build-push-action/releases/tag/v3.3.0). After a few weeks [v3.3.1](https://github.com/docker/build-push-action/releases/tag/v3.3.1), came along switching provenance to be disabled by default (where 4.0.0 came at the same time with the default as enabled).

Failure case would probably be analogous to deploying an image with an incompatible architecture. This, from `docker/django/README.md`, makes it sound like we might have some experience with that.

> `arm64` images cause a lot of trouble if they are deployed to the server and thus should only be pushed without the `latest` tag.

Kubernetes probably failed early and gracefully in that case?

mlissner:
Thanks. This is waiting on https://github.com/freelawproject/courtlistener/issues/2827.

mlissner:
This deployed cleanly, thank goodness. 

Thank you!

Issue #2839, opened 2023-06-28
Links of the visualizations sidebar section of opinion page are incorrect.


cweider:
Instead linking to the visualization (`/visualizations/scotus-mapper/146/bowers-to-obergefell/`), each link is just a relative URL with the query, `?q=obergefell`:

<img align="left" width="291" alt="Screenshot 2023-06-28 at 12 02 25 AM" src="https://github.com/freelawproject/courtlistener/assets/45040/186f32ea-e07e-4769-9ed3-461dd3c6cc07"/>

```
<ul>
        <li>
            <a href="?q=obergefell">
                  Bowers (1986) to Obergefell (2015)
            </a>
        </li>
        <li>
            <a href="?q=obergefell">
                  Same-Sex Marriage: Loving (1967) to Obergefell (2015)
            </a>
        </li>
        <li>
            <a href="?q=obergefell">
                  Screws (1945) to Obergefell (2015) - Civil Rights 1945 to …
            </a>
        </li>
        <li>
            <a href="?q=obergefell">
                  SDP Origins of Obergefell: The Line from Pierce (1925)
            </a>
        </li>
        <li>
            <a href="?q=obergefell">
                  EP Origins of Obergefell: The Line back to Skinner (1942)
            </a>
        </li>
</ul>
```

This is caused by misuse of the `opinions_sidebar.html`, which is built for a `list[OpinionCluster]`, for a `list[SCOTUSMap]`.

PR #2840, opened 2023-06-28
fix(opinion_page): correct visualization sidebar links

cweider:
Misuse of `opinions_sidebar.html`, which expects a `list[OpinionCluster]`, for a `list[SCOTUSMap]` generates the correct text (fortunately), but an incorrect URL. Fix this by uncritically copying the `opinions_sidebar.html` include and adapting for the case of visualizations.

Fixes: #2839

### DIFF
--- a/cl/opinion_page/templates/opinion.html
+++ b/cl/opinion_page/templates/opinion.html
@@ -201,11 +201,22 @@
         {% if cluster.top_visualizations %}
             <div id="visualizations" class="sidebar-section">
                 <h3><span>Visualizations</span></h3>
-                {% url "cluster_visualizations" pk=cluster.pk slug=cluster.slug as cluster_visualizations_url %}
 
-                {% with opinions=cluster.top_visualizations|slice:":5" full_list_url=cluster_visualizations_url|add:"?q"|add:request.META.QUERY_STRING %}
-                    {% include 'includes/opinions_sidebar.html' %}
-                {% endwith %}
+                <ul>
+                    {% for viz in cluster.top_visualizations|slice:":5" %}
+                        <li>
+                            <a href="{{ viz.get_absolute_url }}">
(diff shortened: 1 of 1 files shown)


cl/opinion_page/templates/opinion.html
```diff
@@ -201,11 +201,22 @@ <h3><span>Related Opinions</span></h3>
...
+                            <a href="{{ viz.get_absolute_url }}">
+                                {{ viz.title|default_if_none:"N/A"|safe|truncatewords:10|v_wrapper }}
+                            </a>
+                        </li>
+                    {% endfor %}
+                </ul>
+                <p>
+                    <a href="{% url "cluster_visualizations" pk=cluster.pk slug=cluster.slug %}"
+                       class="btn btn-default">
+                        View All Visualizations
```
cweider, line 217: “View All Visualizations” borrowed from “View All Authorities” above at ln175.

cweider, line 217: Also omitted from this and other links is `?{{ request.META.QUERY_STRING }}`, which is widely used for other sidebar links.

For the “View All” two q’s had been added (`?qq=`) this has got to be a mistake. For the visualization links themselves, there doesn’t appear to be any purpose served by the query.

This said, I haven’t been able to explain to myself the reason for `?{{ request.META.QUERY_STRING }}` on the opinion links. Is it a vestigial thing? part of (now lost) breadcrumb feature? Could it be stripped from all links in a follow-up?

mlissner: Yeah, no point adding the GET parameters to viz links. 

As for the opinion links, I keep those params around in the hopes that one day we'll use them for highighting the search terms in the text of the opinions, but that day may never come. This could certainly be stripped.

mlissner:
I'm pretty surprised you found this at all, but thanks for fixing it. Visualizations aren't a huge feature anymore, but good to shake out their bugs anyway. Merging!

PR #2847, opened 2023-06-29
chore(workflows): bypass git usage by build-push-action

cweider:
When `context` is unspecified, the action will fetch the repository on its own:

> By default, this action uses the Git context, so you don't need to use the actions/checkout action to check out the repository as this will be done directly by  BuildKit.

This might be convenient, but it is not efficient. An earlier step, "Check out CourtListener", has not only retrieved the repository's contents and placed them at `./courtlistener`, but it has done this more efficiently, by fetching the specific ref only.

The default behavior also invites surprises:

> Be careful because *any file mutation in the steps that precede the build step will be ignored, including processing of the `.dockerignore` file* since the context is based on the Git reference.

By specifying a context, the action will use already present copy the content it would otherwise download.


mlissner:
Sorry, I guess I missed this one in June. Setting it to Automerge now!

PR #2848, opened 2023-06-29
chore(workflows): update steps to use new environmental files

cweider:
When run, the lint and build workflows emit the warning: our workflows:
```
The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

The disabling (initially set for 31st May 2023) has been postponed indefinitely. Nevertheless, the deprecated command should be replaced. Here a minor adjustment will suffice - do it.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -29,7 +29,7 @@ jobs:
     - uses: actions/checkout@v3
     - name: Set shortcode
       id: vars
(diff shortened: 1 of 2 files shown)


mlissner (approved):
One weird syntax for another, but presumably progress. Anything that removes a double colon syntax is progress in my book, thanks. 

Setting to auto-merge.

PR #2849, opened 2023-06-29
Add merge_group and see if it makes option appear.

cweider:


cweider:
Test change; should have been destined for *my* repository.

PR #2892, opened 2023-07-11
fix(security): amend CSP for Images and PDFs

cweider:
Adds domains to `connect-src`, `frame-src`, `img-src`, and `object-src`.


cl/settings/project/security.py
```diff
@@ -103,9 +105,15 @@
 CSP_IMG_SRC = (
     "'self'",
     f"https://{AWS_S3_CUSTOM_DOMAIN}/",
+    "https://portraits.free.law",
     "data:",  # @tailwindcss/forms uses data URIs for images.
+    "https://*.stripe.com",
+    "https://i.creativecommons.org",
```
mlissner, line 111: I think we only load that one image from CC. We should just add it to our resources rather than include it in a header for all pages forever, right?

cweider: Yea, thought about that too – easy fix – will do.

PR #2893, opened 2023-07-11
fix(security): prevent Bootstrap modal from remote loading

cweider:
CSP reports revealed that Bootstrap's modal was making an AJAX request for PACER content when presenting `#modal-buy-pacer`. This is because the framework unhelpfully loads the href of the target element, if present. The fix is to set another attribute, `data-remote` on either the source or target of the modal.

This affects all (~30) uses of `modal-buy-pacer`. And so, rather than correct this at the source, we apply `data-remote` to the target.

Fixes: #2882


mlissner:
That's a weird one.

PR #2895, opened 2023-07-12
chore(security): set samesite for JS cookie assignments

cweider:
Through direct and indirect assignments (there are two uses of `cookie-name` used to toggle banners), the following cookies are set:

```
recap_install_plea
buy_on_pacer_modal
no_banner
```

Adjust all to declare cookies to be `samesite=lax`: each are needed in cross-site browsing contexts - e.g. linking directly to a page from outside.


mlissner:
Thank you!

PR #2901, opened 2023-07-17
docker helper scripts

ss108:


mlissner:
Want to add some documentation, too? Otherwise, I'm not sure if/how we're supposed to use these.

ss108:
@troglodite2 Hey Chris, if you have any comments on these updated versions of your helpful Docker scripts that we are looking to get merged in officially, please chime in, but no pressure if you are busy.

docker/scripts/test.sh
```diff
@@ -0,0 +1,5 @@
+
+docker exec cl-doctor pip install x-ray
```
ss108, line 2: Chris, if you do take a look at this PR, can you clarify why this is here? Is x-ray required to be installed in/by cl-doctor for these Selenium tests to run successfully?

docker/scripts/clean.sh
```diff
@@ -0,0 +1,17 @@
...
+# Run it before using the build script. Usage of both of these shell scripts is optional.
+
+# Mostly from here: https://stackoverflow.com/a/39127792/3256077
+
+# stop all running CL containers
+docker-compose -f ../courtlistener/docker-compose.yml stop
+# rm said containers
+docker-compose -f ../courtlistener/docker-compose.yml rm --force
+# rm the locally-stored CL images themselves
+docker rmi $(docker-compose -f ../courtlistener/docker-compose.yml config | awk '/freelawproject/{if ($1 == "image:") print $2;}')
```
ss108, line 15: My understanding is that it's worth removing the images largely because if we're using the `:latest` tag for a lot of stuff, what is labeled as such locally may not match what is up there in the registry. I think some might believe as a matter of like philosophy that it's better to totally delete these things and then pull them fresh than to trust the `pull` command to update them even though it's supposed to do that. 

This was also being done in Chris's original script (via a different set of commands).

ss108:
@mlissner All of these work and are ready for your review, but might as well wait a couple days to see if Chris chimes in.

docker/scripts/clean.sh
```diff
@@ -0,0 +1,17 @@
...
+# pull the latest images
+docker-compose -f ../courtlistener/docker-compose.yml pull
```
mlissner, line 17: I'd take this off of a cleaning script. When you start docker, it'll pull automatically, and it feels weird for a cleaning script to bring in more stuff you may or may not want.

docker/scripts/lint.sh
mlissner, line 1: This feels wrong. It's repeating what we already have, and in theory pre-commit should be figured out to do this instead. Let's just yank it and optionally find a better solution later.

docker/scripts/test.sh
mlissner: It's not. Let's remove it.

docker/scripts/test.sh
mlissner, line 1: This file feels unnecessary too. If you remove the x-ray part, you're left with two commands, and if you remove the tagging part of those commands, it's just the one command that's in the dev guide.

mlissner (changes requested):
Some thoughts..... :)

docker/scripts/clean.sh
ss108: I think you're right that it does not belong in the clean script.

However, I am not sure that running `docker-compose up` actually pulls the latest images by default. I believe that this is only the case if the `pull_policy` config variable is set for a service: https://docs.docker.com/compose/compose-file/05-services/#pull_policy 

It may make sense to set that for certain services. IMO, we should use `build` instead of `image` for `cl-django` and point it at the Dockerfile. Ultimately, then, we can just remove the build script that is included in this PR (as well as remove this line from the clean script).

docker/scripts/lint.sh
ss108: Yeah, sounds good.

docker/scripts/clean.sh
mlissner: I'm happy to try that, sure.

docker/courtlistener/docker-compose.yml
```diff
@@ -72,7 +78,13 @@ services:
...
   cl-django:
     container_name: cl-django
-    image: ${DJANGO_DOCKER_IMAGE:-freelawproject/courtlistener:latest-web-dev}
+    build:
+      context: "../../"
+      dockerfile: "./docker/django/Dockerfile"
+      args:
+        BUILD_ENV: dev
+      target: web-dev
+    image: freelawproject/courtlistener:latest-web-dev
```
ss108, line 87: @mlissner I wasn't 100% sure what to label/name the images; LMK if these need to be changed for `cl-django` and `cl-celery`.

mlissner:
@cweider I'm curious if you have experience here. Do you know the best practice in terms of whether compose files should build vs. pull by default and how tagging ought to work?

docker/courtlistener/docker-compose.yml
cweider: Two things (that we can work out over Slack):

1. `image` might be redundant? What would be necessary for us to rely on `build` entirely.
2. `DJANGO_DOCKER_IMAGE` is removed here, but it persists in `.github/workflows/tests.yml`. There should be a corresponding change?

Perhaps reliance on `build` and removal of `image` could mean that `latest-*-dev` images don’t need to exist and some more thorough simplification can be done.

ss108:
Removed `clean.sh` per Slack convo

mlissner:
I think you and @cweider have a better handle on this, so I'm going to sort of duck out. Chad, do you mind reviewing?

ss108:
@cweider We are getting errors along the following lines during the `Start docker compose` step of the tests workflow: `failed to register layer: write /usr/local/lib/python3.11/site-packages/uvloop/loop.cpython-311-x86_64-linux-gnu.so: no space left on device`

This error arises when pulling one of the remote images, but it's not always the same image, so I don't think any of the images themselves are the cause. Moreover, it's not just happening on this branch, where I have been messing with the `docker-compose` and `tests.yml` files; it is also happening here, which retains the normal `docker-compose.yml` and the normal `tests.yml`: https://github.com/freelawproject/courtlistener/pull/3070 

I tried using Docker Compose v2 syntax, and that didn't fix the problem. 

Currently, this branch has the original `tests.yml` and the new `docker-compose.yml`.

It's unclear whether the changes in this branch are the cause of the problem. On the one hand, changes have been made to the test workflow and what the Compose file does, but on the other: the problem is also happening in a separate branch; this one was passing the test workflow just fine until recently; and reverting the tests.yml here back to the original did not help.

As to the long and now-polluted commit history, I will create a new branch and squash merge this one into it when we get this working again.

ss108:
Closing because we are going to be going with this more comprehensive set of changes: https://github.com/cweider/courtlistener/commit/f17b897dcee4a1ff9bf72a17040b112ebadc4d32

mlissner:
Thanks Samier. The best code is the code you never merge. :)

I'm glad you're pushing this conversation forward so we can really dial this in.

PR #2905, opened 2023-07-17
fix(html): correct form label associations

cweider:
An invisible change to correct warnings that Chrome gives about some label usages on the search and docket pages:
```
Incorrect use of <label for=FORM_ELEMENT>

The label's for attribute doesn't match any element id. This might prevent the browser from correctly autofilling the form and accessibility tools from working correctly.
To fix this issue, make sure the label's for attribute references the correct id of a form field.
```


mlissner:
Set to automerge.

PR #2914, opened 2023-07-19
fix(security): amendments to CSP for `media-src` and `base-url`

cweider:
Two more permissive changes for the policy. These address the outstanding reports that could potentially be caused by CourtListener itself. Those that remain are unreproducible and have not been explainable by any hypothesis apart from browser extensions.


mlissner:
These look good to me, thank you.

Issue #1491, opened 2020-12-02
Escape arrows in opinion HTML


mattdahl:
I was recently perusing Alito's dissent in _Bostock_ (as one does) and noticed that the HTML for his opinion is horribly broken: https://www.courtlistener.com/opinion/4760997/bostock-v-clayton-county/

It would seem that Alito quoted a dictionary containing the string `respectively  designated  male or female <a member of the opposite ~> 2: the` and the `<a` part became interpreted and rendered as the beginning of an `a` tag.

I think this could be fixed by simply escaping the `<` and `>` characters when we import the opinion text?

mlissner:
Wow. Yeah, I know that one of our vulnerabilities is that a court could embed nasty stuff in one of their opinions. I guess it finally happened sort of. I'm not sure that we have a particularly good fix for this. The easy solution, I suppose is to do what you say though, and to start encoding opinion text before putting it in the DB. We can't easily do that for all historical cases (it'd take an ETL script), but we can certainly start doing it going forward. 

Do you want to take a stab at it, @mattdahl? I'm going to fix this one by hand right now.

mlissner:
OK, so the plaintext is not currently considered safe in our HTML (see last if block):

```
<div id="opinion-content"> {# used by Zotero, Juris-M #}
                                {% if sub_opinion.html_with_citations %}
                                  {% if cluster.source == "C" %}
                                    {# It's a PDF with no HTML enrichment #}
                                    <div class="plaintext">{{ sub_opinion.html_with_citations|safe }}</div>
                                  {% else %}
                                    <div class="serif-text">{{ sub_opinion.html_with_citations|safe }}</div>
                                  {% endif %}
                                {% elif sub_opinion.html_columbia %}
                                  <div class="serif-text">{{ sub_opinion.html_columbia|safe }}</div>
                                {% elif sub_opinion.html_lawbox %}
                                  <div class="serif-text">{{ sub_opinion.html_lawbox|safe }}</div>
                                {% elif sub_opinion.xml_harvard %}
                                  <div class="serif-text">{{ sub_opinion.xml_harvard|safe }}</div>
                                {% elif sub_opinion.html_anon_2020 %}
                                  <div class="serif-text">{{ sub_opinion.html_anon_2020|safe }}</div>
                                {% elif sub_opinion.html %}
                                  <div class="serif-text">{{sub_opinion.html|safe}}</div>
                                {% else %}
[...]

PR #2991, opened 2023-08-04
fix(citations): fix encoding of annotated plaintext opinions

cweider:
Eyecite’s `annotate_citations` has an implicit requirement: the content type of `source_text` must be compatible with The `annotations` interpolated into it. In practice, `source_text` should be an HTML fragment.

Here, the output of `annotate_citations` is treated as HTML, but the input `source_text` is never encoded. After annotation, there is no place where encoding could be done without also encoding the annotations themselves; `opinion.source_text` should be encoded (and `<pre>` wrapped, for good measure) as HTML on input to Eyecite.

Fixes: #1491

### DIFF
--- a/cl/citations/annotate_citations.py
+++ b/cl/citations/annotate_citations.py
@@ -1,3 +1,4 @@
+import html
 from typing import Dict, List
 
 from eyecite import annotate_citations, clean_text
@@ -78,16 +79,15 @@ def create_cited_html(
             source_text=opinion.source_text,
             unbalanced_tags="skip",  # Don't risk overwriting existing tags
         )
-    else:  # Else, make sure to wrap the new text in <pre> HTML tags...
+    else:  # Else, present `source_text` wrapped in <pre> HTML tags...
         new_html = annotate_citations(
             plain_text=opinion.cleaned_text,
             annotations=[
                 [a[0], f"</pre>{a[1]}", f'{a[2]}<pre class="inline">']
                 for a in generate_annotations(citation_resolutions)
             ],
-            source_text=opinion.source_text,
+            source_text=f'<pre class="inline">{html.escape(opinion.source_text)}</pre>',
         )
-        new_html = f'<pre class="inline">{new_html}</pre>'
 
     # Return the newly-annotated text
     return new_html
--- a/cl/citations/tests.py
+++ b/cl/citations/tests.py
@@ -158,7 +158,7 @@ class CitationTextTest(SimpleTestCase):
 
             # Id. citation across line break
             ('asdf." Id., at 315.\n       Lorem ipsum dolor sit amet',
-             '<pre class="inline">asdf." </pre><span class="citation '
+             '<pre class="inline">asdf.&quot; </pre><span class="citation '
              'no-link">Id., at 315</span><pre class="inline">.\n       Lorem '
              'ipsum dolor sit amet</pre>'),
 
@@ -173,6 +173,10 @@ class CitationTextTest(SimpleTestCase):
              '<pre class="inline">Lorem ipsum dolor sit amet. U.S. Code </pre>'
              '<span class="citation no-link">§3617.</span><pre class="inline">'
              ' Foo bar.</pre>'),
+
+            # Plaintext with HTML text (see Alexis Hunley v. Instagram, LLC)
+            ('<script async src="//www.instagram.com/embed.js"></script>',
+             '<pre class="inline">&lt;script async src=&quot;//www.instagram.com/embed.js&quot;&gt;&lt;/script&gt;</pre>'),
         ]
 
         # fmt: on


cl/citations/annotate_citations.py
```diff
@@ -78,16 +79,15 @@ def create_cited_html(
...
-    else:  # Else, make sure to wrap the new text in <pre> HTML tags...
+    else:  # Else, present `source_text` wrapped in <pre> HTML tags...
         new_html = annotate_citations(
             plain_text=opinion.cleaned_text,
             annotations=[
                 [a[0], f"</pre>{a[1]}", f'{a[2]}<pre class="inline">']
                 for a in generate_annotations(citation_resolutions)
             ],
-            source_text=opinion.source_text,
+            source_text=f'<pre class="inline">{html.escape(opinion.source_text)}</pre>',
```
mlissner, line 89: This looks fine to me, but I'm confused how we can just wrap stuff in a `pre`, and not face bad consequences. Can you help me, since I haven't looked at this part of the code in forever? I'd have Flooie look, but it's his last day before a vacation...

cweider: The easy answer is basically, what’s sauce for the goose is sauce for the gander: above, in the `source_is_html` case, `source_text` has the same significant difference (HTML with tags) from `plain_text` that the non-HTML case now has.

Within Eyecite, the `SpanUpdater` does some complicated stuff to map the citation indices generated by `get_citations` which are relative to `plain_text` to indices on the HTML `source_text`. It’s a process that sounds fraught, but evidently works well enough.

cweider:
There are three opinions that I know of which could benefit from regeneration after this patch, [4541344](http://hafen.local:8000/opinion/4760997/bostock-v-clayton-county/), [9409282](https://www.courtlistener.com/opinion/9413806/alexis-hunley-v-instagram-llc/), and [4060801](https://www.courtlistener.com/opinion/4283540/rebecca-terrell-and-chandrashekhar-thanedar-v-pampa-independent-school/):

```
for OPINION_ID in 4541344 9409282 4060801
do
  python /opt/courtlistener/manage.py cl_find_citations --doc-id $OPINION_ID
done 
```

Note: [4060801](https://www.courtlistener.com/opinion/4283540/rebecca-terrell-and-chandrashekhar-thanedar-v-pampa-independent-school/), is enormous (for reasons unknown, it is not an opinion, but rather the appellant’s 300 page brief), and this stresses Eyecite and takes a startling amount of time (4 minutes). That said, the page has loads of errors and CSP violations, that could cause false alarms in the future, after this work is forgotten – so regeneration is probably still worth it.

cweider:
Appended related change to escape values interpolated with Python format strings (they’re not Django templates; they require extra care).

cl/citations/annotate_citations.py
```diff
@@ -50,7 +51,7 @@ def generate_annotations(
             ]
         else:  # If successfully matched...
             annotation = [
-                f'<span class="citation" data-id="{opinion.pk}"><a href="{opinion.cluster.get_absolute_url()}">',
+                f'<span class="citation" data-id="{html.escape(str(opinion.pk))}"><a href="{html.escape(opinion.cluster.get_absolute_url())}">',
```
cweider, line 54: `str()` necessary for `option.pk` pushes this into overzealous territory doesn’t it? (I was surprised that `escape` doesn’t coerce the argument to a string itself).

mlissner: opinion.pk will always be an int. Why do this? 

`get_absolute_url` will be a `SlugField` with some numbers. Do we really need to escape it?

cl/citations/annotate_citations.py
mlissner: Cool, thanks. I also didn't see you removed line 90, so now it makes more sense.

cl/citations/annotate_citations.py
cweider: Yea, as you say – both (especially `pk`) are known to contain characters that would cause problems; the argument for it really cultural rather than technical. Routine usage makes it more likely that implementers will do the right thing when they copy paste and make failures to escape more obtrusive. 

But this uncompromising approach only pays off when exceptions are rare and that is very much not CourtListener’s situation today. If the type cast wasn’t necessary, I’d probably stand by the argument for pro forma escaping (since its harmless), but as it is now the extra complexity just raises questions without meaningfully moving the needle on encoding rigor. Gonna drop this commit.

cweider:
Dropped the extra zealous escaping and rebased.

mlissner:
Thanks, looks great.

mlissner:
I love a nice encoding fix.

mlissner:
Thanks for the bash too. I ran it, but it needed ` --index False;` or else it wanted to add all opinions to CL! Not good.

Anyway, this is all set.

PR #2995, opened 2023-08-04
chore(search_results): encode related results URL parameters

cweider:
While the values interpolated into URLs for related searches have recieved HTML escaping, they were not also being encoded for the URL context that they were in.

For example:
```
<a href="/?type=r&amp;q=(&quot;United States v. Ulbricht&quot;) AND docket_id:7195048">
    View 11 Additional Results for this Case
</a>
```

This isn't of much practical significance. I mean, there are lots of actions in rem against things identified by URIs, but, sadly, queries for these do not produce any results. Nonetheless, one of the best guards against slip-ups in encoding where it _is_ important is to perform it consistently - even where it is unimportant.

Also:
- consistently encode `:` as `%3A`
- drop `&amp;` where `&` alone will suffice

### DIFF
--- a/cl/search/templates/includes/search_result.html
+++ b/cl/search/templates/includes/search_result.html
@@ -81,7 +81,7 @@
       <div class="inline-block">
         <span class="meta-data-header">Nature of Suit:</span>
         <span class="meta-data-value">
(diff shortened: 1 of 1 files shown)


mlissner:
I'm not good enough at this stuff anymore to know that this is right as I review it, but I'm entirely ready to take your word for it, and your explanations make sense.

Thanks, merging.

PR #2999, opened 2023-08-06
chore(docker): drop ports from ancillary containers

cweider:
`cl-redis`'s use of `6379` conflicts with `bc2-redis` (BigCases) and frustrates parallel development. There's no need for many container ports to be exposed on the host. It is enough that they are accessible within the `cl_net_overlay` Docker network.


mlissner:
Great, thank you.

PR #3006, opened 2023-08-08
chore(docker): pestore cl-postgresql port exposure

cweider:
Unlike other exposed ports that were removed in #2999, this has non-negligible use in development.


mlissner:
Thank you! Auto-merge set.

PR #3009, opened 2023-08-09
2029 make api documentation more user friendly

dschnelldavis:
- Move REST API TOC menu from left to right, so it matches all the other help pages on the site.
- Move all other REST API items in the left column to the bottom of the main section.
- Add additional REST API TOC links for new items in main section.
- Change REST API "Available Jurisdictions" and "API Change Log" so they link to other pages.


cl/api/urls.py
```diff
@@ -154,6 +154,11 @@
         views.rest_docs,
         name="rest_docs",
     ),
+    re_path(
```
mlissner, line 157:
```suggestion
    path(
```

cl/api/views.py
```diff
@@ -213,6 +213,11 @@ async def deprecated_api(request, v):
     )
 
 
+def rest_change_log(request):
+    context = {"private": False}
+    return render(request, "rest-change-log.html", context)
```
mlissner, line 218: There are tests in `BasicAPIPageTest` that just load these very basic URLs. Mind adding this URL too?

cl/api/templates/rest-docs-vlatest.html
```diff
@@ -13,8 +13,29 @@
 {% endblock %}
 
 {% block content %}
-  <div class="col-sm-7">
+  <div class="col-sm-3">
+
+    {% include "includes/toc_sidebar.html" %}
+
+  </div>
```
mlissner, line 20: This should copy some of the other templates, and has extra whitespace, so:

```suggestion
  <div class="hidden-xs hidden-sm col-md-3">
    {% include "includes/toc_sidebar.html" %}
  </div>
```

cl/api/templates/rest-docs-vlatest.html
```diff
@@ -13,8 +13,29 @@
...
+  <div class="col-sm-9">
```
mlissner, line 21: Again to mimic other pages:

```suggestion
  <div class="col-xs-12 col-md-8 col-lg-6">
```

cl/api/templates/rest-docs-vlatest.html
```diff
@@ -13,8 +13,29 @@
...
+        <h4><code>{{ user.auth_token }}</code></h4>
+        <p>Do not share your API token. Keep it private like a password.
+        </p>
+        <p class="text-right"><a class="btn btn-default" href="{% url "view_api" %}#usage">See Usage</a></p>
+      {% else %}
+        <p><a href="{% url "sign-in" %}?next={{ request.path }}"
+                class="btn btn-primary btn-lg">Sign In To See Your Token</a>
+        </p>
+      {% endif %}
+    </div>
```
mlissner, line 37: Maybe this goes better in the authentication section?

cl/api/templates/rest-docs-vlatest.html
```diff
@@ -501,7 +522,7 @@ <h3 id="idb-data-endpoint">{% url 'fjcintegrateddatabase-list' version='v3' %}</
     </p>
 
 
-    <h3 id="docket-entry-endpoint">{% url "docketentry-list" version="v3" %} <span class="alt">&amp;</span> {% url "recapdocument-list" version="v3" %}</h3>
+    <h3 id="docket-entry-endpoint">{% url "docketentry-list" version="v3" %} &amp; {% url "recapdocument-list" version="v3" %}</h3>
```
mlissner, line 525: Not a fan of fancy ampersands eh?

cl/api/templates/rest-docs-vlatest.html
```diff
@@ -1036,107 +1057,51 @@ <h3 id="visualization-endpoint">{% url "scotusmap-list" version="v3" %}
...
-                class="btn btn-primary btn-lg">Sign In To See Your Token</a>
-        </p>
-      {% endif %}
-    </div>
+    <h2 id="powered-by"><span>Powered By</span></h2>
+    <p> If you are interested, we have a variety of <a href="https://free.law/press-assets/">"Powered By" logos</a> available for your website or application. We appreciate a reference, and it supports our work, but it's not currently required.
+    </p>
+    <a href="https://free.law/press-assets/">
+      <img src="{% static "png/logo-long-w350.png" %}" width="350" height="81">
+    </a>
```
mlissner, line 1070: Ya know, nobody has *ever* used one of these. Let's just remove this idea forever. We can remove the static assets too.

cl/api/templates/rest-docs-vlatest.html
```diff
@@ -1036,107 +1057,51 @@ <h3 id="visualization-endpoint">{% url "scotusmap-list" version="v3" %}
...
 
-    {% include "includes/toc_sidebar.html" %}
+    <h2 id="citegeist-scores"><span>CiteGeist Scores</span></h2>
+    <p>If you are interested in the CiteGeist score for each opinion, it is now available via our <a href="{% url "bulk_data_index" %}#citegeist">bulk data API</a>.
+    </p>
+    <p>This data is not available via the REST API because a single new citation between cases causes a ripple effect through every value in the database.
+    </p>
```
mlissner, line 1076: Let's remove this too. We don't share this file anymore (it was too complex).

mlissner (changes requested):
This looks great. I made a handful of comments and suggestions. One thing that'd also be nice is if you could add this template to the bottom of it, below the footnotes:

image

It's on the bottom of the root API page, and I'm guessing is an easy include to toss in.

Otherwise, this looks like a really nice improvement and a great first PR.

cl/api/templates/rest-docs-vlatest.html
dschnelldavis: They were inconsistent. Some ampersands on the page (and in the menu) were fancy, and others were not. If you like fancy ampersands, I'll be happy to make them _all_ fancy, I just didn't like the mixture.

mlissner: It's all good. :)

cl/api/templates/rest-docs-vlatest.html
dschnelldavis: Done.

cl/api/templates/rest-docs-vlatest.html
dschnelldavis: Done.

cl/api/templates/rest-docs-vlatest.html
dschnelldavis: Done.

mlissner:
When this is ready for a second review, use the "Request Review" feature in the upper right of the PR to flag it for me again. 

In the meantime, I saw that you included the "Please Donate" as a big png file. I meant that the HTML for that screenshot should be included in the API docs. Sorry if that was unclear.

dschnelldavis:
Hmmm. I don't see a "Request Review" feature.

But I do think all the updates have been made, and the PR is ready to go.

mlissner:
It's the circle arrow icon here: 

image

cl/api/templates/rest-docs-vlatest.html
```diff
@@ -1103,5 +1088,22 @@ <h2 id="copyright"><span>Copyright</span></h2>
...
+      <p>Will you support us today with a donation?</p>
+      <div class="row">
+        <div class="hidden-xs col-sm-1 col-md-2 col-lg-3"></div>
+        <div class="col-xs-12 col-sm-10 col-md-8 col-lg-6">
+          <p class="text-center">
+            <a href="{% url "donate" %}?referrer={{ request.path }}" class="btn btn-lg btn-block btn-danger"><i class="fa fa-heart-o"></i> Donate Now</a>
+          </p>
+        </div>
+      </div>
+    </div>
```
mlissner, line 1106: We should avoid repeating ourselves. Mind putting this into an include and then tweaking this and the other page where it appears?

mlissner (changes requested):
Just the one tweak, and then we're ready. Thank you!

mlissner:
Sorry, I didn't catch this before, but tests seem to be broken: 

```
FAIL: test_api_info_page_can_display_different_versions_of_rest_docs (cl.api.tests.BasicAPIPageTest.test_api_info_page_can_display_different_versions_of_rest_docs)
  File "/usr/local/lib/python3.11/unittest/case.py", line 57, in testPartExecutor
    yield
  File "/usr/local/lib/python3.11/unittest/case.py", line 623, in run
    self._callTestMethod(testMethod)
    ^^^^^^^^^^^^^^^^^
  File "/opt/courtlistener/cl/tests/cases.py", line 28, in _callTestMethod
    super()._callTestMethod(method)
    ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/unittest/case.py", line 579, in _callTestMethod
    if method() is not None:
    ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/asgiref/sync.py", line 240, in __call__
    return call_result.result()
    ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
    ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
    ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/asgiref/sync.py", line 306, in main_wrap
    result = await self.awaitable(*args, **kwargs)
    ^^^^^^^^^^^^^^^^^
  File "/opt/courtlistener/cl/api/tests.py", line 119, in test_api_info_page_can_display_different_versions_of_rest_docs
    self.assertContains(response, header)
    ^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/django/test/testcases.py", line 660, in assertContains
    self.assertTrue(
  File "/usr/local/lib/python3.11/unittest/case.py", line 715, in assertTrue
    raise self.failureException(msg)
    ^^^^^^^^^^^^^^^^^
AssertionError: False is not true : Couldn't find 'REST API &ndash; V2' in response
```

Also, this will sound trivial, but I just know `donate.html` will confuse folks. Maybe call it `donate_footer_plea.html`?

dschnelldavis:
Done.

Though it still seems to be failing one of its tests, and I don't know why.

cl/api/templates/rest-docs-vlatest.html
dschnelldavis: Done.

mlissner:
So the test:

```
    async def test_api_info_page_can_display_different_versions_of_rest_docs(
        self,
    ) -> None:
        for version in ["v1", "v2"]:
            response = await self.async_client.get(
                reverse("rest_docs", kwargs={"version": version})
            )
            self.assertEqual(response.status_code, 200)
            self.assertTemplateUsed(response, f"rest-docs-{version}.html")
            header = f"REST API &ndash; {version.upper()}"
            self.assertContains(response, header)
```

Is trying to load the following URLs...

 - /help/api/rest/v1/
 - /help/api/rest/v1/

...and then is checking for the phrase "REST API &ndash; V1" or "REST API &ndash; V2" in the text. I'm not sure what's wrong with the test, but have you tried those URLs? It's a simple test, so I suspect it'll probably be obvious what's wrong when you load the URLs?

dschnelldavis:
Curiouser and curiouser.

- /help/api/rest/v1/ loads the depreciated v1 help file, with the phrase "REST API – V1" at the top of the page, so theoretically it should pass the test just fine.
- /help/api/rest/v2/ loads the latest rest api help file (v3), which does _not_ have phrase "REST API – V2" anywhere on the page, so that explains why it fails the test.

Unfortunately, I've looked in urls.py and views.py, and I can't figure out why the v1 page loads fine, but the v2 URL loads the v3 page instead. So I'm stuck.

Any suggestions?

Also, do versions 1 & 2 of the API still work? If not, would it make sense to just get rid of the REST API v1 & v2 help pages?

mlissner:
I always like to keep the old documentation around just in case people are migrating off it or still have old code that uses it, but it's been a long time. We can remove it without anybody noticing, and just have a template that says the API is deprecated, please use the latest version. 

That said, if I had to guess, I'd say it's probably this that's causing the v3 to be loaded instead of v2:

```
    try:
        return render(request, f"rest-docs-{version}.html", context)
    except TemplateDoesNotExist:
        return render(request, "rest-docs-vlatest.html", context)
```

Could you try debugging that real quick and see if it's failing to find the template for some reason?

mlissner:
Sorry, you're getting a bigger PR than we'd hoped for, but you're climbing the learning curve, right!?

dschnelldavis:
Yes, I'm definitely learning a lot. :-)

I think I fixed it. The V2 help file referenced an included file I deleted. So I got rid of that reference, and hopefully now it will pass the tests.

mlissner (approved):

mlissner:
Of course, darn it, that makes perfect sense. Glad you caught it! Merged! 

This will auto-deploy in about 20 minutes if you refresh the page.

Issue #2029, opened 2022-04-19
Make API Documentation More User Friendly


david-clark-1043:
# CL API Docs Review
I reached out to @mlissner via email for something to help with, and he pointed me to cleaning up the API docs as something that is not being actively worked on. I have done some preliminary research on some other open source projects and their docs to compare. This is a summary of some of those ideas. These ideas can, and should, be split out into individual issues for work, but I figured putting it all in one to start might be useful to get some high-level ideas flowing.  

Problem
Links to Other Docs
Images for Comparison
<b>Ideas</b>  
- Pagination
- Whitespace and Spacing
- Table of Contents Placement
- Table of Contents Nesting
- Getting Started Content
- Docs Project Structure

Conclusion
## Problem
The current [docs](https://www.courtlistener.com/api/rest-info/) are a single page, and (no offense) a little unwieldy. Making the docs more user friendly makes people more likely to use and support the project. More people using it means more great solutions for legal professionals that people seeking legal advice could greatly benefit from.  

## Other Examples

Here is a table of other example docs and their source code. React, Django, and Express also have at least a short guide for contributing to the docs specifically. Bootstrap less so. Bootstrap also includes its docs in the main project similar to CL, while the others have dedicated separate projects (See Point 5)

| Docs | Repo | Guides | Additional Info |
|---|---|---|---|
[React Docs](https://reactjs.org)  | [React Docs Repo](https://github.com/reactjs/reactjs.org)  | [React Contributing to the Docs](https://github.com/reactjs/reactjs.org/blob/main/CONTRIBUTING.md#guidelines-for-text) | N/A |
[Bootstrap Docs](https://getbootstrap.com/docs/4.1/getting-started/introduction/)  | [Bootstrap Docs Repo](https://github.com/twbs/bootstrap/tree/main/site/content/docs) | [Bootstrap Docs Section of Readme](https://github.com/twbs/bootstrap#documentation) | Seems to use [Hugo](https://gohugo.io/documentation/) |
[Django Docs](https://docs.djangoproject.com)  | [Django Docs Repo](https://github.com/django/djangoproject.com)  | [Django Contributing to the Docs](https://docs.djangoproject.com/en/dev/internals/contributing/writing-documentation/) | Appears to use [Read the Docs](https://readthedocs.org/) and [Sphinx](https://www.sphinx-doc.org/en/master/)  
[...]

PR #3025, opened 2023-08-15
docs(all help files): keep menu on screen

dschnelldavis:
On all help pages, affix left menu so it stays on the screen, even as the rest of the page scrolls down.

contributes to issue #2029


mlissner:
@dschnelldavis, can you share two things for this:

1. A gif of it in action (you can upload it to this PR). In general, screenshots or gifs make this kind of PR much easier to review.

2. A link to the documentation for the feature? I want to learn more! 

Thank you!

dschnelldavis:
I don't know how to make an animated .gif. Here's a .png screen shot of the feature.

Essentially, the menu scrolls with the body a little, until it reaches the top of the page, where it stays fixed as you continue to scroll the rest of the page.

dschnelldavis:
The documentation on this feature is very bad (essentially nonexistant). If you Google "bootstrap 3 fixed side menu", this is the best result:
https://stackoverflow.com/questions/19681895/how-to-create-a-sticky-left-sidebar-menu-using-bootstrap-3

Essentially, you add the following to a div around what you want to fix:
```
data-spy="affix" data-offset-top="165"
```
Which says you want to fix it, and how far to let it scroll before it becomes fixed.

Then you add a little CSS:
```
#toc.affix-top {
  position: static;
  width: 250px;
}

#toc.affix {
  position: fixed;
  top: 0;
  width: 250px;
}
```
Which describes how you want it to act before and after it's fixed.

Once I figured it out, implementing it was easy. The only hard part was tracking down the instructions for what to do. (Hence the importance of good help files. :-)

mlissner:
Hm, so I found this documentation for bootstrap:

https://getbootstrap.com/docs/3.4/javascript/#affix

It references `position:sticky`:

https://developer.mozilla.org/en-US/docs/Web/CSS/position#sticky_positioning

Looking at that and playing a bit, I was able to get the menu to stick without tweaking any JS or HTML, just by adding `position:sticky; top: 0;` to the CSS. 

Did you play with that?

dschnelldavis:
The Bootstrap docs you link to would have been helpful while I was implementing it. Nonetheless, it looks like I did it correctly.

I did play with position:sticky a little, but it didn't work for me, so I must have done something wrong.

The current solution uses the bootstrap recommended html attributes, a little CSS, and no JavaScript. So it's not pure CSS, but it's pretty close. If you'd prefer a 100% CSS solution, we can do that instead, but I don't think it will perform any better than what we've already got.

mlissner:
I think a CSS solution is better. We're going to move off of bootstrap at some point soon, so doing less with it is nice. I got it working by adding `position:sticky; top:0` to the whole sidebar. I feel like it could be done better, but it worked pretty easily?

dschnelldavis:
OK, I removed the Bootstrap specific code, and re-implemented it with pure CSS using `position:sticky; top:0`. Functionally, it works the same as before.

mlissner:
Much nicer! Merging, thank you! @flooie is doing similar work in an upcoming PR, so don't be surprised if the way you implemented this gets tweaked soon.

PR #3028, opened 2023-08-17
Update bulk data script

quevon24:
The last backups of July already differentiate between null and empty, but now the problem is that psql by default writes welcome messages or some useful information about what is being executed and if you take a look at the new backups, the first line contains the "SET" message, this is easily solved by removing the first line in order to have the header on the first line.
```

SET
id,pacer_court_id,pacer_has_rss_feed,pacer_rss_entry_types,fjc_court_id,date_modified,in_use,has_opinion_scraper,has_oral_argument_scraper,position,citation_string,short_name,full_name,url,start_date,end_date,jurisdiction,notes
"acca",,,"","","2015-01-09 00:24:40.715+00","t","t","f","113.3","A.C.C.A.","Army Court of Criminal Appeals","Army Court of Criminal Appeals","https://www.jagcnet.army.mil/ACCA",,,"FS",""
"ag",,,"","","2017-01-27 22:35:06.131+00","t","t","f","113.04","Op. Att’y Gen.","Attorneys General","United States Attorneys General","https://www.justice.gov/olc/opinions-main",,,"FS",""
```
 
But by checking the psql [documentation](https://www.postgresql.org/docs/current/app-psql.html) you can prevent psql from printing these messages by passing the -q or --quiet option, this will prevent psql from printing those messages, it will only print the output query, in this case the output is redirected to bzip2 and then this file is copied to s3 using aws cl.

These changes pass the -q option to the command responsible for generating the backup

### DIFF
--- a/scripts/make_bulk_data.sh
+++ b/scripts/make_bulk_data.sh
@@ -28,6 +28,7 @@ PGPASSWORD=$DB_PASSWORD psql \
 	       citation_string, short_name, full_name, url, start_date, end_date,
 	       jurisdiction, notes
 	   ) TO STDOUT WITH (FORMAT csv, ENCODING utf8, HEADER, FORCE_QUOTE *)' \
+	--quiet \
 	--host $DB_HOST \
 	--username $DB_USER \
 	--dbname courtlistener | \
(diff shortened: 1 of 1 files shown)


mlissner:
Great. Thanks for using the long option. It'd be nice have a note in the documentation too, perhaps near the note about nulls vs blanks?

Issue #3020, opened 2023-08-14
CSP Violation in Django REST Framework


mlissner:
I found one! 

If you load this page:

https://www.courtlistener.com/api/rest/v3/

You'll see that there's a CSP violation: 

```
10:31:38.723 Content-Security-Policy: The page’s settings blocked the loading of a resource at inline (“script-src”). v3:284:14
```

Drat. We'll want to fix this. I've noticed on the page, for example, that the OPTIONS button no longer works.

cweider:
Good news: `django-rest-framework` has the fixes already. Bad news: the latest release is a year old and these fixes didn’t make it in. Probable solution is a switch of the dependency from the PyPI package to a (pinned) git ref.

The key changes are: encode/django-rest-framework#7016, encode/django-rest-framework#8783, and encode/django-rest-framework#8784.

mlissner:
Think it's worth asking for a release? I kind of hate pinned versions.

cweider:
Yup, exactly. Lurking in their discussion group, it doesn’t look like a release will be forthcoming. From a [recent discussion in Django REST framework](https://groups.google.com/g/django-rest-framework/c/j-TkbRyiQ1o/m/ZQXaVr2PAQAJ):

> > > > Hello group!
> > > > Does anybody know when the new version of DRF which would support Django 4.2 is expected?
> > >
> > > Hi, i think DRF now support D4.2 https://github.com/encode/django-rest-framework#requirements
> >
> > Master supports it, but the latest tagged version is 3.14 and it doesn't. So my question is when we can expect a new tag (3.15 maybe?)
>
> Hi, if you need install the master/main branch you try this
> pip install https://github.com/encode/django-rest-framework/archive/refs/heads/master.zip
> if you do not want to wait for the tag 3.15
>

mlissner:
Wait, so their readme says it's supported, but the only way to get support is to use `master`??

cweider:
Evidently. I mean the readme _is_ drawn from `master` so it is consistent. 10 months and counting from last release is a problem though. In other news, the built-in schema generation has been deprecated for about as long, but, being unreleased, the deprecation notice has yet to be published!

mlissner:
OK, well, let's do it, I guess.

PR #3038, opened 2023-08-22
build(deps): update django-rest-framework for CSP

cweider:
`django-rest-framework` has recieved several patches for CSP compatibility, but unfortunately these are not included in the latest release (3.14.0) and a new release cannot be expected for a while.

As such, we must replace the semantically versioned PyPI reference with a reference to the latest commit on the master branch of the package's Git Repository (do not use `master` since it could change in unexpected and incompatible ways).

Fixes: #3020


mlissner:
Not my favorite merge, but necessary so we can have secure CSP headers. Thank you. Set for auto-merge.

cweider:
I should add, the log of commits since the latest release is at [encode/django-rest-framework@3.14.0...40eccb0](https://github.com/encode/django-rest-framework/compare/3.14.0...40eccb0).

cweider:
Ok, so the latest version of the `django-rest-framework` is _not_ a drop-in replacement. Updated change to use an earlier commit that contains the relevant CSP changes and none of the changes (affecting permissions and HTTP status codes) causing issues for tests.

Also note the addition of the `inflection` package, which unfortunately became a hard requirement for users of the built-in schema generation. Switching to `drf-spectacular` is looking more compelling!

cweider:
And the change has immaterial change to the generated spec – `listPersons` getting replaced by `listPeople`. Note: these IDs should be unique! `disclosure-typeahead` operation base name needs to be something like `DisclosureTypeaheadPerson` – a task for later.

```
% diff -u <(curl -s https://www.courtlistener.com/api/swagger/) <(curl -s http://localhost:8000/api/swagger/)
--- /dev/fd/11	2023-08-23 10:14:05
+++ /dev/fd/12	2023-08-23 10:14:05
@@ -6722,7 +6722,7 @@
     },
     "https://www.courtlistener.com/api/rest/v3/api/rest/{version}/people/": {
       "get": {
-        "operationId": "listPersons",
+        "operationId": "listPeople",
         "description": "",
         "parameters": [
           {
@@ -9060,7 +9060,7 @@
     },
     "https://www.courtlistener.com/api/rest/v3/api/rest/{version}/disclosure-typeahead/": {
       "get": {
-        "operationId": "listPersons",
+        "operationId": "listPeople",
         "description": "",
         "parameters": [
           {
```

mlissner:
I'm assuming that spec change is a fix, right, so that's good. Thanks for doing the diff. 

The rest sounds good to me. Lucky that we were able to get the CRF fixes without the other stuff.

Once again I rebased and this is set for auto-merge once tests pass.

mlissner (approved):

mlissner:
Sigh. Should have known it'd have conflicts. Mind regenerating the lock file? I should have done the dependabot stuff second, sorry.

PR #3043, opened 2023-08-23
feat(search): Define mapping for Opinion search in ES

ERosendo:
This is the first version of the Opinion document mapping, some comments are below:

this document mapping uses the following parent-child approach:

```
OpinionCluster (Parent)
   OpinionDocument (Child)
   OpinionDocument (Child)
   ...
```

This approach allows us to use `has_child` queries to search for OpinionDocuments and reuse the helper methods we implemented for the RECAP search avoiding duplication of effort.

I've included the same searchable fields we use in Solr in the **`OpinionClusterDocument`** and **`OpinionDocument`** ES mappings. I've also added the `xml_harvard` field to the prepare method of the **`text`** field to make sure this text is searchable in the opinions index.

This PR also tweaks the search template to display the cluster data and its opinions, Here's a screenshot of the new template:

image

When there's only a combined opinion, we avoid showing the header in the result list. Here's a screenshot of a cluster with just a combined opinion:

image

We're using the logic implemented in #3297 to show the first 500 characters of the opinion's text when there are no matching fragments to highlight and we exclude the text field from the search query. 

This PR also tweaks the index command to add support for opinion and opinion clusters. We can use the command like this:

```
manage.py cl_index_parent_and_child_docs --search-type o
```

### Aditional comments

I discovered a bug in the `merge_highlights_into_result` method while testing the highlight for Opinion documents. This helper adds highlighted fragments to each result of the query. However, if the highlight list includes a field and its `.exact` version, and there are no matching terms to highlight in the field, the method sometimes fails to add the returned text. This is a problem when we exclude certain fields from the query (such as `text` and `plain_text`), because we rely on this helper to show the truncated text in the frontend.

I debugged the issue and found that the problem was in the logic that merges the highlights results. This helper checks the `.exact` version of the fields and merges it with the results when highlights are found. It also uses an array called `exact_hl_fields` to track the fields that have already been merged. However, the helper was adding the name of the field to the `exact_hl_fields` array even when it did not merge the highlights. As a result, when the helper tried to add a fragment with no highlights, it skipped that step because it seemed like we'd already merged it, returning a result with missing fields.

I fixed the issue by ensuring that the helper only adds the name of a field to the exact_hl_fields array when it merges the field to the query results.


mlissner:
You're out of date with `main`, my friend. :)

cl/search/templates/includes/search_result.html
```diff
@@ -346,12 +346,12 @@ <h4>
...
     {% if type == SEARCH_TYPES.OPINION or type_override == SEARCH_TYPES.OPINION %}
-      {% if result.solr_highlights.suitNature.0 %}
+      {% if result.suitNature %}
         <div class="inline-block">
           <span class="meta-data-header">Nature of Suit:</span>
           <span class="meta-data-value">
-            <a href="?q=suitNature%3A{{ result.solr_highlights.suitNature.0|striptags|urlencode }}">
-                {{ result.solr_highlights.suitNature.0|safe }}
+            <a href="?q=suitNature%3A{{ result.suitNature|striptags|urlencode }}">
+                {{ result.suitNature|safe }}
```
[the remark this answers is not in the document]
mlissner: Hm, why *do* we need `safe` here?

mlissner:
@albertisfu, first pass on this is yours. Can you please review? I suspect if we've done things well, this should be a pretty straightforward (though huge) one. Thank you guys.

cl/search/documents.py
```diff
@@ -1070,3 +1081,578 @@ def prepare(self, instance):
...
+        if instance.date_filed is None:
+            return
+
+        if isinstance(instance.date_filed, str):
+            datetime_object = datetime.strptime(
+                instance.date_filed, "%Y-%m-%d"
+            )
+            return midnight_pt(datetime_object)
+
+        return midnight_pt(instance.date_filed)
```
albertisfu, line 1338: I believe we no longer need to convert dates to `midnight_pt `since Elasticsearch allows storing naive `datetime` objects, meaning the date is stored at midnight by default without a timezone.

cl/search/documents.py
```diff
@@ -1070,3 +1081,578 @@ def prepare(self, instance):
...
+@opinion_index.document
+class OpinionDocument(OpinionBaseDocument):
+    id = fields.IntegerField(
+        attr="pk",
+        fields={
+            "raw": fields.KeywordField(attr="pk"),
+        },
+    )
+    author_id = fields.IntegerField()
+    type = fields.KeywordField(attr="type")
```
albertisfu, line 1401: I noticed in the Solr mapping, this is a TextField that uses `text_en_splitting_cl`. Here, we have this one as a `KeywordField` and `type_text` as a `TextField`. Do we require both?

cl/search/documents.py
```diff
@@ -1070,3 +1081,578 @@ def prepare(self, instance):
...
+        model = Opinion
+        ignore_signals = True
+
+    def prepare_author_id(self, instance):
+        return getattr(instance.author, "pk", None)
+
+    def prepare_type_text(self, instance):
+        return instance.get_type_display()
+
+    def prepare_local_path(self, instance):
```
albertisfu, line 1437: On RECAP we decided to only do something like:

```
    def prepare_filepath_local(self, instance):
        if instance.filepath_local:
            return instance.filepath_local.name
```

Since checking for the existence of the file before indexing will require millions of queries to S3, we simply index the path without verifying its availability on S3. I believe this approach applies only to Opinions, given its massive index size.

cl/search/documents.py
```diff
@@ -1070,3 +1081,578 @@ def prepare(self, instance):
...
+    def prepare_cluster_child(self, instance):
+        parent_id = getattr(instance.cluster, "pk", None)
+        return {"name": "opinion", "parent": parent_id}
+
+
+@opinion_index.document
+class OpinionClusterDocument(OpinionBaseDocument):
+    court_exact = fields.KeywordField(attr="docket.court_id")
+    absolute_url = fields.KeywordField()
+    caseNameShort = fields.TextField(
```
albertisfu, line 1631: Is this also not required to be indexed in the `OpinionDocument`?

cl/search/documents.py
```diff
@@ -1070,3 +1081,578 @@ def prepare(self, instance):
...
+        attr="case_name_short",
+        analyzer="text_en_splitting_cl",
+        fields={
+            "exact": fields.TextField(analyzer="english_exact"),
+        },
+        search_analyzer="search_analyzer",
+    )
+    non_participating_judge_ids = fields.ListField(
```
albertisfu, line 1639: Also, will this one and `source` be available to be queried only on the `OpinionClusterDocument`, with no combination of child fields?

cl/search/views.py
```diff
@@ -659,6 +690,25 @@ def do_es_search(
...
             )
+            cited_cluster = add_depth_counts(
+                # Also returns cited cluster if found
+                search_data=cd,
+                search_results=paged_results,
+            )
+
+            if cd["type"] in [
+                SEARCH_TYPES.OPINION,
+                SEARCH_TYPES.RECAP,
```
albertisfu, line 701: Why RECAP and Dockets are included here?

albertisfu (dismissed):
This looks good and generally works correctly. There are some comments in the code and a few details that seem to be issues that needs a fix:

- The Precedential Status facets on the Advance Opinion Search pages appear to still be functioning with Solr.
Screenshot 2023-11-14 at 12 29 29

- The query speed is not displayed in the results header:
Screenshot 2023-11-14 at 22 20 10

- The `citation_count` is being indexed properly by signals when using `save()` (like in tests)  but checking the code I found that the place where this field is updated: https://github.com/freelawproject/courtlistener/blob/5389efe9249ce20ad8d79ffe62c211520b618a1d/cl/citations/tasks.py#L219 uses `update()` so I'm afraid that indexing signals won't be triggered for keep this field in sync, we might need to change the approach of updating that field in the code or trigger a `OpinionCluster` save() to get the signal triggered.

- It seems to be an issue related to HL and Citations, when the Citation field is HL, it looks like this:
Screenshot 2023-11-14 at 16 29 05

- Also since Citation is a list of values, it seems that the HL can be also a list:
For instance a query for
`citation:1`
Will return a list of HL like;
`"citation": ["33 state <mark>1</mark>","<mark>1</mark> Yeates 1"]`
Currently only the first HL is displayed.

- HL for the Docket Number does not seem to be working when searching for the docket number.
Screenshot 2023-11-14 at 22 35 15

It does work when using the Docket Number filter:
Screenshot 2023-11-14 at 22 37 00

- There are some tests that I think could be improved, for instance:

```
async def test_can_use_union_query(self) -> None:
    """Does OR queries work"""
    search_params = {"q": "Howard OR Lissner"}
    r = awaitself._test_article_count(search_params, 2, "union query")
    self.assertIn("Howard", r.content.decode())
    self.assertIn("Lissner", r.content.decode()
    self.assertIn("2 Opinions", r.content.decode())
```
The test is correct, however is passing because it's matching the words in the Search box, but in case something changes we won't be able to detect the results are in fact the right ones. 

Screenshot 2023-11-14 at 19 32 47

So, I think in these types of tests, we could match a different term that's not included in the search box, for instance, the docket number.

- Search phrase queries are not being highlighted.
Screenshot 2023-11-14 at 19 36 38

They do work using the case name filter.

- The Opinion Search Feeds continue working with Solr, so they don't match the ES results.

- I noticed the following new fields:
`panel_names, proceduralHistory, posture, syllabus, sha1`
Just to confirm they are required and if so we could add them to `help/search-operators/`?

cl/search/views.py
ERosendo: The `query_citation` object is used within the `search.html` template. Upon examining the original `do_search` method, I noticed that the object is computed for **opinion**, **recap**, and **docket** searches.

https://github.com/freelawproject/courtlistener/blob/9646e5aaf90306d378fd65f106e8c3bf1a13f4a7/cl/search/templates/search.html#L134-L141

https://github.com/freelawproject/courtlistener/blob/9646e5aaf90306d378fd65f106e8c3bf1a13f4a7/cl/search/views.py#L206-L213

cl/search/documents.py
ERosendo: Neither this field nor `source` appears in the list provided on the [Advanced Query Techniques and Operators page](https://www.courtlistener.com/help/search-operators/). I've added both of them because they are included in the SOLR mapping.

cl/search/documents.py
ERosendo: I've added the type_text field to display the “human-readable” value of the field in the frontend, and I included the `type` field to enable its utilization in the API response once we develop that feature.

I considered using [Multi-fields](https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html), but the values of the fields are distinct.

cl/search/documents.py
ERosendo: got it! so we just need to remove the `midnight_pt` helper method, right?

cl/search/documents.py
```diff
@@ -1070,3 +1081,578 @@ def prepare(self, instance):
...
+        },
+        search_analyzer="search_analyzer",
+    )
+    dateReargumentDenied_text = fields.TextField(
+        analyzer="text_en_splitting_cl",
+        fields={
+            "exact": fields.TextField(analyzer="english_exact"),
+        },
+        search_analyzer="search_analyzer",
+    )
```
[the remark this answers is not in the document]
ERosendo: We're just using these fields to allow users to include dates in their full-text queries. These fields were part of the original [snippet](https://github.com/freelawproject/courtlistener/blob/9646e5aaf90306d378fd65f106e8c3bf1a13f4a7/cl/search/templates/indexes/opinion_text.txt).

cl/search/documents.py
```diff
@@ -1070,3 +1081,578 @@ def prepare(self, instance):
...
+    caseName = fields.TextField(
+        analyzer="text_en_splitting_cl",
+        fields={
+            "exact": fields.TextField(analyzer="english_exact"),
+        },
+        search_analyzer="search_analyzer",
+    )
+    caseNameFull = fields.TextField(
```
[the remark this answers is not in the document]
ERosendo: This field is included in the template [file](https://github.com/freelawproject/courtlistener/blob/9646e5aaf90306d378fd65f106e8c3bf1a13f4a7/cl/search/templates/indexes/opinion_text.txt) that we were using to create the content of the text field in the SOLR mapping.

https://github.com/freelawproject/courtlistener/blob/9646e5aaf90306d378fd65f106e8c3bf1a13f4a7/cl/search/templates/indexes/opinion_text.txt#L28-L29

cl/lib/elasticsearch_utils.py
```diff
@@ -762,6 +858,42 @@ def build_es_base_query(
             string_query, join_query = build_full_join_es_queries(
                 cd, child_query_fields, parent_query_fields
             )
+        case SEARCH_TYPES.OPINION:
+            str_query = cd.get("q", "")
+            related_match = RELATED_PATTERN.search(str_query)
+            mlt_query = None
+            if related_match:
+                str_query = RELATED_PATTERN.sub(str_query, "").strip()
```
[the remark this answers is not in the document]
ERosendo: which one?

cl/search/api_views.py
```diff
@@ -11,6 +11,7 @@
     ExtendedPersonESSerializer,
     OAESResultSerializer,
     OpinionClusterSerializer,
+    OpinionESResultSerializer,
```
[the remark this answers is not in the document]
ERosendo: You're right 👍

cl/search/documents.py
ERosendo: understood!

cl/search/documents.py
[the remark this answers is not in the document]
mlissner: I think if you search for "august" and you don't get results, it's your fault and you should use a date filter.

cl/search/views.py
albertisfu: got it!

cl/search/documents.py
albertisfu: Got it, yeah in this case I think we should confirm we don't to enable these fields to be searchable in combination of child fields.

cl/search/documents.py
albertisfu: Oh I see, yeah the `type_text` is now used in the template, so it's required.

cl/search/documents.py
albertisfu: yeah!

cl/search/documents.py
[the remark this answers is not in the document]
albertisfu: oh I see, yeah, in this case is correct to include `case_name_full` too.

cl/lib/elasticsearch_utils.py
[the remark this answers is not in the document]
albertisfu: `str_query`

cl/search/documents.py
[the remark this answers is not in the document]
mlissner: Just generally, I wouldn't consider the current state of things to be the best source of truth. If we haven't done so already, I think we should consider *every* field that makes sense and decide whether we want those included. We've added a LOT of fields to this model without adding them to the search index.

albertisfu:
Ok, I've implemented the following changes and have some remaining questions:

- Added field trackers to `OpinionCluster` and `Opinion`, similar to other ES documents, to trigger update signals only when fields change.
- Introduced the `index_related_cites_fields` task to index `cites` and `citeCount` fields for Opinion documents. I noticed that these fields are updated in the `store_opinion_citations_and_update_parentheticals` task, which uses `bulk_create()` for creating cites and `update()` for updating `citeCount` on clusters. However, these methods don't trigger signals. So to keep these fields in sync, I created the `index_related_cites_fields` task. It updates the `cites` in affected Opinion documents and the `citeCount` for each affected cluster and their child documents in a single bulk request.
- Refactored the `merge_highlights_into_result` method to support highlighting for fields with multiple values. For example, the citation field can return HL for both the exact and normal versions, where each version might have multiple HL values. 
```
    "highlight": {
                        "citation.exact": [
                            "<mark>22</mark> AL 339",
                            "<mark>33</mark> <mark>state</mark> 1"
                        ],
                        "citation": [
                            "22 <mark>AL</mark> 339",
                            "1 <mark>Yeates</mark> 1"
                        ]
                    }
```
The method now extracts HL terms for each version, cleans strings, selects unique strings to avoid repetition (preferring the longer string if one contains the other), and then reintegrates the highlights into these unique strings. Issues with Unicode characters, such as emojis, have also been addressed.
Added tests for these multi-value HLs.
- Tweaked the search_result.html and ES alerts templates to support the new multi-value HL fields using a custom filter.
- Moved Opinions Feeds to ES.
  - The SearchFeed now displays `OpinionClusters` and selects the first opinion text for rendering in the template, similar to the Solr version.
  - `JurisdictionFeed` and `AllJurisdictionsFeed` display `OpinionDocuments` instead of clusters, as in the Solr version. This required adding the `absolute_url` to the `OpinionDocument` mapping so that the cluster's `absolute_url` could be displayed for each opinion.
- Added `court_id` filter for opinions and added a test.
- Made `court_id` and `suitNature` searchable by text query and added tests for these text query fields.
- Document mapping changes:
  - Disable indexing of `type`  and `type_text` fields since they're only used in the template.
  - Disable indexing of `sha1` since it's not used for searching.
  - I added the `opinion_url` fields to the OpinionDocument mapping, as we talked about today, so we can index the `opinion_url` when it's available.
 
### Some questions about Opinion document mapping:

- `caseName` and `caseNameFull` are included in both Clusters and Opinion documents, with `caseName` storing the best case name using the `best_case_name` method. 
```
if obj.case_name:
        return obj.case_name
    elif obj.case_name_full:
        return obj.case_name_full
    else:
        return obj.case_name_short
```
`caseNameShort` is only indexed in Cluster documents. Should these fields be required in both parent and child documents, or both? Considering whether it will be useful for users to query opinions by `case_name_short` or `case_name_full` combined with other cluster fields, given that `caseName` already selects the best case name.

- These are new fields added to both the `OpinionCluster` and `Opinion` document mapping:
`attorney`
`proceduralHistory`
`posture`
`syllabus`
Which were not included in Solr mapping. In Solr, they were searchable by a text query as they were included in the opinion snippet. 
Should these fields be indexed, and if so, should they be searchable? or should I just add them to the `/help/search-operators/` page so users are aware of these fields for advanced queries?

- The following fields are included in the Solr mapping but they're not mentioned in `/help/search-operators/` and they were not included in the snippet.
  - `non_participating_judge_ids`
  - `source`
Should these fields be indexed and searchable? If so, hould they be queryable in combination with child fields (opinions)? Currently, they're included only in the Cluster mapping.

- Regarding `status` and `status_exact`, both store the same value from `get_precedential_status_display()`. `status` is described in /help/search-operators/, and `status_exact` is displayed in the API. Do we need both fields?
In `/help/search-operators/` the description of `status` says: 
> The precedential status of a case. Valid entries for this field are u for Unpublished, p for published, r for relating-to, and i for in chambers opinions.
But we are not storing these values  `u, p, r, i`, should we update the documentation?
  
Finally, some methods related to citations still need to be moved to ES and should be switched once all opinions are indexed: 
`search_db_for_fullcitation`
`get_citing_clusters_with_cache`
`case_name_query`
`reverse_match`

I'll create an issue to remember to do it.

Let me know what you think.

cl/search/templates/includes/search_result.html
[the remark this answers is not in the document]
albertisfu: `safe` is required here to display highlights:
Screenshot 2023-12-19 at 18 26 51

cl/search/documents.py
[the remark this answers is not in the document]
albertisfu: yeah, I added more details and questions in the general comment about new fields.

mlissner:
Question...isn't this:

> This required adding the absolute_url to the OpinionDocument mapping so that the cluster's absolute_url could be displayed for each opinion.

The same as this:

> I added the opinion_url fields to the OpinionDocument mapping, as we talked about today, so we can index the opinion_url when it's available.

? 

> I disabled indexing of type and type_text fields since they're only used in the template.

I think we'll want those fields, actually. People should be able to search for concurrences or dissents or whatever. We'll want to document this too.

> caseNameShort is only indexed in Cluster documents. Should these fields be required in both parent and child documents, or both? Considering whether it will be useful for users to query opinions by case_name_short or case_name_full combined with other cluster fields, given that caseName already selects the best case name.

I don't see any value in caseNameShort. Let's not index that. caseNameFull has some value because it's really long and sometimes has parties in it that aren't elsewhere, so we should keep that. 

> Should these fields be indexed, and if so, should they be searchable? or should I just add them to the /help/search-operators/ page so users are aware of these fields for advanced queries?

Yes, these should be searchable and they should be documented in our blog post as well as the help page. We don't need any UI elements for them though.

> Should [the `non_participating_judge_ids` and `source`] fields be indexed and searchable? If so, should they be queryable in combination with child fields (opinions)? Currently, they're included only in the Cluster mapping.

They're small enough to be basically free, so we might as well include them. If they're on opinion objects in the model, that's where they should be in Elastic. If on the cluster in the model, then on the cluster in Elastic.

For the `status` field:

> The precedential status of a case. Valid entries for this field are u for Unpublished, p for published, r for relating-to, and i for in chambers opinions.

That documentation appears to be wrong, and I like the way it works now. It's better for users to query `status:precedential` or `status:in-chambers` than to query `status:i`. But this implies that we do some basic normalization of the status fields as we add them, so that the multi-word ones get normalized somehow. I'm not sure which is better: 

 - `status:in-chambers`
 - `status:inChambers`

I think the first? But this is no good:

 - `status:"In chambers"`

----

One last thing is that we're making some tweaks to fields that exist and to the values fields contain. I think that's OK, and these fields are boring enough we don't have to do a big post about it, BUT I think we should make sure that we update the alerts accordingly. For example, if an alert currently has `status:p`, we'll need to fix that so it continues working properly, same for `caseNameShort`, which we're eliminating, etc.

Maybe just some code to regex the queries for bad values, and spit out the admin URLs for the alerts so somebody can fix them by hand?

albertisfu:
Thanks for your answers.

> Question...isn't this:
> 
> This required adding the absolute_url to the OpinionDocument mapping so that the cluster's absolute_url could be displayed for each opinion.
> 
> The same as this:
> 
> I added the opinion_url fields to the OpinionDocument mapping, as we talked about today, so we can index the opinion_url when it's available.
> 
> ?

`absolute_url` is the cluster URL included in each `OpinionDocument`. This allows the display of the cluster URL in the `JurisdictionFeed`, which shows opinions instead of clusters.

Additionally, `opinion_url` is a new field to store the unique URL for each opinion. This will enable linking to individual opinions nested within cluster results on the front end once we have separate URLs for each opinion in the future.

 Is that correct?


> I disabled indexing of type and type_text fields since they're only used in the template.
> 
> I think we'll want those fields, actually. People should be able to search for concurrences or dissents or whatever. We'll want to document this too.
> 

Got it, so I'll enable indexing for these fields. Both of them right? So users can query by a `type` that holds a value like "010combined" or the human-readable value "Combined Opinion", correct?

> I don't see any value in caseNameShort. Let's not index that. caseNameFull has some value because it's really long and sometimes has parties in it that aren't elsewhere, so we should keep that.

Great, I'll remove `caseNameShort`.

> Yes, these should be searchable and they should be documented in our blog post as well as the help page. We don't need any UI elements for them though.

Got it, I'll add support for these fields `attorney`, `proceduralHistory`,`posture`,`syllabus` to be searchable in text queries and also add them to the documentation.

> They're small enough to be basically free, so we might as well include them. If they're on opinion objects in the model, that's where they should be in Elastic. If on the cluster in the model, then on the cluster in Elastic.

Correct, so I'll keep these fields: `non_participating_judge_ids`, `source` they're cluster fields and they are currently only added to the `OpinionClusterDocument` in ES. Just confirm that in case a user wants to perform a query that involves one of these fields and an `Opinion` field like the following:

`source:"CR" AND cites:"1234"` It won't return results since `source` won't be indexed into each `OpinionDocument`. Is that not necessary for these two fields?

> That documentation appears to be wrong, and I like the way it works now. It's better for users to query status:precedential or status:in-chambers than to query status:i. But this implies that we do some basic normalization of the status fields as we add them, so that the multi-word ones get normalized somehow. I'm not sure which is better:
> 
> status:in-chambers
> status:inChambers
> I think the first? But this is no good:
> 
> status:"In chambers"
> 

Got it, yeah we're currently indexing the multi-word ones also in Solr, so I'll change it.
About the normalization, I think we could index the one-word values that are used in the model:
```
class PRECEDENTIAL_STATUS:
    PUBLISHED = "Published"
    UNPUBLISHED = "Unpublished"
    ERRATA = "Errata"
    SEPARATE = "Separate"
    IN_CHAMBERS = "In-chambers"
    RELATING_TO = "Relating-to"
    UNKNOWN = "Unknown"
```
So we don't need to migrate these values in DB or make a translation before indexing, does that sounds good?

> One last thing is that we're making some tweaks to fields that exist and to the values fields contain. I think that's OK, and these fields are boring enough we don't have to do a big post about it, BUT I think we should make sure that we update the alerts accordingly. For example, if an alert currently has status:p, we'll need to fix that so it continues working properly, same for caseNameShort, which we're eliminating, etc.
> 
> Maybe just some code to regex the queries for bad values, and spit out the admin URLs for the alerts so somebody can fix them by hand?

Yeha, regarding `status`, we'll need to update the values in the `SearchForm` and also check alerts to update the status values, so they continue working. 

As for `caseNameShort`, it seems unlikely that we have alerts incorporating this value, given that this field was neither documented nor included in filters. However, I will add it to the regex to confirm the absence of alerts with this value.

I will create an issue to identify alerts that require updates and will manually update them.

mlissner:
For opinion_url vs absolute_url (cluster URL), you can just do one field called the absolute_url. For now, that's the same as the cluster URL, but once opinions have their own URLs, we can update it to be better. That should work, right? 

> Both of them right? So users can query by a type that holds a value like "010combined" or the human-readable value "Combined Opinion", correct?

Nah, I don't see the point of that. Could we do the same thing we do with the status field and provide a hyphenated query for users and put the possible values in the documentation? So: `type:concurrence` or `type:combined-opinion`? 

> Is that not necessary for these two fields?

I think that's OK.

> So we don't need to migrate these values in DB or make a translation before indexing, does that sounds good?

Great.

Thank you!

albertisfu:
Thanks, I've applied the requested changes here.

- Removed `opinion_url` from the `OpinionDocument` mapping, only `absolute_url` remained.
- Made `type` an indexed field and translated its values to be indexed as:
```
{
    Opinion.COMBINED: "combined-opinion",
    Opinion.UNANIMOUS: "unanimous-opinion",
    Opinion.LEAD: "lead-opinion",
    Opinion.PLURALITY: "plurality-opinion",
    Opinion.CONCURRENCE: "concurrence-opinion",
    Opinion.CONCUR_IN_PART: "in-part-opinion",
    Opinion.DISSENT: "dissent",
    Opinion.ADDENDUM: "addendum",
    Opinion.REMITTUR: "remittitur",
    Opinion.REHEARING: "rehearing",
    Opinion.ON_THE_MERITS: "on-the-merits",
    Opinion.ON_MOTION_TO_STRIKE: "on-motion-to-strike",
}
```
Instead of the original values used in the model:
```
    COMBINED = "010combined"
    UNANIMOUS = "015unamimous"
    LEAD = "020lead"
    PLURALITY = "025plurality"
    CONCURRENCE = "030concurrence"
    CONCUR_IN_PART = "035concurrenceinpart"
    DISSENT = "040dissent"
    ADDENDUM = "050addendum"
    REMITTUR = "060remittitur"
    REHEARING = "070rehearing"
    ON_THE_MERITS = "080onthemerits"
    ON_MOTION_TO_STRIKE = "090onmotiontostrike"
```
- Removed `caseNameShort` from OpinionDocument mapping.
- Made `attorney, proceduralHistory, posture, syllabus` searchable in text queries and added related tests.
- Updated `status` values, now the indexed values are:

```
    PUBLISHED = "Published"
    UNPUBLISHED = "Unpublished"
    ERRATA = "Errata"
    SEPARATE = "Separate"
    IN_CHAMBERS = "In-chambers"
    RELATING_TO = "Relating-to"
    UNKNOWN = "Unknown"
```
instead of as in Solr.
```
        PUBLISHED, "Precedential"
        UNPUBLISHED, "Non-Precedential"
        ERRATA, "Errata"
        SEPARATE, "Separate Opinion"
        IN_CHAMBERS, "In-chambers"
        RELATING_TO, "Relating-to orders"
        UNKNOWN, "Unknown Status"
```
This required updating the `SearchForm` so that the ES version uses the new values for the status filter. Once we switch Opinion Alerts to work with ES, we will need to look for the old status values and change them to the new ones, ensuring that alerts continue to work correctly. 
Additionally, if we switch to ES Opinion Search before enabling Alerts to work with ES, we will need to introduce a method that takes the new status values in the query and translates them to the old ones, so Alerts that work with Solr can match the documents.

- Documented the new fields in the `help/search-operators/` page:
Screenshot 2023-12-26 at 11 15 04
The `source` field is a long one since it has many values.

albertisfu:
The process to release this one is the following:

Before merging to main:
Set the number of shards and replicas for the Opinion Search index env vars:

`ELASTICSEARCH_OPINION_NUMBER_OF_SHARDS`
`ELASTICSEARCH_OPINION_NUMBER_OF_REPLICAS`

Then merge and create the index:

`search_index --create --models search.OpinionCluster`

If it throws an error `resource_already_exists_exception` it needs to be removed and recreated.

`search_index --delete --models search.OpinionCluster`
`search_index --create --models search.OpinionCluster`

To start the indexing:

`manage.py cl_index_parent_and_child_docs --queue celery --search-type o`

Default values:
`--chunk-size` by default is 100.
`--pk-offset` 0 it'll start from the first opinion.

The `o-es-active` flag should be set as `Everyone: Unknown` so the ES version is only visible to Superusers.

mlissner:
There's a LOT here, much more than I can look at. @ERosendo I'm going to put this on your stack for when you have a moment, and we'll have to get it started when I get back (or perhaps I'll have time to pay attention from a beach). 

Alberto, thanks for the huge PR. I thought opinions would be easy! :sob:

cl/search/templates/includes/search_result.html
```diff
@@ -397,14 +397,14 @@ <h4>
     {% for doc in result.child_docs %}
       {% with doc=doc|get_attrdict:"_source" %}
         <div class="col-md-offset-half">
-          {% if result.child_docs|length > 1 or doc.type != '010combined' %}
+          {% if result.child_docs|length > 1 or doc.type != 'combined-opinion' %}
             <h4>
               <p>{{ doc.type_text }}</p>
```
ERosendo, line 402: We should use the new `render_string_or_list` filter here, right?

cl/lib/elasticsearch_utils.py
```diff
@@ -7,29 +7,46 @@
 from dataclasses import fields
 from datetime import date, datetime
 from functools import reduce, wraps
-from typing import Any, Callable, Dict, List, Literal, Match
+from typing import Any, Callable, Dict, List, Literal, Tuple
```
ERosendo, line 10: We should remove the `Tuple` import, we can use the built-in collection type `tuple` as a generic type instead

mlissner: I haven't been mentioning this, but we should be doing this any time we're adding type hints. I look forward to the day when all of the type imports are removed.

cl/search/views.py
```diff
@@ -441,7 +465,9 @@ def show_results(request: HttpRequest) -> HttpResponse:
                 )
 
             # But give it a fresh form for the advanced search section
-            render_dict.update({"search_form": SearchForm(request.GET)})
+            render_dict.update(
+                {"search_form": SearchForm(request.GET)}, request=request
```
ERosendo, line 469: is this right?

I noticed the `searchForm` class accepts a `request` object now and you've been updating several lines to accommodate this change but this line is different, we're adding the request to the render_dict object instead.

albertisfu: yeah, this was a mistake, fixed it. Thanks for noticing it.

albertisfu:
I've implemented #3586 solution here for ES. So this is ready for a new review.

mlissner:
@ERosendo will you take another look here too, please?

ERosendo:
@mlissner Sure!

cl/settings/third_party/elasticsearch.py
mlissner, line 1: I've set this variables to be 30 shards and 1 replica.

mlissner (approved):
OK, all of the lines in this PR have at least passed by my eyeball. It's a monument of work, congratulations, Alberto and Eduardo — a veritable Chichen Itza in its hugeness. 

If you and Eduardo are ready to merge, let's do this. I can't wait to launch opinions.

albertisfu:
Ok, the failing tests are solved now.

The Solr error was significant since it could disrupt the Solr search.

The issue arose because `get_selected_field_string` was modified in this PR, and it no longer enclosed courts in quotes for the Solr query. Previously it was `"canb" OR "nysd"`, after the change, it became `canb OR nysd`

And the new method for expanding courts for the Solr version, `get_child_court_ids_for_parents`, searches for courts within quotes. Since it didn't find any courts, it passed an empty `court_exact` `fq` query to Solr, which always resulted in a 404 error.

The issue has now been fixed by wrapping again courts in quotes.

**Also:**
- I've added term vectors to all the fields that need to be highlighted for Opinions. Initially, I only added it to `text`, but as observed in one benchmark, it appears that a vector field using `plain` highlighting is faster than using FVH. These fields' values are not too long, so I assume the performance difference won't be too significant.

- Added the following settings:

```
ELASTICSEARCH_CLUSTERS_SIGNALS_ENABLED
ELASTICSEARCH_OPINIONS_SIGNALS_ENABLED
```

By default set to `False`.
So, enabling indexing signals for opinions will be necessary in K8s.

The process for deploying this to production is similar to that of other documents.

Once this is deployed it's required to create the Opinion index.
`search_index --create --models search.OpinionCluster`

If it throws an error resource_already_exists_exception it needs to be removed and recreated. It's supposed this error won't be triggered since signals are disabled by default.

`search_index --delete --models search.OpinionCluster`
`search_index --create --models search.OpinionCluster`

Then we can index Clusters and Opinions.

`manage.py cl_index_parent_and_child_docs --queue celery --search-type o`
`--chunk-size` by default is 100.

And enable indexing signals via the settings.

mlissner:
I'm merging this one because it's certainly mature enough and I want to start the indexer tonight. 

Let's finish one last review of it though, and fix any last tweaks before moving to the next thing. Thank you all! This it's a big one.

mlissner:
This is deployed. I'm now running it with: 

```
./manage.py cl_index_parent_and_child_docs --queue etl_tasks --search-type o --chunk-size 25 --auto-resume
```

PR #3056, opened 2023-08-25
chore(clone_from_cl): Correct spelling in clone_from_cl

cweider:
`search.Opinion` is not a valid type; replace it with `search.OpinionCluster`.


mlissner:
Automerge set, thanks.

PR #3058, opened 2023-08-25
feature(javascript): automate Javascript asset compilation

cweider:
In the absense of an asset build process, compilation and compression of Javascript must be done manually, during development, and the output checked into the repository. This impedes development, clutters reviews, and is an invitation for mismatches between source and compiled assets.

This branch first replaces the manual compiled assets with a phase in the docker build and ad-hoc webpack server with a Docker container, `cl-webpack`, that watches and recompiles assets in development. Then, once done, with building of production assets separated from development assets, the non-production compilations can be made far more legible. In the later patch, development JS is produced unminified and with source maps.


cweider:
To verify, it’s helpful to disable the `cl-webpack` container (see [this example](https://github.com/cweider/courtlistener/commit/71c7282c24e3c903cb6287eb00abf09675f73a93)). Also, to compare current and local versions these commands will do:
```
% diff -u <(curl -s https://storage.courtlistener.com/static/js/react/main.0217b10f42b9.js) <(curl -s http://localhost:8000/static/js/react/main.js)
% diff -u <(curl -s https://storage.courtlistener.com/static/js/react/vendor.ec95755e7d94.js) <(curl -s http://localhost:8000/static/js/react/vendor.js)
```

cweider:
Also, this patch removes `cl/assets/static-global/js/react/*` from source control, but it doesn’t remove them from the working copy (the `cl-webpack` will create them as a matter of course). Sometimes, when transitioning between branches with and without this change, `git` will stumble. `.gitignore` helps this, but doesn’t resolve the trouble.

During the transition, contributors should know that removing the untracked files (`rm cl/assets/static-global/js/react/*`) before checking out old branches will fix their issue. One thing they should _not_ do is restore these to the repository (`git add cl/assets/static-global/js/react/*`) – accidental resurrection of the files is something to watch for in code review.

docker/django/Dockerfile
```diff
@@ -76,6 +78,13 @@ RUN tar xvfz courtlistener.tar.gz && \
 # `python-base` sets up all our shared environment variables
 FROM build-${BUILD_ENV} as python-base
 
+# Collect compiled assets from webpack-build stage.
+# Note: Always produce production code and never produce source-maps
+# (run the devMiddleware for legible output)
+WORKDIR /opt/courtlistener/cl
+RUN npm install
+RUN npx webpack --mode=production --no-devtool
+
```
cweider, line 87: Note: You can see what that would look like in the alternative branch, [javascript-webpack-staged](https://github.com/cweider/courtlistener/commit/8207c6744d8ce74f713f45593655eeacb840ae39)’s version of the [Dockerfile](https://github.com/cweider/courtlistener/commit/8207c6744d8ce74f713f45593655eeacb840ae39#diff-10aefcba0ca0bff173d169c2b0368a7b669159cd9d7b565a9d7ccc020919d41d). Unfortunately, this is not compatible with the repository unarchiving approach used in the `build-prod` step – which would need the change [replace use of repository archive during build-prod](https://github.com/freelawproject/courtlistener/commit/518b7ef542b5620dccbc2bdb09ee9a0e68d19a5a).

mlissner (approved):
This looks good to me and I trust your work. Let's give this a go, and we'll know very quickly if we have an issue.

Thank you.

PR #3073, opened 2023-08-30
chore(docker): use human-readable user name for `celery`

cweider:
Reviewing `Dockerfile.yml`, I thought to myself, who is user 33? The answer is that it is `www-data`:

```
% docker exec -it cl-celery id
uid=33(www-data) gid=33(www-data) groups=33(www-data)
```

This is the same user specified for `rss-scraper` and `retry-webhooks`, which both use the user name. There's no reason for `celery` to take a different approach.


mlissner:
You make a convincing case, but I had *reasons* for doing this years ago when I did. We shall see what kinds of trouble this causes! :)

PR #3106, opened 2023-09-05
chores(workflows): revise docker/build-push-action caching.

cweider:
Experimental support for gha (GitHub Actions) caching was introduced two years ago by moby/buildkit@b9d7315. While it continues to be marked as experimental, it should be stable enough for our purposes.


mlissner:
Nice to see the hack getting removed and things getting better. I know y'all are still testing this, but I'm going to merge because it's better already. Hopefully it fixes the "no space" on disk issue too.

PR #3107, opened 2023-09-05
chore(build): use www-data user for all services

cweider:
It is a best practice for production processes to not run at `root` (also for `root` to be inaccessible via `sudo`). Docker's best practices emphasize this:

> If a service can run without privileges, use `USER`
> to change to a non-root user.

Typically, binding to port `80` requires special privileges and presents a problem, but both `web-dev` and `web-prod` bind on `:8000`; Uniformly use `www-data` user for all Docker images/containers.


mlissner:
Makes sense to me. Let's see if it works.

PR #3111, opened 2023-09-06
feat(build): eliminate latest-web-dev image

cweider:
`docker-compose` and `docker compose` are substantially different. The former corresponds with "Compose V1", which is deprecated, and the later is a rewrite that, significantly, uses the same build system as the other `docker/*` actions and, thereby, the same cache.

Once done, it became apparent that the development images are unnecessary.
- Pushing to `localhost:5000` in workflows isn't needed, because images produced by earlier steps are now available in the `buildx` cache primed by the `build-push-action` and used by `docker compose`.
- With `localhost:5000` removed, `DJANGO_DOCKER_IMAGE` and `CELERY_DOCKER_IMAGE` environment variables are no longer used.
- With no consumption of development images outside of the local environment, there's no reason to publish these images or have a `Makefile` targets to build them.

This change switches from `docker-compose` (V1) to `docker compose` (V2) and changes the configuration of the `cl-django` and `cl-celery` to replace `image` with `build`. It then removes all development image related code, including:

- `.github/workflows/docker-build-dev.yml` workflow
- local registry in the `.github/workflows/tests.yml`
- `development` targets of `docker/django/Makefile`

For debugging it also separates the docker operations of `pull` and `build` into separate steps in the `tests.yml` workflow.

After this change, where implementers would run:
```
% make development --file docker/django/Makefile VERSION=$RANDOM
```
They should use `docker compose` with the `--build` argument:
```
docker compose up --build
```


mlissner:
Looks great. Want to update the wiki too?

PR #3117, opened 2023-09-07
fix(corpous.tests): Stop network requests in test

flooie:
Patches Get Fix List method that was making http requests that could 
cause test to fail for other PRs.


flooie:
Was waiting for tests to pass.  Looks like it-

mlissner:
Thanks. Merged. Don't forget to auto-close the related issues if applicable. :)

PR #3110, opened 2023-09-06
fix(build): use manual installation for python Poetry

cweider:
The Docker image, `python:3.11-slim`, that CourtListener are built from, recently stopped working with the installation script that is downloaded and executed for the Poetry tool:
```
RUN curl -sSL https://install.python-poetry.org/ | python3 -
```
Instead of this script, use the installation with `pip` that is a alternative approach which doesn't have this problem and is more direct than the script.

Also in this branch:
- Removal of surplus `POETRY_HOME` argument
- Remove nonoperative `VENV_PATH` environment variable

Fixes: #3079

PR #3124, opened 2023-09-08
fix(build): link to poetry in /usr/local/bin/

cweider:
After #3110, which manually installed Poetry using `pip`, the binary was no longer in `$PATH` and needed to be addressed directly. This is fine so far as the Dockerfile is concerned, but it is inconvenient when an implementer wants to interact with `poetry` using `docker exec`.

Resolve the issue by creating a `poetry` symlink accessible via `$PATH`.


mlissner:
Great. Auto-merge set, thank you.

PR #2118, opened 2022-06-13
feat(workflows): Add build and push workflow

flooie:
Add docker build and push code for main branches if current version is not on docker hub.

PR #3130, opened 2023-09-10
chore(build): drop use of github archive and merge build-* stages

cweider:
When the archive download approach was introduced (in https://github.com/freelawproject/courtlistener/pull/2002), both image building and deployment were done locally and there was a risk that an unclean working copy could cause WIP code to be deployed.

Today, responsibility for building (https://github.com/freelawproject/courtlistener/pull/2118) and deployment (https://github.com/freelawproject/courtlistener/pull/2169) have moved to GitHub Actions. This is the best practice and has made the archive download safeguard unnecessary.

Change the `build-prod` target to use the Docker build context rather than a downloaded copy of the `courtlistener` repository. Once done, the only remaining difference between the two targets is the `--without dev` option. Eliminate this difference by using `$BUILD_ENV` for the `RUN` instruction and merging the, otherwise identical, stages.


cweider:
Indirectly, addresses #2730 and #2736.

mlissner:
Yeah, that makes sense to me. Merging, thank you. Removing code, imagine that!

PR #3133, opened 2023-09-11
chore(html): update OpenGraph og:image

cweider:
- Use updated, flat design
- Increase resolution
- Include margins

| Before | After |
|--------|--------|
| og_before | og_after |


mlissner:
Thanks. This looks great to me.

PR #3142, opened 2023-09-13
add user to dockercompose file

ss108:
This settings change tells Docker to run the Django app as root, which is needed after the `www-data` change to have the container play more nicely with IDEs (in particular, VSCode, which @ERosendo and I use, but I also ran into the same issue trying it with Fleet).

### DIFF
--- a/docker/courtlistener/docker-compose.yml
+++ b/docker/courtlistener/docker-compose.yml
@@ -104,6 +104,7 @@ services:
       - cl-es
     ports:
       - "8000:8000"
+    user: root
     volumes:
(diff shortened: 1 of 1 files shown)


ss108:
@cweider Can you approve when you get a chance plz

ERosendo:
This works for me. I was able to attach VSCode to the container and used the `pre-commit run --all-files`command.

mlissner:
LGTM, but I'm happy to let Chad make the final determination.

cweider (approved):

ss108:
@mlissner This is ready to merge

mlissner:
Thanks all!

PR #3156, opened 2023-09-16
Delete reference to Amazon Smile

ToaKraka:
Amazon Smile was discontinued in February 2023.


mlissner:
Thanks! This stuff helps.

PR #3158, opened 2023-09-16
chore(workflows): harmonize triggering of Semgrep and CodeQL

cweider:
The Semgrep and CodeQL workflows have similar purposes, but different triggers. A closure looks reveals that these differences are incidental and that each would benefit from borrowing from the other.

- In 475587, CodeQL was made to run on a schedule only. Only running on `main` and not blocking pull requests is reasonable, but schedule alone is a little odd: Semgrep, a similar task runs `push` and `pull_request`.
- In 52b8f5, Semgrep was made to always checkout `main`. The effect of this rule would be that `main` would be checked any time there was a `push` or `pull_request` (i.e. frequently). This is better expressed by a (sufficiently frequent) schedule and `push` or `pull_request` tied specifically to `main`.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,6 +1,12 @@
 name: "CodeQL"
 
 on:
+  push:
+    branches:
+      - 'main'
+  pull_request:
+    branches:
+      - 'main'
   schedule:
     # Execute every day at 2:00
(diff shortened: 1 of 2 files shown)


mlissner:
We can try this. I think the problem with codeql was that it was too slow and was blocking merges without providing much value.

Issue #3152, opened 2023-09-14
Error importing opinions.csv into Postgres: unterminated CSV quoted field


taylorkangbeck:
I'm trying to import the ~200GB of opinions data downloaded from the S3 bucket into my local Postgres database using a COPY statement in psql:

```
\copy public.search_opinion (id, date_created, date_modified, author_str, per_curiam, joined_by_str,
               type, sha1, page_count, download_url, local_path, plain_text, html,
               html_lawbox, html_columbia, html_anon_2020, xml_harvard,
               html_with_citations, extracted_by_ocr, author_id, cluster_id) FROM 'opinions-2023-08-31.csv' WITH (FORMAT csv, ENCODING utf8, header);
```
and it fails on a particular row with the following error:
```
ERROR:  unterminated CSV quoted field                                                                                                                                                   
CONTEXT:  COPY search_opinion, line 507903410: ""4544092","2020-06-25 17:13:53.973335+00","2022-03-09 10:53:56.197499+00","","f","","010combined","7..."
```
I encountered the same error (on the same item with id `4544092`) trying to import the `opinions-2023-07-31.csv` as well. It looks like there's an extra quote in the id field. 

This may be a problem with the data that needs to be fixed. If I'm just doing something wrong or there's an additional option I should pass to COPY FROM, please let me know!

mlissner:
Hey Taylor, sorry about that. We use the COPY TO command to make the CSV, so it *should* be able to do a round trip like you're attempting. One of our developers recently was working with this data, and I believe he got it working. We should document this better. 

@quevon24 do you think you can help? Do you know the commands you used with your `COPY TO`?

taylorkangbeck:
Quick update on this, it seems this problem was specific to `\copy`, which for some reason fails to handle escape characters within field text bodies. Using a regular `COPY FROM` worked. Using hexedit I found that the html of a particular opinion contained `\.` which was causing the error.

taylorkangbeck:
Closing this issue since this is not a problem with courtlistener

mlissner:
That's very interesting, thank you. Did you consider filing it with postgres?

quevon24:
> Quick update on this, it seems this problem was specific to `\copy`, which for some reason fails to handle escape characters within field text bodies. Using a regular `COPY FROM` worked. Using hexedit I found that the html of a particular opinion contained `\.` which was causing the error.

[...]

PR #3174, opened 2023-09-22
Update bulk script

quevon24:
- Bulk script refactored to make it easier to maintain.
- CourtHouse table added to bulk script.
- OriginatingCourtInformation table added to bulk script. (As i mentioned in issue #3152, OriginatingCourtInformation table is referenced in the docket table.)
- Court appeals_to through table added to bulk script.
- Update court table in bulk script.
- Update bulk data release notes.
- Generate shell script to load bulk data and stream the script to S3. [(Example of generated file here!)](https://gist.github.com/quevon24/9671e4e1dec915176106f1d67082664b)


mlissner (changes requested):
Code looks good, but can you add something to the release notes, please? 

Also, in the Table of Contents it says "Old Formats" instead of "Release Notes" could you tweak that please?

mlissner (changes requested):
This looks pretty good to me. One thing I'd suggest is that bash has primitive for loops. Did you try using those instead of repeating the same code a bunch of times?

quevon24:
@mlissner I updated the bulk script to reduce a large part of the code to use loops to stream the data to S3, but also to generate the script to load the bulk data

mlissner (approved):

mlissner:
Thanks Kevin. We missed the bulk generation this month, but we need a howto for this file. What do you think about creating a new discussion topic with a howto and then linking to that from the bulk data documentation? 

Should we wait a month to do that after the bash script has been generated?

quevon24:
Yes, no problem, so people can know how to update the script correctly.

I will continue to investigate to see if something else can be done for the script.

mlissner:
Sorry, what are you investigating regarding the script?

quevon24:
> Sorry, what are you investigating regarding the script?

I mean seeing if the script can be improved even more

quevon24:
> What else did you have in mind? Unless we discover bugs, it seems good to me. Can you set a reminder to do the remaining pieces early next month?

Well, currently we specify the column names, maybe we could just simply copy all columns(I am still thinking the pros and cons of this), which will reduce the size of the arrays in the script.

We could also use the table name as the bz2 filename, e.g. search_docket-2023-08-31.csv.bz2 instead of dockets-2022-08-02.csv.bz2, that also will reduce the variables in the file and reduce the array size.

This:
`declare -a t_4=("search_docket" "$docket_fields" "$dockets_csv_filename")`

will become this:
`declare -a t_4=("search_docket")`

That way we will copy all columns, and use the name of the table as the file name

mlissner:
Gotcha, yeah, those would certainly make the script easier. I think there are column names I excluded though for some reason or other, and, well, we've got the complicated code now, so we might as well stick with it! 

I think we can mark this as done and move to the next thing (but remember to do documentation in a month!)

Issue #3175, opened 2023-09-23
Todo: Add Courthouse Table to Search Model


flooie:
Need to add a courthouse table to search to enable location / state data for courts. 

Can't believe it's been this long and we dont actually know where the Supreme Court of California physically is.

flooie:
Oh - add new jurisdiction ... Federal Military.

flooie:
phew.

PR #3184, opened 2023-09-25
3175 courthouses - Add Courthouse Table to search

flooie:
PR Resolves the following Issues 

Address #3175

This PR
 - Creates Courthouse table in search model
 - Updates Court Model to
    - Add appellate path in the Court table
    - Add Parent/Child relationship for Court table 
       - NOTE: This will allow for grouping courts together that are of the same kind or of the same parent court (ie Boston Municipal Court - is a Massachusetts District Court)
    - Adds two new jurisdictions (Military Trial/ Military Appellate)


flooie:
@mlissner I'm of the mind to ignore the semgrep warning here- as being inconsistent with our normal practice but I could change it.  Otherwise- I think it should be ready for you.

cl/search/models.py
```diff
@@ -1931,6 +1932,9 @@ def tribal_courts(self) -> models.QuerySet:
     def territorial_courts(self) -> models.QuerySet:
         return self.filter(jurisdictions__in=Court.TERRITORY_JURISDICTIONS)
 
+    def military_courts(self) -> models.QuerySet:
+        return self.filter(jurisdictions__in=Court.MIL)
```
mlissner, line 1936: Shouldn't this be:

```suggestion
        return self.filter(jurisdictions__in=Court.MILITARY_JURISDICTIONS)
```

cl/search/models.py
```diff
@@ -2013,13 +2021,31 @@ class Court(models.Model):
...
+        related_name="subcourts",
+        help_text="Parent court for subdivisions",
+    )
+    appellate_courts = models.ManyToManyField(
+        "self",
+        blank=True,
+        symmetrical=False,
+        related_name="appellate_courts_from",
+        help_text="Appellate courts for this court",
+    )
```
mlissner, line 2048: Would this be more clear if the name were `appeals_to` and then related could be `appeals_from`?

cl/search/models.py
```diff
@@ -2144,6 +2170,77 @@ class Meta:
         ordering = ["position"]
 
 
+class Courthouse(models.Model):
```
mlissner, line 2173: For the various fields here, where possible, we should make them more similar to the fields in `AttorneyOrganization` (which also has address information). A few tweaks:

 - Use `USPostalCodeField` and `USZipCodeField`
 - Change the name of the fields to match the fields in `AttorneyOrganization` (address1 instead of address_line_1, etc.)
 - Use TextFields — They perform the same in postgresql and have unlimited length.

And semgrep is right. We shouldn't make text or char fields nullable. We do it sometimes, when we need uniqueness constraints, but very rarely. What makes you think it's our normal practice?

mlissner (changes requested):
A few comments, thank you!

cl/search/models.py
flooie: yes.

cl/search/models.py
flooie: yep.

cl/search/models.py
flooie: Thanks.  I'll switch zip code over - but on second pass thru I think we should actually leave the state codes as is and add in the following options 

OBSOLETE_STATES = (
    ('CM', _('Commonwealth of the Northern Mariana Islands')),  # Is now 'MP'
    ('CZ', _('Panama Canal Zone')),  # Reverted to Panama 1979
    ('PI', _('Philippine Islands')),  # Philippine independence 1946
    # Became the independent COFA states + Northern Mariana Islands 1979-1994
    ('TT', _('Trust Territory of the Pacific Islands')),
)
that is not part of the `USPS_CHOICES` and we have 195 opinions in the Panama Canal Zone in the db.

mlissner: Fun. I think you should be able to split the baby with something like:

```
state = USPostalCodeField(choices=USPS_CHOICES + OBSOLETE_STATE_CHOICES)
```

flooie:
@mlissner thank you for the suggestions.  I believe this should address them.

cl/search/models.py
```diff
@@ -2144,6 +2171,70 @@ class Meta:
...
+        db_index=True,
+        choices=USPS_CHOICES + OBSOLETE_STATES,
+        blank=True,
+    )
+    zip_code = USZipCodeField(
+        help_text="The zip code for the organization, XXXXX or XXXXX-XXXX "
+        "work.",
+        db_index=True,
+        blank=True,
+    )
```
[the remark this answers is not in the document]
flooie: thanks semgrep

cl/search/models.py
mlissner: Looks like you didn't include this change.

cl/search/models.py
```diff
@@ -2144,6 +2171,72 @@ class Meta:
...
+    county = models.TextField(
+        help_text="The county, if any, where the court resides.",
+        blank=True,
+    )
+    state = USPostalCodeField(
+        help_text="The two-letter USPS postal abbreviation for the "
+        "organization w/ obsolete state options.",
+        db_index=True,
+        choices=USPS_CHOICES + OBSOLETE_STATES,
+        null=True,
```
mlissner, line 2217: Why is this still nullable?

cl/search/models.py
flooie: There must be a ghost in git.  I made those changes but will update again

cl/search/models.py
```diff
@@ -2144,6 +2171,72 @@ class Meta:
...
+        blank=True,
+    )
+    zip_code = USZipCodeField(
+        help_text="The zip code for the organization, XXXXX or XXXXX-XXXX "
+        "work.",
+        db_index=True,
+        blank=True,
+        null=True,
+    )
+    country_code = models.CharField(
```
mlissner, line 2227: Wouldn't you know it, localflavor has a useful variable for this that we can put into a `choices`. Check out ISO 3166: https://django-localflavor.readthedocs.io/en/latest/generic/#data

cl/search/models.py
```diff
@@ -2144,6 +2171,72 @@ class Meta:
...
+        choices=USPS_CHOICES + OBSOLETE_STATES,
+        null=True,
+        blank=True,
+    )
+    zip_code = USZipCodeField(
+        help_text="The zip code for the organization, XXXXX or XXXXX-XXXX "
+        "work.",
+        db_index=True,
+        blank=True,
+        null=True,
```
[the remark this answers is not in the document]
flooie: Semgrep suggested it becuase it's a char field not a text.

cl/search/models.py
flooie: I looked for it. But I guess I wouldn't as I failed to find it last week. I'll add that in.

cl/search/models.py
```diff
@@ -2013,13 +2022,31 @@ class Court(models.Model):
...
         max_length=15,  # Changes here will require updates in urls.py
         primary_key=True,
     )
-
+    parent_court = models.ForeignKey(
+        "self",
+        on_delete=models.SET_NULL,
+        blank=True,
+        null=True,
+        related_name="subcourts",
```
mlissner, line 2040: Does `child_courts` make more sense here?

cl/search/models.py
```diff
@@ -2144,6 +2171,72 @@ class Meta:
...
+    """A class to represent the physical location of a court."""
+
+    court = models.ForeignKey(
+        Court,
+        help_text="The court object associated with this Courthouse.",
+        related_name="courts",
+        on_delete=models.CASCADE,
+    )
+    court_seat = models.BooleanField(
+        help_text="Is this the seat of the Court.",
```
mlissner, line 2184:
```suggestion
        help_text="Is this the seat of the Court?",
```

mlissner (changes requested):
A few more comments. Two more:

 - Why did you add indexes to a bunch of the fields?
 - In general, the `help_text` field should be the first argument to a model field (or second when it's a FK or M2M field).

cl/search/models.py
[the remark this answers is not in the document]
mlissner: Semgrep is sort of right, but it shouldn't have triggered. It says that:

> null=True should be set if blank=True is set on non-text fields.

But this is a text field (it's just in disguise).

flooie:
When I looked at the attorney address model that you pointed to I saw it had plenty of indexes and I thought I must have been missing something.  So I followed the lead.

cl/search/models.py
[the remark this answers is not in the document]
flooie: Well great.

cl/search/models.py
flooie: yes.

cl/search/models.py
flooie: ah - I thought you were referring to the related name which I changed.

cl/search/models.py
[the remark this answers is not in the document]
flooie: I had checked that it inherits char field - are you suggesting since the options are all text that it is in practice a text field?

cl/search/models.py
flooie: okay, I dove into this one a bit, but they dont provide a nice tuple associated with the country names.  I think the best option now is just to do the two countries I know we will be adding now.  UK and US ... I suppose there could be other ones but they aren't needed for state opinions I think or need a courthouse object

flooie:
@mlissner back to you.  🏓 

I created a simple two list set for country codes because I didnt want to install Django-countries to do what you were discussing above.

cl/search/models.py
```diff
@@ -2144,6 +2171,70 @@ class Meta:
...
+    )
+    state = USPostalCodeField(
+        help_text="The two-letter USPS postal abbreviation for the "
+        "organization w/ obsolete state options.",
+        db_index=True,
+        choices=USPS_CHOICES + OBSOLETE_STATES,
+        blank=True,
+    )
+    zip_code = USZipCodeField(
+        help_text="The zip code for the organization, XXXXX or XXXXX-XXXX "
```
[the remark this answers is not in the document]
flooie: gonna ignore this.

cl/search/models.py
```diff
@@ -2144,6 +2171,70 @@ class Meta:
...
+        blank=True,
+    )
+    county = models.TextField(
+        help_text="The county, if any, where the court resides.",
+        blank=True,
+    )
+    state = USPostalCodeField(
+        help_text="The two-letter USPS postal abbreviation for the "
+        "organization w/ obsolete state options.",
+        db_index=True,
```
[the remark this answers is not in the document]
flooie: gonna ignore this

cl/search/models.py
mlissner: Well here's a fun fact: Nobody on Github has ever used `ISO_3166_1_ALPHA2_COUNTRY_CODES`. Every usage that comes up in search is from the localflavor package itself.

cl/search/models.py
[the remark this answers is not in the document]
mlissner: Yep. This is a text field, so semgrep is wrong.

mlissner (approved):

cl/search/models.py
mlissner: I filed a bug to see if localflavor would find this useful: https://github.com/django/django-localflavor/issues/499

mlissner:
I merged this so it's in the code, but deploy will have to wait until tomorrow.

flooie:
Thank you

PR #3186, opened 2023-09-26
Update harvard merger

quevon24:
- Add new docket sources required for harvard merger:

```
COLUMBIA_AND_RECAP_AND_HARVARD = 21
COLUMBIA_AND_SCRAPER_AND_HARVARD = 22
COLUMBIA_AND_RECAP_AND_SCRAPER_AND_HARVARD = 23
IDB_AND_HARVARD = 24
RECAP_AND_IDB_AND_HARVARD = 25
SCRAPER_AND_IDB_AND_HARVARD = 26
RECAP_AND_SCRAPER_AND_IDB_HARVARD = 27
COLUMBIA_AND_IDB_AND_HARVARD = 28
COLUMBIA_AND_RECAP_AND_IDB_AND_HARVARD = 29
COLUMBIA_AND_SCRAPER_AND_IDB_AND_HARVARD = 30
COLUMBIA_AND_RECAP_AND_SCRAPER_AND_IDB_AND_HARVARD = 31
```

- Handle "Per curiam" in find_all_judges function
- Handle ZeroDivisionError exception from issue #3182 
- Move functions from harvard_opinions.py to corpus_importer utils
- I added a new parameter to command to skip judge merger: --skip-judge-merger
`manage.py harvard_merge --cluster-id 2442 --skip-judge-merger `


cl/search/models.py
```diff
@@ -320,6 +324,13 @@ class Docket(AbstractDateTimeModel):
         (SCRAPER_AND_HARVARD, "Scraper and Harvard"),
         (RECAP_AND_SCRAPER_AND_HARVARD, "RECAP, Scraper and Harvard"),
         (HARVARD_AND_COLUMBIA, "Harvard and Columbia"),
+        (IDB_AND_HARVARD, "IDB and Columbia"),
```
flooie, line 327: Is this correct?

cl/search/models.py
```diff
@@ -289,6 +289,10 @@ class Docket(AbstractDateTimeModel):
     SCRAPER_AND_HARVARD = 18
     RECAP_AND_SCRAPER_AND_HARVARD = 19
     HARVARD_AND_COLUMBIA = 20
+    IDB_AND_HARVARD = 24
+    RECAP_AND_IDB_AND_HARVARD = 25
+    SCRAPER_AND_IDB_AND_HARVARD = 26
+    RECAP_AND_SCRAPER_AND_IDB_AND_HARVARD = 27
     DIRECT_INPUT = 32
     DIRECT_INPUT_AND_HARVARD = 48
```
flooie, line 297: It would be great to just identify the remaining options here now so we dont have to do this docket source dance anymore .

flooie: Maybe just 21-to 23 and 28 to 31 to complete those scenarios

cl/corpus_importer/utils.py
```diff
@@ -87,9 +198,9 @@ def match_lists(
     :return: Matches if found or False
     """
     # We import this here to avoid a circular import
-    from cl.corpus_importer.management.commands.harvard_opinions import (
-        compare_documents,
-    )
+    # from cl.corpus_importer.management.commands.harvard_opinions import (
+    #     compare_documents,
+    # )
```
flooie, line 203: What is going on here?

cl/corpus_importer/management/commands/harvard_merge.py
```diff
@@ -941,6 +973,11 @@ def add_arguments(self, parser) -> None:
             action="store_true",
             help="A flag to choose to merge only fastcase opinions",
         )
+        parser.add_argument(
+            "--merge-judges",
+            action="store_true",
+            help="Set flag to merge judges",
+        )
```
flooie, line 980: I think a better name would be smart.  

Im guilty of this more than most but we want this flag to be used to skip merging of judge data if enabled.  right?  

so something like `skip_judge_merger` would be more understandable-  or `merge_without_judge_match` to illustrate that we are enabling a merge even if it doesnt match.  ? this would change the store true to store false I think but I think it makes sense

cl/search/models.py
quevon24: I already identified all the required docket sources using the bulk data, we should no longer see the warning in the logs.

cl/corpus_importer/utils.py
quevon24: I forgot to delete this, I moved some functions at corpus_importer users

cl/search/models.py
quevon24: No, my mistake, "IDB and Harvard" is the correct option name

cl/corpus_importer/utils.py
```diff
@@ -1,5 +1,5 @@
 from datetime import date
-from typing import Any, Optional
+from typing import Any, Iterator, List, Optional
```
flooie, line 2: do we need List, I took a quick look thru the PEPs but didnt find the change but I think we dont need titled List here.  list will do

cl/corpus_importer/management/commands/harvard_merge.py
quevon24: maybe `enable-judge-merger`? By default, it is disabled unless you pass the option to the command

flooie: I think my issue with this - is that the default is always to merge judges-  Its more of a flag to enable the completion of the merger without the judge if the judge doesnt match (presumably for typos).   soI think  the flag should reflect that.

cl/corpus_importer/utils.py
flooie: @quevon24 https://peps.python.org/pep-0604/#simplified-syntax

quevon24:
Changes are ready @flooie

cl/corpus_importer/management/commands/harvard_merge.py
```diff
@@ -1002,5 +1037,7 @@ def handle(self, *args, **options) -> None:
         for cluster_id, filepath in cluster_ids:
             logger.info(msg=f"Merging {cluster_id} at {filepath}")
             merge_opinion_clusters(
-                cluster_id=cluster_id, only_fastcase=options["fastcase"]
+                cluster_id=cluster_id,
+                only_fastcase=options["fastcase"],
+                skip_judge_merger=options["skip_judge_merger"],
```
flooie, line 1042: we aren't just skipping judge merger right?  Its let merger complete without judge matching no?

flooie: I dont want to skip the judge merger - if the judges matched.

cl/corpus_importer/management/commands/harvard_merge.py
```diff
@@ -543,11 +554,14 @@ def merge_overlapping_data(
...
-            data_to_update.update(
-                merge_judges(
-                    changed_values_dictionary.get(field_name),
+            if skip_judge_merger:
+                logger.info(
+                    f"Can't merge judges, option disabled, cluster id: {cluster.id}"
+                )
+            else:
+                data_to_update.update(
+                    merge_judges(changed_values_dictionary.get(field_name))
```
flooie, line 563: like here ... if we have skip judge merger we never try to merger judges - I would be more open to this flag - attempting to merge - and then we basically loosened the requirement.

quevon24: I pushed some changes, now, if the flag is set, merge won't be ignored, it will try to merge the judges, but if it fails, it will only log a message instead of raising an exception

flooie (commented):
Thanks @quevon24 

These changes look good to me.

Issue #3187, opened 2023-09-26
Update code to handle military court jurisdictions


flooie:
Previously (and currently), all military courts are grouped under Federal Special.

We are introducing Military Appellate and Military Trial jurisdictions.  We need to account for that in the court picker and anywhere else the code requires.

PR #3188, opened 2023-09-26
feat(search_utils): Update merge_form_with_courts

flooie:
PR addresses issue #3187 

When we switch military courts from Federal Special we need to account for that in the court picker.

This was an easier thing to do as we simply needed to make a small tweak.
Also I think we should eventually identify courts as Article I vs Article III


mlissner:
Thanks. Looks simple enough!

PR #3191, opened 2023-09-27
Add pghistory to CourtHouse model

quevon24:
- Add pghistory to CourthHouse model
- Add pghistory to court appeals_to m2m field


mlissner:
Very nice, thank you Kevin!

PR #3193, opened 2023-09-28
Initial support for ACMS docket reports.

johnhawkinson:
Initial support for ACMS.

This probably needs the juriscraper stub to be committed: https://github.com/freelawproject/juriscraper/pull/730

I do not have a courtlistener test environment available, and I'm focusing on the juriscraper and recap-chrome parts of this work. Can someone else deal with getting this up and running? I hope it is simple and mechanical.


mlissner (changes requested):
This looks pretty good to me, but for a couple things:

1. Lots of repetition with other functions. Do we want to merge these?
2. We need migration files per https://github.com/freelawproject/courtlistener/wiki/Database-migrations

@johnhawkinson, I hear you that you don't have a CL env. set up, so @albertisfu once you've got your elastic PR in, could you adopt this one and see what it takes to get it landed?

albertisfu:
Sure, I'm starting to look at this, I'm catching up on the conversation on Slack and I'll post any questions I have in order to get this landed.

johnhawkinson:
@albertisfu perhaps unwisely, I removed your merge commit and rebased off the `main` branch, because the merge doesn't really belong in the PR. I hope that doesn't mess up your merging.

Anyhow, I just added a relaxation of the `pacer_case_id` constraint as to number of dashes, after a conversation with Mike in #recap.

albertisfu:
I added the missing migrations here and also added the Juriscraper version that has the ACMS stub. Next, I'll add some tests for this new upload but before some comments about after analyzing this new docket report:

From the ACMS docket report, we can extract both the docket data and all the docket entries from the JSON. We could then map this data to the DB model:

### Docket data
"caseId":"e15ebc78-9507-4639-8a61-4bc42e613a66" - > **`pacer_case_id`**

"caseNumber":"23-6364" -> **`docket_number`**

"name":"United States of America v. Surmik" -> **`case_name `**

"caseOpened":"2023-04-19" -> **`date_filed`**

"aNumber":null - > **Not in DB**

"receivedDate":"2023-04-19T13:57:24Z" - > **Not in DB**

"partyAttorneyList":"" **`parties` this would need some parsing in Juriscraper since it's in HTML format.**

 "court":{
         "name":"N.D.N.Y. (SYRACUSE)",
         "identifier":"N.D.N.Y. (SYRACUSE)"
      } **is the `appeal_from` court?**

"caseType":"Criminal, **is this case_type_information?**

 "caseSubType":"Post-Conviction", **do we have this in DB?**

"caseSubSubType":null **do we have this in DB?**

"districtCourtName":null  **Do we have this in DB?**

"feeStatus":"IFP Granted" **Do we have this in DB?**

###  Docket Entries
We can get entries with these fields from the JSON:
  
"endDate":"2023-04-14" →  **DocketEntry `date_filed/date_entered`**

"endDateFormatted":"04/14/2023" →  **DocketEntry `date_filed/date_entered`**

"entryNumber":1, →  **DocketEntry/RECAPDocument `document_number`** 

"docketEntryText":"NOTICE OF CRIMINAL APPEAL, with the district court docket, on behalf of Appellant William Surmik, FILED. [Entered: 04/19/2023 10:26 AM]", →  **DocketEntry `description`**

"docketEntryId":"bde556a7-bdde-ed11-a7c6-001dd806a1fd" → **RECAPDocument `pacer_doc_id`**

"createdOn":"2023-04-19T14:22:56Z", **Not in DB**

"documentCount":2, → **(Number of documents, 0 a minute entry, 1 only main document, 2 or greater main document + attachments, we might not need this.)**

"pageCount":12, → **RECAPdocument `page_count`**

"fileSize":410, **RECAPdocument `file_size`**

"restrictedPartyFilingDocketEntry":false, **Not in DB**

"restrictedDocsAvailable":false, **Not in DB**


Discussing this with @ERosendo we think we could use the `docketEntryId` and store it as the `pacer_doc_id`, however, we'd need to increase its current `max_length` since it's currently `32` and `docketEntryId` are `36` length.

However, we should consider that `docketEntryId` is the same for all the documents that belong to a docket entry, for the main document, and for all their attachments (if there are any.)

Then we should be able to match the document uploads/document availability to the right DocketEntries/RECAPDocuments.
If you click on an entry with only one document we could use the `docketEntryId` which is in the download URL so we can match it with the correct RECAPDocument in CL.

Screenshot 2023-10-02 at 18 28 56


However, the situation is more complicated when there are attachments.
In this case, when you click any of the documents you’d get the same `docketEntryId` in the download URL for all of them. So we can’t use the `docketEntryId` in order to identify the document being downloaded/available in CL.

Screenshot 2023-10-02 at 18 02 36

Until you click one of them and the download page is opened, you’ll get the `selectedDocuments` key that contains the `docketDocumentDetailsId` which is the unique identifier for documents when the entry has more than one document.
Screenshot 2023-10-02 at 18 39 33

So a workaround that @ERosendo suggested for this issue is that we can use the `docketEntryId` +  the `document/attachment number` in order to identify available documents in CL or the document being downloaded.

We think this can work and we can avoid adding an additional field for storing `docketDocumentDetailsId`.

We just need to confirm this approach with more examples, could you provide more examples for this and other courts where this report is available? 


 About Juriscraper, If I understand correctly, we would obtain the JSON from the extension. We could then store this JSON in CL as a file for backup, similar to how we store HTML files. Juriscraper would then parse the PACER JSON and convert it into a format compatible with CL JSON, correct?

johnhawkinson:
> From the ACMS docket report, we can extract both the docket data and all the docket entries from the JSON. We could then map this data to the DB model:

I think most of this is properly in juriscraper, and https://github.com/freelawproject/juriscraper/pull/730 doesn't really attempt to do so, since the parsing code hasn't been written.

> ### Docket data
> "court":{ "name":"N.D.N.Y. (SYRACUSE)", "identifier":"N.D.N.Y. (SYRACUSE)" } **is the `appeal_from` court?**

Yes.

> ### Docket Entries
…
> "endDate":"2023-04-14" → **DocketEntry `date_filed/date_entered`**
> "endDateFormatted":"04/14/2023" → **DocketEntry `date_filed/date_entered`**
> "createdOn":"2023-04-19T14:22:56Z", **Not in DB**

Err, don't we store times now? I thought we did, for items received from RSS.

ACMS may not expose a date_filed/date_entered distinction that I am aware of (I'm not sure—I think Appellate CM/ECF only exposes it in NDA emails, and I haven't seen any from ACMS). But it does appear that the `createdOn` field is consistent with the submission time of the entry and the `Entered` text in the `docketEntryText` reflects the actual time it was entered, accounting for processing delay, which for ACMS can be several minutes (up to fifteen, perhaps, allegedly), and which was basically instantaneous in CMECF. Which is a long way to say I'm not sure which date we should be storing, but the I think we should probably use the DateTime parsed back out of the docket tet.


> "pageCount":12, → **RECAPdocument `page_count`**
> 
> "fileSize":410, **RECAPdocument `file_size`**

I don't think this is what that is. It's certainly not the size in bytes. (I guess it could be in kilobytes?).

> Discussing this with @ERosendo we think we could use the `docketEntryId` and store it as the `pacer_doc_id`, however, we'd need to increase its current `max_length` since it's currently `32` and `docketEntryId` are `36` length.

Hilariously, we could shoehorn it in by removing the hyphens from the GUID before storing it.
That's probably unwise, we should just increase the length.

> However, the situation is more complicated when there are attachments. In this case, when you click any of the documents you’d get the same `docketEntryId` in the download URL for all of them. So we can’t use the `docketEntryId` in order to identify the document being downloaded/available in CL.
…
> We think this can work and we can avoid adding an additional field for storing `docketDocumentDetailsId`.

This is a problem that we have in the district court, too. For instance, `https://ecf.dcd.uscourts.gov/doc1/045010205474` is the main document in DE 101 in https://www.courtlistener.com/docket/61642105/101/freeman-v-herring-networks-inc/, but `045010205474` is also the aggregation of all 8 attachments and the main document.

I guess the problem is a little different here, but related.
I think it would be unwise to throw away the attachment ID information where we have access to it.

But maybe that's not what you're proposing.


> So a workaround that @ERosendo suggested for this issue is that we can use the `docketEntryId` + the `document/attachment number` in order to identify available documents in CL or the document being downloaded.

Saving the docket number doesn't preclude this. I guess we would need a new CL lookup query if we went this route?

> We just need to confirm this approach with more examples, could you provide more examples for this and other courts where this report is available?

What kind of examples do you want? `ca2` and `ca9` are now using this for all new cases as of yesterday, and have been using it for quite some time (a year?) for immigration cases.

https://ca9-showdoc.azurewebsites.us/23-2487 is an example of a case docketed today in the 9th circuit.
https://ca2-showdoc.azurewebsites.us/23-7222?caseId=1003904 is an example of a case docketed today in the 2nd circuit (err, with the questionable caseId addition that the recap extension applies).


> About Juriscraper, If I understand correctly, we would obtain the JSON from the extension. We could then store this JSON in CL as a file for backup, similar to how we store HTML files. Juriscraper would then parse the PACER JSON and convert it into a format compatible with CL JSON, correct?

Correct.

albertisfu:
> Err, don't we store times now? I thought we did, for items received from RSS.
> 
> ACMS may not expose a date_filed/date_entered distinction that I am aware of (I'm not sure—I think Appellate CM/ECF only exposes it in NDA emails, and I haven't seen any from ACMS). But it does appear that the createdOn field is consistent with the submission time of the entry and the Entered text in the docketEntryText reflects the actual time it was entered, accounting for processing delay, which for ACMS can be several minutes (up to fifteen, perhaps, allegedly), and which was basically instantaneous in CMECF. Which is a long way to say I'm not sure which date we should be storing, but the I think we should probably use the DateTime parsed back out of the docket tet.

Yeah, now we can store `time` for docket entries, I was confused due to `endDate` didn't match to `createdOn` so I thought it was something different. But now that you mentioned that `createdOn` is the `date_entered` we could store this `date` and` time` as we do for RSS entries in which the date and time stored are the `date_entered`.
If so, dow we should just ignore `endDate`? 

> Hilariously, we could shoehorn it in by removing the hyphens from the GUID before storing it.
> That's probably unwise, we should just increase the length.

Yeah I agree, it's better to increase the length in `pacer_doc_id`. 

> I think it would be unwise to throw away the attachment ID information where we have access to it.

Got it, so, in that case, we could store the `docketDocumentDetailsId` as a new field in the `RECAPDocument` model, maybe we could call it `pacer_doc_detail_id` if @mlissner agrees. Then we could include this field in the `ProcessingQueue` model, so we can use it to match uploads.

> Saving the docket number doesn't preclude this. I guess we would need a new CL lookup query if we went this route?

Yeah, the docket number doesn't help here. So the idea was to use the `pacer_doc_id` (`docketEntryId`) and tweak the `recap-query` response to return the document/attachment numbers and use them to match available documents by the extension using the `docketEntryId` and `document/attachment` number.

As an alternative, also @ERosendo found that it's possible to get the `docketDocumentDetailsId` from a div in the attachment page that contains `vue` properties with all the document data including the `docketDocumentDetailsId`. He mentioned is a bit hacky but it can work since he already has some code for extracting those properties here: https://github.com/johnhawkinson/recap-chrome/pull/1 
 
> https://ca9-showdoc.azurewebsites.us/23-2487 is an example of a case docketed today in the 9th circuit.
> https://ca2-showdoc.azurewebsites.us/23-7222?caseId=1003904 is an example of a case docketed today in the 2nd circuit (err, with the questionable caseId addition that the recap extension applies).

Thanks, yeah these examples are fine. So if we'll only receive ACMS dockets from courts that use docket entry numbers (like ca2 and ca9), that seems fine. Because if we expect to receive an ACMS where their docket entries don't have numbers. That will be an additional issue we need to handle since in those cases we use the `pacer_doc_id` as an entry number and in these entries, the `pacer_doc_id` will be the `docketEntryId` which is not a number so it can't be stored in `entry_number` which is an integer. 

Let me know your thoughts about these comments so if you agree I can start working on mocking some data for an ACMS docket and adding tests for ingesting the docket/docket entries.

johnhawkinson:
> Let me know your thoughts about these comments so if you agree I can start working on mocking some data for an ACMS docket and adding tests for ingesting the docket/docket entries.

Best is the enemy of good enough. This went live on Sunday and we're missing new appellate cases, so I'm trying to move with alacrity.

> Yeah, now we can store `time` for docket entries, I was confused due to `endDate` didn't match to `createdOn` so I thought it was something different. But now that you mentioned that `createdOn` is the `date_entered` we could store this `date` and` time` as we do for RSS entries in which the date and time stored are the `date_entered`.

That is…not exactly what I said:

> > But it does appear that the `createdOn` field is consistent with the submission time of the entry and the `Entered` text in the `docketEntryText` reflects the actual time it was entered, accounting for processing delay, which for ACMS can be several minutes (up to fifteen, perhaps, allegedly), and which was basically instantaneous in CMECF. Which is a long way to say I'm not sure which date we should be storing, but the I think we should probably use the DateTime parsed back out of the docket te[x]t.

I don't think we should be using `createdOn` at all. But this is a question for the JSON parser and one we can revisit in the future and reparse old data, so we don't have to decide now.

> If so, dow we should just ignore `endDate`?

We should ignore almost all of them and use the date/time parsed out of the Entered text field, unfortunately.

 
> Got it, so, in that case, we could store the `docketDocumentDetailsId` as a new field in the `RECAPDocument` model, maybe we could call it `pacer_doc_detail_id` if @mlissner agrees. Then we could include this field in the `ProcessingQueue` model, so we can use it to match uploads.

As you are probably aware, I could write a treatise on naming.
I think that:
* Nobody should be calling anything `pacer_`* because, "PACER" is too confusing a term that can refer to the generalized single-sign on infrastructure, to the nationwide billing system, to the PACER Case Locator formerly US National Case/Party Index, to individual court-specific CM/ECF instances, and perhaps, as you would be suggesting to ACMS. And to overlaps of any of these things. There is almost always a better more-specific name to use.
* Is this field intended to store ACMS document IDs and but store nothing in Appellate CM/ECF and District (and BK) CM/ECF? If so, it should probably have an `acms_`* name.
* I recognize we have historically named things `pacer_`* where we perhaps should have called them `ecf_`*, and it is nice to not change  <s>horses</s>conventions in midstream.
* I am not sure it's great to choose names based on the AOUSC programmer's implementation names; sometimes that's good for seeing the 1:1 correspondence, but "details ID" doesn't really express that this is a unique ID for a document.

Anyhow, I would call it `acms_document_guid`.

> So if we'll only receive ACMS dockets from courts that use docket entry numbers (like ca2 and ca9), that seems fine. 

We do not know the future with perfect clarity, but I think it is unlikely any other courts will adopt ACMS in the current calendar year, if they ever do (not a foregone conclusion!).

> Because if we expect to receive an ACMS where their docket entries don't have numbers.

It's also not clear that ACMS does or will in the future support the concept of unnumbered docket entries.

> That will be an additional issue we need to handle since in those cases we use the `pacer_doc_id` as an entry number and in these entries, the `pacer_doc_id` will be the `docketEntryId` which is not a number so it can't be stored in `entry_number` which is an integer.

Hrmm. As long as we are changing schema (lengthening `pacer_doc_id` ), maybe we should consider changing this to a character field:

```py
    entry_number = models.BigIntegerField(
        help_text=(
            "# on the PACER docket page. For appellate cases, this may "
            "be the internal PACER ID for the document, when an entry "
            "ID is otherwise unavailable."
        ),
```

I guess we can bridge that gap when we come to it?

johnhawkinson:
Oops, so, I think my force-push/rebase confused Github's tracking of @mlissner's review comments. I tried force-pushing back to d0c76c9509b61245d5b9e101a2eee87c5114c4c5 but that did not make them show up again, so I re-force-pushed back to where we were. Hopefully this confuses no humans, only computers.

> This looks pretty good to me, but for a couple things:

I addressed these in Slack on Sept. 28 but should have done so here:

>     1. Lots of repetition with other functions. Do we want to merge these?

In a perfect world, yes. This maintains the same level of duplication found between appellate dockets and district court dockets, and seems like it is too much, but also it would be work to break it up and abstract it better. Of course the case for that work increases when you have things in triplicate rather than duplicate, but I did not see an easy way for me to tackle that task without a lot more uncertainty than I already have about this code that I'm not in a good position to test.
(Maybe it is unrealistic for me to think I can do recap development without standing up a test courtlistener server to point it at? If so someone should disabuse me of my presumptions.

>     2. We need migration files per https://github.com/freelawproject/courtlistener/wiki/Database-migrations

OK, that one's on Alberto (well, they both are, really).

albertisfu:
> I guess we can bridge that gap when we come to it.

Yeah, we can tweak `entry_number` only if it's required in the future.

Well for now I'll work on increasing the length of `pacer_doc_id` and adding the new field `acms_document_guid`, I'll add the migration files and some tests.

mlissner:
Thank you guys for all the discussion here. I've been mostly waiting for you guys to lead me, but I do have a couple thoughts (none about code, apparently!):

> Best is the enemy of good enough.

Yes, but good enough is the enemy of great. If we've got our eye on this now and have the energy to do so, I'm very much in favor of doing everything the right way so we don't have to come back around and fix things. The sad reality is we're pretty bad at coming around later to fix things, particularly things that are good enough.

> We can revisit in the future and reparse old data,

This freaks me out and I'd like to emphasize that I don't want to do this. One issue is that any data we get has to be re-processed in the order it's received, including across object types. So if we get a docket, then an attachment page, then a docket, we can't just re-process the dockets alone, we'll have to do all three, in the correct order. This is partly because the data can change under us, and partly because it's an assumption baked into the merging code — I don't know what would happen if we ignored this assumption.

There is also a small amount of manual work that gets done in the DB, to, for example, deal with sealed content and other issues like that. When we re-process data, we tend to wipe out those manual changes. 

It's also just work I don't want to do (my time is at an extreme premium and I tend to be the one that has to do it). We should make every effort to get it right, even at a time-delay, so we can avoid having to re-process data.

> force-push/rebase confused Github

Yeah, Github doesn't do well with this. The worst part is, it breaks a feature that lets reviewers see changes since their last review. This is getting annoying enough that I'm getting close to blocking force pushes on all branches.

That's it for me for now. Thank you guys again. I appreciate this moving forward at such a clip.

johnhawkinson:
> Yes, but good enough is the enemy of great.

Is it? In any case, I'm not here to define "good enough," just to say that at least from my perspective, I wouldn't hold up work mocking things based on the level of open questions we have here.

> The sad reality is we're pretty bad at coming around later to fix things, particularly things that are good enough.

I guess this is about issue tracking and assignment and stuff.

> > We can revisit in the future and reparse old data,
> 
> This freaks me out and I'd like to emphasize that I don't want to do this.

I hear you. I could have been a bit more clear — we don't really understand what many of these fields are and whether they have any value (remember they are not displayed to the user in the UI), and we would need database model changes to store them. I think we don't need them, but if we do, there is always a way to come back and get them.
The alternative is adding them to the db model even if we're not clear on what they are used for, or even adding an `extra_shit` JSON blob to the database. I bet your DBA and DQA twins would both love that!

> > force-push/rebase confused Github
> 
> Yeah, Github doesn't do well with this. The worst part is, it breaks a feature that lets reviewers see changes since their last review. This is getting annoying enough that I'm getting close to blocking force pushes on all branches.

Yeah. I thought I was safe because (I thought), your comments were not actually tied to code. I guess I'll try clicking the magic circle arrow red button and see if that resolves the review state (in)consistency.
 
> That's it for me for now.

You did not speak to the T.S. Eliot question:
> > Got it, so, in that case, we could store the `docketDocumentDetailsId` as a new field in the `RECAPDocument` model, maybe we could call it `pacer_doc_detail_id` if @mlissner agrees. Then we could include this field in the `ProcessingQueue` model, so we can use it to match uploads.
> 
> As you are probably aware, I could write a treatise on naming. I think that:
…
> Anyhow, I would call it `acms_document_guid`.


> Thank you guys again. I appreciate this moving forward at such a clip.

Har. I have no point, I just want to say: _JSON Tub Time Machine…countless screaming Argonauts._

mlissner:
Thanks for the comments.

For the field, I think between you and Alberto, I don't have much to add. The name seems fine and reasonable to me. If you guys agree we need it and agree that the name is good, I'm +1 to whatever you decide.

johnhawkinson:
To the limited extent that I can do so with my eyeballs, these two recent changes (37d5c1a7908c8a4ca10c04096da084c3e77d1d7b f5dde3a9bf2ff41374a2230039fba31a06a68f32) look correct, yes, modulo the failed migration test which looks like it's just been dealt with. 👍

johnhawkinson:
> * Added the new `acms_document_guid` field in `AbstractPacerDocument` and `ProcessingQueue`
> Since I added `acms_document_guid` in `AbstractPacerDocument` this new field will be also added to `ClaimHistory` (in addition to `RECAPDocument`) since it depends on `AbstractPacerDocument`, is that correct?  Like if there is a possibility of getting `ClaimHistory` documents for ACMS?  or should we add it to the `RECAPDocument` model directly instead?

Oops, yes, I think you're right to raise this concern.

The `ClaimHistory` only exists in bankruptcy court, and to my knowledge it does not appear in appeals of bankruptcy cases, not at the district court level, not at the Bankruptcy Appellate Panel level, and not at the Court of Appeals level. So I do not think we would ever expect to see an `acms_document_guid` in a ClaimHistory and the field does not belong, and it should probably be added directly to `RECAPDocument`.

(The 'A' in ACMS stands for 'Appellate'.)

mlissner:
Yeah, I second that. We don't need the new field for claims docs.

cl/search/models.py
```diff
@@ -1295,6 +1287,11 @@ class RECAPDocument(AbstractPacerDocument, AbstractPDF, AbstractDateTimeModel):
         ),
         blank=True,
     )
+    acms_document_guid = models.CharField(
+        help_text="The ID of the document in PACER.",
```
johnhawkinson, line 1291: Err, oops. Maybe we should say "The GUID of the document in ACMS."
Sorry for not catching this earlier.

albertisfu:
Correct, I've moved `acms_document_guid` directly into `RECAPDocument` and fixed its help text.

johnhawkinson:
Oops, clearly that failing test was my fault. I just tried a blind fix, but if that doesn't do it, someone with a dev environment should take it on.

I am not clear whether Alberto was planning to do more, or we were waiting for Mike's re-review, or something else. Clarity on who has the next step would be good. Thanks.

mlissner:
Alberto has this on his plate.

johnhawkinson:
> Once the Juriscraper parser is complete, please let me know, and I can update this test to replace the mock data with the actual data returned by this parser.

Yes, I definitely shall. Is there anything blocking this from landing?

albertisfu:
Well, we need to wait for Juriscraper, so we can do a release before we can update the dependency here and merge it into production.

johnhawkinson:
In order to support having `date_filed` returned as a naive `datetime` (without timezone, representing the time in the Court's local time zone), @albertisfu says:

> So we’ll need to tweak the `localize_date_and_time` method to skip naive datetimes, but it’s an easy tweak.

I understand Alberto will add that here.

johnhawkinson:
@albertisfu:
> Once the Juriscraper parser is complete, please let me know, and I can update this test to replace the mock data with the actual data returned by this parser.

Data parsing tests are part of https://github.com/freelawproject/juriscraper/pull/730

albertisfu:
@johnhawkinson I've applied a tweak to avoid converting naive dates from docket entries.

I've also updated the mock data based on one of the examples in Juriscraper. Although I haven't looked in detail the Juriscraper PR, I noticed some potential issues based on the examples:

- The docket `date_filed` field is currently returning a string when it should be a date object.
- The docket entry `description` contains HTML, e.g:
`"description": "<p>NOTICE OF CRIMINAL APPEAL, with district court docket, on behalf of Appellant Mustapha Raji, FILED. [Entered: 10/02/2023 11:17 AM]</p>",`

In CL we don't clean the data from the docket entries, so if we store that data it'll be shown like this:
Screenshot 2023-10-17 at 13 18 45
So it might require to be cleaned in Juriscraper?

Lastly, there is a merge conflict. I've resolved it but haven't pushed the changes yet, as I used rebase and a force push will be required. Is that okay?

cl/lib/timezone_helpers.py
```diff
@@ -31,7 +32,14 @@ def localize_date_and_time(
...
     if isinstance(date_filed, datetime):
-        datetime_filed_local = convert_to_court_timezone(court_id, date_filed)
+        if is_naive(date_filed):
+            datetime_filed_local = localize_naive_datetime_to_court_timezone(
+                court_id, date_filed
+            )
+        else:
+            datetime_filed_local = convert_to_court_timezone(
+                court_id, date_filed
+            )
```
johnhawkinson, line 42: I think this is probably not sufficiently clear to the reader, and the function docstring should be updated.

johnhawkinson:
> @johnhawkinson I've applied a tweak to avoid converting naive dates from docket entries.

Great, thanks. As noted, it might benefit from an update to the docstring, if not a comment.
I guess I can do that, although I'm in an all-day meeting until either 4pm or 6pm Eastern and will look if it's still unchanged after.

> I've also updated the mock data based on one of the examples in Juriscraper. Although I haven't looked in detail the Juriscraper PR, I noticed some potential issues based on the examples:
> 
>     * The docket `date_filed` field is currently returning a string when it should be a date object.

Huh. I'll run that down.

>     * The docket entry `description` contains HTML, e.g:
>       `"description": "<p>NOTICE OF CRIMINAL APPEAL, with district court docket, on behalf of Appellant Mustapha Raji, FILED. [Entered: 10/02/2023 11:17 AM]</p>",`

This is deliberate, as noted in the code:

https://github.com/freelawproject/juriscraper/blob/728461586b8ca4ee8c9c45d7f47c7d86838c3524/juriscraper/pacer/acms_docket.py#L294-L296

> In CL we don't clean the data from the docket entries, so if we store that data it'll be shown like this: Screenshot 2023-10-17 at 13 18 45 So it might require to be cleaned in Juriscraper?

Well, or we can change how we display it, or I can be sad and give up on this dream.
I would prefer to not lose the semantic information and hyperlinks to Court documents.

> Lastly, there is a merge conflict. I've resolved it but haven't pushed the changes yet, as I used rebase and a force push will be required. Is that okay?

Huh, that wasn't there last I check. I am fine with force pushes (where appropriate) but Mike seems to get upset, and I don't know who else is reviewing this. So my opinion may not be controlling, but seems fine.

cl/tests/utils.py
```diff
@@ -57,3 +60,70 @@ def __init__(
...
+            court_id="ca2",
+            appeal_from="S.D.N.Y . (NEW YORK CITY)",
+            case_name="United States of America v. Raji",
+            case_type_information="Criminal, Direct Criminal",
+            date_filed=date(2023, 10, 2),
+            docket_number="23-6364",
+            fee_status="IFP Granted",
+            originating_court_information={
+                "identifier": "S.D.N.Y. (NEW YORK CITY)",
+                "name": "S.D.N.Y . (NEW YORK CITY)",
```
johnhawkinson, line 80: All things being equal, we should not pick test data with this goofy spacing anomaly (`S.D.N.Y .` versus `S.D.N.Y.`).

Maybe it shouldn't be in the juriscraper tests at all, idk! But there exist other data that do not have this defect.

albertisfu:
Great, I've updated the mock data.

Regarding:

> Well, or we can change how we display it, or I can be sad and give up on this dream.
> I would prefer to not lose the semantic information and hyperlinks to Court documents.

Correct. In that case, I believe you want the HTML to be interpreted by the browser, right? To achieve that, we'd need to remove the `safe` filter from the template. I can do that if @mlissner approves.

Additionally, we should clean the HTML in the description before saving it to Elasticsearch to ensure it's properly indexed.

johnhawkinson:
I split your penultimate commit and squashed in the pre-commit comma fix and force-pushed.

> Correct. In that case, I believe you want the HTML to be interpreted by the browser, right?
Ideally, but I'd settle for preserving it and figuring that out in the future.

> To achieve that, we'd need to remove the `safe` filter from the template. I can do that if @mlissner approves.
> 
> Additionally, we should clean the HTML in the description before saving it to Elasticsearch to ensure it's properly indexed.

So, "I'm not a DBA," I always say before making database schema comments, but to me this strongly suggests my initial leanings were correct and we should be saving the HTML description as a separate field from the plain-text description. But maybe that's hard because it's a schema change with othe rramifications.

johnhawkinson:
@albertisfu: 
> The docket `date_filed` field is currently returning a string when it should be a date object.

What, what does this mean? It's returning JSON, how can it return "an actual date object"?

It also looks like it's just about as `string`y as the dates returned by other parsers:

(juriscraper) jhawk@lrr dockets % pwd
/Users/jhawk/src/juriscraper/tests/examples/pacer/dockets
(juriscraper) jhawk@lrr dockets % grep date_filed appellate/ca9_293296.json|head
  "date_filed": "2017-07-26",
    "date_filed": "2017-01-17",
    "date_filed_noa": "2017-07-26",
(juriscraper) jhawk@lrr dockets % grep date_filed acms/ca9_23-2412.json|head
  "date_filed": "2023-09-26",
      "date_filed": "2023-09-26T13:16:00",
      "date_filed": "2023-09-26T13:19:00",
      "date_filed": "2023-09-26T13:16:00",
      "date_filed": "2023-09-26T13:17:00",
      "date_filed": "2023-09-26T13:16:00",
      "date_filed": "2023-09-26T13:16:00",
      "date_filed": "2023-09-27T05:20:00",
      "date_filed": "2023-09-27T07:39:00",

(juriscraper) jhawk@lrr dockets %

albertisfu:
> What, what does this mean? It's returning JSON, how can it return "an actual date object"?
> 
> It also looks like it's just about as stringy as the dates returned by other parsers:

Yeah, this is true for the tests, however, in Python when accessing the metadata through `report.data` it should return a dictionary with valid python objects, well that's how it works for all the other reports.

For instance, the following method is used to convert a text date string to a python `date` object in the docket_report.
https://github.com/freelawproject/juriscraper/blob/8dfd9275294b1a76103a5a0667479016fe3962e7/juriscraper/pacer/docket_report.py#L106


> So, "I'm not a DBA," I always say before making database schema comments, but to me this strongly suggests my initial leanings were correct and we should be saving the HTML description as a separate field from the plain-text description. But maybe that's hard because it's a schema change with othe rramifications.

Yeah, if you want to preserve the HTML it makes sense to have a separate field for the plain-text, so it can be used for indexing purposes. I'm not sure what @mlissner thinks about it.

johnhawkinson:
Okay I guess that means we've identified a hole in the test coverage,
hooray!

johnhawkinson:
> Yeah, this is true for the tests, however, in Python when accessing the metadata through `report.data` it should return a dictionary with valid python objects, well that's how it works for all the other reports.

@albertisfu: Okay, see https://github.com/freelawproject/juriscraper/pull/730#issuecomment-1771929234.

albertisfu:
@johnhawkinson I resolved the conflicts in migrations and updated Juriscraper.

I tested an ACMS docket upload, and the modification to make `date_filed` a `datetime` is working correctly.

However, after resolving the `date_filed` issue, I noticed a new problem in Juriscraper as the upload process advanced to a further stage. The issue arises because the `attorneys` key sometimes contains empty dictionaries {}, which causes issues during the merging process.

```
 "attorneys": [
        {
          "contact": "Direct: 818-784-1126\nEmail: lisapickering@hotmail.com\nLaw Office of Roni P. Deutsch\n16255 Ventura Boulevard\nSuite 610\nEncino, CA 91436",
          "name": "Ms. Lisa Pickering, Attorney",
          "roles": [
            "Retained"
          ]
        },
        {},
        {
          "contact": "Direct: 213-487-9211\nEmail: raming@raminusa.com\nLaw Offices of Ramin Ghashghaei, Inc.\n23371 Mulholland Dr.\nSuite 399\nWoodland Hills, CA 91364",
          "name": "Ramin Ghashghaei, Attorney",
          "roles": [
            "Retained"
          ]
        },
        {}
      ],
```

johnhawkinson:
> I noticed a new problem in Juriscraper as the upload process advanced to a further stage. The issue arises because the `attorneys` key sometimes contains empty dictionaries {}, which causes issues during the merging process.

Fixed in https://github.com/freelawproject/juriscraper/pull/730/commits/3bb6912f6499ba104e130778c88443b0c8d5d1be

There were empty `<div>`s used for (presumably) spacing, and we now skip them.

albertisfu:
> Fixed in https://github.com/freelawproject/juriscraper/commit/3bb6912f6499ba104e130778c88443b0c8d5d1be

Great, I've updated Juriscraper with your latest changes from the stub branch and tested the upload; now it works correctly.

I think the remaining work here involves deciding what to do with the HTML description of entries. Should we create a new field called `html_description` to hold the HTML content, while continuing to use the `description` field for the plain text version?

johnhawkinson:
> I think the remaining work here involves deciding what to do with the HTML description of entries. Should we create a new field called `html_description` to hold the HTML content, while continuing to use the `description` field for the plain text version?

Isn't this a discussion we should be properly having in https://github.com/freelawproject/juriscraper/pull/730?

Anyhow, @mlissner , I think this is your call.
See around https://github.com/freelawproject/courtlistener/pull/3193#issuecomment-1767446162 

On the basis of our Oct. 4 Slack conversation, I preserved HTML in the `description` field.
But maybe it should go in another field.

Let's get a ruling so this can be unblocked.

mlissner:
I'm sorry, John, I'd like to start saving and serving HTML, but I realized a problem that makes it harder, and I think in the interests of getting this done, we should strip it. The problems are as follows at this point:

1. I agree this should probably go in its own field. If we do that, then we need to update the API, search engine, database, website, API docs, etc. That gets to be a lot.

3. When I said this was fine, because we now have a strict CSP policy on CL, I forgot that our API users trust the data we serve to them. I think it'd be bad if we started giving them HTML where we used to give them plain text, and I suspect for many, if we documented it, ("Oh hey, there's potentially risky info in our DB"), they wouldn't be pleased about it. Duplicating the data in the current field plus another one fixes this problem, but it came to mind.

So, if we want to be expeditious, I'd say skipping HTML would be smart.

johnhawkinson:
@mlissner:
> 1. I agree this should probably go in its own field. If we do that, then we need to update the API, search engine, database, website, API docs, etc. That gets to be a lot.

Okay. Fine to have the juriscraper parser emit `description` (as text) and `description_html` which would then be ignored (for now) by CL?

johnhawkinson:
> Okay. Fine to have the juriscraper parser emit `description` (as text) and `description_html` which would then be ignored (for now) by CL?

Done in https://github.com/freelawproject/juriscraper/pull/730/commits/095b5412dad9b5b171a2bf3475099a2847def969

mlissner:
That makes sense to me, thanks John.

cl/search/models.py
```diff
@@ -871,6 +877,8 @@ def pacer_docket_url(self):
 
             if not self.pacer_case_id:
                 return self.pacer_appellate_url_with_caseNum(path)
+            elif self.pacer_case_id.count("-") == 4:
+                return self.pacer_acms_url()
```
johnhawkinson, line 881: Is this really wise? To decide we are following ACMS URL rules because there are exactly four hyphens in the PACER case ID, i.e. it's presumably a GUID? This just feels fragile and like we're overloading fields.

Is the real problem that AbstractPacerDocument lacks a type field? Should we just open an Issue to address that (later)? idk.

ERosendo:
I updated the `help text` for the `acms_document_guid` field in the `ProcessingQueue` model to align with the terminology used in the `RecapDocument` model. I also noticed the `View on Pacer` button was not working properly for this new dockets so I  added a helper function to compute the links to ACMS.

While trying to apply the same logic to generate links for `Buy on Pacer` buttons, I found an unexpected issue. The **download confirmation** page relies on the browser's sessionStorage, resulting in an error message when it's absent. I think it's not possible to reach the download page using links (like in CMECF) and loading the docket report is an essential step to access the download page.

Here are some examples of the links to the download confirmation page:

- https://ca2-showdoc.azurewebsites.us/download-confirmation/654559bd-c36d-ee11-8def-001dd804e4bd?loadEntry=1
- https://ca2-showdoc.azurewebsites.us/download-confirmation/25834aee-d463-ee11-be6e-001dd804e087?loadEntry=1
- https://ca9-showdoc.azurewebsites.us/download-confirmation/1e3304bc-9b63-ee11-be6e-001dd8087c50?loadEntry=1

mlissner:
Yikes, that's not great. @johnhawkinson have you run into that too?

johnhawkinson:
> While trying to apply the same logic to generate links for `Buy on Pacer` buttons, I found an unexpected issue. The **download confirmation** page relies on the browser's sessionStorage, resulting in an error message when it's absent. I think it's not possible to reach the download page using links (like in CMECF) and loading the docket report is an essential step to access the download page.

Yes, it doesn't look particularly good for this functionality, short of manipulating the sessionStorage or monkeypatching the Javascript. Neither of which seem pleasant — they should not block this.

Of course, Appellate CM/ECF is not particularly great about this either, it's not like it has a "View Document" feature like District CM/ECF does. So it's not altogether shocking that if Appellate CM/ECF defined the public-facing functional spec for ACMS, that this might have been omitted.

I had looked at this a few weeks ago without thinking too hard about it, and I tried a bunch of "obvious" variations this evening and nothing seems to work, so I don't think there's a hidden URL endpoint we just don't know about.

mlissner:
Any chance you could write to your contacts there?

cl/search/models.py
ERosendo: > Is the real problem that AbstractPacerDocument lacks a type field?

A type field would be really helpful, it will provide a clear and concise identifier for ACMS docket. Currently, the only means of differentiation is the "pacer_case_id".

> This just feels fragile.

You're right. I think can use the [uuid](https://docs.python.org/3/library/uuid.html) module to check whether the `pacer_doc_id` is a valid GUID or not:

``` python
def is_valid_uuid(uuid_to_test):
    try:
        uuid_obj = UUID(uuid_to_test)
        return True
    except ValueError:
        return False
```

and then:


```python
elif is_valid_uuid(self.pacer_case_id):
      return self.pacer_acms_url()
```

johnhawkinson: > You're right. I think can use the [uuid](https://docs.python.org/3/library/uuid.html) module to check whether the `pacer_doc_id` is a valid GUID or not:

We could, but that's arguably even worse. We don't enforce that requirement anywhere else (deliberately so, in my view!) Look at the tests, e.g. `PacerFetchQueueSerializer` that checks if `attrs.get("pacer_case_id").count("-")` is not equal to `1`. 

If we allow an invalid GUID to be inserted (which we probably should. If `ca9` gives us a docket id of `correct-horse-battery-staple`, who are we to say they are wrong?), and we don't validate against the uuid spec anywhere else, we should not do so here.

Also, according to https://stackoverflow.com/questions/53847404/how-to-check-uuid-validity-in-python#comment108940714_33245493, your test is insufficient, a `ValueError` is not always raised.
It's that kind of subtlety and complexity that argues against that kind of check.

cl/search/models.py
johnhawkinson, line 1: I am not sure that the premise here is correct, that it is better to return a docket report link here than it is to return nothing at all. Ideally we would change the CL docket report template to explain what is going on and make these links look different, both because there is no confirmation step required before incurring charges, and also because they don't directly give the document desired, nor do they clearly set the parameters to limit the report to the minimum cost (which I'm not sure is even possible in ACMS like it is in CM/ECF).

I think the code's structure is also not good, this commit adds another unnecessary level of hierarchy, and the structure of the hierarchy doesn't match the domain, perhaps necessarily so. Ideally we would be take the common cases first, so:

```py
if cmecf_district:
  ...
elif appellate _cmecf:
  …
elif acms:
  …
else:
  fail!
```

but since we have no type field we kind of have to have the district court be the catchall? Still, there is no reason to do:

```py
if acms:
  …
else:
     if cmecf_appellate:
        …
    else:
      # cmecf_district
```

excessive cognitive load for the reader.

ERosendo: > Ideally we would change the CL docket report template to explain what is going on and make these links look different, both because there is no confirmation step required before incurring charges, and also because they don't directly give the document desired, nor do they clearly set the parameters to limit the report to the minimum cost (which I'm not sure is even possible in ACMS like it is in CM/ECF).

We can use a modal to let users know about the issue with the links for the Download Confirmation Page in ACMS. This is roughly what the modal should look like:

```
+-------------------------------------------------------+
|  ACMS Download Page                                   |
|                                                       |
|  This document is filed in the new  Appellate Case    |
|  Management System (ACMS)</strong>, and direct access | 
|  to the Download Page using links (as in CMECF) is    |
|  not possible, clicking this link will take you to    |
|  the Docket Report Page.                              |
|                                                       |
|  This action will incur an additional charge on your  |
|  account, but loading the docket report is an         |
|  essential step to download documents.                |
|                                                       |
|                                 [$ CONTINUE TO ACMS]  |
|                                                       |
+-------------------------------------------------------+
```

Here's an illustrative GIF showing the interaction with the modal:

acms modal

johnhawkinson: > That those are not the same UUIDs, nor are they obviously related:
> 
> ```shell
> jhawk@lrr ~ % uuid -d c67342f3-1484-ee11-8179-001dd8087c50
> jhawk@lrr ~ % uuid -d 95f6e640-1584-ee11-8179-001dd8087c50

err...clearly they **are** related! The first 5 octets are different (and the 5th differs by -0x01?) and the remaining 11 octets are the same. _Hrmmmmmmmmmmmm_.

ERosendo: > ```html
> <body>
> <h2>You have already used your free look. You are being redirected to PACER to access this document.</h2>
> <h3>If you are not redirected, please click on the following <a href="https://ecf.ca9.uscourts.gov/n/beam/servlet/TransportRoom?servlet=CaseSearch.jsp">link</a>.</h3>
> </body>
> </html>
> ```

I thought We could get more data from the link on Free Look Error but it is not helpful. Why redirect users to the **Case Search** page when they are trying to access a specific document? 🤔

johnhawkinson: _"I had high hopes during this metaphorical pre-breakfast roller coaster ride, but they did not succeed."_

mlissner: @flooie is going to go on a quest to see if he can get in touch with the folks at the ninth circuit. 

Failing that, I think my proposal is something like the modal, except it doesn't have the continue to PACER button. Instead, it just tells you that you can't buy individual documents, and tells you to use the buy docket button at the top of the page. 

The advantage of that is that people will get educated quickly, and then will learn not to click the ACMS buttons anymore, so they'll just learn a new workflow.

johnhawkinson: > The advantage of that is that people will get educated quickly, and then will learn not to click the ACMS buttons anymore, so they'll just learn a new workflow.

Why is this an advantage? And do you think that learning will somehow include an ability to distinguish ACMS dockets from CMECF dockets between district courts and appeals courts, or will they just learn "the button is unreliable, I have to spend money for a consistent result."

It doesn't seem like that is what we want.

mlissner: My thought is that we can make the button slightly different, and then clicking it redirects peolple to go download the docket. So imagine a button on ACMS filings that says, "Direct Link Unavailable" and you click it and see the modal I described.

cl/search/models.py
```diff
@@ -1326,6 +1332,11 @@ class RECAPDocument(AbstractPacerDocument, AbstractPDF, AbstractDateTimeModel):
         ),
         blank=True,
     )
+    acms_document_guid = models.CharField(
+        help_text="The GUID of the document in ACMS.",
+        max_length=64,
+        blank=True,
+    )
```
ttys0dev, line 1339: Shouldn't we have a unique constraint that covers `acms_document_guid` in some way to prevent duplicates?

ttys0dev: I think it may be better to use nulls instead of allowing empty strings here as I don't think multi-column unique constraints would work correctly otherwise for cases where `acms_document_guid` is not used.

cl/search/models.py
ERosendo: @mlissner the button on the docket report is going to look like this for ACMS filings:

Screenshot 2023-11-30 at 11 10 38 AM

Upon clicking the button, the following modal window containing additional information will appear:

Screenshot 2023-11-30 at 11 27 33 AM

ERosendo: > Can you easily make the buttons disabled or gray using Bootstrap?

Yes, I think I can use the `.disable` class.

> Sorry to flip-flop, but I had COVID last time I had opinions about this!

don't worry 😄 . I will implement the suggested changes.

mlissner: Looks great, thank you.

cl/search/models.py
mlissner: @johnhawkinson or @albertisfu, either of you have opinions about this?

mlissner (changes requested):
This was surprisingly straightforward to review, assuming that y'all have made all the right decisions on database modeling. I made a handful of comments, none too crazy, I think, and we should be able to merge this once they're all taken care of.

One thing to check: Do we need to add these fields to any of our APIs?

cl/recap/models.py
```diff
@@ -24,6 +28,7 @@ class UPLOAD_TYPE:
     APPELLATE_CASE_QUERY_PAGE = 13
     CASE_QUERY_RESULT_PAGE = 14
     APPELLATE_CASE_QUERY_RESULT_PAGE = 15
+    ACMS_DOCKET_JSON = 16
```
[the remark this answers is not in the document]
johnhawkinson: Why would that be "best"? A PDF is a PDF, we don't do any source-specific parsing of PDFs (do we?), so why would we want a different endpoint because one came from ACMS and one came from District CMECF?

Similarly, we don't have a different upload type for appellate PDFs versus district PDFs versus bankruptcy PDFs.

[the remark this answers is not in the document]
ERosendo: > and attachment pages?

https://github.com/freelawproject/courtlistener/pull/3393 will add the upload type for attachment pages

cl/opinion_page/templates/includes/de_list.html
```diff
@@ -119,14 +128,23 @@
                     {% else %}
                       {% if rd.pacer_url %}
                         <a href="{{ rd.pacer_url }}"
-                          {% if not request.COOKIES.buy_on_pacer_modal and not request.COOKIES.recap_install_plea %}
+                          {% if '-' in rd.pacer_doc_id %}
+                            class="disabled open_buy_acms_modal btn btn-default btn-xs"
```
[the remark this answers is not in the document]
mlissner: Yes, I think we have a class in override.css for that called `cursor`?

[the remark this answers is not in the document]
ERosendo: Oh you're right! I found the [CSS class](https://github.com/freelawproject/courtlistener/blob/53f9618627c67fbaed6b16efb66fb73e5c0b5bda/cl/assets/static-global/css/override.css#L484C1-L487).

cl/recap/tasks.py
```diff
@@ -1073,6 +1082,198 @@ async def process_recap_appellate_docket(pk):
...
+        if newly_enqueued:
+            await sync_to_async(send_alert_and_webhook.delay)(d.pk, start_time)
+    await mark_pq_successful(pq, d_id=d.pk)
+    return {
+        "docket_pk": d.pk,
+        "content_updated": bool(rds_created or content_updated),
+    }
+
+
+async def process_recap_acms_docket(pk):
```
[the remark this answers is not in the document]
ERosendo: yeah it is repeated.  We should keep the code starting on line 1181 because it uses the new `parse_acms_json` helper.

cl/recap/tasks.py
```diff
@@ -1073,6 +1082,198 @@ async def process_recap_appellate_docket(pk):
...
+        d.originating_court_information = og_info
+    await d.asave()
+
+    # Add the JSON to the docket in case we need it someday.
+    # xxx PacerHtmlFiles is ok for JSON?
+    pacer_file = await PacerHtmlFiles.objects.acreate(
+        content_object=d, upload_type=UPLOAD_TYPE.APPELLATE_DOCKET
+    )
+    await sync_to_async(pacer_file.filepath.save)(
+        "docket.html",  # We only care about the ext w/S3PrivateUUIDStorageTest
```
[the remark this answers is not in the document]
mlissner: No, thanks, you're right.

cl/search/models.py
albertisfu: We currently have the following multi-column constraint on `RECAPDocument`:

```
unique_together = (
            "docket_entry",
            "document_number",
            "attachment_number",
        )
```

I believe this constraint will also help prevent duplicates in ACMS attachments.

The `acms_document_guid` will store the `docketDocumentDetailsId`, which can only be found on the ACMS attachment page and the download confirmation page, so is not available for entries in the docket report, so we cannot use it to prevent duplicates in the main documents only to prevent duplicated attachments.

Screenshot 2023-12-07 at 9 43 22

We also get the "attachment_number" in the ACMS attachment page, which is the `documentNumber` in the previous screenshot, so unless it's possible to have multiple attachments with the same `attachment_number` but different `acms_document_guid` I think including `acms_document_guid`  to the multi-column constraint is not required.

cl/search/models.py
```diff
@@ -1262,12 +1270,10 @@ class AbstractPacerDocument(models.Model):
...
     pacer_doc_id = models.CharField(
-        help_text=(
-            "The ID of the document in PACER. This information is "
-            "provided by RECAP."
-        ),
-        max_length=32,  # Same as in RECAP
+        help_text="The ID of the document in PACER.",
+        max_length=64,  # Increased to support storing docketEntryId from ACMS.
         blank=True,
+        db_index=True,
```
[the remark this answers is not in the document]
albertisfu: Yeah, I think I added this index by mistake in this field, here we only need to increase the length to 64. I removed the index.

cl/search/models.py
```diff
@@ -1361,6 +1372,7 @@ class Meta:
                 fields=["filepath_local"],
                 name="search_recapdocument_filepath_local_7dc6b0e53ccf753_uniq",
             ),
+            models.Index(fields=["acms_document_guid"]),
```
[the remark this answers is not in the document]
albertisfu: Yeah, as described in a previous comment, this index is not required here either. Removed.

cl/search/migrations/0024_search_models_acms_update.sql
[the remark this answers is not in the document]
albertisfu: Sure, I have updated the migrations and applied the optimizations:

`0025_search_models_update.py` with regular tables, event tables, and triggers.
`0025_search_models_update.sql` for our replica, regular tables, and event tables, but no triggers.
`0025_search_models_update_customers.sql` only regular tables for customer replicas.

The other migration `0013_processingqueue_update` didn't require these tweaks, since the PQ model doesn't trigger.

Additionally just to be sure, the wiki mentions:
`New columns With DEFAULT values in Postgresql < v11` 
Here we're adding the new `acms_document_guid` to `RECAPDocument` with the default value `blank=True`
But we're not using Postgresql < v11, so this is not an issue anymore, right?

cl/recap/models.py
```diff
@@ -216,6 +228,9 @@ class Meta:
         permissions = (
             ("has_recap_upload_access", "Can upload documents to RECAP."),
         )
+        indexes = [
+            models.Index(fields=["acms_document_guid"]),
+        ]
```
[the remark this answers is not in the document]
mlissner: Great. Close call. We need to be careful at our scale not to make indexes we don't need. They can be expensive.

cl/search/migrations/0024_search_models_acms_update.sql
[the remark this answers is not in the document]
mlissner: Correct. None of our databases is on < v11 these days, thank goodness. It was horrible for a while there.

mlissner:
Anything else outstanding here?

mlissner:
Sorry, this wasn't a great call to action and surely suffers from diffuse responsibility:

> Anything else outstanding here?

Let me try this another way. @ERosendo, @albertisfu, @johnhawkinson can each of you give this your approval or not, so I can merge (or not)?

ERosendo (approved):

albertisfu (approved):
I've updated the branch with main. This seems ready to me!

mlissner:
@johnhawkinson do you want to give this another look or should we just go for it?

ERosendo:
@mlissner :shipit:

mlissner (approved):

mlissner:
Merged! I agree with Eduardo that this is cooked enough. If we need more changes once this hits prod, we'll do that, but let's get this big thing under our belt and move forward.

@blancoramiro, this will need some clever deployment. Let's see if we can do that today or tomorrow, since all other deploys will be blocked until we get it in (but no urgent code is expected).

mlissner:
Thank you all for your contributions here. This one was a real team effort.

Issue #3194, opened 2023-09-28
`xml_harvard` isn't searchable in the opinions index


mlissner:
When we added this field, we didn't update Solr to make sure we index it, so the result is that Solr doesn't have this text. 

In `opinion_text.txt`, we have the following:

```jinja
{# The body of the item (columbia > lawbox > html > plaintext) #}
{% load text_filters %}
{% if item.html_columbia %}
    {{ item.html_columbia|striptags|html_decode }}
{% elif item.html_lawbox %}
    {{ item.html_lawbox|striptags|html_decode }}
{% elif item.html %}
    {{ item.html|striptags|html_decode }}
{% else %}
    {{ item.plain_text }}
{% endif %}
```

But it's missing `xml_harvard`. We need to add that and then we need to re-index anything that has a value in that field. A potential customer identified this issue, and so it'd be nice to get this fixed ASAP.

mlissner:
This is particularly bad on items that only have Harvard text. In that case, the item doesn't even come up for text queries. For example, searching for this case by ID works (1 result):

https://www.courtlistener.com/?q=id%3A6213042&type=o&order_by=score%20desc&stat_Precedential=on

But including almost any keyword at all doesn't (0 results):

https://www.courtlistener.com/?q=id%3A6213042%20private&type=o&order_by=score%20desc&stat_Precedential=on

ERosendo:
This will be fixed after merging https://github.com/freelawproject/courtlistener/pull/3043. I noticed this issue too and i added that field to the prepare method to store the text in the opinion document.

Here's the code for the prepare method:

```python
    def prepare_text(self, instance):
        if instance.html_columbia:
            return html_decode(strip_tags(instance.html_columbia))
        elif instance.html_lawbox:
            return html_decode(strip_tags(instance.html_lawbox))
        elif instance.xml_harvard:
            return html_decode(strip_tags(instance.xml_harvard))
        elif instance.html_anon_2020:
            return html_decode(strip_tags(instance.html_anon_2020))
        elif instance.html:
            return html_decode(strip_tags(instance.html))
        else:
            return instance.plain_text
```

mlissner:
Nice. Alas, I don't think we can wait for that...but when do you think it'll be ready?

ERosendo:
There is only one thing left to do: make `MoreLikeThis` queries work properly. I think I'll be able to finish it today :).

flooie:
yikes.

flooie:
[...]

PR #3220, opened 2023-10-03
fix(opinion_text.txt): Fix harvard search bug

flooie:
PR Fixes issue #3194 

Tweak opinion_text.txt file to include xml_harvard Without this and the future re-index harvard data
is not actually searchable in solr.


flooie:
Once approved, a simple re-index of affected opinion object ids to solr should resolve this bug.

mlissner:
Looks good to me, thanks. Setting for auto-merge.

Issue #3192, opened 2023-09-27
Put (Solr | Elastic) behind a setting for (prod | dev)


mlissner:
It'd sure be nice if we could turn off Solr during dev or when the machine is down. I think we could just have a setting that disables it (`SOLR_DISABLED`?). Once that was in place, anything that calls to Solr could just have an `if/else` statement that runs first. 

The hard part of this will probably be handling the disabled state. Adding the setting is easy, but if Solr is disabled, we'll want to be smart about how the error is displayed. I guess if this is just for developers, we can be lazy and not do a great job. If it's for prod, we'd want to do a better job and show errors to the users. I'm fine with either way. The latter is more work and we could do it later (or never, since we're switching to Elastic, though we'd probably want this for Elastic someday too!).

Anyhow, happy to be pragmatic, I think, is what I'm saying!

PR #3223, opened 2023-10-03
added health check for solr

ss108:


ss108:
@mlissner initial work for #3192 

Adds a health check to Solr. I think it's best to proceed query-by-query in leveraging this functionality because doing so requires writing analogous Django ORM queries to stand in for every Solr query. It's not that crazy, but I think it would be imprudent to try to do all at once.

This PR adds such for opinion page because I need it to help with front-end work for RECAP citations.

mlissner (commented):
Thanks for looking at this. I think since Solr really is going away soon, and since it's generally pretty reliable, it's better to do the really simple thing and just have an ENV for developers to toggle. 

I'd also say that when the ENV's toggled, we should always throw an error or an empty set or something, instead of trying to make it work anyway. I mean, we could do the work of putting lots of "use Solr if available, else use Postgres" code around, but it's just going to get ripped out in a couple weeks. In most cases where such a substitution is possible, we used Solr for performance reasons too, so I'm a little scared that Solr could go down, and then the next thing we know postgres is struggling too.

I don't know — it feels better to keep this simple for now and to maybe do it for Elastic once Solr is ripped out?

Issue #3222, opened 2023-10-03
Coverage page follow up tasks


mlissner:
A few things I"m seeing and thinking about wrt to the new coverage page:

 - [x] **Performnace**: Right now loading the page itself seems to be solved, but loading the graphs is bad. I don't think we can fix it using the database, but fear not, there are solutions:
    - [x] We need to provide a spinner in the button.
    - [x] We can switch to using Solr/Elastic for the queries. I think it can do a court in about half a second, which should be a lot better than the DB.
    - [x] We can cache each court or each API response. Even though Solr will be faster, caches will still help.

 - [x] **Mobile**: The sidebar is missing on mobile, which is too bad, and the graphs don't work. I think we can just hide the graph buttons on mobile (sorry), and maybe we can do a mini-TOC that only shows up on mobile. It could just have State, Federal, etc. headings.

 - [x] **Architecture**: I didn't catch it before now, but the API shouldn't take the name of the court as a parameter. It should use IDs. Names shouldn't be used for this kind of thing.

 - [x] **Content**: We need to add the text from the Google doc.

What else?

flooie:
Performance performance performance -

mlissner:
The latest list:
 - [ ] The API needs to be done as `?courts=abc,foo,bar` (right now it returns 413 responses b/c the requests are too big.
    - [ ] Check performance again after this
 - [ ] Let's collapse the sidebar by default, so it's better on mobile.
 - [ ] The main content should have a margin at right, like other help pages.
 - [ ] Add `class="lead"` to the first paragraph
 - [ ] Remove `if query_parameters` from the view.
 - [ ] Fix links from charts to search results (they do links like: https://www.courtlistener.com/?court=248)

mlissner:
- [ ] Nuke the purple line from the bottom of every chart.
 - [ ] Add scrapers back to the chart
 - [ ] Remove the current info from the coverage page, and replace it with a link (though there may be info on that page that's worth continuing to say?)

mlissner:
- [ ] The X button is weird in Florida:

    ![image](https://github.com/freelawproject/courtlistener/assets/236970/f37f9690-f931-458c-b51a-4d91a870951e)

flooie:
what is weird about the x button?

mlissner:
It's all the way down by the line?

mlissner:
- [ ] Add this to our help page: https://www.courtlistener.com/help/

mlissner:
[...]

PR #3227, opened 2023-10-04
3222 coverage follow up

flooie:
PR to address issue #3222 

Updates for 

1. Performance enhancements using SOLR
2. Mobile - fixes for TOC and charts (hidden)
3. Architecture - dropped sending court names to API 
4. Content: has been added

Did not add spinner (hoping speed resolves that) 
Same for caching.  I think this may be fast enough.

### DIFF
--- a/cl/api/views.py
+++ b/cl/api/views.py
@@ -206,7 +206,7 @@ def coverage_data_opinions(request: HttpRequest):
                         "rows": "1",
                         "start": "0",
                         "fq": f"court_exact:{court_id}",
-                        "fl": "dateFiled",
+                        "fl": "dateFiled,court",
                         "sort": "dateFiled desc",
                     }
                     items_desc = (
@@ -228,7 +228,7 @@ def coverage_data_opinions(request: HttpRequest):
                     ]
                     court_data = {
                         "id": court_id,
-                        "label": court_name,
+                        "label": items_asc[0]["court"],
                         "data": [
                             {
                                 "val": total,
--- a/cl/api/views.py
+++ b/cl/api/views.py
@@ -196,7 +196,7 @@ def coverage_data_opinions(request: HttpRequest):
         group_dict = dict(Court.JURISDICTIONS)
         with Session() as session:
             solr = ExtraSolrInterface(
-                settings.SOLR_OPINION_URL, http_connection=session, mode="r"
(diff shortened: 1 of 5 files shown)


cl/api/views.py
mlissner, line 1: This looks pretty OK to me, but could we move the `if query_parameters` outside the session context handler and change it to an `if not` statement that early aborts the function? That'd help with indentation.

I'd also like one of our search gurus (we have those now!) to take a look. @albertisfu, does the search part of this PR look OK to you?

cl/assets/static-global/css/override.css
```diff
@@ -1564,3 +1564,11 @@ textarea {
...
   height: 100vh;
 }
+
+@media (max-width: 780px) {
+    #toc-container {
+        overflow-y: initial;
+        position: relative;
+        height: min-content;
+    }
+}
```
mlissner, line 1574: Can you explain what this does and maybe post some screenshots, please?

mlissner (commented):
Looks pretty good. Why not add caching though? We know these are still pretty mean queries. Caches would make life a lot faster and are easy to add?

cl/assets/static-global/css/override.css
flooie: This handles the table of contents displaying on top of the other content.  

On mobile we want the table of contents to just work and not be a scrollable list- so we need to stretch it out.

Ill add some photos

flooie: Screenshot 2023-10-04 at 3 49 14 PM

This is attempting to resolve this on mobile 


Screenshot 2023-10-04 at 3 49 29 PM


This removes the sticky sidebar - and the overlapping text issue and makes it much more usable.

cl/api/views.py
flooie: @albertisfu - thanks for taking a look at this.  I look forward to your thoughts.  I can move the if statement around but we really shouldn't be seeing this very often where we need to short circuit the statement

mlissner (approved):

mlissner:
@ERosendo, just a heads up, we added a new function here that hits Solr. I'm guessing you'll want to convert it to an Elastic query.

cl/api/views.py
albertisfu: Yeah, sorry for not having looked at this earlier. I just reviewed it in a call with Mike, and the query seems pretty right to me.

PR #3234, opened 2023-10-06
Add new court fields to clone command

quevon24:
I updated the clone command to copy the two new fields of Court model: 

- parent_court (a fk to Court table)
- appeals_to (m2m to Court table)

I also fixed an error when cloning a docket when referred_to field has a value.


mlissner:
LGTM. Thanks for continuing to evolve this.

PR #3240, opened 2023-10-11
Update harvard merger to merge remaining files

quevon24:
It took me a while to discover the best way to solve this problem to try to merge the remaining 5k~6k files.

This small change improves the process that tries to match the opinions, we now try to make both texts in equal conditions to prove if both are similar or equal removing non-alphanumeric and non-whitespace characters. This is useful because sometimes capitalized or uppercased words can affect the score of both algorithms. For example, in cluster id: 3647775 we have in the content: 

`Concur: LOUGHRAN, Ch. J., LEWIS, CONWAY, DESMOND, THACHER, DYE and FULD, JJ.`

but in json file we have:

`Concur: Loughran, Ch. J., Lewis, Conway, Desmond, Thacher, Dye and Fuld, JJ.`

It's almost the same except from the uppercased words, and that changes the score enough to say that they are not similar or equal. This was unusual and happened in a small part of the files.


flooie:
This seems like a simple clean fix thank you. @quevon24

flooie (commented):
A simple fix for most of the remaining cases.  Thanks

Issue #3134, opened 2023-09-12
Consolidate disparate images using ENTRYPOINT


cweider:
Within [Dockerfile](https://github.com/freelawproject/courtlistener/commit/a2fbbabfd52f99791e2738360996d7c86bc51de3), there are five “leaf-images”: `celery`, `web-dev`, `web-prod`, `rss-scraper`, and `retry-webhooks`. After #3107 and #3130, it’s apparent these images are minimally differentiated. They only vary in the `CMD` instruction used:

```
FROM python-base as celery

USER www-data
CMD celery
    --app=cl worker \
    #...


FROM python-base as web-dev

USER www-data
CMD python /opt/courtlistener/manage.py migrate && \
    python /opt/courtlistener/manage.py createcachetable && \
    python /opt/courtlistener/manage.py runserver 0.0.0.0:8000


FROM python-base as web-prod

USER www-data
CMD gunicorn cl.asgi:application \
    --chdir /opt/courtlistener/ \
    #...


FROM python-base as rss-scraper

USER www-data
CMD /opt/courtlistener/manage.py scrape_rss


FROM python-base as retry-webhooks

USER www-data
CMD /opt/courtlistener/manage.py cl_retry_webhooks
```

This is valid, but it is not the only way to accomplish their purpose of making it possible create differently behaving containers from one common image (here, `python-base`). The other conventional approach is a single image containing a helper script (e.g. `docker-entrypoint.sh`) that can be used to launch the different processes depending for different types of containers. To support this pattern, Docker has the [`ENTRYPOINT`](https://docs.docker.com/engine/reference/builder/#entrypoint) instruction.

In Courtlistener’s case, `ENTRYPOINT` would be especially helpful because it would drastically simplify image management. There’d many fewer images to build, cache, tag, and push (see [`Makefile`](https://github.com/freelawproject/courtlistener/commit/722d8626ffe051ba33aa470fabb047476cf6d4d8)). Chances are that `Makefile` could be replaced with a `build-push-action` after this change.

A less recognized benefit of this change would be that the different processes run on a container’s start would be centralized within `docker-entrypoint.sh`. This makes the behavior more visible as well as accessible for testing.

Making this change requires two coordinated bits of work:
1. Shifting the behavior in `CMD` instructions to an entrypoint script. There are some [best practices](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#entrypoint) that would be worth reading.
[...]

PR #3260, opened 2023-10-15
Docker 3134

ss108:
This PR is for #3134 

Now, instead of using the multi-stage build paradigm to produce several different images from the same codebase which differ only in the `CMD` they run,  we only have one image built from the `courtlistener` codebase. 

In order to run a container providing for the different kinds of functionality we want (e.g. Celery workers as opposed to normal Django web app), we can simply pass an argument when running the container. This decouples the process of building the image from specifics pertaining to running it in a certain way and provides other benefits alluded to in #3134 

Because we are using `docker compose`, the development workflow remains unchanged.


ss108:
@mlissner We still need to modify the configs for each deployment to start/run the specified containers with the appropriate arguments.

My understanding is that, for example, the one for the main Django deployment would look something like this:

```apiVersion: apps/v1
kind: Deployment
metadata:
  name: cl-python
spec:
  replicas: 1
  template:
    spec:
      containers:
      - name: web
        image: freelawproject/courtlistener:latest
        args: ["web-prod"]
```

ss108:
Also, note that now `collectstatic` and the migration check are separate steps. They occur in a temporary pod that self-deletes after an hour (whether or not there is any error). Thus, you can open a shell/terminal session in that pod to run migrations, and you don't need to worry about deleting it--the cleanup will occur automatically.

docker/django/docker-entrypoint.sh
```diff
@@ -0,0 +1,44 @@
...
+
+case "$1" in
+'celery')
+    exec celery \
+        --app=cl worker \
+        --loglevel=info \
+        --events \
+        --pool=prefork \
+        --hostname=prefork@%h \
+        --queues=${CELERY_QUEUES} \
```
mlissner, line 12: How certain are you that this variable will get picked up somehow?

mlissner (commented):
This looks like a *huge* improvement that should bring our build and deploy times down considerably. Thank you! 

One small comment below, and then I'd love to get @cweider's review, if he's able and interested. 

This must have been a tough PR. How did you get so danged far without having access to k8s, for example?

ss108:
> This looks like a _huge_ improvement that should bring our build and deploy times down considerably. Thank you!
> 
> One small comment below, and then I'd love to get @cweider's review, if he's able and interested.
> 
> This must have been a tough PR. How did you get so danged far without having access to k8s, for example?

Thanks. Do you happen to know what the current average times are for build and/or deployment?

Re: k8s, well, there is still one part left to be done that can't be done without access to k8s, which is actually changing the way the containers are run to use the entrypoint commands. At that stage, you can also set any necessary env variables (see other comment).

docker/django/docker-entrypoint.sh
ss108: It basically depends on how you are currently setting these variables. One way to set them would be, depending on how the k8s stuff actually looks:

```kind: Deployment
metadata:
  name: cl-celery
spec:
  replicas: 1
  template:
    spec:
      containers:
      - name: worker
        image: freelawproject/courtlistener:latest
        args: ["celery"]
        env:
        - name: CELERY_QUEUES
          value: "<whatever>"
        
        ```
        
 But yeah, it's up to you. I think this Kubernetes manifest YAML way is probably the easiest, but I think you can also utilize a lot of other AWS secret management options that strike me as unnecessarily complicated here.

ss108:
Cool. Let me know if you need any further help along the way.

> I think the only way to figure this out is to use the Github API to download all of the runs of the task and do some processing on them. Too much work to really mess around with, I fear.

I will look into it.

mlissner:
> I will look into it.

FWIW, I suspect the results will be pretty obvious when this goes live for a week or so. Eyeballing it, you can usually see if things are getting faster, and I think this change should be enough to make a noticeable difference. So I guess, ya know, go for the API if you're interested, but I'm not sure it'll be necessary.

ss108:
Yeah I couldn't find any tools that did this, though I did find one could be modified to do it.

Will probably just average the last 3-5 builds pre-deploy and the first 3-5 builds after, and call it a day lol

mlissner:
I'm going to give this a go. Wish me luck!

PR #3264, opened 2023-10-16
feat(settings): Add a variable to tweak the ES timeout

ERosendo:
This PR adds a variable to tweak the ES connection's read timeout and sets its default value to 30 seconds. 

This PR should help us lower the number of `ConnectionTimeout` exceptions we get (like https://github.com/freelawproject/courtlistener/issues/3262 and https://github.com/freelawproject/courtlistener/issues/3253) while indexing RECAP documents.


albertisfu (approved):
Great, yeah this seems good to me, it should help to reduce the number of tasks that are timing out and we'll see if some of them still require a greater time out.

mlissner:
Thanks. I appreciate an easy PR!

PR #3282, opened 2023-10-18
fix(search): Tweak order of filters applied to the court_string

ERosendo:
This PR tweaks the search_result.html template and fixes the issue related to the highlighting in the court field.

The problem arose because the `nbsp` filter first escaped the HTML content of the string and then marked the output as safe. As a result, the HTML tags used for highlighting were displayed as text. This pull request changes the order of the filters to prevent the escaping of the HTML content.


albertisfu (approved):
Yeah, this tweak solves the issue in the court string in the results headings.
Thanks!

PR #3299, opened 2023-10-24
fix(elasticsearch): Add the max_analyzed_offset setting to the HL query

ERosendo:
This pull request introduces the max_analyzed_offset setting to the highlight query and resolves the issue outlined in https://github.com/freelawproject/courtlistener/issues/3257#issuecomment-1777235639.


mlissner:
Automerge set. Seems like a good solution for now.

mlissner:
Tested and seems to be working.

PR #3301, opened 2023-10-24
fix(elasticsearch): Add filter to remove results with no children doc

ERosendo:
This PR fixes the issue with the filtering checkbox to display only results with PDFs as described in this [comment](https://github.com/freelawproject/courtlistener/issues/3257#issuecomment-1763430663).


albertisfu (approved):
Great, this is correct, the new filter in the parent document query resolves the issue. Now, queries requesting to show only documents with PDFs will exclude dockets that have no documents with PDFs from the results.

mlissner:
No tests, gentlemen?

ERosendo:
@mlissner Sorry! I have updated the test for the "available_only" field.

mlissner (approved):

PR #3317, opened 2023-10-28
Recap cit frontend

ss108:


ss108:
@mlissner 

This can now be reviewed. I cut it off at the "view all authorities" button; that will be suppressed for RECAP Docs right now. The reason is that this is already a lot of changes, and also I wanted to ask what the "blocked" stuff is and how to handle it for RECAP docs:

```{% block sidebar %}
    <div class="col-sm-3" id="sidebar">
        <div class="sidebar-section">
            <h4 class="bottom">
                <i class="fa fa-arrow-circle-o-left gray"></i>
                <a href="{{ cluster.get_absolute_url }}?{{ request.META.QUERY_STRING }}"
                   {% if cluster.blocked %}rel="nofollow"{% endif %}>Back to Opinion</a>
            </h4>
        </div>
    </div>
{% endblock sidebar %}```

![enjoy-recap-auth](https://github.com/freelawproject/courtlistener/assets/5439659/d98884c7-2907-4bcc-b0fa-2e4533fc7f9e)

File: cl/opinion_page/utils.py
Comment (mlissner) [inline cl/opinion_page/utils.py:1]: This looks like an errant file?

File: cl/opinion_page/templates/includes/authorities_section.html [hunk 1]
```diff
@@ -0,0 +1,33 @@
+{% load humanize %}
+{% load extras %}
+{% load humanize %}
+{% load static %}
+{% load text_filters %}
+
+    {# Show cases this case cites #}
+        {% if authorities_context.total_authorities_count > 0  %}
```
mlissner, line 8: Nit: I'd outdent these lines all the way to the left (and subsequent lines accordingly):

```suggestion
{# Show cases this case cites #}
{% if authorities_context.total_authorities_count > 0  %}
```

And I guess while we're at it, we should do two spaces, not four, as per the `.editorconfig` file.

cl/opinion_page/templates/includes/authorities_section.html
```diff
@@ -0,0 +1,33 @@
...
+            <div id="authorities" class="sidebar-section">
+                <h3><span>Authorities ({{ authorities_context.total_authorities_count|intcomma }})</span></h3>
+                <p class="bottom">This opinion cites:</p>
```
mlissner, line 11: Oops. This will need to be tweaked for RECAP docs.

cl/opinion_page/templates/includes/authorities_section.html
```diff
@@ -0,0 +1,33 @@
...
+                            <a href="{{ authority.url }}">
+                                {{ authority.caption|safe|truncatewords:10|v_wrapper }}
+                            </a>
+                            {% if authority.count%}
+                              <span class="small gray">({{ authority.count}}&nbsp;time{{ authority.count|pluralize }})</span>
+                            {% endif %}
+                        </li>
+                    {% endfor %}
+                </ul>
+                {% if authorities_context.view_all_url|length > 0 %}
```
mlissner, line 24: I'm surprised this wouldn't work: 

```suggestion
                {% if authorities_context.view_all_url %}
```

cl/opinion_page/types.py
mlissner, line 1: I don't think this quite fits in a `types.py` file. @ERosendo is good at organizing files in Django projects. Eduardo, do you have a better suggestion?

cl/opinion_page/types.py
```diff
@@ -0,0 +1,80 @@
...
+        total_cit_count: int,
+        cit_records: List[OpinionsCitedByRECAPDocument],
+        request_query_string: str,
+    ):
+        top_authorities = [
+            ViewAuthority.from_recap_cit_record(
+                r, query_string=request_query_string
+            )
+            for r in cit_records
+        ]
```
mlissner, line 73: I _think_ this means that we make a bunch more queries than before, but I could be wrong. It could be that the queries were just better hidden in the template and they're more obvious in the view. I do see that in your screenshot, the page took 20 queries to load, which feels like a lot. I suspect this accounts for a bunch of them.

In any case, now that we see them, I wonder if we could refactor this to use a `select_related` or `prefetch_related` for this and the same pattern you have above?

cl/opinion_page/types.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+@dataclass
+class AuthoritiesContext:
+    total_authorities_count: int
+    top_authorities: List[ViewAuthority]
+    view_all_url: str
+
+    @classmethod
+    def from_opinion_cluster(
```
mlissner, line 45: I wonder if you could merge this and the method below by sniffing the object type passed in. Did you play with that?

mlissner (commented):
I did a quick skim to get things started, and made a few quick comments for you @ss108. I'm really excited to see this. It seems like the next steps are to:

 - Make the show-all button and the page to show its results?
 - Add a list of authorities to each docket as a tab next to the parties?
 - I don't think it's useful to make these searchable, is it?

@ERosendo, can you please make this your fourth priority, after Opinions, ACMS, and the Psycho-cases bot?

Thank you again Samier. So cool to finally see this. Did the new dev DB help?

ERosendo:
@mlissner sure!

ss108:
@mlissner 

> Did the new dev DB help?

Yeah, thanks again for setting it up. It was essential for testing this. 

I'll start tackling the comments tomorrow.

cl/opinion_page/types.py
ss108: It's only 2 queries in the function that gets the citation records for a recap doc: https://github.com/freelawproject/courtlistener/pull/3317/files#diff-a8a8b6a59162fd73eea97d6b21c17656610182629a4302eaaa54d5432b1a9c62R63

Confirmed by expanding the queries debug thingy.

Some other stuff in the view function seems to be generating extra queries that are susceptible to optimization, e.g., https://github.com/freelawproject/courtlistener/blob/main/cl/opinion_page/views.py#L342 (`select_releated`) 

There are some lookups of the citation table by `cluster_id` too, but it's not immediately clear to me where those are coming from tbh

cl/opinion_page/types.py
ss108: @mlissner current version pushed now has an attempt of this, with initial version commented out.

ss108: @mlissner nudge

mlissner: Sorry, I'm leaving this for @ERosendo to review, but he has a few higher priority items on his list first. We're really bogged down here at FLP, but we're digging out. I'm delegating more dev work to others, and we are just really behind on Elastic, which is causing everything else to suffer. 

Anyway, that's TMI, perhaps, but I expect Eduardo will have a minute for this soon.

ss108: Sounds good

cl/opinion_page/types.py
ERosendo: > I think this means that we make a bunch more queries than before, but I could be wrong.

@mlissner You're right! Excluding the authorities section reduces the number of queries required to load the page to just **5**. Here's a screenshot of the Django debug toolbar when we load the page using the code from the `main` branch:

image

Using the `select_related` and `prefetch_related` methods and carefully selecting only the essential fields from the query, I successfully reduced the number of database queries to **7**. Here's a screenshot of the debug toolbar after the refactor:

Screenshot 2023-11-30 at 1 48 17 PM

ERosendo:
Here's a summary of the changes implemented in this pull request:

1. Updates the indentation of the template for the authorities section, It's using two spaces now.
2. Refine the query to get RECAP citations. 
    
     Optimize the ORM query by employing the select_related and prefetch_related methods and selecting only the relevant fields to display the list of citations associated with a particular RECAPDocument.

3. Add properties to query Citations from the RECAP and Docket instances 

    Aligns the citation retrieval methods with corresponding model properties. Additionally, it mirrors the citation handling approach we use in the `Opinions` class, enabling interoperability between instances of these classes for citation retrieval tasks.

4. Remove constructor classes from dataclasses.

    Removes unnecessary complexity by removing the `RECAPDocCitationRecord` dataclass and the class methods from the `ViewAuthority` dataclass. Additionally, enhances the `AuthoritiesContext` dataclass by introducing two init-only fields and replacing the construct method with the `post_init()` method, simplifying object initialization.

5. Adds a new tab to the document page to display the list of authorities cited by the document.

image


6. Adds a new tab to the docket report to display the list of authorities.

image

cl/opinion_page/templates/docket_authorities.html
```diff
@@ -0,0 +1,41 @@
...
+non-profit dedicated to creating high quality open legal information.{% endblock %}
+{% block og_description %}Authorities for {{ title }} — Brought to you by the RECAP Initiative and Free Law Project, a
+non-profit dedicated to creating high quality open legal information.{% endblock %}
+
+{% block nav-authorities %}active{% endblock %}
+
+{% block tab-content %}
+<div class="flex flex-column justify-content-center col-xs-12 col-md-10 col-lg-12">
+  <div class="col-xs-12 col-md-10 col-lg-10">
+    <h3>This Docket cites {{docket.authority_count|intcomma}} opinion{{docket.authority_count|pluralize}}.</h3>
```
mlissner, line 17:
```suggestion
    <h3>The filings in this docket cite {{docket.authority_count|intcomma}} opinion{{docket.authority_count|pluralize}}</h3>
```

cl/opinion_page/templates/docket_authorities.html
```diff
@@ -0,0 +1,41 @@
...
+        </a>
+        <br />
+        <span class="bullet-tail">{{ authority.cited_opinion.cluster.docket.court }}</span>
+        <span class="bullet-tail">{{ authority.cited_opinion.cluster.date_filed }}</span>
+        {% if authority.cited_opinion.cluster.citation_count > 1 %}
+        Also cited by <a href="/?q=cites%3A({{ authority.cited_opinion.cluster.sub_opinions.all|OR_join }})"
+          rel="nofollow">{{authority.cited_opinion.cluster.citation_count|add:"-1" }} other opinion{{authority.cited_opinion.cluster.citation_count|add:"-1"|pluralize }}</a>
+        {% else %}
+        Cited by 0 other opinions
+        {% endif %}
```
mlissner, line 35: I think we can take this out. I'm not sure how helpful it is, and it's a bit confusing.

cl/opinion_page/templates/docket_authorities.html
mlissner, line 1: The code here feels duplicated. Is it?

cl/opinion_page/templates/recap_authorities.html
```diff
@@ -0,0 +1,38 @@
...
+{% load text_filters %}
+
+{% block nav-pdf %}{% endblock %}
+{% block content-pdf %}{% endblock %}
+{% block nav-authorities %}active{% endblock %}
+
+{% block tab-content %}
+  <div role="tabpanel" class="tab-pane active" id="text">
+    <div id="authorities">
+      <h3>This {{authorities_context.doc_type}} cites {{ authorities_context.total_authorities_count|intcomma }} opinion{{ authorities_context.total_authorities_count|pluralize }}.</h3>
```
mlissner, line 12:
```suggestion
      <h3>This {{authorities_context.doc_type}} cites {{ authorities_context.total_authorities_count|intcomma }} opinion{{ authorities_context.total_authorities_count|pluralize }}</h3>
```

cl/opinion_page/templates/recap_authorities.html
```diff
@@ -0,0 +1,38 @@
...
+                  </a>
+                  <br/>
+                  <span class="bullet-tail">{{ authority.cited_opinion.cluster.docket.court }}</span>
+                  <span class="bullet-tail">{{ authority.cited_opinion.cluster.date_filed }}</span>
+                  {% if authority.cited_opinion.cluster.citation_count > 1 %}
+                      Also cited by <a href="/?q=cites%3A({{ authority.cited_opinion.cluster.sub_opinions.all|OR_join }})"
+                                        rel="nofollow">{{ authority.cited_opinion.cluster.citation_count|add:"-1" }} other opinion{{ authority.cited_opinion.cluster.citation_count|add:"-1"|pluralize }}</a>
+                  {% else %}
+                      Cited by 0 other opinions
+                  {% endif %}
```
mlissner, line 31: Again, we can remove this, and we shouldn't repeat this code.

cl/opinion_page/templates/recap_document.html
```diff
@@ -60,6 +60,7 @@ <h3 class="inline"><span>Your Notes</span></h3>
             </div>
         </div>
 
+        {% include "includes/authorities_section.html" with authorities_context=authorities_context%}
```
mlissner, line 63: If we have the tab, we don't need the sidebar.

cl/opinion_page/templates/recap_document.html
```diff
@@ -100,9 +101,16 @@ <h4>Description</h4>
                  aria-controls="profile"
                  role="tab"
                  data-toggle="tab">Text</a></li>
+          {% if authorities_context.total_authorities_count %}
+            <li class="{% block nav-authorities %}inactive{% endblock %}" role="presentation">
+              <a href="{{ authorities_context.view_all_url }}">
+                Authorities
+              </a>
```
mlissner, line 108: Since we already know the count, why not do:

```suggestion
              <a href="{{ authorities_context.view_all_url }}">
                {{ authorities_context.total_authorities_count|intcomma }}Authorit{{ authorities_context.total_authorities_count|pluralize:"y,ies" }}
              </a>
```

mlissner: Actually, sorry, I take this back. Wouldn't it be better and more performant to just do an `exists()` call to see if there are authorities, instead of getting and counting them (like we do on the entries and parties tabs)?

cl/opinion_page/views.py
```diff
@@ -465,6 +487,32 @@ def view_recap_document(
...
+        )
+    else:
+        view_authorities_url = reverse(
+            "view_document_authorities",
+            args=[
+                rd.docket_entry.docket_id,
+                rd.document_number,
+                rd.docket_entry.docket.slug,
+            ],
+        )
```
mlissner, line 508: I think we can do this in the template, can't we? I think that's where we'd normally figure out the URL for something?

cl/opinion_page/views.py
```diff
@@ -465,6 +487,32 @@ def view_recap_document(
...
+    authorities_context: AuthoritiesContext = AuthoritiesContext(
+        citation_record=rd,
+        query_string=request.META["QUERY_STRING"],
```
mlissner, line 511: We usually have this in the template, don't we?

cl/opinion_page/views.py
```diff
@@ -465,6 +487,32 @@ def view_recap_document(
...
+        total_authorities_count=rd.authority_count,
+        view_all_url=view_authorities_url,
+        doc_type="document",
```
mlissner, line 514: This doesn't look like it's ever another value. Do we need this parameter or can we hardcode it in the template?

cl/opinion_page/views.py
```diff
@@ -479,6 +527,87 @@ def view_recap_document(
...
+        )
+    else:
+        view_authorities_url = reverse(
+            "view_document_authorities",
+            args=[
+                rd.docket_entry.docket_id,
+                rd.document_number,
+                rd.docket_entry.docket.slug,
+            ],
+        )
```
mlissner, line 585: I'd suggest moving this to the template too, but if you think it's better here, let's definitely not repeat it. DRY!

cl/opinion_page/views.py
```diff
@@ -479,6 +527,87 @@ def view_recap_document(
...
+        {
+            "rd": rd,
+            "title": title,
+            "og_file_path": og_file_path_override,
+            "note_form": note_form,
+            "private": True,  # Always True for RECAP docs.
+            "timezone": COURT_TIMEZONES.get(
+                rd.docket_entry.docket.court_id, "US/Eastern"
+            ),
+            "redirect_to_pacer_modal": False,
```
mlissner, line 609: Do we need this logic here? Could we remove it?

mlissner (changes requested):
Some comments for you, @ERosendo, thank you. Too much repeated code, but the general ideas and design look good. 

The one thing I didn't feel good about reviewing is the queries in the models. @albertisfu, could you give the changes in models.py a look and see if they look good to you?

albertisfu:
@mlissner @ERosendo 

I checked the new queries in the models and they seem well-optimized.

Just a couple of suggestions:

The property `authorities_with_data` in Docket and RECAPDocument uses almost the same query. So we could avoid repeat code using a helper method like:

```
def build_authorities_query(base_queryset):
    return (
        base_queryset
        .select_related("cited_opinion__cluster__docket__court")
        .prefetch_related(
            "cited_opinion__cluster__citations",
            Prefetch(
                "cited_opinion__cluster__sub_opinions",
                queryset=Opinion.objects.only("pk", "cluster_id"),
            ),
        )
        .only(
            "depth",
            "citing_document_id",
            "cited_opinion__cluster__slug",
            "cited_opinion__cluster__case_name",
            "cited_opinion__cluster__case_name_full",
            "cited_opinion__cluster__case_name_short",
            "cited_opinion__cluster__citation_count",
            "cited_opinion__cluster__docket_id",
            "cited_opinion__cluster__date_filed",
            "cited_opinion__cluster__docket__docket_number",
            "cited_opinion__cluster__docket__court_id",
            "cited_opinion__cluster__docket__court__citation_string",
            "cited_opinion__cluster__docket__court__full_name",
        )
        .order_by("-depth")
    )

```

And then do something like:

```
def authorities_with_data(self):
        base_queryset = OpinionsCitedByRECAPDocument.objects.filter(
            citing_document__docket_entry__docket_id=self.pk
        )
        return build_authorities_query(base_queryset)
```

and

```
    def authorities_with_data(self):
        return build_authorities_query(self.cited_opinions)
```
        
- Also in the Docket authorities section the `authority_count` query is duplicated:

Screenshot 2023-12-05 at 17 28 02

We could avoid these repeated queries by using `@cached_property` instead of `@property`

```
from django.utils.functional import cached_property

@cached_property
def authority_count(self):
     return OpinionsCitedByRECAPDocument.objects.filter(
            citing_document__docket_entry__docket_id=self.pk
        ).count()
```

ERosendo:
Thanks @albertisfu, I'll apply your suggestions!

cl/opinion_page/templates/docket_authorities.html
ERosendo: got it! I'm removing this section.

ERosendo: This is how it looks without this code: 

image


let me know what you think

cl/opinion_page/templates/recap_authorities.html
ERosendo: You're right! I removed this part of the code.

cl/opinion_page/templates/recap_document.html
ERosendo: > Wouldn't it be better and more performant to just do an exists() call to see if there are authorities, instead of getting and counting them (like we do on the entries and parties tabs)?

You're correct! I'm updating this template.

cl/opinion_page/views.py
ERosendo: > Could we remove it?

Absolutely!. I'll refactor this view.

cl/opinion_page/views.py
ERosendo: > I think we can do this in the template, can't we?

You're right! We could move this code into a dedicated method in the `RECAPDocument` class, similar to the `get_absolute_url` method.

cl/opinion_page/views.py
```diff
@@ -1,7 +1,8 @@
 import datetime
 from collections import OrderedDict, defaultdict
+from dataclasses import dataclass
 from itertools import groupby
-from typing import Dict, Union
+from typing import Dict, List, Tuple, Union
```
mlissner, line 5: Just a random comment. None of these are needed anymore in recent versions of Python. Instead of `List[]`, you can just use `list`. Instead of `Union`, you can use `|`, etc.

ERosendo: You're right! I've updated the type hints in a previous commit and we're not using these imports anymore.

PR #3335, opened 2023-11-01
test(search): Add missing context managers to submit delete transactions

ERosendo:
This PR adds the `captureOnCommitCallbacks` context manager to the .delete operation in some tests of the `RECAPSearchTest` class.

Prior to this PR, the `test_show_documents_when_combining_the_is_available_filter` test would occasionally fail, I hypothesize that some documents were not removed from the index before the test began.


ERosendo:
@albertisfu I updated the `test_party_name_and_children_filter` test to use a different name and description, as suggested.

albertisfu (approved):
Thanks, yeah this looks good to me so the test issue should be solved now.

PR #3387, opened 2023-11-18
feat(poetry): Bump Juriscraper to fix things

flooie:
A simple Poetry bump for Juriscraper


flooie:
Saw a quick fix I could implement for Oregon and add Oregon Court of Appeals to CL and updated juriscraper to 2.5.60

PR #3393, opened 2023-11-21
feat(recap): Add suport to parse ACMS attachment pages

ERosendo:
This PR introduces the logic to process attachment pages from ACMS. Here's a breakdown of the key changes:

- **New Upload Type**: A new upload type named `ACMS_ATTACHMENT_PAGE` is added to the `ProcessingQueue` and `pacerhtmlfiles` model.

- A new helper method `process_recap_acms_appellate_attachment` is introduced. This method encapsulates the logic for **parsing**, **processing**, and **merging** ACMS attachment pages.

- The `PacerDocIdLookUpSerializer` class is updated to include the `acms_document_guid` field in its response. This field is crucial for matching documents within the attachment page and generating the recap icon in ACMS.

- Factories are added to mock the dictionary returned by Juriscraper methods. Additionally, tests are included to verify the functionality of the new features.


cl/recap/mergers.py
```diff
@@ -1457,7 +1457,11 @@ async def clean_duplicate_attachment_entries(
         for attachment in attachment_dicts:
             attachment_number = attachment["attachment_number"]
             pacer_doc_id = attachment["pacer_doc_id"]
-            if dupe.pacer_doc_id == pacer_doc_id:
+            acms_document_guid = attachment["acms_document_guid"]
```
ttys0dev, line 1460: Won't this throw a keyerror if `acms_document_guid` is not present in the attachment dict?

I'm thinking...we shouldn't have ACMS duplicates at the moment right? So maybe we should just skip calling `clean_duplicate_attachment_entries` for ACMS attachment pages entirely?

ERosendo: > Won't this throw a keyerror if acms_document_guid is not present in the attachment dict?

You're correct! Thank you so much for pointing out this issue. I will make sure to correct it immediately

> we shouldn't have ACMS duplicates at the moment right? 

You're right. ACMS is a new source of data. 

> So maybe we should just skip calling clean_duplicate_attachment_entries for ACMS attachment pages entirely?

This is a great idea and We can use the `is_acms_attachment` parameter to skip the step.

cl/recap/mergers.py
```diff
@@ -1580,29 +1601,33 @@ async def merge_attachment_page_data(
...
         if (
             court_is_appellate
             and attachment["pacer_doc_id"] == main_rd.pacer_doc_id
+            and not main_rd_to_att
         ):
+            main_rd_to_att = True
             main_rd.document_type = RECAPDocument.ATTACHMENT
             main_rd.attachment_number = attachment["attachment_number"]
+            if "acms_document_guid" in attachment:
+                main_rd.acms_document_guid = attachment["acms_document_guid"]
```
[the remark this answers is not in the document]
ERosendo: > Is it a missing migration?

I apologize for the confusion. When you reviewed this PR, PR #3193 ([freelawproject/courtlistener#3193](https://github.com/freelawproject/courtlistener/pull/github.com)) hadn't been merged yet. This earlier PR included the migration to add the `acms_document_guid` field.

mlissner:
@grossir, can you please review again before I take another look? It's been a month or two and it looks like we've got some changes.

Thank you!

cl/recap/mergers.py
```diff
@@ -1535,9 +1537,27 @@ async def merge_attachment_page_data(
...
+            #
+            # An alternative approach is to employ the filter method in conjunction
+            # with the afirst method. This combination allows for efficient retrieval
+            # of the main RD (record) of a docket entry.
+            main_rd = (
+                await RECAPDocument.objects.select_related(
+                    "docket_entry", "docket_entry__docket"
+                )
+                .filter(**params)
+                .afirst()
```
grossir, line 1555: Are you sure the first will be the desired RD? Note that the `RecapDocument` model has this Meta attribute `ordering = ("document_type", "document_number", "attachment_number")`

grossir (commented):
The changes look fine, I just left a couple of observations

cl/recap/api_serializers.py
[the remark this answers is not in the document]
ERosendo: > Would it be worth it to add validations for ACMS_ATTACHMENT_PAGE and ACMS_DOCKET_JSON ?

Yes! Thanks for catching this.

cl/recap/mergers.py
ERosendo: > Are you sure the first will be the desired RD?

Yes, we use attachment #1 as the `main_rd` for attachment pages. Unlike district court pages, ACMS pages only have attachments. Here's a comparison: 

| District court | ACMS |
| image  | image  |

>  Note that the RecapDocument model has this Meta attribute ordering = ("document_type", "document_number", "attachment_number")

This won't be an issue because ACMS documents within an attachment page share identical `pacer_doc_id`, `document_type` (ATTACHMENT), `document_number` (the entry number), this ensures the queryset is ordered with attachment #1 appearing first.

| pacer_doc_id | document_type | document_number | attachment_number
| ------------- | ------------- |------------- | ------------- |
| 583c6255-3fd2-ed11-b596-001dd80701be | 2 | 29 | 1 |
| 583c6255-3fd2-ed11-b596-001dd80701be | 2 | 29 | 2 |

mlissner:
The model changes are noops. Would it be possible to rename them to something like 123_add_upload_types_noop.py/sql, so that those of us doing deployments (and those inspecting migrations later) can see that they're not changing the DB? I think this is in our dev guide as a best practice, but if not, we should probably add it.

ERosendo:
> Would it be possible to rename them to something like 123_add_upload_types_noop.py/sql

Sure! 

> I think this is in our dev guide as a best practice, but if not, we should probably add it.

You're right, it is described in the "Steps to making new migrations" section of the Database Migrations wiki page.

PR #3395, opened 2023-11-21
feat(poetry): Bump Juriscraper 2.5.69

flooie:
Fixes
- OklaAG
- ConnAppCt
- VI
- PR
- CoastGuard Crim
- CA2 OA
- Conn
- VA
- Ariz App dist 2


flooie:
This is a simple PR to bump juriscraper so we can start running the recent changes and fixes

PR #3434, opened 2023-12-05
Update columbia merger, handle multiple citations from same reporter

quevon24:
A small fix to avoid adding many citations in a cluster from the same reporter


cl/corpus_importer/utils.py
```diff
@@ -633,6 +633,13 @@ def add_citations_to_cluster(cites: list[str], cluster_id: int) -> None:
             cite_type_str = citation[0].all_editions[0].reporter.cite_type
             reporter_type = map_reporter_db_cite_type(cite_type_str)
 
+        if Citation.objects.filter(
+            cluster_id=cluster_id, reporter=citation[0].corrected_reporter()
+        ).exists():
+            # Avoid adding a citation if we already have a citation from the
+            # citation's reporter
+            continue
```
flooie, line 641: Might be worth logging this.

mlissner: Yeah, we definitely should, right? We don't want to miss an item just because it has bad data. In some cases, this could also catch bad data we already have, right?

flooie: yes.  sorry I didnt see your message earlier enough.  Let me make a small tweak and issue a new pr.

PR #3440, opened 2023-12-06
Update columbia merger

quevon24:
I updated the columbia merger with some small changes:

- Early check for valid citations
- Log a message when reporter's citation already exist in cluster
- winnow_case_name function updated: new words added to false_positive_set, new string added to false_positive_strings and improved the process for joining abbreviations


cl/corpus_importer/import_columbia/columbia_utils.py
```diff
@@ -766,3 +770,40 @@ def map_opinion_types(opinions=None) -> None:
...
+            logger.warning(f"Citation parsing failed for {clean_cite}")
+            continue
+
+        if Citation.objects.filter(
+            cluster_id=cluster_id, reporter=citation[0].corrected_reporter()
+        ).exists():
+            # Avoid adding a citation if we already have a citation from the
+            # citation's reporter
+            logger.info(
+                f"Cluster id: {cluster_id} - Reporter's citation already exist in "
```
flooie, line 802: you added this whole function - but does it not repeat again the same code in the add_citations_to_cluster cluster?  

instead of just moving this around I think we just need to add a simple log.

mlissner: That's a very reasonable idea. I'd suggest checking the data to make sure there's not some insane edge case we haven't considered, but assuming there's not, it's the right way to solve this. Don't forget to include a multi-column index when you do!

quevon24:
These changes are no longer necessary as long as we ensure that the cluster match list with the columbia xml are correct.

Originally this solution was proposed because some incorrect citations were found in the clusters, this related to the reporter (in the Harvard data), so as an additional measure it was decided to implement this check to prevent this from happening in the columbia merger, but this should no longer happen because the match list is being adjusted to ensure that the information added to the cluster is correct.

We also do not need to add more words to the winnow_case_name function since it will not be used in the columbia merger.

flooie:
Thanks @quevon24 for closing this and explaining why.

PR #3468, opened 2023-12-14
3147-Add endpoint to handle Neon webhooks

### LINKED ISSUE #3147: Listen to Neon webhook and take actions
When somebody signs up as a member, deletes a membership, or edits a membership, Neon will send a webhook event. CourtListener needs to capture these events and take actions. 

At a minimum (so far):

 - [ ] Stop their recurring donations they may have in CL
 - [ ] We log their Neon ID
 - [ ] We log their membership termination date (because we will turn off special features when that expires)

There will be other things too, surely. This is going to be the complicated part.

ERosendo:
This PR adds an endpoint to handle webhooks from Neon and fixes #3147.

1. Adds a new field to the `UserProfile` model. The name of this new field is `neon_account_id` and allows us to store the unique identifier assigned by Neon CRM to an account record

2. Introduces a new model to store membership data. This new model is called `NeonMembership`

3. Implements a dedicated model to store webhook payloads for effective debugging and problem resolution. Additionally, it also provides a method to truncate the table content storing the webhook payload.

4. Add a client class to use the Neon API.

5. As part of the webhook integration, this PR introduces a new ViewSet and a custom authorization class for enhanced security and control.

### Deployment

Before merging and deploying this PR, please ensure the following environment variables are set:

- `NEON_ORG_ID` and `NEON_API_KEY`: We use both variables to authenticate API requests.
- `NEON_MAX_WEBHOOK_NUMBER`: Sets the max number of records we retain in the log table.
-  `NEON_AUTHORIZATION_TOKEN`: The key we use to authorize incoming webhooks.


cl/donate/models.py
```diff
@@ -164,3 +166,85 @@ def __str__(self) -> str:
...
+    MEMBERSHIP_DELETE = 3
+    MEMBERSHIP_UPDATE = 4
+    TYPES = (
+        (MEMBERSHIP_CREATION, "createMembership"),
+        (MEMBERSHIP_EDIT, "editMembership"),
+        (MEMBERSHIP_DELETE, "deleteMembership"),
+        (MEMBERSHIP_UPDATE, "updateMembership"),
+    )
+    trigger = models.PositiveSmallIntegerField(
+        help_text="The current membership tier of a user within Neon CRM",
```
mlissner, line 183: Is this `help_text` correct?

cl/donate/models.py
```diff
@@ -164,3 +166,85 @@ def __str__(self) -> str:
...
+        choices=TYPES,
+        null=True,
+    )
+    account_id = models.CharField(
+        help_text="Unique identifier assigned by Neon CRM to a customer record",
+        default="",
```
mlissner, line 189: I don't think you need to set the default to blank if you've got `blank=True`, do you?

cl/donate/models.py
```diff
@@ -164,3 +166,85 @@ def __str__(self) -> str:
             self.monthly_donation_amount,
             self.get_payment_provider_display(),
         )
+
+
+class NeonWebhookEvents(models.Model):
```
mlissner, line 171: Shouldn't this be singular? E.g.:
```suggestion
class NeonWebhookEvent(models.Model):
```

cl/donate/models.py
```diff
@@ -164,3 +166,85 @@ def __str__(self) -> str:
...
+    TYPES = (
+        (BASIC, "CL Membership - Basic"),
+        (LEGACY, "CL Legacy Membership"),
+        (TIER_1, "CL Membership - Tier 1"),
+        (TIER_2, "CL Membership - Tier 2"),
+        (TIER_3, "CL Membership - Tier 3"),
+        (TIER_4, "CL Membership - Tier 4"),
+        (TIER_5, "CL Membership - Tier 5"),
+        (PLATINUM, "CL Platinum Membership"),
+    )
```
mlissner, line 223: I think we're going to regret putting these types into code, but I guess we can see what happens for now. Like, Jenifer is going to make a new tier at some point and we're going to have to figure out what to do about it....maybe we'll find that these types should be rows in our DB. We'll see, I guess.

cl/donate/models.py
```diff
@@ -164,3 +166,85 @@ def __str__(self) -> str:
...
+    INVERTED = invert_choices_group_lookup(TYPES)
```
mlissner, line 224: Perhaps:
```suggestion
    TYPES_INVERTED = invert_choices_group_lookup(TYPES)
```

cl/donate/models.py
```diff
@@ -164,3 +166,85 @@ def __str__(self) -> str:
...
+    user = models.OneToOneField(
+        User,
+        related_name="membership",
+        verbose_name="the user linked to the membership",
+        on_delete=models.CASCADE,
+        unique=True,
+    )
+    neon_id = models.CharField(
+        help_text="Unique identifier assigned by Neon CRM to a membership record",
+        default="",
```
mlissner, line 234: Again, I think you don't need this.

cl/donate/models.py
```diff
@@ -164,3 +166,85 @@ def __str__(self) -> str:
...
+        blank=True,
+    )
+    level = models.PositiveSmallIntegerField(
+        help_text="The current membership tier of a user within Neon CRM",
+        choices=TYPES,
+        null=True,
```
mlissner, line 240: A bit surprising that this can be null?

cl/donate/models.py
```diff
@@ -164,3 +166,85 @@ def __str__(self) -> str:
...
+    )
+    content = models.JSONField(  # type: ignore
+        help_text="The content of the payload of the POST request.",
+        blank=True,
+        null=True,
+    )
+
+
+@pghistory.track(AfterUpdateOrDeleteSnapshot())
+class NeonMembership(models.Model):
```
mlissner, line 205: We should include our AbstractDateTimeModel in all our models.

cl/donate/signals.py
```diff
@@ -0,0 +1,17 @@
...
+@receiver(
+    post_save,
+    sender=NeonWebhookEvents,
+    dispatch_uid="truncate_webhook_events",
+)
+def truncate_webhook_events(sender, instance=None, created=False, **kwargs):
+    if not instance.id % settings.NEON_MAX_WEBHOOK_NUMBER:
+        NeonWebhookEvents.objects.filter(
+            pk__lte=instance.id - settings.NEON_MAX_WEBHOOK_NUMBER
+        ).delete()
```
mlissner, line 17: We also want things that could violate our privacy policy, so, something like:

```suggestion
        twelve_weeks_ago = XYZ
        NeonWebhookEvents.objects.filter(
            Q(pk__lte=instance.id - settings.NEON_MAX_WEBHOOK_NUMBER) |
              Q(date_created__lt=twelve_weeks_ago),
        ).delete()
```

cl/users/models.py
```diff
@@ -141,6 +141,11 @@ class UserProfile(models.Model):
         help_text="Sort dockets in descending order by default",
         default=False,
     )
+    neon_account_id = models.CharField(
+        help_text="Unique identifier assigned by Neon CRM to a customer record",
+        default="",
```
mlissner, line 146: One more here, I think.

cl/donate/api_views.py
```diff
@@ -0,0 +1,215 @@
...
+            membership = {
+                "membershipId": data["membershipId"],
+                "membershipName": data["membershipName"],
+            }
+        else:
+            data = webhook_data["data"]
+            membership = {
+                "membershipId": data["id"],
+                "membershipName": data["membershipLevel"]["name"],
+            }
```
mlissner, line 133: A comment explaining this if/else would be helpful. Why are there two formats of the webhook data?

cl/donate/api_views.py
```diff
@@ -0,0 +1,215 @@
...
+        return trigger
+
+    def _store_webhook_payload(self, webhook_data) -> None:
+        trigger = self._map_trigger_value(webhook_data["eventTrigger"])
+        if trigger != NeonWebhookEvents.MEMBERSHIP_DELETE:
+            membership_data = self._get_membership_data(webhook_data)
+        else:
+            membership_data = webhook_data["data"]["membership"]
+        NeonWebhookEvents.objects.create(
+            content=json.dumps(webhook_data, default=str),
```
mlissner, line 177: I think you'll want indent=2 here, no?

cl/donate/api_views.py
```diff
@@ -0,0 +1,215 @@
...
+
+        webhook_data = serializer.validated_data
+        self._store_webhook_payload(webhook_data)
+        match webhook_data["eventTrigger"]:
+            case "createMembership" | "editMembership":
+                self._handle_membership_creation_or_update(webhook_data)
+            case "deleteMembership":
+                self._handle_membership_deletion(webhook_data)
+            case "updateMembership":
+                pass
```
mlissner, line 55: Why do we skip this? Is it the same as `editMembership`? Maybe a comment here?

cl/donate/api_views.py
```diff
@@ -0,0 +1,215 @@
...
+        Retrieves a user record associated with a Neon CRM account ID.
+
+        This method attempts to find a matching Django user in two ways:
+
+        1. first tries to directly query the database for a user whose
+        `neon_account_id` field matches the provided `account_id`.
+        2. If no matching user is found in the database, it fetches the
+        account email address from the Neon API using the `account_id`.
+        It then tries to find a user whose email address matches the
+        retrieved Neon account email.
```
mlissner, line 73: I think if we still can't find such an account, we should create a stub. There's an example of this when anonymous users create donations. 

And later, when the account is created, we should convert the stub to the real thing, such that the donations for the user (or their membership) is ready and waiting for them. The idea is that you should be able to donate first, and then create an account second.

cl/donate/api_views.py
```diff
@@ -0,0 +1,215 @@
...
+            neon_account = client.get_acount_by_id(account_id)
+            users = User.objects.filter(
+                email=neon_account["primaryContact"]["email1"]
+            )
+            if not users:
+                return HttpResponse(
+                    "Error processing webhook, User not found",
+                    status=HTTPStatus.INTERNAL_SERVER_ERROR,
+                )
+            user = users.first()
```
mlissner, line 94: User objects don't have a default ordering that I'm aware of, and you didn't apply ordering in your query up above, so this call to `first()` is undefined in how it behaves. That's no good. 

On top of that, I actually wonder what we *should* do here. People are currently allowed to make more than one account in CL with the same email address, and indeed people *do*. I guess we should ban that eventually, but maybe the best thing to do here is order by most recently logged in, and update that account?

cl/donate/api_views.py
```diff
@@ -0,0 +1,215 @@
...
+            neon_membership.save()
+        except ObjectDoesNotExist:
+            NeonMembership.objects.create(
+                user=user,
+                neon_id=membership_data["membershipId"],
+                level=NeonMembership.INVERTED[
+                    membership_data["membershipName"]
+                ],
+                termination_date=membership_data["termEndDate"],
+            )
```
mlissner, line 202: We want as little in our `try` block as possible, so perhaps:

```suggestion
        try:
            neon_membership = user.membership
        except ObjectDoesNotExist:
            NeonMembership.objects.create(
                user=user,
                neon_id=membership_data["membershipId"],
                level=NeonMembership.INVERTED[
                    membership_data["membershipName"]
                ],
                termination_date=membership_data["termEndDate"],
            )
        else:
            neon_membership.neon_id = membership_data["membershipId"]
            neon_membership.level = NeonMembership.INVERTED[
                membership_data["membershipName"]
            ]
            neon_membership.termination_date = membership_data["termEndDate"]
            neon_membership.save()
```

mlissner (changes requested):
Good stuff. I've made a number of suggestions, but I think you've got the design about right. A few thoughts though:

1. Do we get membership deleted events when memberships expire? If not, should we be monitoring the date that memberships terminate, and should we be removing them when they do?

2. I guess donations to CourtListener don't get you any more, so we don't even need to know that they happened over in Neon. That's a bit weird, but I guess it's OK.

mlissner:
One more big thought: If I make a monthly donation to FLP via Neon, will I get the benefits that a member would get? That is, if I use the regular donate form and donate $100/month, will the webhook catch that and do the right thing? 

@jeniferw, maybe we need to disable monthly donations and only allow monthly donations through members — but if we do that, how do we connect the two forms so that we have:

1. A donate form for regular one-time donations, and
2. A membership form for monthly donations?

Hm, we need to figure that out before we launch these changes, I'm afraid.

cl/donate/models.py
ERosendo: You're right! I'll rename the class.

cl/donate/models.py
ERosendo: No, it's not. I'll fix it.

cl/users/models.py
ERosendo: You are correct. We don't need to provide a default value when setting the 'blank' attribute to True.

ERosendo:
> Will the webhook catch that and do the right thing?
 
@mlissner Membership webhooks won't be triggered by Donation forms on Neon. That's because Neon already provides independent webhook notifications for Donations. In case we want to add Donation webhooks, we would need to implement dedicated endpoints to capture them. These endpoints could use Neon's API to distinguish between one-time and recurring donations, which would enable appropriate actions within our system.

jeniferw:
> > Will the webhook catch that and do the right thing?
> 
> @mlissner Membership webhooks won't be triggered by Donation forms on Neon. That's because Neon already provides independent webhook notifications for Donations. In case we want to add Donation webhooks, we would need to implement dedicated endpoints to capture them. These endpoints could use Neon's API to distinguish between one-time and recurring donations, which would enable appropriate actions within our system.

So I set up what Neon calls "Complimentary Memberships" if someone has donated at least $60 in the past 12 months, they will be members at the appropriate level. 
The only concerns are 2 things - 

1. people that set up $5 per month memberships (which are recorded in Neon as donations) less than a year ago - then the $60 base won't catch them and give them memberships. 
2. We need to migrate the data into Neon for the time frame 10/5 to present - so missing those individuals to check on with webhooks.

Options:
1. I could adjust that to be $5 or more for now. But that means that anyone that makes a $5 one time donation, would get access to alerts for a year. (not ideal).
2. Could we do something in CL/RECAP to identify people who currently have memberships?


This should be less of an issue moving forward - between migrating recurring payment schedules into Neon, and having new memberships signed up for as memberships.
RECAP/CL should link to membership at top - maybe a "Donate to support Free Law Project" in the footer???
Bots.law & FLP should link to donate at top.

cl/donate/models.py
ERosendo: You're right. I'm removing the `default` argument.

cl/donate/models.py
ERosendo: I'm sorry, this field should not be null.

cl/donate/models.py
ERosendo: got it!

mlissner:
OK, I think this is gelling for me, thank you both. I think the way to proceed is:

1. Links from CourtListener encourage memberships whenever benefits are involved, and whenever no context is given. For example:
   - There are some places on CourtListener where we ask for donations, straight up, with some sort of message about what we do. These should be donate links.
   - There are some places where we ask for donations without any context, like in the navbar. These go to the membership page.
   - There are some places where we ask for money in exchange for services like alerts. These link to the membership page.

2. Memberships come with tricky tax stuff (see: https://github.com/freelawproject/free.law/issues/130#issuecomment-1863472742), so to keep things clean, regular donors don't get complimentary memberships (sorry Jenifer). If you're a donor and you're hoping for membership stuff, well, write us a letter and we'll do it by hand.

3. This means no need for CourtListener to receive donation webhooks. All CourtListener cares is having an accurate accounting of the membership information in Neon.

4. Any current monthly donor gets a legacy (or better) membership in CourtListener when we migrate the data. 

5. Features in CourtListener are enabled based being a member. When we land this feature:
    - Features that currently require monthly donations need to be changed to check the membership levels of users instead. That is, if something currently requires a $5/month donation, the code that checks that should be replaced with code that checks if the user has an active membership.
    - This means we need a script to make all of our monthly donors into members in CourtListener on day one. We'll migrate their actual transactions to Neon ASAP, but until that happens, they need to be marked as members so they get the features they're entitled to.
    - The links above need to be updated in this PR (or https://github.com/freelawproject/courtlistener/pull/3475) as well. When this goes live and we run the script to make monthly users into members, we want that to be the end of new monthly donors on CourtListener. 

6. @jeniferw we need to clean up the membership form before this lands, it feels a little drafty. It might not be an issue to just remove any mention of swag until the swag is ready.

I think this approach should cover the bases, keep the IRS happy, and provide a smooth transition for our users. 

What did I miss?

cl/donate/api_views.py
ERosendo: Thank you for your suggestion. I will implement this refactor.

cl/donate/api_views.py
ERosendo: > maybe the best thing to do here is order by most recently logged in, and update that account?

I believe that this is a good idea. I'll update the code.

cl/donate/signals.py
ERosendo: understood!

cl/donate/api_views.py
ERosendo: Thank you for bringing this to my attention.

I used the `dumps` method to address an issue with `datetime` objects because they are not directly JSON serializable and I was getting an error when the code tried to store the payload. This method's `default` argument provides a function that gets called for objects that can't be serialized. In our case, It just converts everything it doesn't know to strings. This approach is fine for serialization but the output is not a human-readable JSON representation (the dumps method adds a bunch of new lines and backslashes to the string).

While reviewing your comments, I realize we can handle `datetime` objects without relying on the `dumps` method. We can pass an `encoder` subclass to the Django [JSONField](https://docs.djangoproject.com/en/5.0/ref/models/fields/#jsonfield) to serialize data types not supported by the standard JSON serializer and Django provides the handy [DjangoJSONEncoder](https://docs.djangoproject.com/en/5.0/topics/serialization/#djangojsonencoder) class, specifically designed to handle `datetime`, `date`, `time`, and `UUID` objects effortlessly.

We just need to tweak the content field in the `NeonWebhookEvent` model like this:

```python
from django.core.serializers.json import DjangoJSONEncoder

class NeonWebhookEvent(AbstractDateTimeModel):
    ...
    content = models.JSONField(  # type: ignore
        encoder=DjangoJSONEncoder,
        help_text="The content of the payload of the POST request.",
    )

```

cl/donate/api_views.py
ERosendo: > I think if we still can't find such an account, we should create a stub.

Got it! I'll use the helper method to create stubs.

cl/donate/api_views.py
ERosendo: > Why are there two formats of the webhook data?

During initial development, I used the provided [documentation](https://developer.neoncrm.com/api/webhooks/membership-webhooks/#createMembership) to create this helper method. However, upon testing the membership form, I discovered a discrepancy between the expected payload and the actual payload received.

According to the webhook documentation, the `membership` key within the `data` object should provide details about membership changes. However, this key was not present in the webhooks I got by `creating/upgrading/downgrading/renewing` memberships during my testing.

I made this helper compatible with both webhook data formats in case the payload structure outlined in the documentation represents a planned schema update.

cl/donate/api_views.py
ERosendo: > Is it the same as editMembership?

No, It's not. I only received the `updateMembership` webhook after upgrading my membership from my account dashboard, which is separate from the admin dashboard.

I noticed that the upgrade process triggers two webhooks almost simultaneously.:

1. One to update the old membership. The `eventTrigger` of this webhooks is  `updateMembership` and the payload contains details of the previous membership record but with an updated 'termEndDate' field. As a result, the status of the previous membership will be changed to expired in Neon.

2. One to create a new membership. The `eventTrigger` of this webhook is `createMembership` and the payload has all the details of the new membership(new `membershipId`, new `level`, and new `termEndDate`).

> Why do we skip this?

I decided to skip the `updateMembership` webhook to prevent a race condition during the upgrade process. We only need the latest information to update the membership record in our database. However, there's a chance that the `createMembership` webhook might be processed before the `updateMembership` webhook, leading to data integrity issues.

ERosendo:
> This means no need for CourtListener to receive donation webhooks. All CourtListener cares is having an accurate accounting of the membership information in Neon.

got it!

> Features in CourtListener are enabled based being a member. When we land this feature:
> 
> Features that currently require monthly donations need to be changed to check the membership levels of users instead. That is, if something currently requires a $5/month donation, the code that checks that should be replaced with code that checks if the user has an active membership.
>
> This means we need a script to make all of our monthly donors into members in CourtListener on day one. We'll migrate their actual transactions to Neon ASAP, but until that happens, they need to be marked as members so they get the features they're entitled to.

I think we can address both changes (automatically grant membership to all monthly donors and update the logic to check memberships instead of donations) in a separate PR. 

> The links above need to be updated in this PR (or https://github.com/freelawproject/courtlistener/pull/3475) as well.

#3475 seems like a more relevant place to update the links.

cl/donate/api_views.py
mlissner: Well that's very handy.

cl/donate/api_views.py
mlissner: Got it. Can you make a comment about this in the code, please to explain why we ignore this? I guess the easiest thing is:

```
# See: https://github.com/freelawproject/courtlistener/pull/3468#discussion_r1433398175
```

:)

mlissner:
> I think we can address both changes (automatically grant membership to all monthly donors and update the logic to check memberships instead of donations) in a separate PR.

Cool. We need this all to happen at the same time:

 - Links to donate become links to the membership form.
 - We add membership info to all our monthly donors in CL.
 - We get the webhook going
 - We delete the old payment system

That way, when we launch all this, it should be pretty smooth for our existing users (who become members), the webhook will be working as new people sign up, and the old payment system happily goes away.

> https://github.com/freelawproject/courtlistener/pull/3475 seems like a more relevant place to update the links.

I agree, yes. Sorry this is a lot of intermixed stuff. If I had more time, I'd sequence it more intelligently, but hopefully it's not too confusing. If it is, let's chat tomorrow.

ERosendo:
> That way, when we launch all this, it should be pretty smooth for our existing users (who become members), the webhook will be working as new people sign up, and the old payment system happily goes away.

understood!

mlissner:
Sounds great. So we'll be able to set the `is_active` flag using `updateMembership` events, and we can just keep old membership records around because why not. Works for me.

ERosendo:
@mlissner I've implemented all of your suggestions, and this pull request is now ready for review. However, after adding the `create_stub_account` helper, CodeQL flagged some possible issues. 

Do you think we should address those issues within this PR?

ERosendo:
@mlissner after merging this PR, we could use the following code to create memberships for all of our monthly donors:

```python

from cl.donate.models import MonthlyDonation, NeonMembership

monthly_donations= MonthlyDonation.objects.filter(enabled=True).distinct('donor_id').all()

for donation in monthly_donations:
    NeonMembership.objects.create(user_id=donation.donor_id, level=NeonMembership.LEGACY)
```

Membership records will be created without `neon_id` and `termination_date` initially. However, webhooks triggered during the migration process should fill those fields.

mlissner:
Code looks good, but we can't land this until the front end is checking membership information instead of whether somebody is a monthly donor. 

As for your script: 

> Membership records will be created without neon_id and termination_date initially. However, webhooks triggered during the migration process should fill those fields.

I'm not sure we'll get webhooks sent when we migrate the data. Maybe so; maybe not, but if we assume not, I do think it's safe to assume that we'll get updates when people terminate their memberships in Neon, so I think even though we'll lack their `neon_id` and their `termination_date`, that should be OK, right?

ERosendo:
>  I do think it's safe to assume that we'll get updates when people terminate their memberships in Neon, so I think even though we'll lack their neon_id and their termination_date, that should be OK, right?

You're correct! The membership records created by the script will enable users to receive benefits immediately and we'll complete the remaining data fields when we get the termination/renewal webhook.

PR #3474, opened 2023-12-14
Add donate banner and remove cruft

mlissner:
I'm being hypocritical and including some clean up in this PR, but the core of it is just to add the new donate banner. Simple stuff, shouldn't be too exciting.


mlissner:
Once this is deployed, we'll need to set `FUNDRAISING_MODE` to True, and redploy `cl-python`.

PR #3481, opened 2023-12-18

johnludwigm:
This PR makes the following refactors:

* Removes f from some constant strings
* Removes unneeded imports
* Makes some parts more Pythonic (dict comprehension) + type-hint

### DIFF
--- a/cl/scrapers/management/commands/cl_report_scrape_status.py
+++ b/cl/scrapers/management/commands/cl_report_scrape_status.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from datetime import date, timedelta
 from operator import itemgetter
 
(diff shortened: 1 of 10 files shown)


johnludwigm:
@mlissner SSDD
image

mlissner:
Oh, I didn't turn that on, I didn't want to think through whatever permissions/changes that'd mean, but we'll get there I imagine. For now you can just flag things for me with a comment, and that's fine.

johnludwigm:
@mlissner done.

cl/scrapers/management/commands/cl_report_scrape_status.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import annotations
```
mlissner, line 1: Why do we need this? Is it just a performance enhancement to put it everywhere?

mlissner (commented):
One quick question...

Thank you!

cl/scrapers/management/commands/cl_report_scrape_status.py
johnludwigm: I don't know about "performance enhancement". It's not there for performance, but rather linting and type-hinting.

Overall, I like to import it whenever I'm type-hinting. [It's probably a best practice, especially for future changes.](https://docs.python.org/3/whatsnew/3.7.html#pep-563-postponed-evaluation-of-annotations)


I think this code will run without it. I can get rid of it if you'd like. @mlissner

mlissner: Yeah, the `__future__` gives me nightmare flashbacks from Py2. Let's remove it, since it looks like it'll be the default in version soon anyway.

johnludwigm:
@mlissner Now it's done.

cl/scrapers/management/commands/cl_report_scrape_status.py
johnludwigm: @mlissner Now it's done.

mlissner (approved):

mlissner:
Thanks. Auto-merge set.

PR #3496, opened 2023-12-25

johnludwigm:
* Drop f from constant strings
* Remove `object` from class declarations. Hasn't been needed since Python 3.
* Get rid of some unneeded list comprehensions


mlissner:
Thanks, LGTM!

PR #3497, opened 2023-12-25

johnludwigm:
* Remove `object` as a superclass. That's the default.
* Replace `super(ClassName, self)` with `super`. That's the default.
* Remove some unneeded list comprehensions.


mlissner:
Thanks. If you're doing this with a linter, it would be great to include linter rules as you go. I've been looking at ruff for this kind of thing, but switching is probably not the best use of time.

johnludwigm:
> Thanks. If you're doing this with a linter, it would be great to include linter rules as you go. I've been looking at ruff for this kind of thing, but switching is probably not the best use of time.

I'm using fixit from Meta, but I disagree with some of the rules, especially around long f strings. I'll include the rules in future Prs.

PR #3499, opened 2023-12-26

johnludwigm:
* Fix some shuffling action that never worked.
* Remove the `object` superclass references.
* Simplify the `super(..., ...)` method calls
* Add some type hints.
* Remove f from some constant strings.

### DIFF
--- a/cl/recap/management/commands/pacer_iquery_scraper.py
+++ b/cl/recap/management/commands/pacer_iquery_scraper.py
@@ -86,7 +86,7 @@ def get_docket_ids() -> Set[int]:
         ) as e:
             logger.warning(
                 "iQuery scraper was unable to get results from Plausible. Got "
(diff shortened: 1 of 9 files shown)


cl/recap/management/commands/pacer_iquery_scraper.py
```diff
@@ -180,7 +180,8 @@ def handle(self, *args, **options):
         )
         # Shuffle the dockets to make sure we don't hit one district all at
         # once.
-        random.shuffle(list(docket_ids))
```
johnludwigm, line 183: This cannot have ever worked.

mlissner (approved):
Good catch with the shuffle, thanks and merging!

johnludwigm:
Fixed mystery issue after merging `main`.

PR #3500, opened 2023-12-26
3126-block the browsing topics feature

### LINKED ISSUE #3126: Block Chrome's new "privacy" sandbox
Mozilla and Apple have come out against it:
 - https://github.com/mozilla/standards-positions/issues/622#issuecomment-1372979100
 - https://github.com/WebKit/standards-positions/issues/111#issuecomment-1359609317

From what I can tell, this new thing just tracks you around the web, and then builds up a model of what you like so that advertisers can learn about you. Super, freaking creepy, and makes me really glad I don't use Chrome.

The good news is, I think we can block it by setting a header on every page, so we should do so:

https://developer.chrome.com/docs/privacy-sandbox/topics/#site-opt-out

That says we have to set:

```
Permissions-Policy: browsing-topics=()
```

Gross and annoying. I'm not sure what tools we have in Django for setting this header. If it requires a middleware, I guess we could provide that to the community as a library folks can install.

ERosendo:
This PR fixes #3126 by adding the [django-permissions-policy](https://github.com/adamchainz/django-permissions-policy) package and disabling the browsing-topics feature.

### DIFF
--- a/cl/settings/django.py
+++ b/cl/settings/django.py
@@ -119,6 +119,7 @@ MIDDLEWARE = [
     "django.contrib.sessions.middleware.SessionMiddleware",
     "django.contrib.auth.middleware.AuthenticationMiddleware",
     "django.middleware.security.SecurityMiddleware",
+    "django_permissions_policy.PermissionsPolicyMiddleware",
     "django.middleware.clickjacking.XFrameOptionsMiddleware",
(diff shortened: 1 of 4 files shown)


mlissner:
Thanks for this. Can you also open a bug to fill in the rest of these features and put it on the volunteer backlog? It's a fairly simple task, but it takes some care.

PR #3525, opened 2023-12-27

johnludwigm:
* Improve class infrastructure.
* Add type hints
* Use more f strings


mlissner (approved):

mlissner:
I think if all you're doing in your PRs for the moment is cleaning up class declarations and f-strings, perhaps it's best to send one or two bigger PRs. Maybe just do one for all class cleanups, and another for f-strings? It'd certainly be easier on the review end of things, if that worked for you, and we'd be done a lot faster?

PR #3546, opened 2024-01-02
feat(search): Adds new constraint using the MD5 hash of the docket no.

ERosendo:


mlissner (approved):
Looks good, thanks. I'm going to try deleting the index now, as a stopgap and hopefully this can wait until I get home to do a proper migration.

mlissner:
This was news to me:

image

ERosendo:
@mlissner I've added the migration to create the index on the MD5 value of the `docket_number` field.

cl/search/migrations/0025_docket_unique_docket_per_court.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+--
+-- Create constraint unique_docket_per_court on model docket
+--
+CREATE UNIQUE INDEX CONCURRENTLY "unique_docket_per_court" ON "search_docket" ((MD5("docket_number")), "pacer_case_id", "court_id");
```
mlissner, line 5: According to [our wiki](https://github.com/freelawproject/courtlistener/wiki/Database-migrations#adding-and-removing-indexes-to-a-column-dont-use-concurrently-by-default), I don't think `CONCURRENTLY` can be used in a transaction, can it?

mlissner (changes requested):
I actually found something we might need to change that I missed before...

cl/search/migrations/0025_docket_unique_docket_per_court.sql
ERosendo: You're right! Sorry. I'm fixing this.

mlissner: A small thing!

mlissner (approved):

mlissner:
This looks ready to go when I get back, thanks. Approved!

ERosendo:
@mlissner I updated the model to use the [HashIndex](https://docs.djangoproject.com/en/5.0/ref/contrib/postgres/indexes/#hashindex) class and also removed the `startwith` filter for the `docket_number` field from the `DocketFilter` class.

mlissner:
@ERosendo, I'm looking at this a bit more carefully, and comparing it with https://github.com/freelawproject/courtlistener/pull/3558/. 

Do you think you can combine them so that the migrations happen in this order:

1. Add Hashindex
3. Add MD5 index and MD5 unique constraint.
4. Remove b-tree index on docket number.
5. Remove old unique constraint.

Or something similar, such that, at least:

1. The MD5 index is made before its constraint is made. 
2. The HashIndex is made before the B-Tree index is deleted.

I would have figured this out by hand, but it's a little messy anyway (as you saw the other  day), and there's already a problem between these two PRs that they both create the MD5 constraint, so we need a little clean up anyway.

Thank you.

ERosendo:
@mlissner Sure! I'll update the migrations

mlissner:
Thanks. I appreciate it. I'm looking forward to getting this taken care of.

ERosendo:
@mlissner I've updated this PR and is now ready for review.

mlissner:
Wonderful, thanks. I'll put it on my backlog to review and deploy.

mlissner (approved):
Looks good, thanks. 

I'm going to deploy this to prod, our replica, and our users on postgres 16. Everybody on postgres 11 will be upgraded to 16 soon and get the changes that way. The people on 14 will just do without, and should be fine until their eventual upgrade to whatever version they get next (18?)

Issue #3543, opened 2024-01-01
Index on docket_number is too big and needs to be made partial


mlissner:
## The Scene

I'm setting up a new replication user on postgres 16.1, and I'm getting a sawtooth pattern as shown in #3509. 

## Findings

At first, I thought this was because I didn't have enough logical slots while creating the new subscriber, but I restarted the replication from scratch and that didn't fix the issue, so I started digging into the logs and found this at the peak of a saw tooth:

> 2024-01-01 00:25:31 UTC::@:[29891]:ERROR: index row size 2824 exceeds btree version 4 maximum 2704 for index "search_docket_docket_number_7642c6c6dbd04704_uniq"
> 2024-01-01 00:25:31 UTC::@:[29891]:DETAIL: Index row references tuple (1922780,7) in relation "search_docket".
> 2024-01-01 00:25:31 UTC::@:[29891]:HINT: Values larger than 1/3 of a buffer page cannot be indexed.
> Consider a function index of an MD5 hash of the value, or use full text indexing.
> 2024-01-01 00:25:31 UTC::@:[29891]:CONTEXT: COPY search_docket, line 66527254
> 2024-01-01 00:25:31 UTC::@:[395]:LOG: background worker "logical replication worker" (PID 29891) exited with exit code 1

One theory about this error is that it is due to btree indexes have a maximum size, which got smaller between versions. This is explained in a [Stackoverflow Q&A](https://stackoverflow.com/questions/70123567/index-row-size-2712-exceeds-btree-version-4-maximum-2704-for-index-while-doing), which points to [Postgres documentation](https://www.postgresql.org/docs/12/release-12.html#id-1.11.6.14.4), which says:

> In new btree indexes, the maximum index entry length is reduced by eight bytes, to improve handling of duplicate entries (Peter Geoghegan)

I'm not entirely convinced by this because we're replicating from v14 to v16 and the change happened when v12 was released, but the error is pretty clear. 

In any case, the solution proposed in the log seems fine: Use an MD5 of the big value in the index.

The index from the log, `search_docket_docket_number_7642c6c6dbd04704_uniq`, is defined in the DB as 

```postgres
courtlistener=> \d search_docket_docket_number_7642c6c6dbd04704_uniq
Index "public.search_docket_docket_number_7642c6c6dbd04704_uniq"
    Column     |          Type          | Key? |  Definition   
---------------+------------------------+------+---------------
 docket_number | text                   | yes  | docket_number
[...]

PR #3558, opened 2024-01-05
feat(search): Drop old indices on the docket_number field

ERosendo:
This PR removes the problematic indices identified in #3543, working in conjunction with #3546 to ensure a smooth transition to the new indices.

We should merge this code only after the new indices have been successfully created.


cl/search/migrations/0026_alter_docket_unique_together_and_more.sql
```diff
@@ -0,0 +1,9 @@
+--
+-- Alter unique_together for docket (0 constraint(s))
+--
+ALTER TABLE "search_docket" DROP CONSTRAINT "search_docket_docket_number_pacer_case_a3184727_uniq";
+--
+-- Alter field docket_number on docket
+--
+DROP INDEX CONCURRENTLY IF EXISTS "search_docket_docket_number_b2afb9d6";
+DROP INDEX CONCURRENTLY IF EXISTS "search_docket_docket_number_b2afb9d6_like";
```
mlissner, line 9: Django did a (dumb? necessary? annoying?) thing years ago and decided to change how indexes are named. As a result, when you create SQL files in a modern version of Django, it generates index names the way that version does, instead of the way the earlier version _did_. 

So if we ran this on our server, which was created with the old version, it wouldn't work because the index names are wrong. 

The correct place to get index names is from the DB itself. You could get these values from the dev DB, from our schema dump in the bulk data directory, or from the details of the issue I filed. :)

mlissner (changes requested):
One issue, please. (Nothing ever easy on an old project.)

cl/search/migrations/0026_alter_docket_unique_together_and_more.sql
ERosendo: Got it!. I'll update the SQL file.

mlissner:
I'm going to close this one. Let's get it merged with #3546 so all migrations happen there in a series of files.

PR #3568, opened 2024-01-05
fix(api): Use TemplateResponse instead of the render method

### LINKED ISSUE #3564: Bulk data page: SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async.
Sentry Issue: [COURTLISTENER-61T](https://freelawproject.sentry.io/issues/4817055210/?referrer=github_integration)

```
SynchronousOnlyOperation: You cannot call this from an async context - use a thread or sync_to_async.
(10 additional frame(s) were not displayed)
...
  File "cl/api/views.py", line 102, in bulk_data_index
    return render(
  File "cl/lib/context_processors.py", line 134, in inject_email_ban_status
    if email_banned.exists():
```

ERosendo:
Fixes #3564.

### DIFF
--- a/cl/api/views.py
+++ b/cl/api/views.py
@@ -8,6 +8,7 @@ from django.conf import settings
 from django.http import HttpRequest, HttpResponse, JsonResponse
 from django.shortcuts import get_object_or_404, render
 from django.template import TemplateDoesNotExist
+from django.template.response import TemplateResponse
 from django.views.decorators.cache import cache_page
 from requests import Session
 from rest_framework import status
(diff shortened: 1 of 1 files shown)


cl/api/views.py
```diff
@@ -99,7 +99,7 @@ def replication_docs(request: HttpRequest) -> HttpResponse:
 async def bulk_data_index(request: HttpRequest) -> HttpResponse:
     """Shows an index page for the dumps."""
     disclosure_coverage = await get_coverage_data_fds()
-    return render(
+    return await sync_to_async(render)(
```
ttys0dev, line 102: Use `TemplateResponse` intead.

ERosendo: got it! thanks.

Issue #3549, opened 2024-01-03
Bug in CL Scrape Opinions w/ HTML scrapes


flooie:
Juriscraper has a method called extract_from_text that is used in scraping opinions without a full compliment of metadata.

See Below: 

This line `metadata_dict = site.extract_from_text(opinion.plain_text)` passes in only plain text - but when HTML or other sources are scraped, plain_text is not generated but HTML is.  So we are never passing in the text of the opinion to generate the proper metadata.  In this scenario - it will only ever return a blank string to update.  

Whoops

```
def update_document_from_text(opinion: Opinion) -> None:
    """Extract additional metadata from document text

    We use this code with BIA decisions. Previously Tax.
    I think it is not unlikely that we will use or need this in the future.

    Use functions from Juriscraper to pull metadata out of opinion
    text. Formerly implemented in only Tax Court, but functional in all
    scrapers via AbstractSite object.

    Note that this updates the values but does not save them. Saving is left to
    the calling function.

    :param opinion: Opinion object
    :return: None
    """
    court = opinion.cluster.docket.court.pk
    site = get_scraper_object_by_name(court)
    if site is None:
        return
    metadata_dict = site.extract_from_text(opinion.plain_text)
    for model_name, data in metadata_dict.items():
        ModelClass = apps.get_model(f"search.{model_name}")
        if model_name == "Docket":
            opinion.cluster.docket.__dict__.update(data)
        elif model_name == "OpinionCluster":
            opinion.cluster.__dict__.update(data)
        elif model_name == "Citation":
            data["cluster_id"] = opinion.cluster_id
            ModelClass.objects.get_or_create(**data)
        else:
            raise NotImplementedError(
                f"Object type of {model_name} not yet supported."
            )
```

PR #3577, opened 2024-01-08
feat(scrapers): support nytrial scrapers

grossir:
Theses changes are needed to support freelawproject/juriscraper#827

- If possible, get court object from child_court field passed by nytrial families of scrapers. Otherwise, default to parent court. This does not alter behavior for other sources. Solves freelawproject/juriscraper#847
- Pass opinion.html to site.extract_from_text, if opinion.plain_text does not exists. Solves #3549
- Add support to update Opinion object from extract_from_text metadata dict.

### DIFF
--- a/cl/scrapers/management/commands/cl_scrape_opinions.py
+++ b/cl/scrapers/management/commands/cl_scrape_opinions.py
@@ -5,6 +5,7 @@ from datetime import date
 from typing import Any, Dict, List, Optional, Tuple, Union
 
 from asgiref.sync import async_to_sync
+from courts_db import find_court_ids_by_name
 from django.core.files.base import ContentFile
 from django.core.management.base import CommandError
 from django.db import transaction
@@ -238,6 +239,26 @@ class Command(VerboseCommand):
         logger.debug(f"#{len(site)} opinions found.")
         added = 0
         for i, item in enumerate(site):
+            # Get court id from child court string passed by nytrial scrapers
+            child_court = None
+            if court_str.startswith("ny") and item.get("child_courts"):
+                child_court_id = find_court_ids_by_name(
+                    item["child_courts"],
+                    bankruptcy=False,
+                    location="New York",
+                    allow_partial_matches=False,
+                )
+                try:
+                    child_court = Court.objects.get(pk=child_court_id[0])
+                except IndexError:
+                    logger.warning(
+                        "Could not get child court id from name %s", court_str
+                    )
+                except Court.DoesNotExist:
+                    logger.warning(
+                        "Court object does not exist for %s", child_court_id[0]
+                    )
+
             # Minnesota currently rejects Courtlistener and Juriscraper as a User Agent
             if court_str in ["minn", "minnctapp"]:
                 headers = site.headers
@@ -303,7 +324,7 @@ class Command(VerboseCommand):
             dup_checker.reset()
 
             docket, opinion, cluster, citations = make_objects(
-                item, court, sha1_hash, content
+                item, child_court or court, sha1_hash, content
             )
 
             save_everything(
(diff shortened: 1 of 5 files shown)


cl/scrapers/management/commands/cl_scrape_opinions.py
```diff
@@ -238,6 +239,23 @@ def scrape_court(self, site, full_crawl=False, ocr_available=True):
         logger.debug(f"#{len(site)} opinions found.")
         added = 0
         for i, item in enumerate(site):
+            # Get court id from child court string passed by nytrial scrapers
+            child_court = None
+            if court_str.startswith("ny") and item.get("child_courts"):
+                child_court_id = find_court_ids_by_name(
+                    item["child_courts"], False, "New York", False
+                )
```
flooie, line 247: I wouldnt mind seeing the parameters names

cl/scrapers/tasks.py
```diff
@@ -57,7 +57,7 @@ def update_document_from_text(opinion: Opinion) -> None:
     site = get_scraper_object_by_name(court)
     if site is None:
         return
-    metadata_dict = site.extract_from_text(opinion.plain_text)
+    metadata_dict = site.extract_from_text(opinion.plain_text or opinion.html)
```
flooie, line 60: Thanks for fixing this.

flooie:
I think you need to include the latest release of courts-db v0.10.21 to poetry. Otherwise I think this is complete

flooie (changes requested):
Update courts-db to v0.10.21 please.

grossir:
I abstracted the behavior for child courts into get_child_court
I tested it locally creating the following courts:
```
from cl.search.models import Court

# Child court with expected parent
Court(id='nysupctnewyork', parent_court=Court.objects.get(pk='nysupct'), position='33.1', jurisdiction='TRT', full_name='New York County, Supreme Court', short_name='Test2').save()

# Child courts with unexpected parents, error should be logged and records saved with court id from scraper
Court(id='nysupctqueens', parent_court=Court.objects.get(pk='nysurct'), position='35.2', jurisdiction='TRT', full_name='Queens County, Supreme Court', short_name='Test3').save()
Court(id='nysupctkings', parent_court=Court.objects.get(pk='nysurct'), position='36.2', jurisdiction='TRT', full_name='Kings County, Supreme Court', short_name='Test3').save()
```
and checking that both logs and DB behaved as expected, and they did

cl/scrapers/utils.py
```diff
@@ -20,7 +21,52 @@
 from cl.lib.microservice_utils import microservice
 from cl.recap.mergers import find_docket_object
 from cl.scrapers.tasks import extract_recap_pdf
-from cl.search.models import Docket, RECAPDocument
+from cl.search.models import Court, Docket, RECAPDocument
+
+
+def get_child_court(child_court_name: str, court: Court) -> Optional[Court]:
```
flooie, line 27: I would not pass the full court object when the string will do.

cl/scrapers/management/commands/cl_scrape_opinions.py
```diff
@@ -302,8 +304,10 @@ def scrape_court(self, site, full_crawl=False, ocr_available=True):
             )
             dup_checker.reset()
 
+            child_court = get_child_court(item.get("child_courts", ""), court)
```
flooie, line 307: something about this is weird to me ...

flooie (approved):
Thanks @grossir  for putting this through its paces.

flooie:
@grossir what say you?

grossir:
From the Sentry event, the child court name was  `'Ct Cl'` and the child court id was `nyclaimsct`
So, it found the "parent" court as child. 
I think this merits a new control in that function, to ensure that the child found actually has a parent.

PR #3593, opened 2024-01-11
feat(favorites): Avoids reloading the tag details page

ERosendo:
This PR tweaks the code of the `onclick` event that we're adding to the untag buttons. 

The new event handler will avoid reloading the page when clicking the untag button, ensuring an uninterrupted flow and it will compute the list title to reflect the number of remaining dockets after the untag operation.

Here's a GIF showing how the button works:

untag-button


mlissner (approved):
Cool, thanks.

Issue #3608, opened 2024-01-16
Docket Authorities don't say case names


mlissner:
Not sure what's up with this, but during a demo today I loaded up this case: 

https://www.courtlistener.com/docket/68116934/authorities/hsuanyeh-law-group-pc-v-winston-strawn-llp/

In the Authorities section, none of the cases show their names. Not great! 

![image](https://github.com/freelawproject/courtlistener/assets/236970/3f1637a9-c7b2-472d-952d-7daa5846cc50)

It also affects the document pages:

![image](https://github.com/freelawproject/courtlistener/assets/236970/516bd00d-e052-4380-ad65-2407bfa06cd3)

ERosendo:
![image](https://github.com/freelawproject/courtlistener/assets/55959657/00e29277-7754-45b3-8fff-5395ec5ba84e)

PR #3616, opened 2024-01-16
fix(search): Adds sync version of the helper method to compute captions

ERosendo:
This PR addresses Issue #3608 by providing a synchronous option for computing Opinion cluster captions

### DIFF
--- a/cl/opinion_page/views.py
+++ b/cl/opinion_page/views.py
@@ -1044,7 +1044,6 @@ async def citation_handler(
     if cluster_count > 1:
         clusters_list = []
         async for cluster in clusters:
-            cluster.caption = await cluster.acaption()
             docket = await Docket.objects.aget(pk=cluster.docket_id)
             cluster.court = await Court.objects.aget(pk=docket.court_id)
             clusters_list.append(cluster)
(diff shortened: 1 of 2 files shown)


mlissner (commented):
The duplicated code that async seems to thrive on sort of makes me crazy. Could you just do an async_to_sync call on it instead?

ERosendo:
> The duplicated code that async seems to thrive on sort of makes me crazy.

I know! 😢 

> Could you just do an async_to_sync call on it instead?

I attempted this approach as my first solution. However, the `async` version of this helper doesn’t use the cached attributes of the queryset and instead creates multiple queries, leading to slow template loading. Here's an example:

```py

# sync version (this will use cached data if it is available)
if self.docket.court_id != "scotus":
    court = re.sub(" ", "&nbsp;", self.docket.court.citation_string)

# async version (this will create queries to the DB)
cluster = await OpinionCluster.objects.aget(pk=self.pk)
docket = await Docket.objects.aget(id=cluster.docket_id)
court = await Court.objects.aget(pk=docket.court_id)
if docket.court_id != "scotus":
    court = re.sub(" ", "&nbsp;", court.citation_string)
```

We can use `select_related` and `prefetch_related` in the async version. However, that will still generate a query. In addition, doing `self.docket` in the `async` method may raise an exception like #3610.

mlissner:
Hm, OK. Well, hopefully we can find a better way to do this that doesn't repeat so much code and that doesn't generate more queries.

PR #3654, opened 2024-01-24
feat(scraper): Add adapter to get binary content

flooie:
Mount SSL adapter to sessions that require it in juriscraper


flooie:
@grossir - would you mind taking a look at this please.  Im not really using juriscraper the way I maybe should but it seemed painless today to do this.  We do the same thing for user agents more or less.

mlissner (commented):
If the only thing you're doing in the make session function is setting the cipher suite, what about having a get_cipher function that returned just that value (or None, if not an issue)?

PR #3663, opened 2024-01-25
feat(github-actions): Add new action

flooie:
Create new action to automate PRs for FLP dependency releases.


.github/workflows/flp-dependencies-pr.yml
```diff
@@ -0,0 +1,33 @@
+name: PR update Free Law Project dependencies
+
+on:
+  workflow_dispatch:
```
mlissner, line 4: What does this mean in lay terms or is there a link I can look at?

.github/workflows/flp-dependencies-pr.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          python-version: '3.x'
+
+      - name: Install Poetry
+        run: |
+          curl -sSL https://install.python-poetry.org | python3 -
+
+      - name: Update FLP dependencies
+        run: |
+          poetry update juriscraper reporters-db courts-db eyecite
```
mlissner, line 24: It'd be nice to alphabetize these.

.github/workflows/flp-dependencies-pr.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v3
+        with:
+          token: ${{ secrets.TOKEN }}
+          commit-message: Update freelawproject dependencies
+          title: Update freelawproject dependencies
+          body: This PR updates the juriscraper package.
```
mlissner, line 32: Can you fix this?

mlissner (changes requested):
Cool! A few questions. 

I think you'll also need to allow all versions of these dependencies in the pyproject.toml file, or else this will not work once you hit the requirements it lays out. 

The [docs](https://python-poetry.org/docs/cli/#update) say: 

> Note that this will not update versions for dependencies outside their version constraints specified in the pyproject.toml file. In other terms, poetry update foo will be a no-op if the version constraint specified for foo is ~2.3 or 2.3 and 2.4 is available.

.github/workflows/flp-dependencies-pr.yml
flooie: I think I need your knowledge about what/generate the tokens to use

flooie:
> Cool! A few questions.
> 
> I think you'll also need to allow all versions of these dependencies in the pyproject.toml file, or else this will not work once you hit the requirements it lays out.
> 
> The [docs](https://python-poetry.org/docs/cli/#update) say:
> 
> > Note that this will not update versions for dependencies outside their version constraints specified in the pyproject.toml file. In other terms, poetry update foo will be a no-op if the version constraint specified for foo is ~2.3 or 2.3 and 2.4 is available.

Yes it won't update the toml  file but it will work because it should update and grab the latest version.  the pyproject isnt locked to a particular version just a greater than version.

.github/workflows/flp-dependencies-pr.yml
flooie: so glad you asked.  

if you do this you get a nice option to just trigger a new workflow on the actions page. 
Screenshot 2024-01-25 at 7 37 12 PM
Screenshot 2024-01-25 at 7 37 22 PM

.github/workflows/flp-dependencies-pr.yml
mlissner: No, I mean the `body` key, which is incorrect. I think the secrets might be fine.

mlissner:
> Yes it won't update the toml file but it will work because it should update and grab the latest version. the pyproject isnt locked to a particular version just a greater than version.

The dependencies _are_ locked to particular versions in pyproject by using caret versions. These are the versions in pyproject:

```
courts-db = "^0.10.22"
eyecite = "^2.5.5"
juriscraper = "^2.5.82"
reporters-db = "^3.2.38"
```

This [means that](https://stackoverflow.com/questions/54720072/dependency-version-syntax-for-python-poetry):

 - courts-db will never auto-update to 0.11.x (zero point oh versions are treated differently)
 - eyecite will never update to 3.x or 2.6.x
 - juriscraper will never update to 3.x or 2.6.x
 - reporters-db will never update to 4.x or 3.3.x

What we should have instead are wildcards, so this action always works.

.github/workflows/flp-dependencies-pr.yml
mlissner: Awesome. Easy!

.github/workflows/flp-dependencies-pr.yml
flooie: great.  ive updated it

mlissner (approved):
LGTM, thanks. Where should we put this in our documentation?

PR #3672, opened 2024-01-26
Update Black to match release

flooie:
PR updates black to latest release version to match GitHub action

Updates the precommit yml and 
Updates the ignore revisions file


flooie:
The failing test - is again the recap async bug.

mlissner:
Cool, thanks for taking care of this. I think `black` is moving towards annual format updates, which is why this is in January?

flooie:
Yep - that is what it seems like from their release notes

PR #3696, opened 2024-01-30
fix(scraper.utils): Add content type check when scraping

flooie:
Check binary content type when scraping opinions. 

Long story short we keep downloading JS files that are meant to be PDFs 
because the courts sometimes just send us that. 

Add a simple check to make sure the response is not a JS file and if it is
throw an error log and don't process it.

### DIFF
--- a/cl/scrapers/utils.py
+++ b/cl/scrapers/utils.py
@@ -202,6 +202,11 @@ def get_binary_content(
             msg = f"EmptyFileError: {download_url}\n{traceback.format_exc()}"
             return msg, None
 
+        # test for captcha redirects
+        if r.headers.get("Content-Type") == "text/javascript":
+            msg = f"RedirectJSError: {download_url}"
+            return msg, None
+
         # test for and follow meta redirects
         r = follow_redirections(r, s)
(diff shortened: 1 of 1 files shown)


cl/scrapers/utils.py
```diff
@@ -202,6 +202,11 @@ def get_binary_content(
             msg = f"EmptyFileError: {download_url}\n{traceback.format_exc()}"
             return msg, None
 
+        # test for captcha redirects
```
grossir, line 205: From the `nev` issue that inspired this PR, I don't think this is triggered by a captcha. Also, the JS extension is gotten from doctor, not directly from the response, so the actual check for `.js` extension should be here
https://github.com/freelawproject/courtlistener/blob/main/cl/scrapers/management/commands/cl_scrape_opinions.py#L139

Check my comment in the [issue](https://github.com/freelawproject/courtlistener/issues/3695#issuecomment-1918107491) for more detail

flooie: Fascinating- any chance you validated that this file was being identified in Doctor as a JS file?

PR #3710, opened 2024-02-01
feat(template): Updates language to use "Case law" instead of "Opinion"

### LINKED ISSUE #3472: Change headers from "Opinions" to "Case Law"
I was demoing CourtListener the other day and I realized that people think about things in terms of "Case Law," not "Opinions".

We should update CourtListener and free.law to say Case Law wherever possible.

@flooie, do you agree?

ERosendo:
This PR closes #3472 by using the word "Case law" instead of "Opinion" in several HTML templates.


mlissner:
Alright, assuming tests pass, this is set for auto-merge. I took mercy on you and made my 50 or so tweaks myself so you wouldn't have to. Thank you regardless for getting this to where I could work on it.

PR #3722, opened 2024-02-02
feat(user): Push new CL users to Neon

### LINKED ISSUE #3145: Push new CL users to Neon
Our donation platform to be, Neon, doesn't charge based on user-count and it is what we will be using to send email campaigns. Therefore, all CL users should be pushed to Neon when they are created. 

It's also worth considering whether to keep Neon in sync when somebody updates certain fields, like their email address (probably, but this is of diminishing returns).

ERosendo:

- Adds helper functions to the NeonClient class to create and update account records.

- Adds a task to send user's data to Neon after registration and one to update the account record after updating the profile data. 

- Removes all the code related to moosend.

- Removes newsletter stuff from the profile page.

- Updates the newsletter section in the base template to add the link to the new Neon form. 

image

Fixes #3145


cl/users/tasks.py
```diff
@@ -22,57 +24,81 @@ def abort_or_retry(task, exc):
...
+        # Neon CRM automatically merges accounts with identical names and
+        # email addresses. This process, called Account Match feature, ensures
+        # consistent data across records. However, If the accounts are very
+        # close (email and phone match but not name) the accounts will be
+        # entered into the Partial Match Queue. This queue allows users to
+        # review these potential matches and decide whether to merge them
+        # manually.
+        raise AssertionError(
+            "There's more than one account using the same email address. "
+            "We should check the Partial Match Queue."
```
mlissner, line 61: This will wind up in sentry. Could we include a detail about the email address that failed in the assertion error? Perhaps:

```suggestion
            "There's more than one account using the same email address. "
            "We should check the Partial Match Queue for: %s" % user.email
```

cl/users/tasks.py
```diff
@@ -22,57 +24,81 @@ def abort_or_retry(task, exc):
...
-        error = j.get("Error", "Unknown error")
-        logger.warning(
-            "Did not complete '%s' action on '%s' in moosend: '%s'",
-            action,
-            email,
-            error,
-        )
+    new_account_id = neon_client.create_account(user)
+    profile.neon_account_id = new_account_id
+    profile.save()
```
mlissner, line 75: To make you code more self-documenting, I'd put this under an if statement: 

```
if num_neon_accounts_found == 0:
    # No account found, create one
```

cl/users/views.py
```diff
@@ -538,6 +523,8 @@ def register(request: HttpRequest) -> HttpResponse:
                     urlencode(redirect_to),
                     urlencode(user.email),
                 )
+                if not settings.DEVELOPMENT:
+                    create_neon_account.delay(user.pk)
```
mlissner, line 527: It occurs to me that when somebody registers for an account, we can't trust their email address. We can only trust it after it has been confirmed. We want to prevent information leakage. 

For example, imagine:

1. Bob makes a donation in Neon and is assigned user number 4242.
2. Malory wants to see if Bob is a donor to FLP.
3. Malory creates an account in CL using Bob's email address.
4. Malory goes to /profile/my-donations/ and sees a link to "My Neon Portal", with an `href` of "neon.com/user/?id=4242". If Bob had not donated, the `href` would have been "neon.com/login/".
5. Now Malory knows, because of the ID in the link, that Bob is a donor.

So, simple solution, don't do this until after somebody has confirmed their email address, and make sure that we handle it properly when they change (AND confirm) their email address too.

cl/users/tasks.py
mlissner: Actually, instead of kicking this out to the developers to deal with, let's have our development person/team do it. To accomplish that, let's send Jenifer an email:

 - To: donate@free.law
 - Subject: [Action Needed]: Please merge multiple accounts found for %s % user.email
 - Body: 

    > A new user signed up for CourtListener and we tried to match their email address to Neon. Alas, it didn't work because the email address `{ user.email }` appears is in Neon more than once.
    > 
    > Please take the following actions to fix this situation:
    > 
    > 1. Please review the "Partial Match Queue" in Neon, as documented here:
    > 
    >     https://...
    > 
    >     In that queue, you should find this user and you should be able to merge their records. 
    > 
    > 2. Once that's completed, please add the Neon account ID of the merged record to CourtListener at the following link:
    > 
    >     https://www.courtlistener.com/admin/user/xxx
    > 
    > Thank you for keeping everything tidy and good! 
    > 
    > The bots at CourtListener

mlissner (changes requested):
Just a few things. Thanks!

cl/users/tasks.py
mlissner: @jeniferw, I wanted to highlight this for you. We're tweaking CourtListener so that when somebody signs up for an account on CL, their information gets pushed to Neon, and so that when they update their email address on CL, it gets updated on Neon.

We're also tweaking it so that if somebody *already* has a Neon account when they sign up on CL, we'll get the Neon ID and add it to CL. That should work most of the time, except when that person has multiple accounts in Neon's "Partial Match Queue". In that case, it'll need human review and fixing, so CL will send the above email to donate@free.law. 

Does this all sound good / look good / seem right to you? Now's the best time to chime in with tweaks you'd like.

mlissner:
Future work:
 - [ ] Migrate current subscribers to Neon.
 - [ ] Remove newsletter field from model.
 - [ ] Optionally: Unsubscribe people when they delete CL accounts.

cl/users/tasks.py
ERosendo: Thanks for the suggestion. I will update the code accordingly.

ERosendo:
@mlissner I can remove the `newsletter` field from the `UserProfile` model in this PR, would that be okay?

mlissner:
Yes. Jenifer and I can do an export before merging and then we'll be all set without it.

cl/users/tasks.py
```diff
@@ -22,57 +26,89 @@ def abort_or_retry(task, exc):
...
+    max_retries=3,
+    interval_start=5,
+    ignore_result=True,
 )
-def update_moosend_subscription(self: Task, email: str, action: str) -> None:
-    """Subscribe or unsubscribe email address to moosend mailing list.
+def create_neon_account(self: Task, user_id: int) -> None:
+    """
+    Checks for existing Neon CRM accounts using the user's email address
+    and handles account creation or raises AssertionError when multiple
```
mlissner, line 38: Sorry, looks like this is out of sync with the tweak below. Don't you wish this were Java, where it's fail to compile if you didn't correctly handle all exceptions, and therefore you wouldn't need to note what it raises in your docstring??? :p

cl/users/templates/emails/neon_multiple_accounts.txt
```diff
@@ -0,0 +1,20 @@
+A new user signed up for CourtListener and we tried to match their email address
+to Neon. Alas, it didn't work because the email address {{ user.email }} appears
+is in Neon more than once.
+
+Please take the following actions to fix this situation:
+
+i. Please review the "Partial Match Queue" in Neon, as documented here:
```
mlissner, line 7:
```suggestion
A new user signed up for CourtListener and we tried to match their email address
to Neon. Alas, it didn't work because the email address '{{ user.email }}' appears
is in Neon more than once.

Please take the following actions to fix this situation:

1. Please review the "Partial Match Queue" in Neon, as documented here:
```

mlissner (changes requested):
Two small comments, please, then I'll merge, but before I can deploy I need to export the data for Jenifer.

ERosendo:
@mlissner Thanks for your feedback. I've implemented the changes you recommended.

mlissner:
Jenifer and I will migrate data on Thursday, then merge this.

mlissner (approved):

mlissner:
This is merged and deployed. Thank you.

PR #3723, opened 2024-02-02
chore(build): drop redundant docker image build step

cweider:
With #3134, the celery and django containers began to use the same image, making the celery-specific job unproductive.


mlissner:
Omg.

PR #3733, opened 2024-02-03
refactor(deploy): replace Makefile with `build-push-action`

cweider:
With the Makefile radically simplified by #3260, the work that it does can easily be done by a `docker/build-push-action` step in the deployment workflow. Beyond than simplicity, the use of the action makes build caching available and will trim a decent amount of time from the build job.

In this change, the method of referring to the image is changed: instead of a tag (e.g. `c29346b-prod`) the digest of the uploaded image is used. The digest is an unambiguous and immutable reference to an image, where a tag can be changed. For this reason, it's a recommended for deployments (see https://cloud.google.com/kubernetes-engine/docs/concepts/about-container-images).

Also in this change is the use of `docker/metadata-action`. It is configured to produce the same tags as before (`c29346b-prod`), but it is capable of doing much more. A future direction is to replace the `sha`-based tag with either a build date or build id (`$RUN_ID-$RUN_ATTEMPT`). For now we only introduce `labels`, which attaches creation date and the full sha SHA to the built images.


mlissner:
This looks good to me, but I'd love some more eyes since it could affect the dev workflow. I'm going to flag it in our internal slack channel for folks to look at.

mlissner:
@blancoramiro, this one got stalled a month+ ago because I didn't feel confident to merge it. Can you take a look?

cweider:
Pushed-up commit that brings this up-to-date with the `main` as it is now. The deploy workflow hasn’t change much, just a few adjustments for subsequently added services, `cl-celery-prefork-es-sweep`, `cl-send-rt-percolator-alerts`, `cl-es-sweep-indexer`.
```
% git diff b3f3aef...origin/main -- .github/workflows/docker-build.yml
```

PR #3738, opened 2024-02-05
3695 content types scraping opinions

flooie:
Adds support for checking content types 

This is a compliment to https://github.com/freelawproject/juriscraper/pull/921

which adds expected content types to OpinionSite object


flooie:
I seem to be having more bad luck around tests - different tests unrelated to changes to juriscraper appear to fail.

PR #3740, opened 2024-02-06
refactor(alerts): Reorganize sections in the docket alert email template

### LINKED ISSUE #3112: Move alert info higher in alert emails
This is a simple reorganization to make our alert emails more skimmable in mail clients. A user suggests:

> Might it be possible to put some of the header information in Docket Alerts – especially “Use notes and tags to organize and share the cases you follow. Learn more” – below the substance rather than above?  When one is looking at one’s mailbox and sees the text of the current message in the preview pane, it’s helpful to have the substance immediately visible, without the need for scrolling, so one can quickly delete e-mail in which one isn’t interested (e.g., because it refers to minor procedural filings).

He's right. We should do this.

ERosendo:
This commit repositions the notes section in the `docket_alert_email` template files, placing it after the list of filings. This change enhances the scannability of our alert emails within mail clients. 

Fixes #3112

### DIFF
--- a/cl/alerts/templates/docket_alert_email.html
+++ b/cl/alerts/templates/docket_alert_email.html
@@ -60,27 +60,6 @@
     </p>
 
     <hr style="background: #ddd; color: #ddd; clear: both; float: none; width: 60%; height: .1em; margin: 0 0 1.45em; border: none;">
-
-    {% if notes or tags %}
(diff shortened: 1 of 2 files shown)


mlissner:
Easy and impactful. Best kind, thanks.

PR #3744, opened 2024-02-06
Update columbia merger

quevon24:
Improve process to clean docket numbers
Fix process to update docket number
Update function to merge long text fields like posture or syllabus


cl/corpus_importer/management/commands/columbia_merge.py
```diff
@@ -370,8 +370,12 @@ def merge_docket_data(docket_data: dict, cluster: OpinionCluster) -> None:
     data_to_update = {}
 
     if docket_data["docket_number"]:
-        merge_docket_numbers(cluster, docket_data["docket_number"])
-        cluster.docket.refresh_from_db()
+        new_docket_number = merge_docket_numbers(
```
flooie, line 373: can you provide an example for this?  why is this being added

cl/corpus_importer/utils.py
```diff
@@ -315,23 +315,51 @@ def clean_docket_number(docket_number: str) -> str:
...
+    # remove File in File CV-03 0402585
+    docket_number = re.sub("File ", "", docket_number)
+    # remove judgement affirmed in "31130. Judgment affirmed."
+    docket_number = re.sub("Judgment affirmed.?", "", docket_number)
+    # remove cause transferred in "31342. Cause transferred."
+    docket_number = re.sub("Cause transferred.?", "", docket_number)
+    # remove parentheses/brackets in (AC 16691)
+    docket_number = re.sub("[()]", "", docket_number)
+    # remove comma in No. 28,049
+    docket_number = docket_number.replace(",", "")
```
flooie, line 341: it may be useful to reduce these to numbers for comparison sake but we dont want to remove strings from docket numbers in all cases

cl/corpus_importer/management/commands/columbia_merge.py
quevon24: There are clusters without docket number, in some of the matches from the full list, for example this cluster: https://www.courtlistener.com/opinion/1439900/go/ and this file and docket number  "E2000-02272-COA-R3-CV" from file matched tennessee\court_opinions\documents\4c52f83b11e551bb.xml

The function could be reduced to that if there is no docket number in the system then save the one from the file.

cl/corpus_importer/utils.py
quevon24: No problem, i can leave that in another separate function, since it could be useful to us in the future

flooie:
Let's shelve this for now - this code doesnt affect our current import.

PR #3754, opened 2024-02-09
build(deps): bump cryptography/pyopenssl versions

cweider:
Dependabot has had trouble generating the patch addressing CVE-2023-50782. This accomplishes the update.

See https://github.com/freelawproject/courtlistener/network/updates/784272861


mlissner:
Thanks

Issue #1616, opened 2021-03-15
Swagger (OpenAPI) schema contain wrong URLs


elibrin:
The Swagger (OpenAPI) schema contain wrong URLs to the API.
For example: https://www.courtlistener.com/api/rest/v3/api/rest/{version}/dockets/

Step to reproduce:
1.  Browse to:  https://www.courtlistener.com/api/swagger/
2.  See the links in the returned json.
3.  Link points to wrong location. and
4.  {version} is in the path

cite-reader:
The duplication of the `/api/v3/` fragment isn't the only issue with the paths object. Field names of the paths object (that is, the urls at which the API endpoints may be found) are not meant to be absolute URLs: "A relative path to an individual endpoint. The field name MUST begin with a slash. The path is appended (no relative URL resolution) to the expanded URL from the [Server Object](https://spec.openapis.org/oas/v3.0.2#serverObject)’s url field in order to construct the full URL." ([OpenAPI specification version 3.0.2, § 4.7.8 Paths Object](https://spec.openapis.org/oas/v3.0.2#paths-object)). _That_ issue is a result of encode/django-rest-framework#7631; the issue is marked as stale, but the maintainers seemed amenable to receiving a PR that fixes it in their last comment, so I'll see if I can get that fixed myself.

PR #3770, opened 2024-02-13
chore(api): remove OpenAPI/Swagger integration

cweider:
The `django-rest-framework`'s built-in support for OpenAPI schema generation has been deprecated (encode/django-rest-framework#8773). `drf-spectacular` is the recommended alternative, but CourtListener's schema has always been experimental. Moreover the current integration is crashing and has other bugs that prevent its effective use. Outright removal, rather than replacement is the best way forward.

Fixes: #3653, #1616

### DIFF
--- a/cl/api/templates/rest-docs-vlatest.html
+++ b/cl/api/templates/rest-docs-vlatest.html
@@ -43,7 +43,7 @@
 
     <pre class="scrollable">curl -X OPTIONS "{% get_full_host %}{% url "api-root" version="v3" %}"</pre>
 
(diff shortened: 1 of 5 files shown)


cweider:
As-is change doesn’t mention removal in the REST API change log (or affect version), but it could:
https://github.com/freelawproject/courtlistener/blob/ab5e6a45c2237ed373b040670a79c10dc93f8efb/cl/api/templates/rest-change-log.html#L31

mlissner (approved):

mlissner:
Thanks Chad. Good to have this cleaned up. Setting for auto-merge!

Issue #3228, opened 2023-10-04
Enhance the Profile > Docket Alerts page


mlissner:
A user just asked that we enhance the Docket Alert page that people see in their profile. Mine looks like this:

![image](https://github.com/freelawproject/courtlistener/assets/236970/fadfd49d-aa90-4365-bd86-9114acbad7d8)

They ask that we add more metadata to the page. Things like:

> "Date Filed", "Assigned To", "Cause" etc.

We could do this pretty easily by adding a second line to each alert, and I could see it being quite helpful. Good idea.

mlissner:
Related: https://github.com/freelawproject/courtlistener/issues/4271

mlissner:
Another comment, from Slack: 

![image](https://github.com/user-attachments/assets/16080c9a-9136-45d0-a1b7-22a3c589db00)

mlissner:
Another similar request: 

![image](https://github.com/user-attachments/assets/683be087-2f02-48f9-aea3-567ae50afbca)

mlissner:
I followed up with each of the people that reported this. All set!

PR #3796, opened 2024-02-15
feat(alerts): Add unsubscribe headers to emails

### LINKED ISSUE #3617: Add Unsubscribe headers to emails (Deadline May 31st)
Right now, we have unsubscribe links in our emails which Google (impressively?) identifies:

image

Clicking that takes you to:

image

And clicking *that* takes you to the CourtListener unsubscribe link. 

That's neat, but, alas, starting in June it won't be good enough, and Google/Yahoo! will punish us if we don't provide unsubscribe headers like these:

image

At least this is what [AWS's documentation](https://aws.amazon.com/blogs/messaging-and-targeting/an-overview-of-bulk-sender-changes-at-yahoo-gmail/) says.

So we should add these headers. I'm guessing this will be pretty simple in our custom email router thing.

ERosendo:
This PR adds unsubscribe headers for docket and search alerts and fixes #3617. Key changes include:

- A new endpoint to unsubscribe from docket alerts with one click
- Updates the `disable_alert` method to support post requests
- Tweaks tests to check the new endpoints and email headers.
- Adds the `List-Unsubscribe-Post` and `List-Unsubscribe` headers to docket and search alert emails


cl/alerts/tests.py
mlissner, line 1: This looks fine, but I wouldn't hard-code URLs even in tests. Using `reverse` is still best.

cl/alerts/urls.py
```diff
@@ -38,4 +39,9 @@
         toggle_docket_alert_confirmation,
         name="toggle_docket_alert_confirmation",
     ),
+    re_path(
+        "alert/docket/one_click_unsubscribe/([a-zA-Z0-9]{40})/",
```
mlissner, line 43: I don't think we ever use underscores in URLs, so:

```suggestion
        "alert/docket/one-click-unsubscribe/([a-zA-Z0-9]{40})/",
```

cl/alerts/views.py
```diff
@@ -60,12 +63,16 @@ def delete_alert_confirm(request, pk):
...
+@csrf_exempt
+@require_http_methods(["GET", "POST"])
 def disable_alert(request, secret_key):
     """Disable an alert based on a secret key."""
     alert = get_object_or_404(Alert, secret_key=secret_key)
     prev_rate = alert.rate
     alert.rate = Alert.OFF
     alert.save()
+    if request.method == "POST":
+        return HttpResponse("You have been successfully unsubscribed!")
```
mlissner, line 75:
```suggestion
    if request.method == "POST":
        # Mail clients send POSTs; an ugly response is OK.
        return HttpResponse("You have been successfully unsubscribed!")
```

mlissner (commented):
This looks about right to me. A couple little suggestions, but, really, Alberto is our expert, so let's see if he catches anything I missed. LGTM though, thank you!

ERosendo:
@mlissner thanks for your feedback.

albertisfu (commented):
The code looks great!

However, I'm concerned about Search Alerts. A Search Alert can contain many alerts in a single email.

Screenshot 2024-02-16 at 19 07 21

So, if we only aim to comply with the new header, it seems sufficient to select the first search alert: `alert = hits[0][0]`. However, in terms of functionality, this will only unsubscribe users from the first alert in an email, leaving the rest active. 
Maybe creating a view to handle the unsubscription from all the alerts contained in an email could solve this issue?

mlissner:
That's a great point, but I don't think unsubscribing from all alerts is what users would want either. 

It looks like we can get away with doing [RFC2069](https://datatracker.ietf.org/doc/html/rfc2369) without RFC8058, by just providing a URL that responds to a GET request. If we do that, we can present the user's subscriptions _for that email_, and they can opt out to whatever they want. I'm imagining a GET url that just takes multiple secret keys as a parameter.

We can also be clever and do one-click (RFC8058) if there's only one alert triggered in the message, and do the old school way above (RFC2069) if there's more than one alert triggered in the message. 

I think we wound up putting some sort of captcha (or other JS thing) on the unsubscribe link we currently have, so spam bots wouldn't trigger it accidentally.

Related to that, reading [RFC8058](https://datatracker.ietf.org/doc/html/rfc8058), it mentions the threat that a troublemaker would unsubscribe everybody from everything. Let's make sure that's not possible by adding some rate limiters to all of these views.

ERosendo:
@mlissner I've updated this PR. Key changes include:

- Adds a rate-limiting decorator to the view that handles unsubscription from alerts with one click.

- Implements a captcha for anonymous users accessing the view to disable search alerts to enhance security. Logged-in users bypass the captcha.

- Refines the logic we use to create headers for the search alert emails:
   - If there's a single alert triggered in the message, we use a one-click unsubscribe link (compliant with RFC8058)
  
   - If multiple alerts are triggered in the message, we use the `List-Unsubscribe` header and provide a URL that responds to a GET request. This new URL takes multiple secret keys as a parameter and displays the user's subscriptions for that email message in a list. Here's a gif showing the new page to manage search alert subscriptions: 
   
   
Screen Recording 2024-02-27 at 9 37 45 AM

cl/alerts/management/commands/cl_send_alerts.py
```diff
@@ -63,10 +65,32 @@ def send_alert(user_profile, hits):
...
+            "List-Unsubscribe-Post": "List-Unsubscribe=One-Click",
+            "List-Unsubscribe": f"<https://www.courtlistener.com{disable_url}>",
+        }
+    else:
+        params = {"keys": [hit[0].secret_key for hit in hits]}
+        query_string = urllib.parse.urlencode(params, doseq=True)
+        alert_list = reverse("disable_alert_list")
+        headers = {
+            "List-Unsubscribe": f"<https://www.courtlistener.com{alert_list}?{query_string}>",
+        }
```
mlissner, line 84: This is a quibble, but how about:

```suggestion
    headers = {}
    query_string = ""
    if len(hits) == 1:
        alert = hits[0][0]
        unsubscribe_path = reverse(
            "one_click_disable_alert", args=[alert.secret_key]
        )
        headers["List-Unsubscribe-Post"] = "List-Unsubscribe=One-Click",
    else:
        params = {"keys": [hit[0].secret_key for hit in hits]}
        query_string = urlencode(params, doseq=True)
        unsubscribe_path = reverse("disable_alert_list")
    headers["List-Unsubscribe"] = f"<https://www.courtlistener.com{unsubscribe_path}?{query_string}>"
```

cl/alerts/tasks.py
```diff
@@ -1,4 +1,5 @@
 import copy
+import urllib.parse
```
mlissner, line 2: Some comment as above. :)

cl/alerts/templates/includes/manage-alert-table-row.html
```diff
@@ -0,0 +1,32 @@
...
+      {% if alert.alert_type == "o" %}
+        Case Law
+      {% else %}
+        Oral Argument
+      {% endif %}
+    </p>
+  </td>
+  <td>
+    <p class="bottom">
+      {{ alert.name }}
```
mlissner, line 13: I think we should probably at least make this a link, like in `/profile/alerts/`. 

If we want to go wild though (and perhaps we should), could you also check how similar the two tables are, and if it makes sense to do so, merge them into a common code base? And also take note of the requested changes in this issue: https://github.com/freelawproject/courtlistener/issues/3228

Sorry, I guess this was a landmine!

cl/alerts/templates/manage_search_alerts.html
```diff
@@ -0,0 +1,82 @@
...
+{% block content %}
+<div class="col-sm-2"></div>
+<div class="col-sm-8">
+  <div class="text-center">
+    <h1 class="mb-2">Manage your search alerts subscriptions</h1>
+  </div>
+  <p class="lead">
+    Right here on this page, disable any alerts that clutter your inbox and keep only the ones that matter most.
+    Stay informed, stay focused, and enjoy a more personalized experience. Don't worry, disabling is temporary – you can
+    always re-enable them later.
```
mlissner, line 27: This is simpler (but assumes we add more editing functionality):

```suggestion
    The alert email you received contained results for multiple alerts. Use this page to disable or edit any alerts that are cluttering your inbox.
```

cl/alerts/templates/search_alert_confirmation.html
mlissner, line 1: Sorry if I missed it, but can you share a screenshot of this page too?

cl/alerts/views.py
```diff
@@ -60,16 +64,116 @@ def delete_alert_confirm(request, pk):
...
+    """Disables a specified alert within an HTMX-powered page."""
+    if not request.META.get("HTTP_HX_REQUEST"):
+        return HttpResponseNotFound(
+            "Your attempt to disable the alert was unsuccessful."
+        )
+
+    alert = get_object_or_404(Alert, secret_key=secret_key)
+    alert.rate = Alert.OFF
+    alert.save()
+    # Mail clients send POSTs; an ugly response is OK.
```
mlissner, line 90: This comment doesn't make sense here, does it?

cl/alerts/views.py
mlissner, line 1: We need ratelimiters on any of these that rely on secret keys and allow anonymous users to change data.

mlissner (changes requested):
Looks pretty good to me architecturally. A handful of comments, but I'd love to have Alberto take a look too.

cl/alerts/management/commands/cl_send_alerts.py
```diff
@@ -1,5 +1,6 @@
 import datetime
 import traceback
+import urllib.parse
```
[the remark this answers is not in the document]
ERosendo: You're right! Thanks for bringing this to my attention.

cl/alerts/management/commands/cl_send_alerts.py
ERosendo: Your suggestion definitely looks better. Thank you, I'll go with that.

cl/alerts/templates/manage_search_alerts.html
ERosendo: I'll update the template. Thanks

cl/alerts/templates/search_alert_confirmation.html
ERosendo: Sorry, I forgot to attach a screenshot of the page. Here's a GIF for reference.

Screen Recording 2024-02-28 at 12 22 41 AM

cl/alerts/views.py
ERosendo: You're right, this comment doesn't make sense. I'll remove it.

cl/alerts/templates/search_alert_confirmation.html
mlissner: Hmm, I thought we had a hidden captcha for this on our other unsubscribe page?

ERosendo: > Hmm, I thought we had a hidden captcha for this on our other unsubscribe page?

We have a similar one on the webpage that allows you to unsubscribe from docket alerts.

Screen Recording 2024-02-28 at 12 45 16 AM

Should we add a hidden captcha to the new page that displays a list of all the search alerts included in the email?

ERosendo:
Here's a list of recent changes::

- Added rate limiting to all views allowing users to unsubscribe from or disable alerts. 

- Refactored the code for the tables in `/profile/alerts/` and the new page to manage search alerts from emails. @mlissner was right and both tables were identical. 

- Updated the import statement for the `urlencode` method.

- Refined the message on the `manage_search_alerts` template

cl/alerts/templates/search_alert_confirmation.html
albertisfu: yeah, the captcha in these views was hidden and only shown when the request potentially came from a bot. Now, the challenge is always displayed for anonymous users. After reviewing the `hcaptcha` [documentation](https://docs.hcaptcha.com/invisible/), it appears the functionality has changed. There is a new option called Passive that, in combination with invisible, can replicate the previous behavior of only showing the challenge to risky users. However, this new option is available only to Enterprise users. If we are an Enterprise user, we'll just need to go to the site settings and select the mode 99.9% Passive.

Screenshot 2024-02-28 at 17 21 59

mlissner: Hm, we don't have Pro or Enterprise and I'm not inclined to pay for it, so I guess this is the way it'll be. I did turn down our captcha difficulty, so perhaps that'll hide them some of the time. I'll reach out about non-profit discounts though and see if they offer anything.

mlissner:
Thanks Eduardo. That's it for me, once you and Alberto are happy.

cl/alerts/management/commands/cl_send_alerts.py
```diff
@@ -63,10 +65,30 @@ def send_alert(user_profile, hits):
...
+        unsubscribe_path = reverse(
+            "one_click_disable_alert", args=[alert.secret_key]
+        )
+        headers["List-Unsubscribe-Post"] = "List-Unsubscribe=One-Click"
+    else:
+        params = {"keys": [hit[0].secret_key for hit in hits]}
+        query_string = urlencode(params, doseq=True)
+        unsubscribe_path = reverse("disable_alert_list")
+    headers["List-Unsubscribe"] = (
+        f"<https://www.courtlistener.com{unsubscribe_path}?{query_string}>"
```
albertisfu, line 81: This line is adding an unnecessary `?` when there is only one search alert in the email. I think it doesn't affect the request, but we could just remove it.

```
List-Unsubscribe-Post: List-Unsubscribe=One-Click
List-Unsubscribe: <https://www.courtlistener.com/alert/one-click-disable/secret_key/?>
```

albertisfu (commented):
Great, the changes here are pretty good. 
I tested them, and they work as expected.

Just one small suggestion related to a URL.

And one question:
Is it not required to also add a captcha for each button in this view when the user is not authenticated, similar to `disable_alert`, which disables only one alert?

Screenshot 2024-02-28 at 20 32 16

mlissner:
> Is it not required to also add a captcha for each button in this view 

Hm. We get a pretty specific header that the view checks for when we click it. Would a bot send the same one?

cl/alerts/management/commands/cl_send_alerts.py
ERosendo: Thanks for the suggestion. I updated the f-string to avoid adding the question mark when there isn't a query string.

mlissner:
Wonderful. Thank you both, these will be nice QOL improvements.

PR #3799, opened 2024-02-16
Exclude combined opinions from columbia merger

quevon24:
This small change excludes combined opinions when we try to map and merge the opinions in the columbia merger.

We need to exclude those opinions because we have splitted opinions in the columbia dataset.


quevon24:
For this PR the change is minimal and avoids conflicts when trying to match combined opinions from cl when the xml has multiple opinions. The only thing that could be improved is the FILED_TAGS list which contains the valid text strings to identify a date as date_filed

PR #3810, opened 2024-02-20
fix(users): Tweaks the admin link in the neon multi-account email

ERosendo:
This PR fixes the admin URL in the `neon_multiple_accounts.txt` email template


mlissner:
I'm realizing that if we used `reverse` in this, we wouldn't have had this bug.

Next time; merging. :)

Issue #3811, opened 2024-02-20
Opinion document content has not been extracted


grossir:
I identified 10197 opinions which have a `download_url` but no extracted content, using the Dev DB. The record had neither of the following content fields, but had a download URL:
- plain_text
- html
- html_lawbox
- html_columbia
- html_anon_2020

The list of ids can be gotten using the query at the end against the prod DB. Maybe a custom script should be written to reprocess our backup files

As for the cause of the failed extraction, it depends on the source. For example, 99% (3028) of `bia` errors happen in a single day 2022-01-13, which makes me think it was an ingestion script error. The documents did have available text inside

For `ncctapp` and `moctapp` and `nc` all the empty documents have local paths starting with `bin/`. Even when a local inspection shows they are pdfs. Perhaps `doctor` is not prepared to extract those?. [Example](https://www.courtlistener.com/opinion/4897514/state-v-medlin/)

Some of the other top sources may have an explanation. The courts with lower volume of errors may be to a one-time failure on the doctor service. However, there seems to be a distinct increase of errors in late 2018
![image](https://github.com/freelawproject/courtlistener/assets/14970769/ad672f07-9470-47cf-a56b-e3516b5bb058)

Most contentless records are clustered in a few sources. 

|court_id        | count|first_date_created |last_date_created |
|:---------------|-----:|:------------------|:-----------------|
|bia             |  3039|2021-12-31         |2022-01-27        |
|bva             |  1720|2015-08-21         |2021-11-03        |
|okla            |  1046|2022-04-19         |2023-10-24        |
|texapp          |   506|2014-09-03         |2023-10-14        |
|oklacivapp      |   374|2022-04-14         |2023-10-19        |
|ncctapp         |   373|2021-04-21         |2022-04-19        |
|moctapp         |   371|2021-04-21         |2022-03-29        |
|arizctapp       |   332|2014-08-01         |2022-06-02        |
|nyappdiv        |   312|2018-12-26         |2023-09-20        |
|oklacrimapp     |   228|2022-04-14         |2023-09-21        |
|nc              |   216|2021-04-21         |2022-03-18        |
|del             |   162|2018-12-27         |2020-07-27        |
|ca7             |   157|2018-12-26         |2023-09-07        |
|pasuperct       |   138|2018-03-15         |2022-08-09        |
|nyappterm       |   111|2018-12-27         |2020-06-29        |
|ohioctapp       |    74|2018-12-26         |2023-06-13        |
[...]

PR #3814, opened 2024-02-20
fix(scrapers.tasks.extract_doc_content): replace warning for logger.error

grossir:
Will help solve #3811

This will help us actually see opinions where the text was not extracted,
 instead of just ignoring them. Even more, the previous code was
 `logging.warning`, when it should have been `logger.warning`


flooie:
looks good to me

Issue #3823, opened 2024-02-26
Update Backscraper commands


flooie:
Update CL Back Scraper 

We recently updated juriscraper to allow for dynamic backscraping instead of hard coded dates or parameters.  

We need to update the command for backscraping to enable this new feature.

PR #3829, opened 2024-02-27
feat(cl_scrape_opinions): support backscraper kwargs

grossir:
Solves #3823

Added `backscrape_start` and `backscrape_end` optional command line arguments, and passed them to cl_back_scrape_opinions


cl/scrapers/management/commands/cl_scrape_opinions.py
```diff
@@ -230,6 +230,18 @@ def add_arguments(self, parser):
             default=False,
             help="Disable duplicate aborting.",
         )
+        parser.add_argument(
```
quevon24, line 233: I think this arguments should be added in the backscraper command to keep the scraper separate from the backscraper. With something like:

```
        super().add_arguments(parser)
        parser.add_argument(
            "--backscrape-start",
            dest="backscrape_start",
            help="Starting value for backscraper iterable creation. "
                 "Each scraper handles the parsing of the argument,"
                 "since the value may represent a year, a string, a date, etc.",
        )
        parser.add_argument(
            "--backscrape-end",
            dest="backscrape_end",
            help="End value for backscraper iterable creation.",
        )
```

cl/scrapers/management/commands/cl_scrape_opinions.py
```diff
@@ -346,7 +358,7 @@ def scrape_court(self, site, full_crawl=False, ocr_available=True):
             # Only update the hash if no errors occurred.
             dup_checker.update_site_hash(site.hash)
 
-    def parse_and_scrape_site(self, mod, full_crawl):
+    def parse_and_scrape_site(self, mod, full_crawl, _, __):
```
quevon24, line 361: Here pass the options instead of adding the params from the backscraper

```
    def parse_and_scrape_site(self, mod, options):
        site = mod.Site().parse()
        self.scrape_court(site, options["full_crawl"])
```

cl/scrapers/management/commands/cl_scrape_opinions.py
```diff
@@ -384,7 +396,12 @@ def handle(self, *args, **options):
                 i += 1
                 continue
             try:
-                self.parse_and_scrape_site(mod, options["full_crawl"])
```
quevon24, line 387: self.parse_and_scrape_site(mod, options)

cl/scrapers/management/commands/cl_back_scrape_opinions.py
```diff
@@ -10,16 +12,28 @@ def parse_and_scrape_site(
         self,
         mod: AbstractSite,
         full_crawl: bool,
+        backscrape_start: Optional[str] = None,
```
quevon24, line 15: Here remove these params and use the options from cl_scrape_opinions.py as param

```
def parse_and_scrape_site(
        self,
        mod: AbstractSite,
        options: dict,
    ) -> None:
```

cl/scrapers/management/commands/cl_back_scrape_opinions.py
```diff
@@ -10,16 +12,28 @@ def parse_and_scrape_site(
...
+                range to be backscraped
+        :param backscrape_end: end value for backscraper range
         :return: None
         """
         court_str = mod.__name__.split(".")[-1].split("_")[0]
         logger.info(f'Using court_str: "{court_str}"')
-        for site in site_yielder(mod.Site().back_scrape_iterable, mod):
+        for site in site_yielder(
+            mod.Site(
+                backscrape_start=backscrape_start,
```
quevon24, line 32:
```
mod.Site(
                backscrape_start=options["backscrape_start"],
                backscrape_end=options["backscrape_end"],
            ).back_scrape_iterable,
            mod,
```

quevon24 (changes requested):
I think we should keep the scraper separate from the backscraper overwriting what is necessary in the backscraper code

grossir:
Thanks for the review @quevon24 , I agree with you. I made the changes and tested it again. Please give it another look

quevon24 (approved):
all good, let's merge it

PR #3837, opened 2024-02-29
feat(corpus importer): Update method to check sealed entries

### LINKED ISSUE #3836: TypeError: 'NoneType' checking for sealing entries when pacer_doc_id is None
The problem here is that we should avoid checking if the entry is sealed in minute entries or in any other case where the `pacer_doc_id` is None.

Sentry Issue: [COURTLISTENER-6SC](https://freelawproject.sentry.io/issues/5022514445/?referrer=github_integration)

```
TypeError: 'NoneType' object is not subscriptable
(2 additional frame(s) were not displayed)
...
  File "cl/recap/tasks.py", line 2279, in process_recap_email
    is_docket_entry_sealed(epq.court_id, pacer_case_id, pacer_doc_id)
  File "cl/corpus_importer/tasks.py", line 1780, in is_docket_entry_sealed
    return report.is_entry_sealed(case_id, doc_id)
```

Filed by: @albertisfu

ERosendo:
This PR updates the `is_docket_entry_sealed` method to gracefully handle situations where the doc_id argument is None.

Fixes #3836

### DIFF
--- a/cl/corpus_importer/tasks.py
+++ b/cl/corpus_importer/tasks.py
@@ -1759,7 +1759,9 @@ def is_pacer_doc_sealed(court_id: str, pacer_doc_id: str) -> bool:
     return False
 
 
-def is_docket_entry_sealed(court_id: str, case_id: str, doc_id: str) -> bool:
+def is_docket_entry_sealed(
+    court_id: str, case_id: str, doc_id: str | None
+) -> bool:
     """Check if a docket entry is sealed, querying the download confirmation
     page in PACER. If a receipt is returned the docket entry is not sealed,
     otherwise is sealed.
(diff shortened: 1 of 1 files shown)


albertisfu (approved):
Great, this seems good! 

With this change, we assume that no minute entry can be sealed, since it is not possible to query their receipt page.

mlissner:
Hm, but I'm afraid minute entries _can_ be sealed. I think to deal with that we may only have a few options:

1. Not putting minute entries into the docket, leaving it super incomplete.

2. Checking every minute entry by attempting to buy it on PACER, costing lots of money.

4. Put every minute entry up without checking, which has been fine so far.

4. Add some kind of permissions layer to every minute entry we receive from recap.email, so that they can only be seen by their uploader until we receive the same content from a public source.

5. Don't post minute entries onto dockets when received by recap.email, but send them out via to select users via email alerts and webhooks. The users would only be ones where the recap.email address in their account matches the recipient of the PACER email with the minute entry.

Other ideas? This is a frustrating development.

Shoot.

PR #3846, opened 2024-03-01
fix(search): Updates homepage.html to show number of opinions

ERosendo:
This PR fixes the issue with displaying the count of precedential opinions on the homepage.html template.


albertisfu (approved):
LGTM!

PR #3859, opened 2024-03-05
Update disclousre apis

flooie:
This PR adds better filtering mechanisms for the remainder of the disclosure API endpoints.

Previously you could really only do stuff on 

Agreements, and Investments.  

Now we should be able to filter gifts and debts, positions and agreements and more.  

Some tests and factories were added to prove things are working nicely


mlissner (approved):
Nice. Setting for auto-merge. Appreciate the John Oliver. It'll give me an excuse to say hi to them again.

PR #3864, opened 2024-03-06
feat(cl_scrape_opinions): support nested objects scraper

grossir:
Related to freelawproject/juriscraper#883

Main difference is that the nested objects supports OpinionClusters, which may have multiple Opinions. Current data flow is based on a single opinion per item

- abstracted a few functions to share code with new nested object approach
- changed `update_or_create_docket` so it accepts most of docket attributes as arguments
- added support to create new objects, such as OriginatingCourtInformation


flooie:
What do we need to do to move this forward?

grossir:
Part of this was implemented on the clustersite integration

PR #3884, opened 2024-03-14
feat(citations): Refactor code related to citations

ERosendo:
This pull request addresses several refactoring improvements and bug fixes within the citation views. Here's a summary of the key changes:

- Cleans up some unused code in the `citation_redirector` view. This PR removes an extra reverse call that is not being used anywhere.

- During the asynchronous migration of the `attempt_reporter_variation` view, we missed that one of the potential responses was using the `render` method, which is incompatible with asynchronous operations. This PR addresses this issue by replacing the `render` method with the `TemplateResponse` constructor, ensuring proper functionality within the asynchronous context.

- Similar to the previous point, during the migration of the `reporter_or_volume_handler` view, we missed that tthe code responsible for handling pagination was not working properly. The current implementation of the citation tool fails to gracefully handle two scenarios: 1) a valid page is given but it has no objects, and 2) a non-integer page value is provided. This PR addresses the issue by wrapping all pagination logic with the sync_to_async decorator and introducing a dedicated test to prevent future regressions.

- Introduces a new reusable selector function to retrieve Opinion clusters using citation strings. This avoids code duplication and promotes maintainability across the new API endpoint and potentially other parts of the codebase. Additionally, the query is optimized by using the `prefect_selected` method to efficiently get court data within the same operation, enhancing overall performance.

### DIFF
--- a/cl/citations/utils.py
+++ b/cl/citations/utils.py
@@ -4,8 +4,11 @@ from django.apps import (  # Must use apps.get_model() to avoid circular import 
     apps,
 )
 from django.db.models import Sum
+from django.template.defaultfilters import slugify
+from django.utils.safestring import SafeString
 from eyecite.models import FullCaseCitation
 from eyecite.utils import strip_punct
+from reporters_db import VARIATIONS_ONLY
 
 QUERY_LENGTH = 10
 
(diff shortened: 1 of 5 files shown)


mlissner (approved):
Looks clean!

PR #3902, opened 2024-03-19
feat(citations): Adds an API endpoint to look up citations

### LINKED ISSUE #3337: Make and document a proper citation look up API
Now that we've put all kinds of bot control features in place, our old lookup tool at `/c/` doesn't work as well as it used to. 

We should make a proper API for this that works basically the same way. You send it some text, it:

1. Finds a citation, and:
  - tells you the cluster ID for it (status 200), OR
  - tells you it could be several clusters, which it presents as JSON (status 300).

OR

2. It fails to find the citation in our data set:
 - status 400 and an empty JSON object

OR

3. It fails to parse a citation:
 - status 400, and an error message

OR

4. The request is too long:
 - status 414, and an error message

Hopefully this could be a simple little project? The URL could live at /api/rest/v3/citation-lookup/, and I wouldn't imagine it using DRF.

ERosendo:
This PR introduces a new API endpoint for citation lookup. It allows searching by string or using a `volume-reporter-page` format. Additionally, it closes #3337.

Here's a summary of the key changes:

- Introduces a new endpoint to look up citations. Users will be able to access it via a GET request to `/api/rest/v3/citation-lookup/`. To use this endpoint, The user must include the citation data in the request's querystring. Here are the supported fields:

     - `text_citation`: Textual representation of the citation. 
     - `reporter`: reporter name.
     - `volume`: Volume number.
     - `citation_page`: Specific page number within the volume.

- The pagination of this new endpoint is handled by the MediumAdjustablePagination class.

- The PR adds a new serializer to validate request parameters for citation lookups. The serializer enforces the following rules: 
    - **Mutually exclusive fields**: Users must provide either a text citation or a reporter name, but not both. This ensures clarity in the search intent. 
    - **Volume is required for reporter lookups**: Unlike the web citation tool, this API endpoint only focuses on retrieving clusters using the citation data provided by users. To achieve this, including the volume parameter is mandatory when using the `reporter` name. This combination (reporter and volume) directly targets the relevant set of opinion clusters. This might differ from the web tool where users can search by reporter only and get available volumes.

- Reuses the `OpinionClusterSerializer` class to provide richer data than just opinion IDs. Additionally, this allows users to dynamically select a specific subset of fields when querying the endpoint.

- The existing test cases for the citation_redirector view have been adapted to verify the new API endpoint replicates the view's behavior.

### DIFF
--- a/cl/api/urls.py
+++ b/cl/api/urls.py
@@ -144,6 +144,13 @@ router.register(
     basename="membership-webhooks",
 )
 
+# Citation lookups
+router.register(
+    r"citation-lookup",
+    citations_views.CitationLookupViewSet,
+    basename="citation-lookup",
+)
+
 API_TITLE = "CourtListener Legal Data API"
 
 
@@ -153,11 +160,6 @@ urlpatterns = [
         include("rest_framework.urls", namespace="rest_framework"),
     ),
     re_path(r"^api/rest/(?P<version>[v3]+)/", include(router.urls)),
-    re_path(
-        r"^api/rest/(?P<version>[v3]+)/citation-lookup/",
-        citations_views.CitationLookupView.as_view(),
-        name="citation_lookup",
-    ),
     # Documentation
     path("help/api/", views.api_index, name="api_index"),
     path("help/api/jurisdictions/", views.court_index, name="court_index"),
--- a/cl/api/urls.py
+++ b/cl/api/urls.py
@@ -5,6 +5,7 @@ from rest_framework.routers import DefaultRouter
 from cl.alerts import api_views as alert_views
 from cl.api import views
 from cl.audio import api_views as audio_views
+from cl.citations import api_views as citations_views
 from cl.disclosures import api_views as disclosure_views
 from cl.donate import api_views as donate_views
 from cl.favorites import api_views as favorite_views
@@ -152,6 +153,11 @@ urlpatterns = [
         include("rest_framework.urls", namespace="rest_framework"),
     ),
     re_path(r"^api/rest/(?P<version>[v3]+)/", include(router.urls)),
+    re_path(
+        r"^api/rest/(?P<version>[v3]+)/citation-lookup/",
+        citations_views.CitationLookupView.as_view(),
+        name="citation_lookup",
+    ),
     # Documentation
     path("help/api/", views.api_index, name="api_index"),
     path("help/api/jurisdictions/", views.court_index, name="court_index"),
--- a/cl/citations/api_serializers.py
+++ b/cl/citations/api_serializers.py
@@ -5,34 +5,34 @@ from cl.search.api_serializers import OpinionClusterSerializer
 
 
 class CitationAPIRequestSerializer(serializers.Serializer):
-    text_citation = serializers.CharField(required=False)
+    text = serializers.CharField(required=False)
     reporter = serializers.CharField(max_length=100, required=False)
     volume = serializers.IntegerField(required=False)
     page = serializers.CharField(required=False)
 
     def validate(self, data):
         reporter = data.get("reporter")
-        text_citation = data.get("text_citation")
+        text = data.get("text")
         volume = data.get("volume")
         page = data.get("page")
-        citation_or_reporter_provided = any([reporter, text_citation])
+        citation_or_reporter_provided = any([reporter, text])
 
         # make sure users provide either a reporter or a text citation.
         if not all([len(data), citation_or_reporter_provided]):
             raise ValidationError(
                 {
                     "non_field_errors": [
-                        "Either 'text_citation' or 'reporter' is required."
+                        "Either 'text' or 'reporter' is required."
                     ]
                 }
             )
 
-        # Enforces mutually exclusive fields: reporter and text_citation.
-        if all([reporter, text_citation]):
+        # Enforces mutually exclusive fields: reporter and text.
+        if all([reporter, text]):
             raise ValidationError(
                 {
                     "non_field_errors": [
-                        "Invalid request. Provide either a 'text citation' or 'reporter', not both."
+                        "Invalid request. Provide either a 'text' or 'reporter', not both."
                     ]
                 }
             )
--- a/cl/citations/api_serializers.py
+++ b/cl/citations/api_serializers.py
@@ -2,7 +2,7 @@ from rest_framework import serializers
 from rest_framework.serializers import ValidationError
 
 
-class CitationRequestSerializer(serializers.Serializer):
+class CitationAPIRequestSerializer(serializers.Serializer):
     text_citation = serializers.CharField(required=False)
     reporter = serializers.CharField(max_length=100, required=False)
     volume = serializers.IntegerField(required=False)
--- a/cl/citations/api_serializers.py
+++ b/cl/citations/api_serializers.py
@@ -1,0 +1,42 @@
+from rest_framework import serializers
+from rest_framework.serializers import ValidationError
+
+
+class CitationRequestSerializer(serializers.Serializer):
+    text_citation = serializers.CharField(required=False)
+    reporter = serializers.CharField(max_length=100, required=False)
+    volume = serializers.IntegerField(required=False)
+    citation_page = serializers.CharField(required=False)
+
+    def validate(self, data):
+        reporter = data.get("reporter")
+        text_citation = data.get("text_citation")
+        volume = data.get("volume")
+        citation_or_reporter_provided = any([reporter, text_citation])
+
+        # make sure users provide either a reporter or a text citation.
+        if not all([len(data), citation_or_reporter_provided]):
+            raise ValidationError(
+                {
+                    "non_field_errors": [
+                        "You must either provide a 'text citation' or 'reporter'."
+                    ]
+                }
+            )
+
+        # Enforces mutually exclusive fields: reporter and text_citation.
+        if all([reporter, text_citation]):
+            raise ValidationError(
+                {
+                    "non_field_errors": [
+                        "Invalid request. Provide either a 'text citation' or 'reporter', not both."
+                    ]
+                }
+            )
+
+        # Make sure users provide a volume when they use the reporter
+        # field to lookup citations.
+        if reporter and not volume:
+            raise ValidationError({"volume": ["This field is required."]})
+
+        return data
--- a/cl/citations/api_serializers.py
+++ b/cl/citations/api_serializers.py
@@ -12,6 +12,7 @@ class CitationRequestSerializer(serializers.Serializer):
         reporter = data.get("reporter")
         text_citation = data.get("text_citation")
         volume = data.get("volume")
+        page = data.get("citation_page")
         citation_or_reporter_provided = any([reporter, text_citation])
 
         # make sure users provide either a reporter or a text citation.
@@ -34,9 +35,14 @@ class CitationRequestSerializer(serializers.Serializer):
                 }
             )
 
-        # Make sure users provide a volume when they use the reporter
-        # field to lookup citations.
-        if reporter and not volume:
-            raise ValidationError({"volume": ["This field is required."]})
+        # Make sure users provide a volume and a page when they use the
+        # reporter field to look up citations.
+        if reporter and not all([volume, page]):
+            errors = {}
+            if not volume:
+                errors["volume"] = ["This field is required."]
+            if not page:
+                errors["citation_page"] = ["This field is required."]
+            raise ValidationError(errors)
 
         return data
--- a/cl/citations/api_serializers.py
+++ b/cl/citations/api_serializers.py
@@ -19,7 +19,7 @@ class CitationRequestSerializer(serializers.Serializer):
             raise ValidationError(
                 {
                     "non_field_errors": [
-                        "You must either provide a 'text citation' or 'reporter'."
+                        "Either 'text_citation' or 'reporter' is required."
                     ]
                 }
             )
--- a/cl/citations/api_serializers.py
+++ b/cl/citations/api_serializers.py
@@ -6,13 +6,13 @@ class CitationRequestSerializer(serializers.Serializer):
     text_citation = serializers.CharField(required=False)
     reporter = serializers.CharField(max_length=100, required=False)
     volume = serializers.IntegerField(required=False)
-    citation_page = serializers.CharField(required=False)
+    page = serializers.CharField(required=False)
 
     def validate(self, data):
         reporter = data.get("reporter")
         text_citation = data.get("text_citation")
         volume = data.get("volume")
-        page = data.get("citation_page")
+        page = data.get("page")
         citation_or_reporter_provided = any([reporter, text_citation])
 
         # make sure users provide either a reporter or a text citation.
@@ -42,7 +42,7 @@ class CitationRequestSerializer(serializers.Serializer):
             if not volume:
                 errors["volume"] = ["This field is required."]
             if not page:
-                errors["citation_page"] = ["This field is required."]
+                errors["page"] = ["This field is required."]
             raise ValidationError(errors)
 
         return data
--- a/cl/citations/api_serializers.py
+++ b/cl/citations/api_serializers.py
@@ -1,6 +1,8 @@
 from rest_framework import serializers
 from rest_framework.serializers import ValidationError
 
+from cl.search.api_serializers import OpinionClusterSerializer
+
 
 class CitationAPIRequestSerializer(serializers.Serializer):
     text_citation = serializers.CharField(required=False)
@@ -46,3 +48,22 @@ class CitationAPIRequestSerializer(serializers.Serializer):
             raise ValidationError(errors)
 
         return data
+
+
+class CitationAPIResponseSerializer(serializers.Serializer):
+    citation = serializers.CharField()
+    start_index = serializers.IntegerField()
+    end_index = serializers.IntegerField()
+    status = serializers.IntegerField()
+    error_message = serializers.CharField(required=False)
+    clusters = serializers.SerializerMethodField(required=False)
+
+    def get_clusters(self, obj):
+        if "clusters" not in obj:
+            return []
+
+        return OpinionClusterSerializer(
+            obj["clusters"],
+            many=True,
+            context={"request": self.context["request"]},
+        ).data
--- a/cl/citations/api_serializers.py
+++ b/cl/citations/api_serializers.py
@@ -52,6 +52,7 @@ class CitationAPIRequestSerializer(serializers.Serializer):
 
 class CitationAPIResponseSerializer(serializers.Serializer):
     citation = serializers.CharField()
+    normalized_citations = serializers.ListField(child=serializers.CharField())
     start_index = serializers.IntegerField()
     end_index = serializers.IntegerField()
     status = serializers.IntegerField()
--- a/cl/citations/api_views.py
+++ b/cl/citations/api_views.py
@@ -5,6 +5,7 @@ from asgiref.sync import async_to_sync
 from django.db.models import QuerySet
 from django.http import HttpResponse
 from django.template.defaultfilters import slugify
+from django.utils.safestring import SafeString
 from rest_framework.exceptions import NotFound, ValidationError
 from rest_framework.mixins import CreateModelMixin
 from rest_framework.request import Request
@@ -16,7 +17,7 @@ from cl.citations.utils import SLUGIFIED_EDITIONS, get_canonicals_from_reporter
 from cl.search.api_serializers import OpinionClusterSerializer
 from cl.search.models import OpinionCluster
 from cl.search.selectors import get_clusters_from_citation_str
-from django.utils.safestring import SafeString
+
 
 class CitationLookupViewSet(CreateModelMixin, GenericViewSet):
     queryset = OpinionCluster.objects.all()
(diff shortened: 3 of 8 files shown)


cl/api/urls.py
```diff
@@ -152,6 +153,11 @@
         include("rest_framework.urls", namespace="rest_framework"),
     ),
     re_path(r"^api/rest/(?P<version>[v3]+)/", include(router.urls)),
+    re_path(
+        r"^api/rest/(?P<version>[v3]+)/citation-lookup/",
+        citations_views.CitationLookupView.as_view(),
+        name="citation_lookup",
+    ),
```
mlissner, line 160: Did you try registering this endpoint, like the examples above?

cl/citations/api_serializers.py
```diff
@@ -0,0 +1,42 @@
...
+        text_citation = data.get("text_citation")
+        volume = data.get("volume")
+        citation_or_reporter_provided = any([reporter, text_citation])
+
+        # make sure users provide either a reporter or a text citation.
+        if not all([len(data), citation_or_reporter_provided]):
+            raise ValidationError(
+                {
+                    "non_field_errors": [
+                        "You must either provide a 'text citation' or 'reporter'."
```
mlissner, line 22:
```suggestion
                        "Either 'text_citation' or 'reporter' is required."
```

cl/citations/api_serializers.py
```diff
@@ -0,0 +1,42 @@
...
+                    "non_field_errors": [
+                        "Invalid request. Provide either a 'text citation' or 'reporter', not both."
+                    ]
+                }
+            )
+
+        # Make sure users provide a volume when they use the reporter
+        # field to lookup citations.
+        if reporter and not volume:
+            raise ValidationError({"volume": ["This field is required."]})
```
mlissner, line 40: I hate to say it, but I don't think there's a use case for getting all of the cases in a reporter-volume dyad. I think for this endpoint it's better to require either text or page/volume/reporter, but not reporter/volume.

cl/citations/tests.py
```diff
@@ -17,6 +19,12 @@
 )
 from factory import RelatedFactory
 from lxml import etree
+from rest_framework.status import (
+    HTTP_200_OK,
+    HTTP_300_MULTIPLE_CHOICES,
+    HTTP_400_BAD_REQUEST,
+    HTTP_404_NOT_FOUND,
+)
```
mlissner, line 27: I actually prefer pulling these from the `http` library, since they're there. If you want to include that refactor across the entire project in this PR, I won't stop you.

cl/citations/api_views.py
```diff
@@ -0,0 +1,209 @@
...
+from cl.citations.api_serializers import CitationRequestSerializer
+from cl.citations.exceptions import MultipleChoices
+from cl.citations.utils import get_canonicals_from_reporter
+from cl.search.api_serializers import OpinionClusterSerializer
+from cl.search.models import OpinionCluster
+from cl.search.selectors import get_clusters_from_citation_str
+
+SLUGIFIED_EDITIONS: dict[str, str] = {
+    str(slugify(item)): item for item in EDITIONS.keys()
+}
```
mlissner, line 24: If you're going to put this here, you should replace it at `opinion_pages.views:1146` too. DRY.

cl/citations/api_views.py
```diff
@@ -0,0 +1,209 @@
...
+        self.full_text_citation = data.get("text_citation", None)
+        self.reporter = data.get("reporter", None)
+        self.volume = data.get("volume", None)
+        self.page = data.get("citation_page", None)
+
+        if self.full_text_citation:
+            citations = eyecite.get_citations(self.full_text_citation)
+            if not citations:
+                raise NotFound(
+                    f"Unable to find a citation withing the provided text"
```
mlissner, line 50:
```suggestion
                    f"No citations found in 'text_citation'"
```

cl/citations/api_views.py
```diff
@@ -0,0 +1,209 @@
...
+                )
+
+            if citations[0].groups:
+                c = citations[0]
+                self.reporter = slugify(c.groups["reporter"])
+                self.volume = c.groups["volume"]
+                self.page = c.groups["page"]
+            else:
+                raise NotFound(
+                    f"The provided text is not a valid citation. Please review it and try again"
```
mlissner, line 60: Should this be a different status code for invalid input, say?

Also:
```suggestion
                    f"Invalid citation."
```

cl/citations/api_views.py
```diff
@@ -0,0 +1,209 @@
...
+                )
+
+        self.reporter_slug = slugify(self.reporter)
```
mlissner, line 63: Didn't we slugify this value on line 55 already (`self.reporter = slugify(c.groups["reporter"])`)?

cl/citations/api_views.py
```diff
@@ -0,0 +1,209 @@
...
+
+        # Look up the reporter to get its proper version (so-2d -> So. 2d)
+        proper_reporter = SLUGIFIED_EDITIONS.get(self.reporter_slug, None)
+        if not proper_reporter:
+            proper_reporter = self._attempt_reporter_variation()
+
+        if proper_reporter and self.volume and self.page:
+            return self._citation_handler(request, proper_reporter)
+        else:
+            return self._reporter_volume_handler(request, proper_reporter)
```
mlissner, line 73: As mentioned above, I think we can remove this branch of the code and the `self._reporter_volume_handler` method.

cl/citations/api_views.py
```diff
@@ -0,0 +1,209 @@
...
+
+        Returns:
+            str: canonical name for the reporter.
+        """
+        potential_canonicals = get_canonicals_from_reporter(self.reporter_slug)
+
+        if len(potential_canonicals) == 0:
+            # Couldn't find it as a variation. Give up.
+            raise NotFound(
+                f"Unable to find Reporter with abbreviation of '{self.reporter}'"
```
mlissner, line 91:
```suggestion
                f"Unable to find reporter with abbreviation of '{self.reporter}'"
```

mlissner: Also, will this value be slugified? Do we need to store the raw value so we can return it properly in this error message?

cl/citations/api_views.py
```diff
@@ -0,0 +1,209 @@
...
+            )
+
+        elif len(potential_canonicals) > 1:
+            # The reporter variation is ambiguous b/c it can refer to more than
+            # one reporter. Abort with a 300 status.
+            raise MultipleChoices(
+                f"Found more than one possible reporter for '{self.reporter}'"
+            )
```
mlissner, line 99: Since we're getting rid of support for reporter/volume dyads, I think we can return serialized results for this, but before we do, we need to talk about what the shape of the response should be, in particular when we make it possible to find more than one citation in a block of text. 

I _think_ we need to get to a place where a response is more like this (very rough idea!):

```
POST: 
    text="In 22 U.S. 44 J. Kagan referenced 334 ASDF 1 and 1 P. 4"
RESPONSE:
[
   {
      "citation": "22 U.S. 44",
      "status": 300,
      "clusters: {...2 clusters...}
    },
    {
      "citation": "334 ASDF 1",
      "status": 404,
      "clusters": {},
    },
    {
      "citation": "1 P. 4",
      "status": 200,
      "clusters": {...1 cluster object...},
    }
]
```

What do you think of that? And then the HTTP status for the API call itself could indicate what it usually does (validation errors, bad URL, etc.)?

cl/citations/api_views.py
```diff
@@ -0,0 +1,209 @@
...
+            HttpResponse: An HTTP response object containing a list of matching
+                opinion clusters.
+        """
+        citation_str = " ".join([str(self.volume), reporter, self.page])
+        clusters, cluster_count = async_to_sync(
+            get_clusters_from_citation_str
+        )(reporter, str(self.volume), self.page)
+
+        if cluster_count == 0:
+            raise NotFound(f"Unable to Find Citation '{ citation_str }'")
```
mlissner, line 129:
```suggestion
            raise NotFound(f"Citation not found: '{ citation_str }'")
```

cl/citations/api_views.py
```diff
@@ -0,0 +1,209 @@
...
+        return self._show_paginated_response(request, cases_in_volume)
+
+    def _show_paginated_response(
+        self, request: Request, clusters: QuerySet[OpinionCluster]
+    ) -> HttpResponse:
+        """
+        Generates a paginated HTTP response containing opinion clusters.
+
+        This method takes a queryset of opinion clusters and an HTTP request
+        object.It then paginates the queryset based on the request parameters
```
mlissner, line 177:
```suggestion
        object. It then paginates the queryset based on the request parameters
```

cl/citations/tests.py
```diff
@@ -1727,3 +1735,222 @@ def test_get_graph_component(self):
                     sorted(output),
                     f"Got incorrect result from get_graph_component for inputs (expected {output}): {inputs}",
                 )
+
+
+class CitationLookUpApiTest(
```
mlissner, line 1740: I think I'm calling for some pretty major changes, so I haven't reviewed the tests yet, but if a review would help at this stage, let me know and I'm happy to find a minute.

mlissner (changes requested):
I guess I'm back to my old habits today, making lots of comments, sorry. Big picture, I made a comment about the general shape of the response, which I think needs to change now, rather than later. 

The other thing I think is missing is documentation. Do you want to put that in a separate PR or in here?

ERosendo:
> I guess I'm back to my old habits today, making lots of comments, sorry.

No problem. I like comments 😃. I'll start addressing them in a bit.

cl/api/urls.py
ERosendo: No, I didn't because the `CitationLookupView` inherits from `GenericAPIView`, which isn't compatible with the default router. However, updating the view to work with the router is a straightforward process.

cl/citations/api_serializers.py
ERosendo: got it! I will remove this code and the `_reporter_volume_handler` method.

cl/citations/api_views.py
ERosendo: > Should this be a different status code for invalid input

You're correct. We should use status code 400 becuase the request has malformed data

cl/citations/api_views.py
ERosendo: Thanks for the suggestion. I've updated the error message 👍

cl/citations/api_views.py
ERosendo: Thanks for catching this! We should avoid using slugify on line 55. The purpose of having separate `reporter` and `reporter_slug` variables is to store the raw reporter name in `reporter` so we can return it properly in error messages for better user clarity. I'm adding a commit to fix this 👍

cl/citations/api_views.py
ERosendo: > will this value be slugified? 

No. The error message will display the reporter name provided by the user.. 

> Do we need to store the raw value so we can return it properly in this error message?

Yes, I think it's a good idea to store the raw value. Most citation helpers use a slugified version of the reporter name for internal purposes, but this could be confusing in error messages. That's why we have separate variables for the reporter name and its slug.

cl/citations/api_views.py
ERosendo: > Since we're getting rid of support for reporter/volume dyads, I think we can return serialized results for this

do you mean returning all the clusters that match the reporter and its potential canonical? 

> I think we need to get to a place where a response is more like this

I like the idea of including the citation, its status, and any related clusters in the response. This approach offers flexibility for handling multiple lookups. However, I noticed that pagination seems to be missing,  wouldn't pagination be helpful for large result sets?

mlissner: Yes, if we get to the point where we have multiple matches for a citation, we should return them all.

I didn't think about pagination. I wonder if it's really needed. We're not providing the text of the opinions, so these responses should be pretty small. Maybe we can get away with not doing it? 

Another thing, would it be helpful to provide start and finish indexes for the citations that we find? Something like:

```
{
  "citation": "1 P. 22",
  "start_index": 32,
  "end_index": 39,
  ...
}
```

ERosendo: > Yes, if we get to the point where we have multiple matches for a citation, we should return them all.

got it! 

> Maybe we can get away with not doing it?

You're right about the data of the opinions. Let's move forward with a non-paginated implementation.

> would it be helpful to provide start and finish indexes for the citations that we find?

I like this idea! I'll start working on this tweak.

ERosendo:
@mlissner I've applied all your suggestions. Here's a summary of the latest changes:

- Removes pagination from citation lookup API and renames the `citation_page` field. Since pagination won't be implemented for this endpoint, it's no longer necessary to differentiate between a `page` within a volume and a `page` of the result set. Therefore, the `citation_page` field has been renamed to `page` for clarity.

- Updates the `CitationLookup` view parent classes so we can register the endpoint using the Default router(like the other API endpoints) and handle POST requests instead of GET requests. This change allows users to submit more complex data through the request body, enhancing the data submission process.

- Removes the `_reporter_volume_handler` method and enforces the `page` field as required in the serializer. This means this new endpoint will accept either text or page/volume/reporter. Reporter and volume combinations alone are no longer supported.

- Introduces a new serializer class named `CitationAPIResponseSerializer` to format API responses and use the schema from https://github.com/freelawproject/courtlistener/pull/3902#discussion_r1531271170

- Refactors the `CitationLookupViewSet` class to handle multiple citations in the `text_citation` field.

- Refines error messages for empty request payloads and improves code organization by moving the `slugified_editions` constant to `utils.py`

- Updates test files to use the `http` library to check HTTP status codes and the new response format.

cl/citations/tests.py
```diff
@@ -1727,3 +1731,316 @@ def test_get_graph_component(self):
...
+        clusters = first_citation["clusters"]
+        self.assertEqual(len(clusters), 2)
+        for cluster in clusters:
+            self.assertIn(
+                cluster["absolute_url"],
+                [
+                    handy_citation.cluster.get_absolute_url(),
+                    haw_citation.cluster.get_absolute_url(),
+                ],
+            )
```
mlissner, line 1866: I'm imagining how this looks as I go, but I think this is exactly right.

cl/citations/tests.py
```diff
@@ -1727,3 +1731,316 @@ def test_get_graph_component(self):
...
+        r = await self.async_client.post(
+            reverse("citation-lookup-list", kwargs={"version": "v3"}),
+            {"reporter": "F2d", "volume": "56", "page": "9"},
+        )
+        self.assertEqual(r.status_code, HTTPStatus.OK)
+        data = json.loads(r.content)
+        self.assertEqual(len(data), 1)
+
+        first_citation = data[0]
+        self.assertEqual(first_citation["citation"], "56 F2d 9")
```
mlissner, line 1942: I'm wondering if our users would find it easier to debug if we returned our normalized citations in an additional key. Like, they sent F2d, so the `citation` key is `56 F2d 9`, but we fixed their input and got `56 F.2d 9`, which we put in `normalized_citation`? 

Beyond debugging, I bet people would appreciate this as a way of fixing slopping RAGs, so they give cleaner citations.

Easy to do?

cl/citations/tests.py
```diff
@@ -1727,3 +1731,316 @@ def test_get_graph_component(self):
...
+
+    async def test_can_handle_full_citation_within_text(self) -> None:
+        """Do we get redirected to the correct URL when we pass in a full
+        citation?"""
+        text_citation = (
+            "Reference to Lissner v. Saad, 56 F.2d 9 11 (1st Cir. 2015)"
+        )
+        r = await self.async_client.post(
+            reverse("citation-lookup-list", kwargs={"version": "v3"}),
+            {"text_citation": text_citation},
```
mlissner, line 1980: I'm sorry, but I don't think I like the POST key to be `text_citation`. A couple reasons:

1. It's wordy. Seems like something simpler is possible.
2. We can have more than one citation, but it implies it's singular. 

Could we just have the key be `text` instead?

mlissner (changes requested):
Some more comments for you, Eduardo. I think it's good, but I think we can make it a *bit* better in a few spots, hopefully without too much more trouble. 

I also am pretty sure I know what the JSON output and HTTP status code would be after you've made these changes (assuming you agree with them), but can you please share a response (JSON and HTTP code) it would give for a text that has:

1. Multiple citations in it.
2. At least one citation that is ambiguous.
3. At least one citation that cannot be found.
4. At least one case with a reporter like F2d, which lacks the correct spacing, etc.
5. At least one citation that's simple and works nicely.

That'll give me one final sanity check, and I think we can probably also use that as a starter for the documentation.

Thank you!

cl/citations/tests.py
ERosendo: > Could we just have the key be text instead?

sure! I'll tweak the serializer.

cl/citations/tests.py
ERosendo: @mlissner I have a question regarding ambiguous citations. These citations typically have multiple potential reporters. Should the `normalized_citation` field display both potentially valid citations?

for example, the reporter `H.` is ambiguous and a request to the API endpoint might return clusters matching both `Handy` or `Haw`. Should the `normalized_citation` key include both possibilities separated by a comma?

The complete example looks like this:

```
POST: 
   {"reporter": "H.", "volume": "1", "page": "150"}
RESPONSE:
[
   {
      "citation": "1 H. 150",
      "status": 200,
      "normalized_citation": "1 Handy 150,  1 Haw 150"
      "clusters: {...2 clusters...}
    },
]
```

cl/citations/tests.py
```diff
@@ -1727,3 +1731,316 @@ def test_get_graph_component(self):
...
+            reverse("citation-lookup-list", kwargs={"version": "v3"}),
+            {"text_citation": "§ 97-29-63"},
+        )
+        self.assertEqual(r.status_code, HTTPStatus.OK)
+        data = json.loads(r.content)
+
+        first_citation = data[0]
+        self.assertEqual(first_citation["status"], HTTPStatus.BAD_REQUEST)
+        self.assertIn("Invalid citation", first_citation["error_message"])
+
```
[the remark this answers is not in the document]
ERosendo: > I think the easiest thing to do is to ignore non-case law citations.

Alright, I'll add code to check the citation type. This comment aligns with the [suggestion](https://github.com/freelawproject/courtlistener/pull/3902#discussion_r1538542038) to use a type check to refine the list of citations and include only opinion citation.

cl/citations/api_views.py
```diff
@@ -0,0 +1,235 @@
...
+        citation_serializer.is_valid(raise_exception=True)
+
+        # Get query parameters from the validated data
+        citations = []
+        data = citation_serializer.validated_data
+        full_text_citation = data.get("text_citation", None)
+        if full_text_citation:
+            citation_objs = eyecite.get_citations(full_text_citation)
+            if not citation_objs:
+                raise NotFound(f"No citations found in 'text_citation'.")
```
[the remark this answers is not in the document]
ERosendo: got it!

cl/citations/tests.py
mlissner: That makes sense to me, but I'd use a list and change the key to `normalized_citations`, plural. That'd work, right?

ERosendo: > I'd use a list and change the key to normalized_citations

got it! 

> That'd work, right?

Yes! we can do that.

ERosendo:
Here are some examples:

------------------------------------------------------------------------------------------

#### 1. Text including only non-caselaw citations:

`POST /citation-lookup/`

##### Body

```json
{
    "text": "Cal. Prob. Code § 141. John M. Balkin, Nested Opposition, 99 Yale L.J. 1669 (1990) (reviewing John M. Ellis, Against Deconstruction (1989)). 42 U.S.C. § 405(a)."
}
```
##### Response
```json
{ }
```

------------------------------------------------------------------------------------------

#### 2. Text with a few non-caselaw citations, an ambiguous citation, a citation that cannot be found, and one that works nicely.

`POST /citation-lookup/`

##### Body

```json
{"text": "Steven G. Calabresi & Kevin H. Rhodes, The Structural Constitution: Unitary Executive, Plural Judiciary, 105 Harv. L. Rev. 1155, 1158 (1992). 1 H. 150. U.S. Const. amend. XIII, § 2. Kootenai Env't All., Inc. v. Panhandle Yacht Club, Inc., 671 P.2d 1085 (Idaho 1983). Hudspeth v. State, 1 Ark. 20 (Ark. 1837)."}
```
##### Response

```json
[
    {
        "citation": "1 H. 150",
        "normalized_citations": [
            "1 Handy 150",
            "1 Haw. 150",
            "1 Hill 150"
        ],
        "start_index": 142,
        "end_index": 150,
        "status": 300,
        "clusters": [ ... 2 clusters ...]
     },
    {
        "citation": "671 P.2d 1085",
        "normalized_citations": [
            "671 P.2d 1085"
        ],
        "start_index": 239,
        "end_index": 252,
        "status": 404,
        "error_message": "Citation not found: '671 P.2d 1085'",
       "clusters": []
    },
    {
        "citation": "1 Ark. 20",
        "normalized_citations": [
            "1 Ark. 20"
        ],
        "start_index": 286,
        "end_index": 295,
        "status": 200,
         "clusters": [ ... 1 cluster ... ]
    }
]
```
------------------------------------------------------------------------------------------

#### 3. Request with a typo in the reporter field.

`POST /citation-lookup/`

##### Body

```json
{
    "reporter": "US",
    "volume": 1,
    "page": 2
}
```
##### Response
```json
[
    {
        "citation": "1 US 2",
        "normalized_citations": [
            "1 U.S. 2"
        ],
        "start_index": 0,
        "end_index": 7,
        "status": 404,
        "error_message": "Citation not found: '1 US 2'",
        "clusters": []
    }
]
```

------------------------------------------------------------------------------------------

#### 3. Request with an invalid reporter

`POST /citation-lookup/`

##### Body

```json
{
    "reporter": "test-reporter",
    "volume": 1,
    "page": 2
}
```

##### Response

```json
[
    {
        "citation": "1 test-reporter 2",
        "normalized_citations": [],
        "start_index": 0,
        "end_index": 17,
        "status": 404,
        "error_message": "Unable to find reporter with abbreviation of 'test-reporter'",
        "clusters": []
    }
]
```

mlissner:
Those examples look fantastic. 

A couple little thoughts...

In this one:

```
[
    {
        "citation": "1 US 2",
        "normalized_citations": [
            "1 U.S. 2"
        ],
        "start_index": 0,
        "end_index": 7,
        "status": 404,
        "error_message": "Citation not found: '1 US 2'",
        "clusters": []
    }
]
```
Am I counting wrong or shouldn't he end_index be number 5 (or maybe 6), but not 7?

In this one:

```
[
    {
        "citation": "1 test-reporter 2",
        "normalized_citations": [],
        "start_index": 0,
        "end_index": 17,
        "status": 404,
        "error_message": "Unable to find reporter with abbreviation of 'test-reporter'",
        "clusters": []
    }
]
```

404 means not found...maybe a 400 ("Bad Request") would be better here? 

And one last thought, I think: Did we do anything to prevent people from uploading WAAY too much content or did we decide we'd do that once we can check performance in prod?

mlissner (approved):
Looks great. Just a couple tweaks from my last comment. The rest looks solid to me!

ERosendo:
> Am I counting wrong or shouldn't he end_index be number 5 (or maybe 6), but not 7?

You're right! sorry! it looks like I mixed up the response with another request. I've just double-checked that the API returns 6 in the end_index field.

image


> 404 means not found...maybe a 400 ("Bad Request") would be better here?

I agree. I'll update the code. 

> Did we do anything to prevent people from uploading WAAY too much content or did we decide we'd do that once we can check performance in prod?

This PR doesn't currently include a mechanism to prevent requests with excessive content.  Should we limit the maximum number of characters in the `text` field using CitationAPIRequestSerializer class?

mlissner:
Cool. Let's go ahead and wait on the length check. We'll know what kind of trouble we've got once we launch this thing. :)

mlissner:
Auto-merge set! Looking forward to playing with this.

ERosendo:
🚀 Thanks @mlissner 👍

PR #3905, opened 2024-03-21
Use the Http library to handle HTTP status 

ERosendo:
This PR updates views, tests, and custom commands throughout the project to consistently use the  [HTTPStatus class](https://docs.python.org/3/library/http.html#http.HTTPStatus) from the standard [http python module](https://docs.python.org/3/library/http.html#module-http). Previously, these components relied on importing status codes from the `status` module included with [Django REST framework](https://www.django-rest-framework.org/api-guide/status-codes/).

### DIFF
--- a/cl/alerts/tests.py
+++ b/cl/alerts/tests.py
@@ -1,6 +1,7 @@
 import json
 from collections import defaultdict
 from datetime import datetime, timedelta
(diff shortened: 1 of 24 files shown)


mlissner (approved):
Holy cow that took more work than I expected. Good to have it done though.

PR #3914, opened 2024-03-26
fix(template): Updates the newsletter link

ERosendo:
This PR updates the newsletter link in the base HTML template and confirmation account email template to make sure it directs users to the correct location.


mlissner:
Nice one, thank you.

PR #3915, opened 2024-03-26
Management command to import citations from a csv file

quevon24:
As I mentioned in previous meetings, I consider that it is easier to match the citations with the datasets that we have from the local environment and only generate a csv to load the new citations in the clusters. This should be relatively fast and reuses existing code.

This command will be responsible for loading the citations that come in the csv with the following format (no header row, column 1: cluster id, column 2: citation to add):

```
"2155423", "2003 WL 22508842"
"7903720","520 A.2d 234"
"7903715","520 A.2d 233"
```

Here is a sample file to test the command:
[updated_sample.csv](https://github.com/freelawproject/courtlistener/files/15297717/updated_sample.csv)

We need to place the file in cl/assets/media/sample.csv, clone the clusters and then run the command.

**How to run the command using the csv file:**
```
docker exec -it cl-django python /opt/courtlistener/manage.py clone_from_cl --type search.OpinionCluster --id 1904175 7903720 7903715 7903719 7903933
docker exec -it cl-django python /opt/courtlistener/manage.py import_citations_csv --csv /opt/courtlistener/cl/assets/media/updated_sample.csv
```


quevon24:
@mlissner 

**Westlaw**

7206737 rows with parallel citation
1063191 rows without parallel citation

From the 7206737 rows:

**Citations to add to CL: 2,133,707** (matched using one citation of the citations, same filed date, same court and case name matched >= 90%)
Both citations exist: 1,643,083 rows
Can’t match using any of the citations: 981,403 rows
Possible matches: 2,448,544 rows (there is not enough precision in the match using the dataset data to ensure at least 90% similarity, this number includes incorrect matches)

Of the 2,133,707 citations to be added, 1,020,864 are from Westlaw and the remaining 1112,843 are from other reporters.

At the moment the 1063191 rows without parallel citation were ignored since only the case data would have to be used to try to find a match (case name, filed date, docket number and court) and later verify if the citation we have is in the system or not, in this case the precision has to be higher to ensure a correct match.

**Lexis**

In total there are 14,194,271 rows in the dataset, the number of citations in each row may vary.

288,117 rows doesn’t have a case name
167,463 rows doesn’t have citations
10,918,806 rows are valid (with more than one citation, we need at least one citation to try to match it to a cluster and add the rest if they are not already there)
 ~25M citations in the 10,918,806 rows, if we have at least two citations in each row and use one to match the cluster and find the match, that leaves us with approximately 12.5M citations available to add

4,467,182 rows matched using one citation, same filed date, same court id and case name matched >= 90%
**6,608,508 citations to add** to CL from ~4.4M rows
3,367,005 rows that we couldn’t match
3,084,619 rows with possible matches (not enough precision in the match or bad matches)

_Update 05/02/2024:_ ~440K new citations to add using filed date, court id and matched case name >= 90% (this process is more exhaustive because we can get more than 100 results for possible cases for a single row from the dataset using only the court id and filed date, we could get more citations but this will take some time, could also apply this for rows that only have one citation) 

**~7M new citations** in total from dataset

quevon24:
@grossir please could you take a look at this PR

quevon24:
@grossir i updated the code to use skiprows and nrows to set start and end row and set rows limit to process.

I removed the header in the csv and updated the sample file to test the command

cl/citations/management/commands/import_citations_csv.py
```diff
@@ -0,0 +1,128 @@
...
+        options["csv"],
+        names=["cluster_id", "citation_to_add"],
+        delimiter=",",
+        skiprows=start_row,
+        nrows=end_row,
+    )
+
+    # Replace nan in dataframe
+    data = data.replace(np.nan, "", regex=True)
+    logger.info(f"Found {len(data.index)} rows in csv file: {options['csv']}")
```
grossir, line 57: My editor complains about using f-strings on logger calls. Check https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/logging-fstring-interpolation.html

cl/citations/management/commands/import_citations_csv.py
```diff
@@ -0,0 +1,128 @@
...
+The csv file must have the following structure, the file should not have a header row:
+
+"2155423", "2003 WL 22508842"
+"7903720","520 A.2d 234"
+"7903715","520 A.2d 233"
+
+How to run the command:
+manage.py import_citations_csv --csv /opt/courtlistener/cl/assets/media/wl_citations_1.csv
+
+Note: If --limit is greater than --end-row, end row will be ignored
```
grossir, line 13: Seems like now limit and end-row are doing basically the same, so why not just keep one of them?

grossir (commented):
I tested this and it works fine, I think it is ready to merge

 I just left some minor comments about the arguments and about a pylint complaint

cl/citations/management/commands/import_citations_csv.py
quevon24: i think more than anything it is for the user's ease to differentiate what each argument does

grossir (approved):
Feel free to merge this PR!

mlissner:
Nice one! For the record, a zillion people are asking for this right now. Very nice milestone, Kevin!

quevon24:
> Member

About this, @blancoramiro is in charge of running it, right? so that I can send the files to him and we can execute it when you order it

mlissner:
Yeah. I want to get Bill's eyes on this too before we run it, since it could cause so much damage if there are any wrong assumptions, but do you have instructions for running it that Ramiro will need?

quevon24:
@grossir Here are two sample files so you can validate or refute that the matches are correct and that the citations we want to add are correct or not.

To avoid any bias I took 200 rows at random from the correct matches from each dataset valid results, the only additional information that each dataset contains is the source path, the cluster_id with which the data of the row and the citation to be added matched, all the other data come from the dataset.

The first sample: 
[dataset_1_sample.csv](https://github.com/user-attachments/files/15571293/dataset_1_sample.csv)
column cluster_id is the cluster that matched the data in the row, citations_to_add contains the citations that come from the row data and we need to add to the cluster_id

The second sample:
[dataset_2_sample.csv](https://github.com/user-attachments/files/15571298/dataset_2_sample.csv)
column cluster_id is the cluster that matched the data in the row, Parallel Cite column contains the citation that we need to add to the cluster

The first goal is to validate that the cluster_id does correspond to the column data (not all the column data will match because in some cases there is no data such as docker number, the court is of a higher or lower level than the one we have in CL, the names are not exactly the same, there are abbreviations, longer or shorter names, etc.)

The second goal is to verify that the citation we want to add is correct

As a suggestion you can directly use the CL API to obtain all the necessary information and use the necessary magic to match or unmatch the results

@mlissner about the instructions that ramiro requires, @blancoramiro you can save all the log messages generated from the execution of the command? These will be helpful to us because in case we need to delete something we just have to parse the log to get the ids of the citations that were created. Or i could modify the command so that it saves the log in a file and it would only be a matter of downloading it from the container.

In case you need a larger sample, I can generate it without problems.

blancoramiro:
> @mlissner about the instructions that ramiro requires, @blancoramiro you can save all the log messages generated from the execution of the command? These will be helpful to us because in case we need to delete something we just have to parse the log to get the ids of the citations that were created. Or i could modify the command so that it saves the log in a file and it would only be a matter of downloading it from the container.

Hey @quevon24! I can save the log with no problem. Normally what I do is capture the stdout/stderr and save them to a file locally in my computer. I prefer to store it locally just as an extra precaution in case the pod dies for any reason.

grossir:
@quevon24 I reviewed all the matches for both files, and they all look OK

flooie:
Thanks @grossir

cl/citations/management/commands/import_citations_csv.py
```diff
@@ -0,0 +1,128 @@
...
+
+        if not os.path.exists(options["csv"]):
+            print(f"Csv file: {options['csv']} doesn't exist.")
+            return
+
+        data = load_citations_file(options)
+        if not data.empty:
+            process_csv_data(data)
+        else:
+            print("CSV file empty")
```
flooie, line 128: I think this would be better with a logger

cl/citations/management/commands/import_citations_csv.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+
+    def handle(self, *args, **options):
+        if options["start_row"] and options["end_row"]:
+            if options["start_row"] > options["end_row"]:
+                print("--start-row can't be greater than --end-row")
+                return
+
+        if not os.path.exists(options["csv"]):
+            print(f"Csv file: {options['csv']} doesn't exist.")
```
flooie, line 121: same here

flooie: ie. a logger instead of a print please

cl/citations/management/commands/import_citations_csv.py
```diff
@@ -0,0 +1,128 @@
...
+    if options["limit"]:
+        end_row = options["limit"]
+
+    data = pd.read_csv(
+        options["csv"],
+        names=["cluster_id", "citation_to_add"],
+        delimiter=",",
+        skiprows=start_row,
+        nrows=end_row,
+    )
```
flooie, line 53: I think this start_row end_row assigning of variable doesnt make any sense to me.  Doesnt this mean that we are always putting None into the skip rows and nrows variable?

PR #3916, opened 2024-03-27
feat(citations): Handle Errors on the web citation tool

### LINKED ISSUE #2747: Catch and ignore errors on the Citation Lookup tool
This is a new issue since the work to the citation redirector that landed yesterday. It's a crasher that we caught in Sentry. 

@troglodite2, do you have a moment to take a look? So far, this has happened only once. Let me know if you need more info from Sentry and I can share details. Unfortunately, I can't share the Sentry issue due to private variables though (like user info).

------

Sentry Issue: [COURTLISTENER-3TH](https://freelawproject.sentry.io/issues/4168216754/?referrer=github_integration)

```
NoReverseMatch: Reverse for 'citation_redirector' with keyword arguments '{'reporter': 'ARB/11/20'}' not found. 3 pattern(s) tried: ['c/(?P<reporter>[^/]+)/\\Z', 'c/(?P<reporter>.*)/(?P<volume>\\d{1,10})/$', 'c/(?:(?P<reporter>.*)/(?P<volume>\\d{1,10})/(?P<page>.*)/)?$']
  File "django/core/handlers/exception.py", line 55, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 197, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "cl/opinion_page/views.py", line 1120, in citation_homepage
    reverse("citation_redirector", kwargs=kwargs)
  File "django/urls/base.py", line 88, in reverse
    return resolver._reverse_with_prefix(view, prefix, *args, **kwargs)
  File "django/urls/resolvers.py", line 828, in _reverse_with_prefix
    raise NoReverseMatch(msg)
```

ERosendo:
This PR fixes #2747. Here are the key changes of this PR:

- Updates the `CitationRedirectorForm` class to handle two distinct input scenarios identified from [COURTLISTENER-4SW](https://freelawproject.sentry.io/issues/4439604480/?project=5257254) events. The first scenario involves input strings containing forward slashes (/), like in this event: [event-fe6daae9196b4eb68d5e38e8d8d9d821](https://freelawproject.sentry.io/issues/4439604480/events/fe6daae9196b4eb68d5e38e8d8d9d821/?project=5257254), while the second involves input strings containing URLs like the following event: [event-e6fbf3df9f1946ea8cb5998278d0e510](https://freelawproject.sentry.io/issues/4439604480/events/e6fbf3df9f1946ea8cb5998278d0e510/?project=5257254).

- During development of a previous change, I identified an issue with the citation homepage template. The template wasn't displaying form errors correctly. To address this, I modified the HTML template to ensure proper rendering of errors included in the form object.

    As an example, the following screenshot shows the validation error displayed when a user attempts to submit a URL within the citation tool's input field:

image

- In the same issue, there is another sentry error ([COURTLISTENER-4V7](https://freelawproject.sentry.io/issues/4449227335/?project=5257254&query=is%3Aunresolved+volume&referrer=issue-stream&statsPeriod=90d&stream_index=0&utc=true)) that occurred when a user submitted a statute citation instead of an opinion citation. This PR tweaks the `citation_redirector` view to use Eyecite classes to identify the citation type provided by the user and display a user-friendly error message when a non-opinion citation is detected. Here's a screenshot of the error message: 

image

### DIFF
--- a/cl/opinion_page/forms.py
+++ b/cl/opinion_page/forms.py
@@ -56,6 +56,14 @@ class CitationRedirectorForm(forms.Form):
 
     def clean_reporter(self):
         data = self.cleaned_data["reporter"]
+        # Flag reporter text for potential URLs
+        if bool(re.search(r"http[s]?://\S+", data)):
+            raise ValidationError(
+                (
+                    "URLs are not allowed in this field. "
+                    "Please remove the URL and try again."
+                )
+            )
         # Sanitize reporter text: remove slashes
         return data.replace("/", " ")
 
--- a/cl/opinion_page/forms.py
+++ b/cl/opinion_page/forms.py
@@ -1,5 +1,4 @@
 import logging
-import re
 from datetime import datetime
 from typing import Any, MutableMapping
 
--- a/cl/opinion_page/forms.py
+++ b/cl/opinion_page/forms.py
@@ -54,11 +54,6 @@ class CitationRedirectorForm(forms.Form):
         required=False,
     )
 
-    def clean_reporter(self):
-        data = self.cleaned_data["reporter"]
-        # Sanitize reporter text: remove slashes
-        return data.replace("/", " ")
-
 
 class DocketEntryFilterForm(forms.Form):
     ASCENDING = "asc"
--- a/cl/opinion_page/forms.py
+++ b/cl/opinion_page/forms.py
@@ -56,14 +56,6 @@ class CitationRedirectorForm(forms.Form):
 
     def clean_reporter(self):
         data = self.cleaned_data["reporter"]
-        # Flag reporter text for potential URLs
-        if bool(re.search(r"http[s]?://\S+", data)):
-            raise ValidationError(
-                (
-                    "URLs are not allowed in this field. "
-                    "Please remove the URL and try again."
-                )
-            )
         # Sanitize reporter text: remove slashes
         return data.replace("/", " ")
 
--- a/cl/opinion_page/forms.py
+++ b/cl/opinion_page/forms.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from datetime import datetime
 from typing import Any, MutableMapping
 
@@ -52,6 +53,11 @@ class CitationRedirectorForm(forms.Form):
         ),
         required=False,
     )
+
+    def clean_reporter(self):
+        data = self.cleaned_data["reporter"]
+        # Sanitize reporter text: remove slashes
+        return data.replace("/", " ")
 
 
 class DocketEntryFilterForm(forms.Form):
--- a/cl/opinion_page/templates/volumes_for_reporter.html
+++ b/cl/opinion_page/templates/volumes_for_reporter.html
@@ -15,8 +15,8 @@
     No Cases Found for {{ volume_names.0 }} (404) – CourtListener.com
   {% elif volumes %}
     {{ reporter }}, {{ volume_names.0 }} – CourtListener.com
-  {% elif not_opinion_citation %}
-    Not an opinion citation - CourtListener.com
+  {% elif no_citation_found %}
+    No Citations Detected (400) - CourtListener.com
   {% elif cases.paginator.count %}
     Vol. {{ volume }} of {{ volume_names.0 }} ({{ reporter }}) – CourtListener.com
   {% endif %}
@@ -64,10 +64,10 @@
         {% endfor %}
       </div>
     </div>
-  {% elif not_opinion_citation %}
+  {% elif no_citation_found %}
     <div class="col-xs-12 col-md-10 col-lg-8">
-      <h1>Citation Type Mismatch</h1>
-      <p>We identified a potential citation within the text you provided, it's not formatted in the typical style used for judicial opinions.</p>
+      <h1>No Citations Detected</h1>
+      <p>There seems to be a problem with your input! We couldn't find any citations within the text.</p>
       <p>Please try looking up a different piece of text.</p>
       <p class="v-offset-above-2">
         <a href="{% url "citation_homepage" %}"
--- a/cl/opinion_page/templates/volumes_for_reporter.html
+++ b/cl/opinion_page/templates/volumes_for_reporter.html
@@ -15,6 +15,8 @@
     No Cases Found for {{ volume_names.0 }} (404) – CourtListener.com
   {% elif volumes %}
     {{ reporter }}, {{ volume_names.0 }} – CourtListener.com
+  {% elif not_opinion_citation %}
+    Not an opinion citation - CourtListener.com
   {% elif cases.paginator.count %}
     Vol. {{ volume }} of {{ volume_names.0 }} ({{ reporter }}) – CourtListener.com
   {% endif %}
@@ -61,6 +63,15 @@
           </div>
         {% endfor %}
       </div>
+    </div>
+  {% elif not_opinion_citation %}
+    <div class="col-xs-12 col-md-10 col-lg-8">
+      <h1>Citation Type Mismatch</h1>
+      <p>We identified a potential citation within the text you provided, it's not formatted in the typical style used for judicial opinions.</p>
+      <p>Please try looking up a different piece of text.</p>
+      <p class="v-offset-above-2">
+        <a href="{% url "citation_homepage" %}"
+          class="btn btn-primary btn-lg">Look Up Another Citation</a></p>
     </div>
   {% elif cases.paginator.count %}
     {# This lists the cases in a given reporter & volume #}
--- a/cl/opinion_page/views.py
+++ b/cl/opinion_page/views.py
@@ -1111,31 +1111,6 @@ async def citation_redirector(
     This uses the same infrastructure as the thing that identifies citations in
     the text of opinions.
     """
-    # "reporter" can be a reporter or a full citation.  If it is a full
-    # citation then we will get the reporter, volume, and page from
-    # eyecite.
-    #
-    # By adding this extra test we can keep the rest of the logic untouched
-    #
-    if not volume and not page:
-        citations = eyecite.get_citations(reporter)
-        if citations and citations[0].groups:
-            # check whether the parsed object is an opinion citation
-            if not isinstance(
-                citations[0], (FullCaseCitation, ShortCaseCitation)
-            ):
-                return await throw_404(
-                    request,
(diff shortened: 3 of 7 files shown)


cl/opinion_page/views.py
```diff
@@ -1119,12 +1120,21 @@ async def citation_redirector(
...
-            c = citations[0]
+            # check whether the parsed object is an opinion citation
+            if not isinstance(
+                citations[0], (FullCaseCitation, ShortCaseCitation)
+            ):
+                return await throw_404(
+                    request,
+                    {"not_opinion_citation": True, "private": False},
+                )
+
```
mlissner, line 1131: Can't we do better here? Like, if the first citation found is not the correct type, grab the next until we find a good one? 

This input should work:

```
§102 USC 222 is the statute that was discussed in 22 U.S. 444
```

The statute should be found, ignored, and then then the case citation should be used.

cl/opinion_page/templates/volumes_for_reporter.html
```diff
@@ -62,6 +64,15 @@ <h4><a href="{% url "citation_redirector" reporter volume %}">{{ volume }}</a></
...
     </div>
+  {% elif not_opinion_citation %}
+    <div class="col-xs-12 col-md-10 col-lg-8">
+      <h1>Citation Type Mismatch</h1>
+      <p>We identified a potential citation within the text you provided, it's not formatted in the typical style used for judicial opinions.</p>
+      <p>Please try looking up a different piece of text.</p>
+      <p class="v-offset-above-2">
+        <a href="{% url "citation_homepage" %}"
+          class="btn btn-primary btn-lg">Look Up Another Citation</a></p>
+    </div>
```
mlissner, line 75: I think we can yank this if we loop through all the found citations as I mention in my other comment. Then, if we don't find a citation, we just say that's the issue.

cl/opinion_page/forms.py
```diff
@@ -53,6 +54,19 @@ class CitationRedirectorForm(forms.Form):
...
+        # Flag reporter text for potential URLs
+        if bool(re.search(r"http[s]?://\S+", data)):
+            raise ValidationError(
+                (
+                    "URLs are not allowed in this field. "
+                    "Please remove the URL and try again."
+                )
+            )
+        # Sanitize reporter text: remove slashes
+        return data.replace("/", " ")
```
mlissner, line 68: I'm surprised we need to do this. Can't eyecite just ignore bad input and move on? Like, what if my input is:

```
The https://  protocol was discussed in 22 U.S. 44
```

Shouldn't that work just fine? I thought eyecite would be really good at dealing with that?

mlissner (changes requested):
I'm sorry, I'm not sure I like your approach here. 

Also, do you think you can make the error message (if we still need it), the same width as the input box? I think that'd look a lot better. 

Finally, is there any way we can get more than one error at a time in the form? If not, can we ignore the loop in the template and just put the one error in a `<p>` tag?

cl/opinion_page/forms.py
ERosendo: > Can't eyecite just ignore bad input and move on?

Yeah Eyecite can handle the input, but we're adding a sanitization step for the reporter field because that string is not directly passed to the `get_citations` method. The `citation_homepage` view is the one that handles the form submission and the view redirects users to another view(`citation_redirector`) where the actual analysis of the input value happens.

If the value of the 'reporter' field contains forward slashes ('/'), the redirection logic in the 'citation_homepage' view fails because none of the URL patterns used for the 'citation_redirector' view allow slashes.

while writing this explanation, I realized there's a simpler solution😞. Instead of sanitizing the input text in the clean method, we can update the `citation_homepage` view to use the `get_citations` method from Eyecite. This way, the view will only redirect to the `citation_redirector` view when the input contains citations. By doing this, we can rely on Eyecite to help us sanitize the input text and we can remove `clean_reporter` method.

cl/opinion_page/views.py
ERosendo: Thank you for pointing this out. I'll implement the same approach we use for the API. 👍

cl/opinion_page/templates/volumes_for_reporter.html
ERosendo: You're right. I'm removing this.

cl/opinion_page/forms.py
ERosendo: Here's an update about this comment:

I've tried using Eyecite alone to handle the form submission within the `citation_homepage` view. While this approach works well for most scenarios, there's a specific limitation. The current implementation of the tool allows users to search by reporter name only. For instance, if a user enters "U.S." in the search bar, we should display all volumes associated with the **United States Supreme Court Reporter**. Unfortunately, Eyecite is not able to identify the reporter's name in such cases and returns an empty list.

if we still want to allow users to search by reporter name only, we need to send the text they enter to the `citation_redirection` view. However, we must ensure the text is sanitized and doesn't contain slashes. Therefore, we should keep the `clean_reporter` method around.

cl/opinion_page/views.py
ERosendo: Here's a gif showing how the filtering logic works:

filter out non-caselaw

cl/opinion_page/forms.py
mlissner: Does it get simpler if we stop supporting search for just the reporter?

ERosendo: > Does it get simpler if we stop supporting search for just the reporter?

Code-wise, If we ditch the reporter search feature, we can axe the `clean_reporter` method too (it's only like 3 lines of code anyway!) and that's pretty much it.

OTOH, supporting the reporter search feature makes the logic to show a "no citation found" error as you suggested in [here](https://github.com/freelawproject/courtlistener/pull/3916#discussion_r1544050430) a little bit confusing because an empty list from Eyecite isn't enough to display the message(The input might be a reporter name). We would only be able to show it when Eyecite finds citations within the text and after filtering the list, we find that none of them is a case law citation 🤔 

Here's a table summarizing the behavior of the new `no-citation-found` error with and without reporter search support:

| Input | Eyecite Ouput  | Reporter Search On |  Reporter Search Off
| ------------- | ------------- | ------------- | ------------- |
| a bc (Random text)  | Empty list  |  bad-reporter-error | no-citation-found(new error) |
| U.S. (Reporter name)  | Empty list  | List of volumes | no-citation-found(new error) |
|  Iowa Code § 602.1614. | List with a LawCitation |  no-citation-found(new error) | no-citation-found(new error) |
|  §102 USC 222, 44 Vand. L. Rev. 1041 (1991) | List with a LawCitation and a JournalCitation | no-citation-found(new error) | no-citation-found(new error) |
| 613 N.W.2d 120  | List with a CaseCitation |  Clusters or no-match-error | Clusters or no-match-error |

ERosendo: I think we just need to refine the error messages and we can still support reporter search:

| Input | Eyecite Ouput  | Response |  
| ------------- | ------------- | ------------- | 
| a bc (Random text)  | Empty list  |  **No reporter or citation error(replaces the bad-reporter error)** | 
| U.S. (Reporter name)  | Empty list  | List of volumes | 
|  Iowa Code § 602.1614. | List with a LawCitation |  **no-caselaw-citation-found(new error)** | 
|  §102 USC 222, 44 Vand. L. Rev. 1041 (1991) | List with a LawCitation and a JournalCitation | **no-caselaw-citation-found(new error)** | 
| 613 N.W.2d 120  | List with a CaseCitation |  Opinion page, List of Clusters or no-match-error |


- By refining the "bad-reporter" error, we can provide the user with clearer feedback. The new message can indicate that the input either doesn't contain any valid citations or a recognizable reporter name.

- The new error message you suggested [here](https://github.com/freelawproject/courtlistener/pull/3916#discussion_r1544050430) could inform users that their input might contain citations, but not for legal opinions. We can refine the existing *Citation Type Mismatch* error introduced by this PR.

mlissner: I think looking up just reporters really isn't very useful, and it adds complexity to the system, so I don't think it's worth it even if you could make it work. I'm also thinking that the cleaning step feels wack-a-mole. Like, we can take care of slashes and `https`, but eventually we'll find some other funky thing we have to deal with.

ERosendo: Alright! I'll remove the code to stop supporting reporter searches via the search bar. However, we can still access all volumes associated with a reporter using the URL. Thanks for your input! 👍 🚀

mlissner: Perfect. Thank you!

cl/opinion_page/templates/volumes_for_reporter.html
ERosendo: Alright. This is the new error:

image

mlissner: That'll work!

ERosendo: Thanks!

cl/opinion_page/forms.py
ERosendo: Here's a GIF showing how the citation search works with the new logic:

Screen Recording 2024-03-29 at 5 19 08 PM


I tested the following scenarios:

- Random Text 
- Reporter name
- Non-caselaw citations:
   - Iowa Code § 602.1614
   - §102 USC 222, 44 Vand. L. Rev. 1041 (1991)
- Caselaw citation: 
   - 613 N.W.2d 12
- Listing the volumes using the URL
- A citation with no volume

ERosendo:
@mlissner Thanks for your comments. I've applied all the changes and [here's](https://github.com/freelawproject/courtlistener/pull/3916#discussion_r1544888664) a GIF showing how the tool is working with the new logic. 

> do you think you can make the error message (if we still need it), the same width as the input box? 

I removed this section from the template because we don't need it after the latest changes.

cl/opinion_page/forms.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import re
```
mlissner, line 2: Looks like you can remove this now, right?

mlissner (approved):
Looks great. One tweak, then feel free to merge, and it's the weekend!

cl/opinion_page/forms.py
ERosendo: You're right!

Issue #3847, opened 2024-03-01
Replace `ErrorLog` call for `logger.error` calls so they are visible on Sentry


grossir:
We have an `ErrorLog` Django Model and DB table that is used to capture problem with `get_binary_content` on both cl_scrape_opinions and cl_scrape_oral_arguments

https://github.com/freelawproject/courtlistener/blob/b62611d1554d44f432b88b6f481509cdd76a25fb/cl/scrapers/models.py#L30

However, this makes it necessary to check 2 places: Sentry and the ErrorLog table. Also, not everyone has access to the ErrorLog table in the prod DB, so it may mask issues, such as freelawproject/juriscraper#948

I propose replacing this ErrorLog, or adding a parallel `logger.error` call that would be sent to Sentry

PR #3939, opened 2024-04-03
feat(cl.scrape_opinions): send get_binary_content failures to sentry

grossir:
Solves #3847

We are not deleting the ErrorLog creation, simply duplicating it to Sentry for better visibility


cl/scrapers/management/commands/cl_scrape_opinions.py
```diff
@@ -258,7 +258,8 @@ def scrape_court(self, site, full_crawl=False, ocr_available=True):
                 method=site.method,
             )
             if msg:
-                logger.warning(msg)
+                fp = [f"{court_str}-unexpected-content-type"]
+                logger.error(msg, extra={"fingerprint": fp})
```
flooie, line 262: can we use something better than `fp` as a variable name

flooie (approved):
Looks good to me

PR #3943, opened 2024-04-04
feat(citation): Add throttling to the citation API endpoint

ERosendo:
This PR introduces the following key changes: 

- Enforces the maximum allowed characters for the `text` key in CitationAPIRequestSerializer to 64,000. Requests exceeding this limit will now be rejected.

-  Enhances API citation filtering with a new helper method. This method streamlines the process by efficiently removing both non-case-law citations and citations lacking volume and page information. This function replaces the less specific `filter_out_non_case_law_citations` method in the `CitationLookupViewSet` class.

- Introduces a new environment variable, `MAX_CITATIONS_PER_REQUEST`, to define the maximum number of citations the API will look up in a single request. This variable defaults to 250.

- Enhances throttling for citation lookups in the API by introducing two changes to the settings module:
   - Adds a new rate to the DEFAULT_THROTTLE_RATES dictionary. This allows us to define a default rate limit for regular API users.
   - Introduces a new dictionary, CITATION_LOOKUP_OVERRIDE_THROTTLE_RATES. This enables us to define specific throttle rates for certain users, overriding the default rate.

- Implements a new throttling mechanism specifically for citation lookups.

   - Introduces the `CitationCountRateThrottle` class, inheriting from `ExceptionalUserRateThrottle`. This ensures consistency with throttling logic used in other API endpoints.
   
   - Uses caching to track the number of citations a user has looked up through the endpoint.
   
   - Enhances the allow_request method with a two-step check:
      - It first verifies the user's citation lookup count using the cache and applies rate limiting if necessary.
      - It then performs the standard request check from the parent class.
 
   - Provides informative error messages for throttled users: The CitationCountRateThrottle class uses a custom response when throttling occurs. This response includes two key pieces of information:
       - `error_message`: This clarifies that the request was throttled due to exceeding the rate limit.
       - `wait_until`: This provides the user with the earliest time they can make another request without being throttled again, formatted in ISO 8601 standard 

      ```
      {
          "error_message": "Request was throttled.",
          "wait_until": "2024-04-09T14:25:05.896574+00:00"
      }
     ```


cl/citations/tests.py
```diff
@@ -2161,3 +2195,54 @@ async def test_can_extract_all_citations_within_text(self) -> None:
...
+        ten_citations = "56 F.2d 9, " * 10
+        r = await self.async_client.post(
+            reverse("citation-lookup-list", kwargs={"version": "v3"}),
+            {"text": ten_citations},
+        )
+
+        self.assertEqual(r.status_code, HTTPStatus.TOO_MANY_REQUESTS)
+        data = json.loads(r.content)
+        self.assertEqual(data["error_message"], "Request was throttled.")
+        self.assertIn("wait_until", data)
```
mlissner, line 2248: I think it'd be helpful to actually have an assertion for the correct value of `wait_until`. I think if you use [time machine](https://adamj.eu/tech/2020/06/03/introducing-time-machine/), you can accomplish things like this? I think we already have it as a dependency?

cl/api/utils.py
```diff
@@ -346,6 +349,148 @@ def allow_request(self, request, view):
...
+            )
+            if remaining_citation < max_num_citations or not idx:
+                datetime_obj = datetime.fromtimestamp(
+                    self.history[idx][-1], timezone.utc
+                )
+                soonest_time = datetime_obj.isoformat()
+
+        raise Throttled(
+            detail={
+                "error_message": "Request was throttled.",
```
mlissner, line 488: We'd have to update line 474 to change `_` to `unit`, but if we did, we could do:

```suggestion
                "error_message": f"Too many requests (allowed rate: {max_num_citations}/{unit}).",
```

cl/api/utils.py
```diff
@@ -346,6 +349,148 @@ def allow_request(self, request, view):
...
+        if not text:
+            # Since the 'text' key is missing from the request, the user is
+            # likely trying to retrieve opinions using a reporter, volume,
+            # and page combination. This approach allows looking one citation
+            # at a time.
+            return 1
+
+        citation_objs = filter_out_non_case_law_and_non_valid_citations(
+            eyecite.get_citations(text)
+        )
```
mlissner, line 381: I don't suppose there's a way to do this in the request and pass it to the throttle, so we only have to do it once? I think not, but I thought I'd ask.

mlissner (commented):
This looks about right. A few things that crossed my mind though:

1. I haven't totally engaged with the logic of your throttling code because I think Alberto will do a better job. It's clearly quite tricky, but the things I'm looking for are:
 
     - Do you hit the cache / DB as few times as possible? 

        I think you check the cache once and then add to it once. Great.

    - Does it parse more than necessary or do other slow things a lot?

        I'm not so sure here. I made a comment for you.

     - Does it work?

        Again, I'm not so sure. I think we need more tests to make me feel more confident. Can you think about tests that might help? I made one suggestion, but I suspect you'll have other ideas. 

1. We still need to update the documentation to pull from settings. Do you want that here or in a separate PR? Either way, we could get clever and include personal overrides in the documentation if we wanted to. If we did that, we'd want to say in the default documentation something like:

    > 250 citations will be looked up per request.

    And in a special person's documentation, it could say:

    > The default throttle is 250 citations per request, but your account allows 500.

    Or whatever, but if we can do that, people might appreciate it. If it's hard, or not worth it, that's OK too.


1. Finally, is there a deployment process we need to do to put the correct variables in place before this launches? 

Thank you again! Excited to get this launched.

cl/api/utils.py
```diff
@@ -346,6 +349,148 @@ def allow_request(self, request, view):
...
+        """
+        Checks the settings for a custom citations API rate limit.
+
+        If the authenticated user has a custom rate limit set in the settings,
+        it parses that value. Otherwise, it parses the default rate limit.
+
+        Args:
+            request: The request object with the user's data.
+        """
+        default_rate = self.THROTTLE_RATES["citations"]
```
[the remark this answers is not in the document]
ERosendo: This setting is currently configured within the application itself and is not defined by an environment variable. It's similar to the `anon` and `user` rates, which are also hardcoded in the `rest_framework.py` file.  

Would you prefer to manage this rate with an environment variable instead?

cl/citations/api_views.py
```diff
@@ -60,6 +63,16 @@ def create(self, request: Request, *args, **kwargs):
                     "end_index": end_index,
                 }
 
+                if idx == settings.MAX_CITATIONS_PER_REQUEST:
+                    citations.append(
+                        {
+                            **citation_data,
+                            "status": HTTPStatus.TOO_MANY_REQUESTS,
+                            "error_message": "Too many requests.",
```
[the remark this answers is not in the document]
ERosendo: Thanks for catching this! I'll update the error message

cl/citations/tests.py
```diff
@@ -2161,3 +2195,54 @@ async def test_can_extract_all_citations_within_text(self) -> None:
...
+        self.assertEqual(len(data), 11)
+
+        last_citation = data[-1]
+        self.assertEqual(last_citation["citation"], "139 U.S. 601")
+        self.assertEqual(last_citation["status"], HTTPStatus.TOO_MANY_REQUESTS)
+        self.assertEqual(last_citation["error_message"], "Too many requests.")
+
+    @patch(
+        "cl.api.utils.CitationCountRateThrottle.get_citations_rate",
+        return_value=(20, 60),
```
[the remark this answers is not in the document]
ERosendo: The `get_citations_rate` method currently returns a `Tuple[int, int]`. This is because it parses the citation rate before returning the value. We can refactor this code to return the rate as a `string`, making it consistent with the other Throttle classes in DRF. The [get_rate](https://github.com/encode/django-rest-framework/blob/9864c47018238dd9f8477df0e4ab2ef62a24fd2b/rest_framework/throttling.py#L82-L95) method in the SimpleRateThrottle class returns the rate as a string the includes the value and the unit( e.g., "20/s", "60/m", "250/h")

cl/api/utils.py
ERosendo: Thanks for the suggestion, I'll tweak the `get_citations_rate` method so we can include the rate in the error message

cl/api/utils.py
```diff
@@ -346,6 +349,148 @@ def allow_request(self, request, view):
...
+
+        Returns:
+            int: The number of citations as an integer.
+        """
+        validated_data = view.validate_request_data(request)
+        text = validated_data.get("text", None)
+        if not text:
+            # Since the 'text' key is missing from the request, the user is
+            # likely trying to retrieve opinions using a reporter, volume,
+            # and page combination. This approach allows looking one citation
```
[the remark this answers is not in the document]
ERosendo: Thanks for your suggestion.

cl/citations/tests.py
ERosendo: > I think we already have it as a dependency?

Yes, It's already a dependency. I'll update the test.

cl/api/utils.py
ERosendo: Actually, I believe there's a way to compute this list of citations just once. Since the `get_citation_count_from_request` method has access to both the request object and the view object, we can store the list of citations as an attribute of the view instance. This way, we can avoid redundant computations whenever a request is allowed.

The new `get_citation_count_from_request` helper will look like this: 

```python
def get_citation_count_from_request(self, request, view) -> int:
    ...
    citation_objs = filter_out_non_case_law_and_non_valid_citations(
         eyecite.get_citations(text)
    )
    view.citations = citations_objs
    return len(citations_objs)
```


Then, in the API view, we'll apply the following tweaks:


```diff
def create(self, request: Request, *args, **kwargs):
   ...
   if text:
-       citation_objs = eyecite.get_citations(text)
-       citation_objs = filter_out_non_case_law_and_non_valid_citations(
-               citation_objs
-       )
-       if not citation_objs:
+       if not self.citations:
            return Response([])
         
-       for idx, citation in enumerate(citation_objs):
+.      for idx, citation in enumerate(self.citations):
            ...

```

Let me know what you think

cl/citations/api_views.py
```diff
@@ -60,6 +63,16 @@ def create(self, request: Request, *args, **kwargs):
...
+                        }
+                    )
+                    break
```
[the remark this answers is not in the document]
ERosendo: > we parse all the citations and give useful info about all of them, even if we don't look them up. Parsing is fast. Might as well return it

You're absolutely right. Thanks for catching that!  I'll update the code to return all citations and make sure any citation after the maximum allowed(`MAX_CITATIONS_PER_REQUEST`) has a status 429, indicating "Too many citations requested."

cl/api/utils.py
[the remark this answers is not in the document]
mlissner: Got it. Yeah, it's fine, you're right.

cl/citations/tests.py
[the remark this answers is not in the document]
mlissner: Hm, I'm even more confused now, because looking at `get_citations_rate`, it looks like it either does:

```
default_rate = self.THROTTLE_RATES["citations"]
```
Which returns a `str` like `"60/min"`, or it gets overridden, which does:

```
        custom_rate = settings.REST_FRAMEWORK[
            "CITATION_LOOKUP_OVERRIDE_THROTTLE_RATES"
        ].get(request.user.username, None)
```

In that case, it returns something like, say, `"1000/min"`, no? I don't see where it returns either a Tuple or anything other than a string?

cl/api/utils.py
mlissner: I'd be confused how the view object got a `citations` attribute, but otherwise it makes sense. Maybe just add a comment in there:

```
# self.citations is set by the throttle class. This is a performance
# enhancement to avoid parsing the citations twice, first in the 
# throttle and then again here.
if not self.citations:
``` 

Works for me?

cl/citations/tests.py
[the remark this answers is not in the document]
ERosendo: > I don't see where it returns either a Tuple or anything other than a string.

I recently updated this helper function in commit [c1858a0](https://github.com/freelawproject/courtlistener/pull/3943/commits/c1858a075d39be5e6e078949b8243176aa340539). This change simplifies the function's output by returning a single string value instead of a tuple. Previously, the helper function returned the value from calling `self.parse_rate(custom_rate or default_rate)` and self.parse_rate() returns tuples.

cl/api/utils.py
ERosendo: > Maybe just add a comment in there

great! I'll do that. Thanks 👍

cl/citations/tests.py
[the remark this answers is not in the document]
mlissner: Oh, whew, I'm not crazy. Good.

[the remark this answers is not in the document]
ERosendo: > I'm not crazy.

No, you're not 😄. It's my fault. I should've added a comment saying I updated the code because your initial comment made me realize that returning strings makes a lot more sense for this particular helper function.

ERosendo:
@mlissner Thanks for your comments. I've updated the code to address all your comments. Additionally, I've updated the documentation in this PR to pull data from the settings and include personal overrides.

cl/citations/tests.py
```diff
@@ -2209,40 +2210,181 @@ async def test_can_limit_max_citations_per_request(
...
+        self.assertEqual(
+            last_citation["error_message"], "Too many citations requested."
+        )
 
     @patch(
         "cl.api.utils.CitationCountRateThrottle.get_citations_rate",
-        return_value=(20, 60),
+        return_value="20/m",
     )
     async def test_can_rate_limit_user_based_on_citation_count(
```
mlissner, line 2240: This is a very big test. Worth splitting it up?

mlissner (approved):
One last thought from me, but if you and Alberto like this one, I'm all set. :)

cl/api/utils.py
```diff
@@ -346,6 +349,154 @@ def allow_request(self, request, view):
...
+        for idx in reversed(range(len(self.history))):
+            remaining_citation = sum(
+                citation_count for citation_count, _ in self.history[:idx]
+            )
+            if remaining_citation < max_num_citations or not idx:
+                datetime_obj = datetime.fromtimestamp(
+                    self.history[idx][-1], timezone.utc
+                )
+                soonest_time = datetime_obj.isoformat()
+
```
albertisfu, line 491: Shouldn't the `soonest_time` indicate the closest time a user can make a new request after history records that exceeded their quota have expired?  However, during my testing, it appears the `wait_until` is not aligned with this behavior. Instead of showing the earliest time a user can submit a new request, it displays the latest time after all the user's history records have expired.

If my assumption is correct this can be easy fixed by breaking the for loop once `soonest_time` is found.

```suggestion
break
```

cl/citations/tests.py
```diff
@@ -2161,3 +2196,195 @@ async def test_can_extract_all_citations_within_text(self) -> None:
...
+            ten_citations = "56 F.2d 9, " * 10
+            r = await self.async_client.post(
+                reverse("citation-lookup-list", kwargs={"version": "v3"}),
+                {"text": ten_citations},
+            )
+            self.assertEqual(r.status_code, HTTPStatus.OK)
+            data = json.loads(r.content)
+            self.assertEqual(len(data), 10)
+
+            # Ten more citations, This request should be allowed
```
albertisfu, line 2257: Aligned with my previous comment regarding `soonest_time`, the test does not appear to catch the issue of displaying `soonest_time` because requests are executed within the same `test_date` context.

Therefore, we could employ a second `time_machine` context for different requests to capture the time differences. For example, here we can create `test_date_next`, which would be a few seconds later than `test_date`.

cl/citations/tests.py
```diff
@@ -2161,3 +2196,195 @@ async def test_can_extract_all_citations_within_text(self) -> None:
...
+        )
+
+    @patch(
+        "cl.api.utils.CitationCountRateThrottle.get_citations_rate",
+        return_value="20/m",
+    )
+    async def test_can_rate_limit_user_based_on_citation_count(
+        self, get_rate_mock, throttle_logic_mock
+    ) -> None:
+        throttle_logic_mock.return_value = "citation_throttle_test"
```
albertisfu, line 2243: In the event of a test failure, it appears that the throttle key retains records from the previous execution, making it difficult to debug the actual issue. Therefore, it would be a good idea to clear the history records before the test begins.

albertisfu (commented):
This looks great. I have reviewed and tested the different throttling mechanisms introduced in this PR, and they work correctly.

I only found a possible discrepancy regarding the `wait_until` time informed to the user.

Additionally, I some comments related to the tests.

cl/api/utils.py
ERosendo: You're right!: Thanks for catching that.

cl/citations/tests.py
ERosendo: Thanks for the suggestion. I'll update the `setUp` method.

cl/citations/tests.py
ERosendo: Yes, splitting this test is a good idea.

ERosendo:
@albertisfu Thanks for your comments. I've updated the code to address the issue with the loop in the `throttle_request` method and the tests. I've split the lengthy test to check the throttling logic into three smaller, more focused tests to improve readability,

albertisfu (approved):
Perfect! The changes are great, and this one looks ready to me!

PR #3946, opened 2024-04-05
3945 citation api documentation

### LINKED ISSUE #3945: Document the new citation look up API
I've been working on this here: 

https://docs.google.com/document/d/1rotEkmhwCQa-jojuHJHe8PS8W3HaZ4C9urJxxPmOABo/edit

Should be pretty straightforward at this point. "Just" have to get the HTML looking good and add a few more examples.

mlissner:
This PR fixes #3945 by adding a new page to our documentation and linking to it in various places. This PR should be merged after it has been updated as below and after the throttle PR has landed.

This PR is nearly done, but needs the following tweaks:
 - References to throttle speeds and limitations should use settings once they are available, instead of being hardcoded in the HTML.
 - There's an XXX where it needs an example of a throttled API request with more than 250 items.
 - A simple test is a good idea, just in case, to ensure the new page continues loading into the future, forever.

In addition, this PR:
 - Fixes a widowed header in a donate plea.


mlissner:
I made a guess at the correct documentation for a request with > 250 citations since a customer wants this documentation. Auto-merge is set. 

After this merges, we still need to clean up whatever I got wrong in my guess, and then to do the other couple of tasks.

PR #3963, opened 2024-04-11
feat(search): Adds command to compute token count

### LINKED ISSUE #3958: Develop script to approximate a few token counts
We need a script to figure out a few token counts:

1. How many tokens are in the text of the RECAP Archive? Let's take random 10,000 items from the DB where is_available=True and there is text in the plaintext column. Let's count the tokens, and then extrapolate from there to the amount that is in the entire corpus.

3. We need a per-page token count from RECAP so we can see if buying the biggest documents makes economic sense these days. It just might. 

    Again, a random sample of documents would work well and we can do the simple thing: "The doc is ten pages, it has 15,000 words, therefore there are 1,500 words per page."

2. How many tokens are in case law? Same as above, but we have a few things to do before measuring tokens:
     - Find the best field. For this, we can probably use `html_with_citations`. If it's blank, then we can use the fields in the order explained in the API: https://www.courtlistener.com/help/api/rest/#opinion-endpoint
     - Remove the HTML. 

----

Counting tokens should be done using `tiktoken`, OpenAI's tokenizer, and the `cl100k_base` encoding model. A simple example of this is here:

https://stackoverflow.com/questions/75804599/openai-api-how-do-i-count-tokens-before-i-send-an-api-request/75804651#75804651

Another is here:

https://github.com/openai/openai-cookbook/blob/main/examples/How_to_count_tokens_with_tiktoken.ipynb

ERosendo:
This PR introduces a new command, cl_compute_token_count, to calculate the total number of tokens within the Recap Archive and Opinions collections.

The command accepts an optional `percentage` argument (float between 0 and 100) that allows you to specify the portion of the data to be sampled for the calculation. The default value is set to 0.1, which translates to roughly 12,000 valid documents for Recap(documents with `plain_text` and `page_count`) and 9,500 for Opinions.

Example Usage:

```
manage.py cl_compute_token_count
```

or 

```
manage.py cl_compute_token_count --percentage 0.2
```

This PR fixes #3958


cl/search/management/commands/cl_compute_token_count.py
```diff
@@ -0,0 +1,205 @@
...
+        self.stdout.write(
+            f"Total number of tokens in the recap archive: {intword(total_token_in_recap)}"
+        )
+
+        opinion_queryset = get_opinions_random_dataset(percentage)
+        self.stdout.write("Starting to retrieve the random Opinion dataset.")
+        for opinion in opinion_queryset.iterator():
+            text = get_clean_opinion_text(opinion)
+            count = get_token_count_from_string(text)
+            token_count.append(count)
```
mlissner, line 182: Don't you need to clear the `token_count` variable before you start appending to it again?

ERosendo: you're right!. Thanks for catching that.

mlissner (approved):

PR #3971, opened 2024-04-12
tests(alerts): Adds test for sending emails with surrogate pairs

### LINKED ISSUE #3969: Cannot send emails with non-ascii characters
There are actually four issues in Sentry that this affects. I'll link them all to this one, which has the most instances. We need to fix this quickly because it's affecting our alerts. 

I'm not sure what caused this, but you can easily reproduce it by using the link here: 

https://www.courtlistener.com/reset-password/

The error started about 20 hours ago, when I merged the token counting PR: 

https://github.com/freelawproject/courtlistener/pull/3963/files

It came with a new dependency for tiktoken, but no other dependency updates as I understand the poetry.lock file to work.

So:
 - Why did this start happening?
 - Does it affect others (if so, can we fix it upstream)?
 - How do we fix it as quickly as possible?
 - How do we get the upstream fix in place (assuming we make one/there is one)?

Sentry Issue: [COURTLISTENER-71K](https://freelawproject.sentry.io/issues/5180196366/?referrer=github_integration)

```
UnicodeEncodeError: 'utf-8' codec can't encode characters in position 48-50: surrogates not allowed
(13 additional frame(s) were not displayed)
...
  File "cl/alerts/tasks.py", line 348, in send_alert_and_webhook
    connection.send_messages(messages)
  File "cl/lib/email_backends.py", line 226, in send_messages
    email.send()
```

ERosendo:
Ensures email functionality supports Unicode characters beyond basic multilingual plane.

This PR fixes #3969


mlissner:
Auto merge is set.

Issue #3953, opened 2024-04-08
Enable eyecite with vectorscan in prod (instead of using Python's re module)


mlissner:
Amazingly, in prod we're not using the CPU-optimized version of eyecite, which uses vectorscan: 

https://github.com/VectorCamp/vectorscan

We should do that. It's faster! It saves energy! Let's do that. I think we just need to enable/install a dependency, but the eyecite docs should have the details.

mlissner:
Over in https://github.com/freelawproject/eyecite/pull/177, we bumped to hyperscan 0.7.7, which switches to vectorscan and thereby enables support for ARM processors. That might be enough to make vectorscan work, but I'd like to be sure we're using it.

mlissner:
Yeah, looks like we need to enable the correct tokenizer: https://github.com/freelawproject/eyecite/?tab=readme-ov-file#tokenizers

PR #3976, opened 2024-04-16
feat(citations): Implements the Hyperscan Tokenizer

ERosendo:
This PR addresses issue #3953 by integrating Hyperscan as a new dependency. The get_citations method is then updated to use the capabilities of Hyperscan's tokenizer, enabling faster and more efficient citation detection within the application.


mlissner (approved):
Easy enough! Looks good to me. When this deploys, it might break things, since our tests are being run on Intel processors (I assume), and prod is on ARM. So we'll roll it out, see what happens, and roll back if we need to.

mlissner:
Auto-merge enabled.

mlissner:
Looks like this is fixed, I believe. Do you mind closing all the Sentry issues, please?

ERosendo:
@mlissner Sure!

PR #3990, opened 2024-04-22
feat(citations): Adds the LoggingMixin to the citation API

ERosendo:
This PR implements logging for the `CitationLookupViewSet` by making it inherit from `LoggingMixin` class. This improves traceability within the viewset.


mlissner:
Automerge set, thank you. Easy enough!

mlissner:
We should do more mixins.

Issue #4018, opened 2024-05-03
Error 403 in development environment when uploading files to S3


quevon24:
@grossir  told me that he was having a 403 error when uploading files

`An error occurred (403) when calling the HeadObject operation: Forbidden`

I was reviewing the problem with @blancoramiro and we discovered that to use temporary credentials[ it is also required to pass the session token to validate the credentials. ](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html)

Temporarily, the solution is to add an extra variable to the .env file.

`AWS_DEV_SESSION_TOKEN="XXXXXXXX"`

and add a line to cl/settings/third_party/aws.py

```
if DEVELOPMENT:
    AWS_ACCESS_KEY_ID = env("AWS_DEV_ACCESS_KEY_ID", default="")
    AWS_SECRET_ACCESS_KEY = env("AWS_DEV_SECRET_ACCESS_KEY", default="")
    AWS_SESSION_TOKEN = env("AWS_DEV_SESSION_TOKEN", default="") # <-------- add this line

```

and then reload .env file:
`docker-compose -f docker/courtlistener/docker-compose.yml up -d`

@blancoramiro told me that he is going to look for an alternative to this so that we don't have to constantly modify the environment variables in development environment.

blancoramiro:
I created PR [4019](https://github.com/freelawproject/courtlistener/pull/4019/files) to show some changes that could possibly help help with issue.

In the file cl/settings/third_party/aws.py the variables `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` but I couldn't find  any reference to them anywhere in the repo. Boto does not use those variables, it uses the ones in the environment: [boto3 authentication](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#passing-credentials-as-parameters). I tested this locally.

Knowing that docker-compose is only used locally, I suggest adding the lines in the PR to the compose file so that you could set the variables in the following way:

- Obtain the AWS environment variables from the SSO screen:
![test](https://github.com/freelawproject/courtlistener/assets/82423284/70e8b771-f24b-4688-abf8-2a8762dd5025)

- Paste them into the console.
- Run docker compose up.

The AWS environment variables will be then available inside the `cl-django` and `cl-celery` containers and boto3 will use them.

@quevon24  We could test this to see if works properly locally.

Thank you!

grossir:
[...]

PR #4019, opened 2024-05-03
4018 Error 403 in development environment when uploading files to S3

blancoramiro:
This PR is related to issue #4018 

- Removes unused apparently unused AWS variables from settings
- Explicitly sets AWS environment variables in courtlistener docker compose file.


quevon24:
@mlissner the change looks good, it is easy to use, we just need to paste the environment variables provided by aws (there is a copy button in the portal) into the host terminal:

```
export AWS_ACCESS_KEY_ID="xxxx"
export AWS_SECRET_ACCESS_KEY="xxxx"
export AWS_SESSION_TOKEN="xxxx"
```

and then update the env variables in the container:

`docker-compose -f docker/courtlistener/docker-compose.yml up -d
`

As the keys expire every 8 hours, this would be necessary to do it daily only if you are going to use s3 file upload

cl/settings/third_party/aws.py
```diff
@@ -3,15 +3,7 @@
...
 DEVELOPMENT = env.bool("DEVELOPMENT", default=True)
 
-
 # S3
-if DEVELOPMENT:
-    AWS_ACCESS_KEY_ID = env("AWS_DEV_ACCESS_KEY_ID", default="")
-    AWS_SECRET_ACCESS_KEY = env("AWS_DEV_SECRET_ACCESS_KEY", default="")
-else:
-    AWS_ACCESS_KEY_ID = env("AWS_ACCESS_KEY_ID", default="")
-    AWS_SECRET_ACCESS_KEY = env("AWS_SECRET_ACCESS_KEY", default="")
```
mlissner, line 13: Hm, these are used in prod. Are you sure we can remove them safely?

blancoramiro: I am pretty sure boto is picking them up from the environment vars.
But just to be sure I just reverted this change. Too risky.

mlissner:
A few notes to get this over the line, since it's one of your first CourtListener PRs, @blancoramiro:

1. We need you to sign the contributor thing (see comment above).
2. Tests aren't passing yet, when they're fixed, please mark it as ready for review (it's still a draft), and you can set me as the reviewer.

cl/settings/third_party/aws.py
quevon24: > Hm, these are used in prod. Are you sure we can remove them safely?

mmm that is a good question, we can probably leave it as I suggested at https://github.com/freelawproject/courtlistener/issues/4018#issue-2277950289

mlissner:
Should we put this in the dev documentation?

blancoramiro:
Yes. Probably [here](https://github.com/freelawproject/courtlistener/wiki/Getting-Started-Developing-CourtListener#developer-installation)

Also not sure if it's better to receive a warning when you run the builds or if we want to setup some default bogus values for these variables.


```
WARNING: The AWS_ACCESS_KEY_ID variable is not set. Defaulting to a blank string.
WARNING: The AWS_SECRET_ACCESS_KEY variable is not set. Defaulting to a blank string.
WARNING: The AWS_SESSION_TOKEN variable is not set. Defaulting to a blank string.
```

Maybe something like this:

```
    environment:
      - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-readdocumentation}
      - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY:-readdocumentation}
      - AWS_SESSION_TOKEN=${AWS_SESSION_TOKEN:-readdocumentation}
```

mlissner:
¿Porqúe no los dos? We can do warnings and also make the values a pointer to the documentation. Hell, you could make them bit.ly links or something, no? 

> Yes. Probably [here](https://github.com/freelawproject/courtlistener/wiki/Getting-Started-Developing-CourtListener#developer-installation)

Looks right to me.

blancoramiro:
Sadly it looks like it's one or the other, at least with out adding something else like a script that checks the vars before docker-compose. There might be a way that I am missing though.
[Interpolation](https://docs.docker.com/compose/environment-variables/env-file/#interpolation)

I am thinking that it'll be best to remove the warning since it might lead contributors to think that they *need* those in place to work when probably not.

Added [this](https://github.com/freelawproject/courtlistener/wiki/Getting-Started-Developing-CourtListener#aws) to the documentation. I would appreciate a review, not a 100% happy with it.

Do we habe a bit.ly account? I think it's safe to add the full url, but maybe not as pretty.

mlissner:
Eh, let's do the full URL and forget bit.ly, that's probably better anyway.

If we have to choose one, removing the warning seems fine to me too. 

I'm not a fan of having to set and export env's. Don't we usually set things in our .env file?

blancoramiro:
> Eh, let's do the full URL and forget bit.ly, that's probably better anyway.
> If we have to choose one, removing the warning seems fine to me too.

I'll create the PR.

> I'm not a fan of having to set and export env's. Don't we usually set things in our .env file?

The issue is that now since we are using the SSO and using temporary credentials adding them to the .env file could be more troublesome than just login, copying the creds with the `copy` button, pasting them into the terminal and then running docker compose.

The SSO provides an easy way copy the credentials ready to be pasted and export the variables:

creds


Plus this way no keys are stored on disk.

mlissner:
Got it!

PR #4025, opened 2024-05-05
fix(citations): Display court name on multiple results page

### LINKED ISSUE #4013: `court_data` not set on multiple results page, leaving blank spot in website
If you go here: 

https://www.courtlistener.com/c/us/488/984/

You'll see blanks where we should have the court names:

image

Looking in the template, that value is supposed to come from the `court_data` attribute: 

```
{{ cluster.court_data }} |
```

And that attribute in the view is supposed to be set with:

```
        clusters = (
            OpinionCluster.objects.filter(citation=citation_str)
            .select_related("docket")
            .prefetch_related(Prefetch("docket__court", to_attr="court_data"))
        )
```

But somehow it's not working. We should fix this.

ERosendo:
This PR fixes #4013 by tweaking the `get_clusters_from_citation_str` method and updating the `citation_redirect_info_page.html` template. 

I identified the reason why the court name wasn't showing in the template. The prefetch_related approach stores cached data in a separate attribute on the parent model instance. In this case, the parent was the Docket model, not the Cluster. As a result, the court_data attribute wasn't available within the Cluster instances. To access the court data, we'd need to use: 

```python
cluster.docket.court_data
```

However, for a simple one-to-one relationship like the one between `Docket` and `Court`, this approach feels clunky. Since the court query is straightforward, using `select_related` would be more efficient.

This PR also updates the tests related to the multiple results page to check the court name is displayed correctly for all results.

### DIFF
--- a/cl/opinion_page/templates/citation_redirect_info_page.html
+++ b/cl/opinion_page/templates/citation_redirect_info_page.html
@@ -145,7 +145,7 @@
                             <a href="{{ cluster.get_absolute_url }}"
                                {% if cluster.blocked %}rel="nofollow"{% endif %}>{{ cluster.caption|safe|v_wrapper }}</a>
                             <br>
-                            {{ cluster.court_data }} |
+                            {{ cluster.docket.court }} |
                             {{ cluster.date_filed }}
                         </li>
                     {% endfor %}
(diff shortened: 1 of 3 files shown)


mlissner:
Nice. Simple enough. Set for auto-merge!

PR #4058, opened 2024-05-20
Update clone command

quevon24:
I made some changes to the clone command to warn the user and finish execution if CL_API_TOKEN is not set.

It now also checks that the API token is valid and finish execution immediately if it is not.

And finally i updated the docstrings.


cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -80,10 +81,31 @@ def __init__(self, message: str) -> None:
 
 
 def get_id_from_url(api_url: str) -> str:
-    """Get the PK from an API url"""
+    """Get the PK from an API url
+
+    :param api_url: api url with a pk
+    :return: list of opinion cluster objects
```
mlissner, line 87: This doesn't seem right, is it?

mlissner (approved):
One little thing, then let's merge!

cl/scrapers/management/commands/clone_from_cl.py
quevon24: It's already fixed and merged

mlissner: Thank you Kevin! I appreciate your continued enhancements to this script. It seems like it's becoming a popular one!

PR #4065, opened 2024-05-22
feat(settings): Adds variable to set the private storage bucket name

ERosendo:
This PR adds a new variable `AWS_PRIVATE_STORAGE_BUCKET_NAME` to the settings. This allows us to configure a separate storage bucket specifically for development purposes. This way, when developing new features or debugging existing ones, we can avoid accidentally uploading files to the production bucket, improving development safety and data security.

### DIFF
--- a/cl/lib/storage.py
+++ b/cl/lib/storage.py
@@ -3,6 +3,7 @@ import os
 import uuid
 from typing import Dict, Optional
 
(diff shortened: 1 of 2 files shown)


mlissner (approved):
LGTM. Thanks for the quick solution!

PR #4075, opened 2024-05-25
feat(recap): Adds logic to check court_id to the fetch_docket method

### LINKED ISSUE #4074: Add logic to check the court_id within the `fetch_docket` method
While debugging a sentry report regarding the `PacerLoginException`, I noticed that users can potentially attempt to purchase appellate dockets, which are currently not supported.

The API documentation for recap-fetch specifies three ways to identify the desired docket:

1. Sending `docket_id`
2. Sending `pacer_case_id` and `court`
3. Sending `docket_number` and `court` 

While methods 2 and 3 involve sending the court ID, allowing verification of whether it's a district/bankruptcy court (supported), method 1 bypasses this check and the subsequent purchase workflow(chain of tasks) lacks court type validation, enabling users to initiate purchases for unsupported appellate dockets.

ERosendo:
This PR introduces logic to the fetch_docket method to validate the court_id parameter. This aligns with the validation logic already implemented in the PacerFetchQueueSerializer when a user submits a court_id in the request.

Additionally, the included tests ensure that the status of the PacerFetchQueue record is updated appropriately when dealing with appellate docket records.

This PR fixes #4074


mlissner (commented):
Did you consider doing this in the `PacerFetchQueueSerializer.validate` method, where we usually check if the court is OK?

ERosendo:
> Did you consider doing this in the PacerFetchQueueSerializer.validate method, where we usually check if the court is OK?

I didn't 😢, sorry!. Thanks for pointing that out, Adding the extra logic to the validate method to avoid repetition makes perfect sense. I'll update the PR to change the approach.

ERosendo:
@mlissner I've updated this PR and it's ready for another review.

mlissner (approved):
Looks great, thanks!

PR #4088, opened 2024-05-31
Adds new RECAP endpoint from Doctor

flooie:
Add Import code for starting RECAP into OPINIONs


cl/search/migrations/0031_alter_opinion_type_alter_opinioncluster_source_and_more.py
mlissner, line 1: If these migrations don't change the DB, they should have a _noop.py suffix (it's in the docs).

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,124 @@
...
+def import_opinions_from_recap(court=None, total_count=0):
+    """Import recap documents into opinion db
+
+    :param court: Court ID if any
+    :param total_count: The number of new opinions to add
+    :return: None
+    """
+    count = 0
+    if not court:
+        courts = Court.objects.filter(jurisdiction=Court.FEDERAL_DISTRICT)
```
mlissner, line 27: What about bankruptcy?

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,124 @@
...
+    else:
+        courts = Court.objects.filter(id=court)
```
mlissner, line 29: Using pk is better b/c `id` is a reserved word in Python:

```suggestion
        courts = Court.objects.filter(pk=court)
```

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,124 @@
...
+        cluster = (
+            OpinionCluster.objects.filter(docket__court=court)
+            .order_by("-date_filed")
+            .first()
+        )
+        if not cluster:
+            print("No clusters for this court")
+            logger.info(f"Court {court} has no opinion clusters")
+
+            continue
```
mlissner, line 41: Misplaced newline here, but can you explain why we skip courts with no cluster objects? Seems like we'd particularly want to add content to such courts, no?

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,124 @@
...
+            docket_entry__docket__date_filed__lte=cluster.date_filed,
+            is_available=True,
+            is_free_on_pacer=True,
+        ).order_by("docket_entry__docket__date_filed")
+
+        chunk_size = 100
+        paginator = Paginator(documents, chunk_size)
+        for page_number in paginator.page_range:
+            chunk = paginator.page(page_number).object_list
+            for recap_document in chunk:
```
mlissner, line 53: Can't you just do `documents.iterator()`?

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,124 @@
...
+                docket = Docket.objects.get(
+                    docket_entries__recap_documents__id=recap_document.id
+                )
+                if "-cv-" not in docket.docket_number:
+                    logger.info(f"Skipping non civil opinion")
+                    continue
+
+                ops = Opinion.objects.filter(sha1=recap_document.sha1)
+                if ops.count() > 0:
+                    print("Skipping previously imported opinion")
```
mlissner, line 63: Logger, not print, right? Also, I'd suggest logging the ID or something on each line.

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,124 @@
...
+                        case_name=docket.case_name,
+                        case_name_short=docket.case_name_short,
+                        docket=docket,
+                        date_filed=recap_document.docket_entry.date_filed,
+                        source=SOURCES.RECAP,
+                    )
+                    Opinion.objects.create(
+                        cluster=cluster,
+                        type=Opinion.TRIAL_COURT,
+                        plain_text=response.json()["content"],
```
mlissner, line 83: Do we need a check to make sure doctor worked before creating this object? Something like `response.raise_for_status()`?

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,124 @@
...
+        paginator = Paginator(documents, chunk_size)
+        for page_number in paginator.page_range:
+            chunk = paginator.page(page_number).object_list
+            for recap_document in chunk:
+                docket = Docket.objects.get(
+                    docket_entries__recap_documents__id=recap_document.id
+                )
+                if "-cv-" not in docket.docket_number:
+                    logger.info(f"Skipping non civil opinion")
+                    continue
```
mlissner, line 59: You can't trust the dashes in docket numbers. They can be various types because people are nuts. I'd just check for cv, and I'd do case normalization to:

```suggestion
                if "cv" not in docket.docket_number.lower():
                    logger.info(f"Skipping non civil opinion") <---Also include an ID here, right?
                    continue
```

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,124 @@
...
+        ).order_by("docket_entry__docket__date_filed")
+
+        chunk_size = 100
+        paginator = Paginator(documents, chunk_size)
+        for page_number in paginator.page_range:
+            chunk = paginator.page(page_number).object_list
+            for recap_document in chunk:
+                docket = Docket.objects.get(
+                    docket_entries__recap_documents__id=recap_document.id
+                )
```
mlissner, line 56: This is the same on the backend, but more Django-ish:

```suggestion
                docket = recap_document.docket_entry.docket
```

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,124 @@
...
+            print("No clusters for this court")
+            logger.info(f"Court {court} has no opinion clusters")
+
+            continue
+        documents = RECAPDocument.objects.filter(
+            docket_entry__docket__court=court,
+            docket_entry__docket__date_filed__lte=cluster.date_filed,
+            is_available=True,
+            is_free_on_pacer=True,
+        ).order_by("docket_entry__docket__date_filed")
```
mlissner, line 47: This ordering will kind of hurt the performance. Do you need it just to make multiple runs the same? If so, perhaps just order by `id`?

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,124 @@
...
+            .first()
+        )
+        if not cluster:
+            print("No clusters for this court")
+            logger.info(f"Court {court} has no opinion clusters")
+
+            continue
+        documents = RECAPDocument.objects.filter(
+            docket_entry__docket__court=court,
+            docket_entry__docket__date_filed__lte=cluster.date_filed,
```
mlissner, line 44: Do you mean `gte` or even `gt`? I think you're getting the latest cluster in the court up above, and then you want to get RECAP docs that are newer than that, right? 

And if you do `gte` instead of `gt`, do you run the risk of duplicates on the first date?

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,124 @@
...
+        courts = Court.objects.filter(id=court)
+    for court in courts:
+        logger.info(f"Importing RECAP documents for {court}")
+        cluster = (
+            OpinionCluster.objects.filter(docket__court=court)
+            .order_by("-date_filed")
+            .first()
+        )
+        if not cluster:
+            print("No clusters for this court")
```
mlissner, line 38: Print or logger?

mlissner (changes requested):
Cool. Looks like the right idea. I made a bunch of comments, but one thing that I didn't see is the eyecite check in the content?

cl/search/migrations/0031_alter_opinion_type_alter_opinioncluster_source_and_more.py
flooie: indeed ill fix that

cl/corpus_importer/management/commands/recap_into_opinions.py
flooie: We have one court, the District Court for District of Orleans that has no opinions.  I think I was trying to avoid issues from historical courts.  

https://en.wikipedia.org/wiki/United_States_District_Court_for_the_District_of_Orleans

cl/corpus_importer/management/commands/recap_into_opinions.py
flooie: yes you are right  its a better approach

cl/corpus_importer/management/commands/recap_into_opinions.py
flooie: yeah that makes sense to me

cl/corpus_importer/management/commands/recap_into_opinions.py
mlissner: In that case, it's better to be explicit and exclude it from your query. In a year that code will still be clear of its intentions.

cl/corpus_importer/management/commands/recap_into_opinions.py
flooie: I think we can exclude the filtering completely- as long as we make sure to exclude newly imported documents from recap above.

cl/corpus_importer/management/commands/recap_into_opinions.py
flooie: yes - sorry this was a holdover from testing

cl/corpus_importer/management/commands/recap_into_opinions.py
flooie: logger

flooie:
@mlissner -- I think I've incorporated your suggestions.  Please feel free to review it

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,123 @@
...
+        if not cluster:
+            logger.info(f"No clusters for this {court}")
+            continue
+
+        documents = RECAPDocument.objects.filter(
+            docket_entry__docket__court=court,
+            docket_entry__docket__date_filed__gt=cluster.date_filed,
+            is_available=True,
+            is_free_on_pacer=True,
+        )
```
mlissner, line 46: Small thing, but some ordering here is a good idea. which is why I suggested `id`. Postgres doesn't given any guarantees about the order of the content unless you explicitly request it, so  two runs could in theory have totally different ordering.

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,123 @@
...
+
+        for recap_document in documents.iterator():
+            docket = Docket.objects.get(
+                docket_entries__recap_documents__id=recap_document.id
+            )
```
mlissner, line 51: Looks like this isn't fixed yet:

```suggestion
            docket = recap_document.docket_entry.docket
```

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,123 @@
...
+        logger.info(f"Importing RECAP documents for {court}")
+        cluster = (
+            OpinionCluster.objects.filter(docket__court=court)
+            .exclude(source=SOURCES.RECAP)
+            .order_by("-date_filed")
+            .first()
+        )
+        if not cluster:
+            logger.info(f"No clusters for this {court}")
+            continue
```
mlissner, line 39: It would be better to exclude these courts explicitly so we know in the future which courts weren't included. 

Beyond that, in the 4-12 entmoot notes, we also said we'd exclude dcd? https://docs.google.com/document/d/1E0w_swX44qWers_1tPxu0hI9GcQtuBjc6RHksotTzw4/edit#heading=h.aysr7ee06hkb

mlissner (changes requested):
A couple little things that slipped through, but we still need to check for citations with eyecite, right?

cl/corpus_importer/management/commands/recap_into_opinions.py
flooie: sounds good

cl/corpus_importer/management/commands/recap_into_opinions.py
flooie: but it doesnt hurt to add it

cl/corpus_importer/management/commands/recap_into_opinions.py
flooie: I assume order here doesnt matter - but will add it because im sure it will help with the ingest

cl/corpus_importer/management/commands/recap_into_opinions.py
flooie: fixed

flooie:
@mlissner 🏓

flooie:
@mlissner tests are back to passing after the eyecite bump.  Thanks @quevon24 for your insight

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -0,0 +1,127 @@
...
+                continue
+
+            response = async_to_sync(microservice)(
+                service="recap-extract",
+                item=recap_document,
+                params={"strip_margin": True},
+            )
+
+            response.raise_for_status()
+            cites = eyecite.get_citations(response.json()["content"])
```
mlissner, line 69: I think you'll want to use `filter_out_non_case_law_citations`, and the HYPERSCAN_TOKENIZER, for speed.

mlissner (approved):
OK, last thing, then feel free to merge!

cl/corpus_importer/management/commands/recap_into_opinions.py
flooie: yep

PR #4097, opened 2024-06-05
Update recap-into-op with retry 

flooie:
Add retry functionality for recap-extract micro service call

### DIFF
--- a/cl/corpus_importer/management/commands/recap_into_opinions.py
+++ b/cl/corpus_importer/management/commands/recap_into_opinions.py
@@ -3,9 +3,16 @@ from asgiref.sync import async_to_sync
 from django.core.management import BaseCommand
 from django.db import transaction
 from eyecite.tokenizers import HyperscanTokenizer
+from httpx import (
+    HTTPStatusError,
+    NetworkError,
+    RemoteProtocolError,
+    TimeoutException,
+)
 
 from cl.citations.utils import filter_out_non_case_law_citations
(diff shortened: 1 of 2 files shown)


cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -80,7 +109,9 @@ def import_opinions_from_recap(court=None, total_count=0):
                 # Ex. 42\u2009U.S.C.\u2009§\u200912131 \u2009 is a small space
                 # fallback to regular citation match
                 logger.warning(
-                    f"Hyperscan failed for {recap_document}, trying w/o tokenizer"
+                    "Hyperscan failed for {}, trying w/o tokenizer".format(
+                        recap_document
+                    )
```
mlissner, line 114: I think this has the same problem as before, in that it runs the .format() command when the code is run even if we're not outputting at the warning level, say. I believe the right way to do this is: 

```suggestion
                    "Hyperscan failed for %s, trying w/o tokenizer", recap_document)
```

mlissner (approved):
Cool. Thanks for tweaking the logging. I don't think you quite fixed it, but it's little stuff, so merging.

cl/corpus_importer/management/commands/recap_into_opinions.py
flooie: thanks

PR #4102, opened 2024-06-07
feat(audio.transcribe): use OpenAI API whisper model for STT

grossir:
- audio.Audio model changes: rename stt_google_response; add stt_source
- Many files needed update due to renaming of field stt_google_response
- new model: audio.AudioTranscriptionMetadata to hold model outputs for future diarization use
- add openai dependency
- new command: audio.transcribe, for now limited to audio files with size lesser than 25MB
- added tests for new command
- clone_from_cl: add support to clone audio files, including local_mp3_path. Useful for testing new command


grossir:
To test it I cloned Audio objects from the DB and then ran the transcribe command
Take note to input the proper Token / Keys in the commands below
I couldn't test the API limits since I couldn't get celery to work in parallel in my env, I would appreciate help with that

I measured the time taken to process each audio, will upload shortly some stats about that, also some comments about the effect of passing the case name as prompt

```
docker exec -i cl-django /bin/bash -c "export CL_API_TOKEN=??? && python manage.py clone_from_cl --type search.Docket --id 7030987 65399068 3042549 16579230 18440194 3042546 7865747 59714553 66737158 66635300 34340198 16966475 16956184 18473600 18667139 15967694 15035687 16735537 7807052 65762153 61629320 16377479 5645183 16507096 63232589 4423234 4425129 15015420 2599920 2605660 4485061 2605676 2603258 2599043 50362451 66616963 3008243 4483514 60412928 16399764 65346134 2591620 2632113 6325729 14853452 62620096 3017240 2630886 4426352 17056727 2624713 6332200 66832257 18511014 4546458 2625724 3018661 18716515 60686336 67517872 65737134 6149118 61458935 66821309 65587316 63231064 6241286 59707905 59898414 6134046 7856378 7839101 7849853 7846788 7846982 7839044 3079798 60379863 8041876 7979513 6324293 16504494 4481089 66614929 2601788 2601777 16276382 6058117 15898711 2600464 2606068 16208168 6341153 16189345 16176412 59721823 3082987 3084417 6383719 3087987 16248094 3089930 3088803 16278959 3083965 6220754 17125335 2645137 3010678 15006775 8127709 4609876 65740042 --add-audio-files"
```

```
docker exec -i cl-django /bin/bash -c "export OPENAI_API_KEY=??? && export CL_API_TOKEN=f9587de9d3b298200f94c1352c148cb131f285a2 && python manage.py transcribe"
```

cl/audio/tasks.py
```diff
@@ -45,3 +51,55 @@ def upload_audio_to_ia(self, af_pk: int) -> None:
         af.save()
     else:
         increment_failure_count(af)
+
+
+@app.task
```
mlissner, line 56: We can do some automatic retries here.

cl/audio/tasks.py
```diff
@@ -45,3 +51,55 @@ def upload_audio_to_ia(self, af_pk: int) -> None:
...
+                language="en",
+                response_format="verbose_json",
+                timestamp_granularities=["word", "segment"],
+                prompt=audio.case_name,
+            )
+        except openai.RateLimitError:
+            # implement backoff strategy
+            # More about openai errors
+            # https://github.com/openai/openai-python?tab=readme-ov-file#handling-errors
+            pass
```
mlissner, line 85: Celery makes catching/retrying pretty easy. Look around in the code for a bunch of examples.

scripts/make_bulk_data.sh
```diff
@@ -125,7 +125,7 @@ oralarguments_fields='(
 	       id, date_created, date_modified, source, case_name_short,
 	       case_name, case_name_full, judges, sha1, download_url, local_path_mp3,
 	       local_path_original_file, filepath_ia, ia_upload_failure_count, duration,
-	       processing_complete, date_blocked, blocked, stt_status, stt_google_response,
+	       processing_complete, date_blocked, blocked, stt_status, stt_transcript,
```
mlissner, line 128: We should add the new field here too, right?

cl/audio/management/commands/transcribe.py
```diff
@@ -0,0 +1,97 @@
...
+            help="Model to use for extraction",
+        )
+        parser.add_argument(
+            "--rpm",
+            type=int,
+            default=50,
+            help="Requests Per Minute, a rate limit in the model API",
+        )
+        parser.add_argument(
+            "--retry",
```
mlissner, line 32: Maybe `--retry-failures` would be more clear?

cl/audio/management/commands/transcribe.py
```diff
@@ -0,0 +1,97 @@
...
+
+        valid_count = 0
+        for audio in Audio.objects.filter(query):
+            # Validate that audio can actually be processed
+            try:
+                # Following the previous comment about Audio.duration
+                # only check durations that may cause a problem
+                if audio.duration and audio.duration > 3500:
+                    size_mb = audio.local_path_mp3.size / 1_000_000
+                    if size_mb > 25:
```
mlissner, line 72: I think I wouldn't bother with all the duration stuff up above and would just do this. Maybe you can just do a check to see if the duration is border line? 

```
get the file
check the duration
if it's not obviously short:
    get the file length
    if too long
       warn; continue
do_processing
```

cl/audio/management/commands/transcribe.py
```diff
@@ -0,0 +1,97 @@
...
+            except FileNotFoundError:
+                # Triggered when local_path_mp3 does not exist
+                logger.warning(
+                    "Audio id %s has no local_path_mp3, needs reprocessing",
+                    audio.pk,
+                )
+                continue
+
+            valid_count += 1
+            transcribe_from_open_ai_api(audio.pk)
```
mlissner, line 88: If you want to do many at once, you have to send the task to celery:

```suggestion
            transcribe_from_open_ai_api.delay(audio.pk)
```

But be careful, that'll send them all to Celery almost instantly. 

A good trick to use in combination with this is the `throttle_task` decorator, which you can use to configure a specific throttling rate. (But beware, it's kind of a pain to remove tasks from the celery queue, so if you use `delay` with `throttle_task` you'll wind up with everything on the queue immediately, BUT slowly processing through. (You can always stop celery in dev by killing the celery pod, of course.)

mlissner (commented):
Some comments for you, Gianfranco, including one to fix your celery issues!

cl/audio/tasks.py
grossir: I found that we have 2 options to handle API errors (including the rate limiting). 
- using  `openai.OpenAI` client `max_retries`
- using celery

Perhaps you prefer one of them

With celery. Note that in this example I am missing some OpenAI API error types, but the general idea is there:
```
@app.task(bind=True, max_retries=5)
def transcribe_from_open_ai_api(audio_pk: int):
    with openai.OpenAI(max_retries=0) as client:
        try:
            transcript = client.audio.transcriptions.create(
                file=file,
                model="whisper-1",
                language="en",
                response_format="verbose_json",
                timestamp_granularities=["word", "segment"],
                prompt=audio.case_name,
            )
        # sleep at least one minute before retrying
        except openai.RateLimitError:
            self.retry(countdown=60*(1+self.request.retries))
        except (openai.APIConnectionError):
            self.retry(countdown=60)
        except openai.UnprocessableEntityError:
            audio.stt_status = Audio.STT_FAILED
            audio.save()
            logger.warning("UnprocessableEntityError for audio %s", audio_pk)
            return
            
         ....
```
------
With the python `openai` package. They handle [retries](https://github.com/openai/openai-python?tab=readme-ov-file#retries) for "Connection errors, 408 Request Timeout, 409 Conflict, 429 Rate Limit, and>=500 Internal errors". It also checks the explicit header `x-should-retry` from the server : 
https://github.com/openai/openai-python/blob/54a5911f5215148a0bdeb10e2bcfb84f635a75b9/src/openai/_base_client.py#L679-L712

```
@app.task
def transcribe_from_open_ai_api(audio_pk: int):
    audio = Audio.objects.get(pk=audio_pk)
    audio_file = audio.local_path_mp3
    file = (audio_file.name.split("/")[-1], audio_file.read(), "mp3")

    logger.info("Starting transcription for audio %s", audio_pk)

    with openai.OpenAI(max_retries=5) as client:
        try:
            transcript = client.audio.transcriptions.create(
                file=file,
                model="whisper-1",
                language="en",
                response_format="verbose_json",
                timestamp_granularities=["word", "segment"],
                prompt=audio.case_name,
            )
        except openai.UnprocessableEntityError:
            audio.stt_status = Audio.STT_FAILED
            audio.save()
            logger.warning("UnprocessableEntityError for audio %s", audio_pk)
            return
        except openai.OpenAIError as e:
            # Sending to Sentry errors that are not auto-retried
            # or that have failed many times
            capture_exception(e)
            return
```

The problem with the packaged retry is that at most it waits for 60 seconds, which is the lower limit of waiting that I expect us to need.

mlissner: Definitely better to use the celery retry. You don't want your celery workers sleeping because that would stop other work from proceeding.

grossir:
@mlissner I think this PR is ready for another review. 
I added tests for the most important parts, I refactored some code for easier testing and reading, and I tested this in my local environment with concurrent celery workers, it behaves good

----

I was reading our [wiki](https://github.com/freelawproject/courtlistener/wiki/Working-with-Celery-on-CourtListener) for Celery use, and to get concurrent workers I had to update `CELERY_WORKER_CONCURRENCY` in `cl/settings/third_party/celery.py` and use 
`docker exec -it cl-celery celery -A cl worker -c 10 -E`, 
with `-c 1` I got linear transcriptions.

Perhaps something to add to the wiki?

mlissner:
That'd be a great addition to the wiki, thanks @grossir. 

Did you see that we moved to tier 3? I think that means we can tweak the throttling here, right? 

The PR looks good to me, but since it's about to spend a ton of money and it's a one-shot thing, let's get a second review too. @ERosendo, can you please take a look today and merge if you like it too? 

Once merged, let's **NOT** run this until I say so. It's a big, expensive script, so let's be really careful with it.

ERosendo (changes requested):
The code looks great overall! I've left some comments suggesting some minor refactoring opportunities.

cl/audio/migrations/0008_audiotranscriptionmetadata_and_more.py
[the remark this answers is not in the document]
grossir: Thanks for the detailed example

> Create SQL File for customers
I think this part is undocumented in the wiki, I will add it!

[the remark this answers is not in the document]
ERosendo: > Thanks for the detailed example

No problem.

> I think this part is undocumented in the wiki, I will add it!

Information about customer data migration can be found in the [Migrating event tables and trigger section ](https://github.com/freelawproject/courtlistener/wiki/Database-migrations#migrating-event-tables-and-triggers) 😄

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -387,9 +390,12 @@ def clone_docket(
 
             if add_docket_entries:
                 clone_docket_entries(session, docket.pk)
+            print("before audio file")
```
ERosendo, line 393: Do we need this print statement, or did you add it for debugging purposes and accidentally include it in the commit?

I noticed the other print statements you added to notify users about adding audio files and also one to display the docket data. Those seem helpful.

grossir: Those were some debugging statements, sorry, will remove them

ERosendo (approved):

cl/scrapers/management/commands/clone_from_cl.py
ERosendo: Thanks @grossir

cl/audio/migrations/0008_audiotranscriptionmetadata_and_more.py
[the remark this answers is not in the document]
grossir: The wiki has all the reasoning why we would create the files, but I think the step by step you listed is better explained. Also, the wiki is missing the convention about adding the `"_customers.sql"` suffix to one of the files

------
 Kinda unrelated, but in the same wiki:
 > This isn't the worst crime, but sometimes Django can be pretty dumb. For example, if you convert a text field to be blank=True, that'll create a migration that sets DEFAULT='', followed immediately by DROP DEFAULT. That doesn't do anything except confuse things, so the first rule of good migrations is: "Migrations should do something."

I think creating a column with DEFAULT='' and then dropping the default,  is because the field is designed to have the NOT NULL constraint but no default, so by using that DEFAULT you populate it with an empty string on creation, and then you drop the DEFAULT.

Anyway, it is somewhat confusing that Django will initialize a `TextField` with an empty string, even if the developer doesn't declare it on instance creation, and even if it has no default. Took me a while to understand it

grossir:
@mlissner, about

> Did you see that we moved to tier 3? I think that means we can tweak the throttling here, right?

We can control it with the `--rpm` argument. `rpm` for "Requests per minute"

blancoramiro:
Hello @mlissner Looking at the migrations it looks like the way to go is the following. Only `audio_audio` is replicated but we probably want to keep everything consistent. 

Create `audio_audiotranscriptionmetadata` table on replicas:

```
ALTER TABLE "audio_audiotranscriptionmetadata" ADD COLUMN "audio_id" integer NOT NULL CONSTRAINT "audio_audiotranscrip_audio_id_22f57b06_fk_audio_aud" REFERENCES "audio_audio"("id") DEFERRABLE INITIALLY DEFERRED; SET CONSTRAINTS "audio_audiotranscrip_audio_id_22f57b06_fk_audio_aud" IMMEDIATE;
```

Add columns to replicas:

```
ALTER TABLE "audio_audio" ADD COLUMN "stt_source" smallint NULL;ALTER TABLE "audio_audio" ADD COLUMN "stt_transcript" text DEFAULT '' NOT NULL;ALTER TABLE "audio_audio" ALTER COLUMN "stt_transcript" DROP DEFAULT;

ALTER TABLE "audio_audioevent" ADD COLUMN "stt_source" smallint NULL;
ALTER TABLE "audio_audioevent" ADD COLUMN "stt_transcript" text DEFAULT '' NOT NULL;
```


Apply migrations on production with `makemigrations`:
```[ ] 0008_audiotranscriptionmetadata_and_more```


Drop the columns on the replicas:
```
ALTER TABLE "audio_audio" DROP COLUMN "stt_google_response" CASCADE;
ALTER TABLE "audio_audioevent" DROP COLUMN "stt_google_response" CASCADE;
```

Any comments and feedback is appreciated. Thank you!

cl/audio/migrations/0008_audiotranscriptionmetadata_and_more.py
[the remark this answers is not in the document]
mlissner: Definitely feel free to fix the wiki. Thanks for highlighting these tweaks.

mlissner:
@blancoramiro, that sounds good to me. Should work well, since there's not a ton of new data flowing in these migrations.

PR #4103, opened 2024-06-10
fix(docs): Update docket alert example to improve webhook documentation

ERosendo:
This PR updates the docket alert example on the webhook documentation page. 

I noticed a slight difference in the endpoint the example is using to create docket alerts. The correct endpoint is `/api/rest/v3/docket-alerts/` instead of `/api/rest/v3/docket-alert/`. Additionally, this endpoint uses the field `docket` instead of `docket_id` to specify the docket you want to subscribe to.


albertisfu (approved):
Good catch! LGTM

mlissner:
Yowza. Thanks for fixing this. We should actually use the `{% url %}` template thing to avoid this, honestly. I'm using it exclusively in the documentation updates I'm working on.

PR #4108, opened 2024-06-12
feat(corpus_importer): Adds query to create manifest file for OA

ERosendo:
This PR introduces an enhancement to the `make_aws_manifest_files` command. It now enables the creation of manifest files specifically designed for importing all OA records.

Here's an example:

#### 1. Import OA records using the defaults:

```bash
python manage.py make_aws_manifest_files --record-type oa --bucket-name dev-bulk-customer-data
```

The previous commands will use the prod database to get the records. Manifest files will be limited to 1 million records each and each Lambda invocation will process a batch of 100 records at a time


mlissner (approved):
Looks good to me. @grossir  are there any where clauses that would make sense? 

Feel free to merge when you both are ready.

ERosendo:
@grossir I've updated the query to exclude records that weren't fixed by your scripts.

Let me know what you think.

grossir:
Looks good @ERosendo . This will cover the bad data cases we know of:
- no `local_file_mp3` 
- local_file_mp3 exists but is corrupted (very short `duration` and/or bad `download_url`)

ERosendo:
let's :shipit:

PR #4113, opened 2024-06-14
feat(audio.transcribe): add optional limit argument

grossir:
By default, the command tries to transcribe all audio_audio rows With --limit LIMIT, it will only try to transcribe LIMIT audio files


mlissner (approved):

mlissner:
Nice. Set for auto-merge.

PR #4117, opened 2024-06-14
fix(audio.transcribe): mark failed when BadRequestError is raised

grossir:
Set `audio.stt_status = Audio.STT_FAILED` on BadRequestError. The error has a message like
'The audio file could not be decoded or its format is not supported' So as to avoid naive retries of bad audio files


ERosendo:
LGTM :shipit:

PR #4123, opened 2024-06-19
fix(recap->op): fix count trigger

flooie:
Update logging to verify celery process started for slow db queries
Improve court filtering code to allow for skip-until restarts for courts
Add error logging for courts without clusters

### DIFF
--- a/cl/corpus_importer/management/commands/recap_into_opinions.py
+++ b/cl/corpus_importer/management/commands/recap_into_opinions.py
@@ -51,8 +51,8 @@ def extract_recap_document(rd: RECAPDocument) -> Response:
 
 
 def import_opinions_from_recap(
-    court_str: Optional[str] = None,
-    skip_until: Optional[str] = None,
+    court_str: str | None  = None,
+    skip_until: str | None = None,
     total_count: int = 0,
     queue: str = "batch1",
     db_connection: str = "default",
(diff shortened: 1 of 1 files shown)


cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -1,5 +1,9 @@
+from typing import Optional
```
ERosendo, line 1: In Python 3.10 and later, you can express optional types more concisely using union syntax (`type | None`) instead of relying on the `typing.Optional `import.

For example, instead of:

```python
from typing import Optional

def function_1(value: Optional[int]) -> str:
    # ...
```

You can write:
```python
def get_value(value: int | None) -> str:
  # ...
```

cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -47,25 +51,35 @@ def extract_recap_document(rd: RECAPDocument) -> Response:
 
 
 def import_opinions_from_recap(
-    court=None,
-    total_count=0,
-    queue="batch1",
+    court_str: Optional[str] = None,
+    skip_until: Optional[str] = None,
```
ERosendo, line 55:
```suggestion
    court_str: str | None  = None,
    skip_until: str | None = None,
```

cl/corpus_importer/management/commands/recap_into_opinions.py
flooie: yes thank you - I thought I included that from your branch

ERosendo:
:shipit:

ERosendo (approved):

PR #4126, opened 2024-06-19
fix(audio.transcribe): tweak retry behavior

grossir:
- Add "dont-retry-task" argument to command
- Handle retries in except clause, no longer using auto_retry_for
- add logger.info call for number of retries
- add logger.error call when audio with status STT_COMPLETE as enqueued for transcription
- now sending all exceptions to Sentry

In general, these changes allow to better control and monitor the requests, since we were seeing more requests than expected


ERosendo (approved):
Looks good :shipit:

PR #4129, opened 2024-06-21
fix(audio.transcribe): add new stt_status choices

grossir:
- Add status  STT_HALLUCINATION - add function transcription_was_hallucinated based on relationship of unique words and audio duration - add tests for the function
- Add status STT_FILE_TOO_BIG: - add test for updating the status on audio_can_be_processed_by_open_ai_api
- Add status STT_NO_FILE - add test for status update on audio_can_be_processed_by_open_ai_api
- Add migrations for the changes
- Add a return on handle_open_ai_transcriptions task to avoid UnboundLocalError


grossir:
The logic for `transcription_was_hallucinated` can be explained with a graph. Those points below the red line would be tagged as hallucinations, and they indeed are. 

A line / region boundary approach is somewhat equivalent to a series of if clauses that take into account both duration and unique words, but shorter to define. There are some false negatives. I could make the tolerance smaller, but risk having false positives. I think the decision will depend on what is preferable: to have a hallucination tagged as an OK transcript; or to have an OK transcript tagged as a hallucination. 

image

ERosendo (approved):

PR #4137, opened 2024-06-26
fix(audio.transcribe): add argument to retry different stt_status

grossir:
- replace --retry-failures with --status-to-process: allows to retry different types of failures
- handle STT_HALLUCINATION by not sending the case name as prompt. This should work for a good part of the hallucinations
- small fix to clone_from_cl: there was a bug when copying audios with stt_transcript

### DIFF
--- a/cl/audio/management/commands/transcribe.py
+++ b/cl/audio/management/commands/transcribe.py
@@ -1,8 +1,6 @@
 import argparse
 import time
 
-from django.db.models import Q
-
 from cl.audio.models import Audio
 from cl.audio.tasks import transcribe_from_open_ai_api
 from cl.lib.command_utils import VerboseCommand, logger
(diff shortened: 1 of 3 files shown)


grossir:
Tested this by 

- cloning 2 records with stt_status = STT_HALLUCINATION
       - records were retried without the promp, to good results
       
- 2 records with stt_status STT_NO_FILE
       - 1 now had the file, after running a correction script, got transcripted successfully
       - the other did not, wasn't even enqueued for transcription
       
```
python manage.py clone_from_cl --type search.Docket --id 68516864 67512327 67634850 68826691 --add-audio-files"
```

cl/audio/management/commands/transcribe.py
```diff
@@ -61,12 +61,12 @@ def handle_open_ai_transcriptions(options) -> None:
...
     requests_per_minute = options["rpm"]
 
-    if options.get("retry_failures"):
-        query = Q(stt_status=Audio.STT_FAILED)
-    else:
-        query = Q(stt_status=Audio.STT_NEEDED)
+    status_code = options.get("status_to_process")
+    for code, descr in Audio.STT_STATUSES:
+        if code == status_code:
+            logger.info("Querying Audio.stt_status = %s. %s", code, descr)
```
mlissner, line 68: This feels like a lot of code just for a better `desc` in the log line.

mlissner (approved):
Looks good, setting to merge with one little comment.

Thank you!

cl/audio/management/commands/transcribe.py
grossir: It made sense to work with the integer values, because they are exposed in the Audio class and straightforward to validate, but it makes the user experience worse. Not the best looking, but we can alleviate that a little with these lines

PR #4141, opened 2024-06-27
feat(audio): Add logic to transcribe big files

ERosendo:
This PR registers a new service to downsize audio files and updates the transcribe command to remove logic that previously prevented the processing of large files. 

The PR also introduces a new parameter named `max-audio-duration` to the transcription process. During testing, I noticed that the API could successfully transcribe **compressed** audio files under 2 hours (7200 seconds) in length. However, attempts to transcribe files exceeding this duration resulted in `InternalServerError` exceptions, even for files well below the 25MB file size limit imposed by the API. The `max-audio-duration` parameter provides a mechanism to filter out audio files based on their length.

I Identified **1,778 records** with status 4 that are under 2 hours long. By downsizing those files before sending them into the API, we have a strong chance of successfully getting transcription for all of them.

Here's an example demonstrating how to use the new `max-audio-duration` parameter:

```bash
manage.py transcribe --limit 300 --rpm 100 --queue batch3 --max-audio-duration 7200 --status-to-process 4
```

### DIFF
--- a/cl/audio/management/commands/transcribe.py
+++ b/cl/audio/management/commands/transcribe.py
@@ -67,6 +67,8 @@ def handle_open_ai_transcriptions(options) -> None:
             logger.info("Querying Audio.stt_status = %s. %s", code, descr)
 
     queryset = Audio.objects.filter(stt_status=status_code)
+    if options["max_audio_duration"]:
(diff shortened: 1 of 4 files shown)


mlissner (approved):
Looks good enough. Let's give it a shot!

PR #4144, opened 2024-06-27
Update scrape pacer free documents command

quevon24:
This is an update of the command which will allow us to sweep pacer to obtain the missing free documents, to achieve this new cronjobs will be added to execute the sweep weekly, monthly, quarterly in order to be able to obtain any document that has been added on dates after your filling date.

```
manage.py scrape_pacer_free_opinions --action=do-weekly
manage.py scrape_pacer_free_opinions --action=do-monthly
manage.py scrape_pacer_free_opinions --action=do-quarterly
```

In addition to this we have another action with which we seek to obtain older documents, this will be executed from 1950-05-12 until the day the command was executed.

`manage.py scrape_pacer_free_opinions --action=do-all`

In addition to this, new options were added to indicate the start and end date of the query, in addition to being able to indicate the court that you want to scrape using the courtlistener court_id, this is very useful in case we identify a gap.

`manage.py scrape_pacer_free_opinions --action=get-report-results --courts txsd --date-start=06/21/2024 --date-end=07/09/2024
`
The scraped html file obtained from the command is also saved when it is executed daily(as has been done until now), we require this juriscraper PR for this to give it a nice filename: https://github.com/freelawproject/juriscraper/pull/1075.

Here my only question would be if we should also save the html files when the sweep is executed?


flooie:
This is still a draft

quevon24:
> This is still a draft

Yes, I'm still reviewing the problem, I'm going to update the issue with my findings

flooie:
@quevon24 looking forward to this - thanks for the hard work

mlissner:
> Here my only question would be if we should also save the html files when the sweep is executed?

Yes. We should save all HTML files going forward.

Question: Is it easier just to have start and end date parameters and just let our cronjob pass the correct dates? That'd simplify a bit, wouldn't it?

quevon24:
> Question: Is it easier just to have start and end date parameters and just let our cronjob pass the correct dates? That'd simplify a bit, wouldn't it?

mmm, more than easy I think it is better, by default to do the scraping from the beginning, the dates are automatically generated in the command, from 1950 to today (or from a specified start date to today), having the start date option for example would allow us to resume the command in some specific point in case something fails (getting something unexpected or perhaps the pod could die?)

and since we don't really know if there is any pattern to know when documents are marked as free opinions, being able to decide in which date range to execute the command can save us from having to reprocess a larger date range, 
assuming we identify the gap some other way

mlissner:
Right, so I'm saying the only thing you need is the date parameters, not the do-weekly, do-monthly, or do-quarterly parameters, nor the do-all.

quevon24:
> Right, so I'm saying the only thing you need is the date parameters, not the do-weekly, do-monthly, or do-quarterly parameters, nor the do-all.

True, although how would you pass the dates directly to the command?  using a shell script?

mlissner:
We can write the shell script to do the date math pretty easily. I think I'd just do three scripts:

1. Just set date-start to 1950 and let it rip.
2. Run daily: Set it to start and end seven days ago. Something like:

    ```
    manage.py scrape_whatever \
      --date-start `date -d "7 days ago" +"%Y-%m-%d"`
      --date-end `date -d "7 days ago" +"%Y-%m-%d"`
    ````

3. Run daily: Set it to start and end today: 

    ```
    manage.py scrape_whatever \
      --date-start `date +"%Y-%m-%d"`
      --date-end `date +"%Y-%m-%d"`
    ````

That'd work nicely, right?

cl/corpus_importer/tasks.py
```diff
@@ -867,18 +888,12 @@ def upload_to_ia(
...
-    status: int, court_id: str, d: date
-) -> Optional[int]:
-    court_id = map_pacer_to_cl_id(court_id)
+def mark_court_done_on_date(log_id: int, status: int) -> Optional[int]:
     try:
-        doc_log = PACERFreeDocumentLog.objects.filter(
-            status=PACERFreeDocumentLog.SCRAPE_IN_PROGRESS, court_id=court_id
-        ).latest("date_queried")
+        doc_log = PACERFreeDocumentLog.objects.get(pk=log_id)
     except PACERFreeDocumentLog.DoesNotExist:
```
flooie, line 894: how are we marking courts done without identifying the court used

quevon24: the PACERFreeDocumentLog model contains the court that was scraped or is being scraped, this object is created in an early process and we only care about updating it if it is in progress, successful or failed.

We create it in mark_court_in_progress function

cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
```diff
@@ -299,9 +388,43 @@ def add_arguments(self, parser: argparse.ArgumentParser) -> None:
...
+                "Error: date-end must be specified to use date-start option."
+            )
+            return
+
+        if options["date_start"] and options["date_end"]:
+            if options["date_start"] > options["date_end"]:  # type: ignore
+                logger.info(
+                    "Error: date-end must be greater or equal than date-start option."
+                )
+                return
```
flooie, line 427: I think you are missing a scenario 

        date_start = options.get("date_start", None)
        date_end = options.get("date_end", None)

        if date_start:
            if not date_end or date_end < date_start:
                logger.error("Check dates: both dates must be present and correct")
                return
        elif date_end:
            logger.error("Must have start date")
            return

maybe this would work

cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
```diff
@@ -204,7 +282,18 @@ def get_pdfs(options: OptionsType) -> None:
     q = cast(str, options["queue"])
     index = options["index"]
     cnt = CaseNameTweaker()
-    rows = PACERFreeDocumentRow.objects.filter(error_msg="").only("pk")
+    rows = PACERFreeDocumentRow.objects.filter(error_msg="")
```
flooie, line 285: whats with the empty error msg here?  I see it's the same as before but its weird right?

flooie:
this is good but needs a refactor to remove duplicate code.  Lots of cleanup and simplification will also make this more readable.  We met over slack to identify some of the areas to do that.  looking forward to getting this code thru soon.

flooie (changes requested):
A good refactor of the get_and_save_free_document_report and the argparse code should help make this more readable and reduce the CRUD.  Also we can reduce the number of db queries for every run

PR #4148, opened 2024-06-27
feat(deploy): Sync env vars before deployment

blancoramiro:
This new step will apply any update to the Courtlistener environment variables (secret or not) before launching any pod or running any restart.

  Step will first add an annotation which will trigger an event that the external-secrets operator will pick up and start a sync with the AWS secret that holds all the environment variables, and then it'll wait for the secret to have a synced status.

  Wait has a limit of 30 seconds which if it's reach, the command will exit with an error status.


mlissner:
Great! Can you do this for bots.law too, please?

Issue #4143, opened 2024-06-27
Missing pacer free documents


quevon24:
Bill recently realized that some free pacer documents are missing (opinions) even though the scraper continues to run every day.

I've been digging to see why this is happening.

Manually inspecting in "mad" i  found these missing documents:

5/18/2024 – 6/17/2024: 7 documents missing
3/18/2024 – 4/17/2024: 5 documents missing
2/18/2024 – 3/17/2024: ~11 documents missing
2/18/2024 – 3/17/2024: ~22 documents missing

I have been testing the courtlistener command and directly in juriscraper and it works perfectly, it does not lose documents.

I already checked a database dump and i couldn't find any failed rows related to the results above and the only way a PACERFreeDocumentRow row can be deleted from is if the entry has been created in courtlistener so I thought it might be on courtlistener (although it would be strange if recap didn't mark it) but I couldn't find them in the system.

Also verified that the command also merges correctly the docket entry if the docket already exist.

In my local environment I was seeing this exception MaxRetriesExceededError which is related to the celery task, i had to rerun the command to get all the documents into my local instance. I've already been investigating sentry but I couldn't find any logs related to this, so maybe the problem is not there.

The only thing I can think of is that those documents for some reason were added after the scraper was run but I didn't see any timestamp or similar to be able to verify this.

I'm still thinking about why this happened, any suggestions are welcome

quevon24:
> What would happen if the court blocked us for a day or for part of a night?

It is a good suggestion, I already checked what you mentioned, assuming that we are blocked I would expect to receive a 403 code or some other error code, we use the raise_for_status() function which would throw us an exception when trying to generate the report, I checked in sentry but I did not see any error related to this, then it is unlikely to happen

mlissner:
Have you seen the messages in Sentry about "Unable to get free documents between X and Y" or similar? It's an error we get pretty regularly?

quevon24:
> Have you seen the messages in Sentry about "Unable to get free documents between X and Y" or similar? It's an error we get pretty regularly?

Yes, I had already seen the error, Ramiro already shared the logs with me to see if the problem is there

quevon24:
[...]

PR #4158, opened 2024-07-01
Add exc_info param to pacer free document command

quevon24:
This small PR is related to #4143.

It seems that some of the problems are only happening on the server since I have not been able to replicate them locally, the problem is that the logged messages have not been enough to be able to infer what the error is since only something like:

```
2024-06-20 02:14:22.154	ERROR Failed to get free document references for wawd between 2024-06-15 and 2024-06-20 due to PACER 6.3 bug..
2024-06-20 02:14:10.649	INFO Attempting to get latest document references for wawd between 2024-06-15 and 2024-06-20
```

As you can see the message does not say much and there is no traceback available.

The change only passes the exc_info parameter to obtain the traceback but keeping the message.


flooie:
@quevon24 looks good to me

PR #4162, opened 2024-07-02
feat(recap->op): Enable bankruptcy opinions

flooie:
Add tweak to importer to enable FB jurisdictions.


cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -137,6 +139,14 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def add_arguments(self, parser):
+        parser.add_argument(
+            "--jurisdiction",
+            help="Select Jurisdiction",
+            type=str,
+            default="FD",
```
ERosendo, line 146: There's room for improvement in the `jurisdiction` argument for the command. We should tighten things up by restricting the allowed values to a predefined list. Luckily, we already have a list of valid jurisdictions available in `Court.JURISDICTIONS`.

```suggestion
            default=Court.FEDERAL_DISTRICT,
            choices=[c[0] for c in Court.JURISDICTIONS],
```

This change will not only strengthen the code's logic but also provide clearer guidance to users when they run the command with the `--help` flag.

```bash
options:
  -h, --help            show this help message and exit
  --jurisdiction {F,FD,FB,FBP,FS,S,SA,ST,SS,TRS,TRA,TRT,TRX,TS,TA,TT,TSP,SAG,MA,MT,C,I,T}
  --court COURT         Specific court ID to import - skip if want to ingest everything
```

ERosendo (approved):

flooie:
thanks @ERosendo - I'm making a counter proposal because I don't want to have more jurisdiction choices than makes sense.

ERosendo:
@flooie LGTM :shipit:

PR #4184, opened 2024-07-09
Add new icons

mlissner:
This adds favicons per the instructions here: 

https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs

The idea is to simplify what we had while still being generally compliant. A few things:

 - New favicon design. The scales even when simplified were too complex, so I just went with a single node. The red works in light or dark backgrounds.
 - As the sizes of the images goes up, I added more complexity to the icons since we'd have more space.
 - The manifest file requires that you figure out the hash of the file in advance, since the file cannot be generated by Django, so I did that.

Hopefully this will be good enough to satisfy most or all use cases.


flooie (approved):
Looks good to me

mlissner:
Let's see how it goes!

PR #4186, opened 2024-07-09
fix(imgs): Correct manifest links

mlissner:


mlissner:
Not sure why the hashes are wrong, but it's OK. These are from S3 and should be correct.

PR #4187, opened 2024-07-09
fix(imgs): New favicons

mlissner:


flooie (approved):
I like this a lot more!  I mentioned in the other one that it appears to have more lip in the smaller image but perhaps thats not real.

mlissner:
It could be not real, but it could still be the way your eye sees it. Let's keep an eye on it and see if we notice it going forward. (I don't see it right now.)

Issue #4188, opened 2024-07-10
Enable juriscraper loggers 


grossir:
Currently juriscraper `logger` calls do not work, including the `logger` calls from DupChecker, which imports the juriscraper logger https://github.com/freelawproject/courtlistener/blob/9f1c2c4dda03ffea1d637d6b1bf975aed7ca551a/cl/scrapers/DupChecker.py#L1

This makes filling gaps and accouting for duplicates more difficult than it should be. Probably would be useful for other scraper debugging tasks

The solution can be as simple as enabling a console logger for juriscraper
`"juriscraper": {"handlers": ["console"], "propagate": True},`
https://github.com/freelawproject/courtlistener/blob/9f1c2c4dda03ffea1d637d6b1bf975aed7ca551a/cl/settings/project/logging.py#L65

grossir:
The updated logging config from the PR  allow us to see the reason why an opinion was skipped
```
INFO 2024-07-09 18:15:07,753 (/opt/courtlistener/cl/scrapers/DupChecker.py press_on): "Duplicate found on date: 2024-07-09, with lookup value: 517fe87a981dbf171397afc9d9bc0b70d293cb7d"

DEBUG 2024-07-09 18:15:07,753 (/opt/courtlistener/cl/scrapers/management/commands/cl_scrape_opinions.py scrape_court): "Skipping opinion."

```

PR #4189, opened 2024-07-10
feat(logging): enable juriscraper loggers on console handler

grossir:
Solves #4188

All juriscraper files use a logger called "Logger", whose specific name must be added to the `loggers` dict for it to be used.

Adding "juriscraper" module level logger in case other loggers are defined


cl/settings/project/logging.py
```diff
@@ -78,6 +78,16 @@ def skip_unreadable_post(record):
...
+        "juriscraper": {
+            "handlers": ["console"],
+            "propagate": True,
+            "level": "DEBUG",
+        },
+        "Logger": {
+            "handlers": ["console"],
+            "propagate": True,
+            "level": "DEBUG",
+        },
```
mlissner, line 90: I'm pretty bad at loggers, but I think my understanding is that by putting the `juriscraper` logger here, CL will log out juriscraper logs? 

What's the `Logger` logger for? 

Do we need DEBUG for both of these? That feels like it could (and should) be noisy. Better to stick with INFO?

mlissner (commented):
Thanks for taking the initiative here to fix things. I'm really bad at loggers, so I'm asking a few basic questions. (Whenever I have to mess with the logging configuration, it just feels like a mystery box.)

cl/settings/project/logging.py
grossir: The `Logger` is a named logger from `make_default_logger` in `juriscraper/lib/log_utils.py`. It is the only logger used in all of juriscraper, so we can just delete the "juriscraper" logger entry.
https://github.com/freelawproject/juriscraper/blob/aff4c45150d9649871aa9aed5889fd3c074ba776/juriscraper/lib/log_tools.py#L16-L23

[Usually](https://docs.python.org/3/library/logging.html#logger-objects) loggers are defined using `logger = logging.getLogger(__name__)`  which gives it a name with the full module path, and allows for granularity in selecting what we log, but we don't do that in juriscraper

About it logging everything, since the only handler defined for it is the "console" handler, defined as a          `class": "logging.StreamHandler"`, I expect it only to appear in the console where the command is being executed, which is the use case we want to cover, to have all the logs when we are debugging by running something manually. Via the same reasoning, we set the log level to "DEBUG"

mlissner: Wouldn't console logging also add this to all of our server logs? I think they just log to the console too?

grossir:
@mlissner can you check this again?

I pushed changes to "activate" the juriscraper logger only on VerboseCommand AND only when the --verbosity is manually passed to a high level. So, the default behaviour will be as has been, but the juriscraper loggers can be activated manually

An example without --verbosity
```
courtlistener/docker/courtlistener (enable_juriscraper_loggers)$ docker exec -it cl-django python /opt/courtlistener/manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.state.alaska --backscrape-start=08/30/2023 --backscrape-end=09/03/2023
INFO 2024-07-22 09:55:22,783 (/opt/courtlistener/cl/scrapers/management/commands/cl_scrape_opinions.py handle): "Starting up the scraper."
INFO 2024-07-22 09:55:31,193 (/opt/courtlistener/cl/scrapers/management/commands/cl_scrape_opinions.py handle): "The scraper has stopped."
```

An example with --verbosity
```
/courtlistener/docker/courtlistener (enable_juriscraper_loggers)$ docker exec -it cl-django python /opt/courtlistener/manage.py cl_back_scrape_opinions --courts juriscraper.opinions.united_states.state.alaska --backscrape-start=08/30/2023 --backscrape-end=09/03/2023 --verbosity 3
INFO 2024-07-22 09:55:41,551 (/opt/courtlistener/cl/scrapers/management/commands/cl_scrape_opinions.py handle): "Starting up the scraper."
INFO 2024-07-22 09:55:41,570 (/opt/courtlistener/cl/scrapers/management/commands/cl_back_scrape_opinions.py parse_and_scrape_site): "Using court_str: "alaska""
INFO 2024-07-22 09:55:41,571 (/usr/local/lib/python3.12/site-packages/juriscraper/opinions/united_states/state/alaska.py _download_backwards): "Backscraping for range 2023-08-30 2023-09-03"
INFO 2024-07-22 09:55:41,571 (/usr/local/lib/python3.12/site-packages/juriscraper/AbstractSite.py _download): "Now downloading case page at: https://appellate-records.courts.alaska.gov/CMSPublic/Home/Opinions?isCOA=False"
INFO 2024-07-22 09:55:49,793 (/usr/local/lib/python3.12/site-packages/juriscraper/opinions/united_states/state/alaska.py _process_html): "HTML was not downloaded from source page. Should retry"
WARNING 2024-07-22 09:55:49,795 (/usr/local/lib/python3.12/site-packages/juriscraper/AbstractSite.py _check_sanity): "juriscraper.opinions.united_states.state.alaska: Returned with zero items."
INFO 2024-07-22 09:55:49,795 (/usr/local/lib/python3.12/site-packages/juriscraper/AbstractSite.py _check_sanity): "juriscraper.opinions.united_states.state.alaska: Successfully found 0 items."
DEBUG 2024-07-22 09:55:49,807 (/opt/courtlistener/cl/scrapers/management/commands/cl_scrape_opinions.py scrape_court): "#0 opinions found."
DEBUG 2024-07-22 09:55:49,807 (/opt/courtlistener/cl/scrapers/management/commands/cl_scrape_opinions.py scrape_court): "juriscraper.opinions.united_states.state.alaska: Successfully crawled 0/0 opinions."
INFO 2024-07-22 09:55:49,808 (/opt/courtlistener/cl/scrapers/management/commands/cl_scrape_opinions.py handle): "The scraper has stopped."

```

cl/lib/command_utils.py
```diff
@@ -17,6 +17,10 @@ def handle(self, *args, **options):
             logger.setLevel(logging.INFO)
         elif verbosity > 1:
             logger.setLevel(logging.DEBUG)
+            # This will make juriscraper's logger, called "Logger"
+            # accept most logger calls.
+            juriscraper_logger = logging.getLogger("Logger")
```
mlissner, line 22: Should we rename this with something better?

grossir: The naming convention for loggers it to define them like this `logging.getLogger(__name__)`, which gives them the module name, but the change is done on juriscraper [first](https://github.com/freelawproject/juriscraper/blob/main/juriscraper/lib/log_tools.py), and then here

I think it shouldn't have any side effect, so let's update it

grossir: This is corrected now, including the juriscraper side, please check again @mlissner

mlissner:
Cool. Thanks for working to get this right!

PR #4206, opened 2024-07-16
feat(auth):Uses Ada to redirect users upon login

### LINKED ISSUE #3933: Use Ada to redirect users upon login
A while back, I removed redirect-on-login functionality from CourtListener because it was exposing a vulnerability in Python. I reported the Python vulnerability, and it was fixed:

https://bugs.python.org/issue43882

But it was kind of a mess, honestly. The volunteer that fixed it didn't check if the WHATWG test suite passed properly, and just focused on the small vuln that I reported. I'm pretty sure `urlparse` is still parsing things incorrectly and probably still vulnerable, so I didn't reinstate our redirect-on-login functionality even once Python was "fixed."

Now, when people login, they just go to the homepage, and bots.law inherited this functionality. Lame.

There is, however, hope. A new library called Ada actually passes the WHATWG test suite, is fast, has Python bindings, and seems really great: 

https://github.com/ada-url/ada?tab=readme-ov-file

Maybe we should use it and make it so redirect-on-login works again. 

But before we do that, I wanted to see if Django would integrate Ada directly, so I opened an discussion about that here:

https://groups.google.com/g/django-developers/c/Zp9mKpBn4u8

If we get a :+1: from them, we should just add it to Django, wait for that, and then fix our redirect system.

If it gets a :-1:, we should just use Ada ourselves to fix our redirect system.

ERosendo:
This PR fixes #3933 by adding Ada as a dependency and creating a custom LoginView class that validates and sanitizes the redirect URL after a successful login.

Key changes in this PR:

- Adds https://github.com/ada-url/ada-python as a dependency

- Adds a new helper method to parse and validate the redirect **URLs** using Ada

- Introduces a custom `LoginView` class called `SafeRedirectLoginView` that inherits from Django's built-in LoginView but adds an extra layer of security by ensuring the redirect URL submitted by the login form is safe.


cl/opinion_page/views.py
```diff
@@ -292,8 +292,13 @@ async def redirect_og_lookup(request: HttpRequest) -> HttpResponse:
...
     """
-    file_path = get_redirect_or_404(request, "file_path")
-
+    # Since implementing Ada, the `get_redirect_or_404` method now returns a
+    # valid path for redirection, always prefixed with a slash (/). However,
+    # FileField values cannot have a leading slash because it breaks how they
+    # interact with MEDIA_ROOT.
+    # To ensure compatibility, we're striping the leading slash from the
+    # returned redirect path before using it to retrieve the RD record.
+    file_path = get_redirect_or_404(request, "file_path").lstrip("/")
```
mlissner, line 301: If we do the changes above, this becomes:

```suggestion
    file_path = get_redirect_or_abort(request, "file_path", throw_404=True).lstrip("/")
```

I think that's OK?

cl/users/views.py
```diff
@@ -828,3 +833,37 @@ class RateLimitedPasswordResetView(PasswordResetView):
...
+        if not redirect_to:
+            return settings.LOGIN_REDIRECT_URL
+
+        try:
+            cleaned_url = parse_url_with_ada(redirect_to)
+        except ValueError:
+            return settings.LOGIN_REDIRECT_URL
+
+        is_safe = is_safe_url(redirect_to, self.request)
+        return cleaned_url if is_safe else settings.LOGIN_REDIRECT_URL
```
mlissner, line 869: With my changes, this becomes, I think:

```suggestion
        return get_redirect_or_abort(self.request, self.redirect_field_name)
```

I think you can even slim down your comments, since they should be explained adequately by the one function you're running.

pyproject.toml
```diff
@@ -114,6 +114,7 @@ tiktoken = "^0.6.0"
 hyperscan = "^0.7.7"
 openai = "^1.31.1"
 seal-rookery = "^2.2.3"
+ada-url = "^1.15.0"
```
mlissner, line 117: Can you do alphabetical order, please?

mlissner (approved):
I think a small refactor will help a lot here. Let me know what you think. If you like the changes, this looks good to merge to me.

cl/assets/templates/base.html
```diff
@@ -196,8 +196,15 @@ <h1>You did not supply the "private" variable to your template.
...
-                       tabindex="205">Sign in / Register</a></li>
+                {% if request.path != "/sign-out/" %}
+                  <li>
+                    <a href="{% url "sign-in" %}?next={{request.path}}?{{get_string|urlencode}}{% if results %}page={{results.number}}{% endif %}"
+                       tabindex="204">Sign in / Register</a>
+                  </li>
+                {% else %}
+                    <li><a href="{% url "sign-in" %}"
+                           tabindex="205">Sign in / Register</a></li>
+                {% endif %}
```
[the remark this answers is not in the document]
ERosendo: You're right! I'll update the template

cl/lib/url_utils.py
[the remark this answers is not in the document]
ERosendo: love this refactor!!

ERosendo:
@mlissner I've applied your suggestions. The PR is ready for another review.

Thanks.

mlissner (approved):
Great.

Issue #4193, opened 2024-07-10
`extract_from_text` of scrapers in `united_states_backscrapers` folder is never used


grossir:
`Site.extract_from_text` is used  in the `extract_doc_content` task https://github.com/freelawproject/courtlistener/blob/481c24f05d0c895c6067059ff090b1cb065c6ca7/cl/scrapers/tasks.py#L177

To get the Site object, it uses a lookup by court_id over all juriscraper modules. 
https://github.com/freelawproject/courtlistener/blob/481c24f05d0c895c6067059ff090b1cb065c6ca7/cl/lib/juriscraper_utils.py#L7-L32

The ordering of the list makes it so it will never pick up a `united_states_backscrapers` Site object, since they always come after the regular scraper. 

Saw this while working on freelawproject/juriscraper#1001 because `extract_from_text` was never executed

The full module list:
```python
['juriscraper.AbstractSite',
 'juriscraper.DeferringList',
 'juriscraper.OpinionSite',
 'juriscraper.OpinionSiteLinear',
 'juriscraper.OpinionSiteLinearWebDriven',
 'juriscraper.OpinionSiteWebDriven',
 'juriscraper.OralArgumentSite',
 'juriscraper.OralArgumentSiteLinear',
 'juriscraper.OralArgumentSiteLinearWebDriven',
 'juriscraper.WebDriven',
 'juriscraper.fdsys',
 'juriscraper.fdsys.FDSysSite',
 'juriscraper.fdsys.scrape_court_names',
 'juriscraper.fdsys.utils',
 'juriscraper.fdsys.utils.get_metadata',
 'juriscraper.fdsys.utils.get_xpath_results',
 'juriscraper.lasc',
 'juriscraper.lasc.fetch',
 'juriscraper.lasc.http',
 'juriscraper.lib',
 'juriscraper.lib.cookie_utils',
 'juriscraper.lib.date_utils',
 'juriscraper.lib.diff_tools',
 'juriscraper.lib.exceptions',
 'juriscraper.lib.html_utils',
 'juriscraper.lib.importer',
 'juriscraper.lib.judge_parsers',
 'juriscraper.lib.log_tools',
 'juriscraper.lib.models',
 'juriscraper.lib.network_utils',
 'juriscraper.lib.string_utils',
 'juriscraper.lib.test_utils',
 'juriscraper.lib.utils',
 'juriscraper.opinions',
 'juriscraper.opinions.opinion_template',
[...]

PR #4212, opened 2024-07-16
fix(juriscraper_utils): pass full module path as optional argument

grossir:
Solves #4193

Now passing the full module string through cl_scrape_opinions.scrape_court -> extract_doc_content -> update_document_from_text.

This will solve getting the wrong Site.extract_from_text method, when there is more than 1 scraper file for each court id. For example, as it happens with mass

### DIFF
--- a/cl/lib/juriscraper_utils.py
+++ b/cl/lib/juriscraper_utils.py
@@ -4,15 +4,20 @@ import pkgutil
 import juriscraper
 
 
-def get_scraper_object_by_name(court_id):
+def get_scraper_object_by_name(court_id: str, juriscraper_module: str = ""):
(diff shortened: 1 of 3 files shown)


grossir:
I tested this agains the mass backscraper that was recently merged, and where I originally found the problem. It works, in the sense that it now calls the backscraper's extract_from_text, but it has a [bug](https://freelawproject.sentry.io/issues/5613358996/?project=5620212&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=90d&stream_index=0)

flooie (approved):
Thanks for catching this.  This looks great.

PR #4223, opened 2024-07-19
Bulk data export script bugfixes

hopperj:
Adding missing tables, setting import order to avoid foreign key constraint violations, and changing quote character for CSV data to '`' so it doesn't cause parsing errors with quotes in the XML data column.

### DIFF
--- a/scripts/make_bulk_data.sh
+++ b/scripts/make_bulk_data.sh
@@ -195,18 +195,6 @@ people_db_person_race_fields='(
 people_db_person_race_csv_filename="people-db-races-$(date -I).csv"
 
 
-fjcintegrateddatabase_fields='(
-	       id, dataset_source, date_created, date_modified, office, docket_number, origin, date_filed,
-		   jurisdiction, nature_of_suit, title, section, subsection, diversity_of_residence, class_action,
-		   monetary_demand, county_of_residence, arbitrarion_at_filing, arbitration_at_termination,
-		   multidistrict_litigation_docket_number, plaintiff, defendant, date_transfer, transfer_office,
-		   transfer_docket_number, transfer_oprigin, date_terminated, termination_class_action_status,
-		   procedural_progress, disposition, nature_of_judgement, amount_recieved, judgment, pro_se,
-		   year_of_tape, circuit_id, district_id, nature_of_offense, version
-	   )'
-fjcintegrateddatabase_csv_filename="recap_fjcintegrateddatabase-$(date -I).csv"
-
-
 search_opinioncluster_panel_fields='(
 	       id, opinioncluster_id, person_id
 	   )'
--- a/scripts/make_bulk_data.sh
+++ b/scripts/make_bulk_data.sh
@@ -195,18 +195,6 @@ people_db_person_race_fields='(
 people_db_person_race_csv_filename="people-db-races-$(date -I).csv"
 
 
-people_db_attorneyorganization_fields='(
-	       id, date_created, date_modified, lookup_key, name, address1, address2, city, state, zip_code
-	   )'
-people_db_attorneyorganization_csv_filename="people_db_attorneyorganization-$(date -I).csv"
-
-
-people_db_attorney_fields='(
-	       id, date_created, date_modified, name, contact_raw, phone, fax, email
-	   )'
-people_db_attorney_csv_filename="people_db_attorney-$(date -I).csv"
-
-
 docket_fields='(
 	       id, date_created, date_modified, date_cert_granted, date_cert_denied, date_argued,
 		   date_reargued, date_reargument_denied, case_name_short, case_name, case_name_full, slug,
@@ -233,25 +221,6 @@ fjcintegrateddatabase_fields='(
 fjcintegrateddatabase_csv_filename="recap_fjcintegrateddatabase-$(date -I).csv"
 
 
-people_db_role_fields='(
-	       id, role, date_action, attorney_id, docket_id, party_id, role_raw
-	   )'
-people_db_role_csv_filename="people_db_role-$(date -I).csv"
-
-
-people_db_attorneyorganizationassociation_fields='(
-	       id, attorney_id, attorney_organization_id, docket_id
-	   )'
-people_db_attorneyorganizationassociation_csv_filename="people_db_attorneyorganizationassociation-$(date -I).csv"
-
-
-search_docketentry_fields='(
-	       id, date_created_date_modified, date_filed, entry_number, description, docket_id,
-		   pacer_sequence_number, recap_sequence_number
-	   )'
-search_docketentry_csv_filename="search_docketentry-$(date -I).csv"
-
-
 search_opinioncluster_panel_fields='(
 	       id, opinioncluster_id, person_id
 	   )'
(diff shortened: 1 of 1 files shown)


mlissner (changes requested):
Thanks! At a glance, this looks like it's doing a couple things:

1. Fixing/tweaking the escape character. @quevon24 will know if this is a good idea. 

3. Adding a bunch of tables. I'm happy to make sure we have all the financial disclosure tables, but we don't release the party and PACER filings tables in the bulk data. Would you mind tweaking the PR to remove those?

hopperj:
> Thanks! At a glance, this looks like it's doing a couple things:
> 
> 1. Fixing/tweaking the escape character. @quevon24 will know if this is a good idea.
> 2. Adding a bunch of tables. I'm happy to make sure we have all the financial disclosure tables, but we don't release the party and PACER filings tables in the bulk data. Would you mind tweaking the PR to remove those?

Certainly happy to, I didn't realize. To make sure I understand, do you mean all the `disclosures_*` tables?

mlissner:
Yes, exactly, thank you!

hopperj:
@mlissner, updated pushed! I'm happy to make further changes if you spot any other problems. When you're satisfied and merge this in, would it be possible for you to run the export script to update the public S3 datafiles?

scripts/make_bulk_data.sh
```diff
@@ -184,112 +184,121 @@ politicalaffiliation_fields='(
...
-# disclosures_investment
-investment_fields='(
-	       id, date_created, date_modified, page_number, description, redacted,
-	       income_during_reporting_period_code, income_during_reporting_period_type,
-	       gross_value_code, gross_value_method,
-	       transaction_during_reporting_period, transaction_date_raw,
-	       transaction_date, transaction_value_code, transaction_gain_code,
-	       transaction_partner, has_inferred_values, financial_disclosure_id
+
+people_db_attorneyorganization_fields='(
```
mlissner, line 198: This is another table we deliberately don't export. Mind yanking?

scripts/make_bulk_data.sh
```diff
@@ -184,112 +184,121 @@ politicalaffiliation_fields='(
...
 
-# disclosures_position
-disclosures_position_fields='(
-	       id, date_created, date_modified, position, organization_name,
-	       redacted, financial_disclosure_id
+
+people_db_attorney_fields='(
+	       id, date_created, date_modified, name, contact_raw, phone, fax, email
+	   )'
+people_db_attorney_csv_filename="people_db_attorney-$(date -I).csv"
```
mlissner, line 207: And this one too.

scripts/make_bulk_data.sh
```diff
@@ -184,112 +184,121 @@ politicalaffiliation_fields='(
...
-disclosures_reimbursement_fields='(
-	       id, date_created, date_modified, source, date_raw, location,
-	       purpose, items_paid_or_provided, redacted, financial_disclosure_id
+
+search_docketentry_fields='(
+	       id, date_created_date_modified, date_filed, entry_number, description, docket_id,
+		   pacer_sequence_number, recap_sequence_number
 	   )'
-disclosures_reimbursement_csv_filename="financial-disclosures-reimbursements-$(date -I).csv"
+search_docketentry_csv_filename="search_docketentry-$(date -I).csv"
```
mlissner, line 252: Or these, sorry.

mlissner (changes requested):
A few more, if you don't mind?

scripts/make_bulk_data.sh
hopperj: Done!

mlissner:
Sorry to go back and forth on this. It's hard reading the diff and the comments. I didn't catch at first that you removed a few tables. Any reason for that? 

Ideally, we'd just want to add missing tables, without removing the ones we already have, no?

hopperj:
Agreed. When I asked if you meant all the disclosures_*` I thought you meant all, not just any of the ones I had added. I just put them back in now. The delta now is +4 tables in total from original to this version.

scripts/make_bulk_data.sh
```diff
@@ -184,12 +184,54 @@ politicalaffiliation_fields='(
...
+		   date_reargued, date_reargument_denied, case_name_short, case_name, case_name_full, slug,
+		   docket_number, blocked, court_id, assigned_to_id, cause, date_filed, date_list_filing,
+		   date_terminated, filepath_ia, filepath_local, jurisdiction_type, jury_demand,
+		   nature_of_suit, pacer_case_id, referred_to_id, source, assigned_to_str, view_count,
+		   date_last_index, appeal_from_str, appellate_case_type_information,
+		   appellate_fee_status, panel_str, originating_court_information_id, mdl_status,
+		   filepath_ia_json, ia_date_first_change, ia_needs_upload, ia_upload_failure_count,
+		   docket_number_core, idb_data_id
+	   )'
+dockets_csv_filename="search_docket-$(date -I).csv"
```
mlissner, line 209: This is a repeat of line 44, no?

scripts/make_bulk_data.sh
```diff
@@ -184,12 +184,54 @@ politicalaffiliation_fields='(
...
+	       id, dataset_source, date_created, date_modified, office, docket_number, origin, date_filed,
+		   jurisdiction, nature_of_suit, title, section, subsection, diversity_of_residence, class_action,
+		   monetary_demand, county_of_residence, arbitrarion_at_filing, arbitration_at_termination,
+		   multidistrict_litigation_docket_number, plaintiff, defendant, date_transfer, transfer_office,
+		   transfer_docket_number, transfer_oprigin, date_terminated, termination_class_action_status,
+		   procedural_progress, disposition, nature_of_judgement, amount_recieved, judgment, pro_se,
+		   year_of_tape, circuit_id, district_id, nature_of_offense, version
+	   )'
+fjcintegrateddatabase_csv_filename="recap_fjcintegrateddatabase-$(date -I).csv"
+
```
mlissner, line 222: And his is on 69?

mlissner (changes requested):
Hmmm...maybe one more round...?

scripts/make_bulk_data.sh
hopperj: Got it! changes pushed.

scripts/make_bulk_data.sh
```diff
@@ -259,37 +275,45 @@ disclosures_debt_fields='(
...
-declare -a t_28=("disclosures_debt" "$disclosures_debt_fields" "$disclosures_debt_csv_filename")
+declare -a t_7=("search_originatingcourtinformation" "$originatingcourtinformation_fields" "$originatingcourtinformation_csv_filename")
+
+declare -a t_8=("search_docket" "$docket_fields" "$dockets_csv_filename")
+declare -a t_9=("search_opinioncluster" "$opinioncluster_fields" "$opinioncluster_csv_filename")
+declare -a t_10=("search_opinioncluster_panel" "$search_opinioncluster_panel_fields" "$search_opinioncluster_panel_csv_filename")
+declare -a t_11=("search_opinioncluster_non_participating_judges" "$search_opinioncluster_non_participating_judges_fields" "$search_opinioncluster_non_participating_judges_csv_filename")
+
+declare -a t_12=("search_opinion" "$opinion_fields" "$opinions_csv_filename")
+declare -a t_13=("search_opinion_joined_by" "$search_opinion_joined_by_fields" "$search_opinion_joined_by_csv_filename")
```
mlissner, line 296: I noticed that you added three tables above, but that `NUM_TABLES` went up by four. Looking at the diff, I think you forgot to add this table up above?

hopperj: I don't think I follow what you mean. The tables added are: `people_db_race`, `search_opinioncluster_panel`, `search_opinioncluster_non_participating_judges`, and `search_opinion_joined_by`. I've reordered tables to account for foreign key constraints, so when the load-bulk-data script runs it (hopefully) won't encounter any errors.

mlissner: Right, but `search_opinion_joined_by` isn't defined above, so this line of the script would crash?

mlissner (approved):
OK, this looks right. The last thing I want to check, sorry, is whether the change to quoting makes sense to the dev that last worked on this. I'll get him to chime in, and if he approves, let's merge.

mlissner:
@quevon24 can you review the quoting change in this and see what you think?

hopperj:
Sounds good, thanks @mlissner. @quevon24, for context of why I'm requesting it change - there is a column in `opinions` that contains full XML which uses double quote for attribute values and can be hundreds of lines long. This makes parsing/ingesting back into postgres extremely difficult/impossible. When ingesting this file postgres threw syntax errors related to this. I chose the '`' character simply because I expect it won't appear in the data values, although there are probably other choices that would work fine too.

quevon24 (approved):
@hopperj good idea to change the quote character, the idea of ​​forcing the quote character was to be able to differentiate the value that goes in each column, since that caused errors when trying to import it. A character that is difficult to appear will help reduce possible errors. 

Thank you for noticing that some tables were missing in the schema 👍

Everything looks good @mlissner

mlissner:
Great. Set for auto-merge! Thank you @hopperj for the enhancements and getting it through review!

mlissner:
Just a heads up, there's another hurdle, I fear, that may be coming once we export this: https://github.com/freelawproject/courtlistener/issues/4241

PR #4266, opened 2024-08-01
feat(recap): Use de_seq_num for when processing recap emails

ERosendo:
This PR updates the process_recap_email task to extract the `pacer_seq_no` from the email and pass it to relevant download functions.

This PR depends on https://github.com/freelawproject/juriscraper/pull/1099


albertisfu (commented):
The tweak looks good for downloading the main document in an email.

However, I think we could also pass the `de_seq_num` for downloading email attachments within the method `get_and_copy_recap_attachment_docs`, since the attachment report also includes the `de_seq_num` (`pacer_seq_no`).

Additionally, I saw that you included the `de_seq_num` for regular downloads without a magic number, but I believe it is not being used.

For instance, I tried to buy this document, which requires the `de_seq_num` to work via the fetch API:
https://ecf.mdb.uscourts.gov/doc1/0920?pdf_header=&caseid=783601&de_seq_num=57

The fetch failed for the `RECAPDocument` due to the lack of `de_seq_num`.

Therefore, we also need to update the method `download_pacer_pdf_by_rd` to include the `de_seq_num`. However, I am not sure if this improvement is within the scope of this PR because we don't store the `de_seq_num` in the `RECAPDocument` model.

ERosendo:
> However, I am not sure if this improvement is within the scope of this PR because we don't store the de_seq_num in the RECAPDocument model.

@albertisfu You're right, We don't have that information within the RECAPDocument model. However, I think we store it in the `DocketEntry` model (`pacer_sequence_number` field)

albertisfu (approved):
Great! Yeah, you're right. The `pacer_sequence_number` stored in the DocketEntry is the same for all the entry RDs, including attachments. So, it is correct to use it to fetch RD documents.

Also, attachments from emails now properly use the `de_seq_num` when available.

Thanks, this one looks ready to me!

mlissner:
freelawproject/juriscraper#1099 is merged. Is it deployed? We shouldn't merge until then, right?

ERosendo:
> Is it deployed?

@mlissner Yes, I included the changes in v2.6.13 and updated CL dependencies. We can merge this PR :shipit:

PR #4275, opened 2024-08-02
feat(neon): Adds retry logic for Neon account creation 

### LINKED ISSUE #4204: HTTPError 500 while creating Neon Account
Only an instance of this issue has been recorded within the past 30 days. I think we just need a retry for this request. 

This error does not impact account creation in CL and should not affect the membership creation process in Neon.

---
Sentry Issue: [COURTLISTENER-6NZ](https://freelawproject.sentry.io/issues/4961230492/?referrer=github_integration)

```
HTTPError: 500 Server Error:  for url: https://api.neoncrm.com/v2/accounts/
  File "cl/users/tasks.py", line 80, in create_neon_account
    new_account_id = neon_client.create_account(user)
  File "cl/lib/neon_utils.py", line 148, in create_account
    response.raise_for_status()
```

---
Filed by @ERosendo

ERosendo:
This PR fixes #4204 by adding retry logic for the `create_neon_account` tasks on `INTERNAL_SERVER_ERROR`.

While debugging issue #4204, I noticed an unexpected behavior in the `_get_member_record` method. When matching accounts that also have stubs with identical email addresses, membership data was incorrectly assigned to the stub account.

Investigation revealed that the `order_by('-last_login')` clause was causing stub accounts (with null last_login values) to be prioritized over accounts with actual login history due to PostgreSQL's default null sorting behavior. 

> By default, null values sort as if larger than any non-null value; that is, NULLS FIRST is the default for DESC order, and NULLS LAST otherwise.

To address this, I've modified the `_get_member_record` method to use Django's `nulls_last` ordering for the `last_login` field, ensuring that accounts with recent logins are correctly prioritized. Additionally, I've included unit tests to verify this behavior and prevent regressions.


mlissner (approved):
Nice. Never saw the `nulls_last` thing before. Neat.

mlissner:
Set for auto-merge, but deployments are currently blocked.

Thank you.

Issue #4281, opened 2024-08-06
Change how we select custom headers for binary content downloading


grossir:
The default headers for downloading binary content are `{"User-Agent": "CourtListener"}` . For some courts, we use `site.headers`, due to their pages requiring extra headers
https://github.com/freelawproject/courtlistener/blob/1a71ef336fd9cb700c45b1c0a7711aea445322b6/cl/scrapers/management/commands/cl_scrape_opinions.py#L281-L284

We need to support a bunch of New York courts (freelawproject/juriscraper#1042), due to them requiring a token sent in the headers. Instead of adding their court_ids to the list, and of taking care of populating the special variable `site.headers`, we can clean up the interaction

On Juriscraper:

1. wherever it is used, `site.headers` is basically a copy of `site.request["headers"]`. So, we can simply use the latter
2. to let the caller know it should use the `site.request["headers"]`, we can just set a flag `needs_special_headers` to true on the actual site that needs it

This makes the usage self documenting on Juriscraper, by having the `needs_special_headers` flag with default value False and a comment explaining it, instead of having to comment everywhere about `self.headers` usage

On Courtlistener:

```python
if site.needs_special_headers:
    headers = site.request["headers"]
else:
    headers = {"User-Agent": "Courtlistener"}
```

@flooie

PR #4282, opened 2024-08-07
refactor(scrapers.utils.get_binary_content): expect site.needs_special_headers

grossir:
Solves #4281

- delete `headers` argument from get_binary_content signature and calls
- handle site.needs_special_headers to decide headers to use


grossir:
Note that the juriscraper PR needs to be merged first, since it creates the default attribute. If this is merged and deployed before that one, all scrapers will break

flooie:
@grossir - I updated juriscraper so you should be able to update it in this PR to see if it resolves the failing test.

grossir:
@flooie I updated the dependency and it's passing tests now

flooie (approved):
Simple PR, Looks good to me

PR #4285, opened 2024-08-07
Escaping quote character for data export (Issue 4283)

hopperj:
This patch is to address issue #4283. It escapes both the quote characters as well as the intended quote character.


mlissner (approved):
Seems gross in all the right ways. Let's give it a try, thanks!

Issue #4228, opened 2024-07-20
Link to administrative agency dockets


v-anne:
Some administrative agencies like the NLRB have dockets for their proceedings.

It would be great if RECAP could link to them.

Here's an example case this could benefit, with the relevant modification highlighted.
https://www.courtlistener.com/docket/68549109/apple-v-nlrb/
<img width="519" alt="image" src="https://github.com/user-attachments/assets/b5f85a5b-2599-428d-853b-cee03a07710a">

I think cl/opinion_page/templates/docket_tabs.html is the relevant file to modify. Please correct me if wrong.

`{% if og_info %}
      <h3 class="v-offset-above-3">Originating Court Information</h3>
      <hr class="top">
      {% if docket.appeal_from or docket.appeal_from_str %}
        <span class="meta-data-header">Appealed From:</span>
        <span class="meta-data-value">
          {% if docket.appeal_from %}
            {{ docket.appeal_from.short_name }}
          {% elif docket.appeal_from_str %}
            {{ docket.appeal_from_str }}
          {% endif %}
            {% if og_info.docket_number %}
              {% if docket.appeal_from %}
                (<a
                href="/?type=r&amp;docket_number={{ og_info.docket_number }}&amp;court={{ docket.appeal_from.pk }}"
                rel="nofollow"
                data-toggle="tooltip"
                data-placement="right"
                title="Search for this docket number in the RECAP Archive.">{{ og_info.docket_number }})</a>
              {% else %}
                {% if 'NLRB' in docket.appeal_from_str %}
                  (<a href="https://www.nlrb.gov/case/{{ og_info.docket_number }}"
                      target="_blank"
                      rel="nofollow"
                      data-toggle="tooltip"
                      data-placement="right"
                      title="View this case on the NLRB website.">{{ og_info.docket_number }}</a>)
                 {% else %}
                ({{ og_info.docket_number }})
                {% endif %}
              {% endif %}
            {% endif %}
        </span>
      {% endif %}`

mlissner:
That's a fun idea. +1 from me, but it does seem hard to get right. If you're interested in helping with this, we could export a list of the "Appealed From" values so you could analyze them and make whatever links are possible?

v-anne:
Yes, please.

I think the following agencies are very feasible:

- NLRB
- FERC
- FCC

Others might be a bit more difficult:
- EPA (need to link to Federal Register entry I think)
- SEC
- CFTC

[...]

PR #4293, opened 2024-08-08
Linkify administrative citations

v-anne:
This is addressing issue #4228, linking to administrative agency dockets.

Some cases in the federal court system are filed directly in the courts of appeals, bypassing the district courts. Often in those cases, a federal regulation or an administrative agency adjudication is being challenged. This change helps link to some administrative agency dockets, allowing for users to see the background of a case.

Right now, support for citations to rules published in the Federal Register is supported, as is the NLRB docket format. Other smaller agencies will be accounted for in a future update.


mlissner:
Looks like MyPy is unhappy with your typing in the Linting Action. We'll want to fix that before we merge. Your branch is also a few commits behind `main`, but that's easier to fix.

I'll put this on Eduardo's backlog to review. 

Thank you for the hard work!

v-anne:
@mlissner, I caught my mistake. All fixed. Should be good to merge!

ERosendo:
LGTM 👍

ERosendo (approved):

cl/lib/model_helpers.py
```diff
@@ -487,3 +487,36 @@ def suppress_autotime(model, fields):
...
                     "auto_now_add"
                 ]
+
+
+def linkify_orig_docket_number(agency: str, og_docket_number: str) -> str:
+    """Make an originating docket number for an appellate case into a link (MVP version)
+
+    :param agency: The administrative agency the case originated from
+    :param og_docket_number: The docket number where the case was originally heard.
+    :returns: A linkified version of the docket number for the user to click on, or the original if no link can be made.
```
mlissner, line 497:
```suggestion
    """Make an originating docket number for an appellate case into a link (MVP version)

    **NOTE: These links are presented to users and should be subject to strict security checks.**

    For example, each regex should be carefully written so it accepts only the narrowest of
    matches. The risk is that:

      - Mallory uploads a bad document via the RECAP APIs (these are open APIs).
      - The code here parses that upload in a way to create a redirect on the federalregister.gov
        website. 
      - federalregsiter.gov has an open redirect vulnerability (these are common).
      - The user clicks a link on our site that goes to federalregister.gov, which redirects the
        user to evilsite.com (b/c evilsite.com got through our checks here).
      - The user is tricked on that site into doing something bad.
  
    This is all quite unlikely, but we can ensure it doesn't happen by being strict about 
    the inputs our regular expressions capture.

    :param agency: The administrative agency the case originated from
    :param og_docket_number: The docket number where the case was originally heard.
    :returns: A linkified version of the docket number for the user to click on, or the original if no link can be made.
```

mlissner:
Looks good. I added a big note about the danger of this code so that we can keep that in mind as we work on it. Looks fine for now, but want to be careful here, since it could cause trouble. 

Thank you @v-anne for the great contribution!

PR #4263, opened 2024-07-30
Enhance Docket Model to better mirror PACER

flooie:
We need to make some changes to the docket model to include more granular docket number data. 

This includes adding a few fields that are captured partially in docket number 

- office_code
- case type 

and brand new fields 

- defendant_number 
- judge_initials

and finally adding 

- parent_docket

a self-referential foreign key relationship within the docket model that will enable parent/child relationship of dockets with more than one defendant.  This is needed so we can begin to group dockets together in the same manner that pacer does it

PR #4300, opened 2024-08-12
fix(admin): Add parent_docket field as input for editing

ERosendo:
I noticed the admin panel was experiencing slow loading times and occasional failures when accessing docket data. I started debugging the issue and noticed it was primarily caused by excessive database queries generated by the newly added `parent_docket` ForeignKey field (introduced in #4263 ).

Django's admin interface typically uses **select-box interfaces** for `ForeignKey` fields, which can negatively impact performance due to the loading of all related instances. This PR enhances efficiency by adding the `parent_docket` field to the `raw_id_fields` list. This change replaces the **dropdown** with an **input** field, reducing the number of database queries required and resulting in a noticeable improvement in admin panel responsiveness.

### DIFF
--- a/cl/search/admin.py
+++ b/cl/search/admin.py
@@ -285,6 +285,7 @@ class DocketAdmin(CursorPaginatorAdmin):
         "referred_to",
         "originating_court_information",
         "idb_data",
+        "parent_docket",
     )
 
     def save_model(


mlissner:
Thanks, that's helpful. I could imagine how loading 60M items into the interface would cause issues.

PR #4303, opened 2024-08-13
feat(cl_back_scrape_citations): command to scrape citations

grossir:
- Adds new command to back scrape citations from courts that publish the citations in the same website where we scrape their opinions
- Adds tests for the new command
- Refactors `get_binary_content` to be re-used by the new command and to reduce copy pasted code
- create cl.scrapers.exceptions file to hold exceptions raised when ingesting a single case
- use the exceptions to bubble errors to the main loop, to avoid returning break / continue flags
- refactor DupChecker to raise errors
- refactor get_binary_content to raise errors
- refactor cl_scrape_oral_arguments to account for changes
- adapted DupChecker and ContentType tests to changes
- refactors logger.calls to use lazy formatting

Applies to a specific family of court's sites described here:
https://github.com/freelawproject/juriscraper/issues/858#issuecomment-2248873504


grossir:
We can test this by cloning a few `md` cases from 2022, for which we do not have citations into the local db, and running the citation backscraper. We expect:

1. that the cases that do exist get their citations ingested without duplicating the existing clusters
2. that the cases that do not exist get backscraped into clusters together with their citations

https://www.courtlistener.com/opinion/9355831/tapestry-inc-v-factory-mut-insurance/
https://www.courtlistener.com/opinion/9353492/carver-v-state/
https://www.courtlistener.com/opinion/9351006/x

```
manage.py clone_from_cl --type search.OpinionCluster --id 9355831 9353492 9351006
manage.py cl_back_scrape_citations --courts juriscraper.opinions.united_states.state.md --backscrape-start=2022 --backscrape-end=2022
```

grossir:
@mlissner can you review this? 

When writing the tests I realized that we don't have a backscraper for `scotus_slip`, so I tested this using `md`. I will implement that backscraper shortly

cl/scrapers/management/commands/cl_back_scrape_citations.py
```diff
@@ -0,0 +1,127 @@
...
+                super().scrape_court(site, full_crawl=True)
+            else:
+                logger.info("Run the backscraper to collect missing opinions")
+
+    def citation_is_duplicated(
+        self, citation_candidate: Citation, cluster: OpinionCluster, cite: str
+    ) -> bool:
+        """Checks for exact or reporter duplication of citation in the cluster
+        Inspired on corpus_importer.utils.add_citations_to_cluster
+        """
```
mlissner, line 102: I think we'd benefit by having a better docstring here. For example, it returns True/False, but what do those values represent?

cl/scrapers/management/commands/cl_back_scrape_citations.py
```diff
@@ -0,0 +1,127 @@
...
+
+        # We don't have these opinions. Since we are backscraping, if the citation
+        # exists, it will be in the case dictionary, and will be saved in a
+        # regular ingestion process
+        if missing_opinions:
+            # It is easy to ingest a filtered list of cases for OpinionSiteLinear
+            # but not for plain OpinionSite
+            if hasattr(site, "cases"):
+                site.cases = missing_opinions
+                super().scrape_court(site, full_crawl=True)
```
mlissner, line 93: This isn't super elegant, is it? I think this approach means we scrape the site again? Why not use the code from `super().scrape_site` to save the object back around line 59?

cl/scrapers/management/commands/cl_back_scrape_citations.py
```diff
@@ -0,0 +1,127 @@
...
+                if not citation_candidate:
+                    continue
+
+                if self.citation_is_duplicated(
+                    citation_candidate, cluster, cite
+                ):
+                    continue
+
+                logger.info("Saving citation %s for cluster %s", cite, cluster)
+                citation_candidate.save()
```
mlissner, line 83: Is there an opportunity here for a race condition? I.e., could the citation get added between lines 77 and 83? If so, I think it would crash, right? Maybe we should catch that exception?

cl/scrapers/utils.py
```diff
@@ -155,30 +155,33 @@ def get_extension(content: bytes) -> str:
...
+) -> Optional[bytes | str]:
     """Downloads the file, covering a few special cases such as invalid SSL
     certificates and empty file errors.
 
     :param download_url: The URL for the item you wish to download.
     :param site: Site object used to download data
-    :param method: The HTTP method used to get the item, or "LOCAL" to get an
-    item during testing
+
     :return: Two values. The first is a msg indicating any errors encountered.
```
mlissner, line 165: This needs updating too, please.

cl/scrapers/tests.py
```diff
@@ -626,37 +633,94 @@ def setUp(self):
...
+        opinion = OpinionFactory(sha1=self.hash, cluster=self.cluster)
+
+    def test_citation_scraper(self):
+        """Test if citation scraper creates a citation or ignores duplicates"""
+        cmd = "cl.scrapers.management.commands.cl_back_scrape_citations"
+        with mock.patch(f"{cmd}.sha1", side_effect=self.hashes):
+            with mock.patch(
+                f"{cmd}.get_binary_content", return_value="placeholder"
+            ):
+                cl_back_scrape_citations.Command().scrape_court(self.mock_site)
```
mlissner, line 723: As of Python 3.1, you can do multiple with statements, so, perhaps:

```suggestion
        with mock.patch(f"{cmd}.sha1", side_effect=self.hashes), \
             mock.patch(f"{cmd}.get_binary_content", return_value="placeholder"):
            cl_back_scrape_citations.Command().scrape_court(self.mock_site)
```

Eh, not a huge difference, and maybe Black will mess it up, but worth considering, perhaps.

mlissner (changes requested):
Looks good. A few thoughts for you, but this will be very nice to have.

cl/scrapers/management/commands/cl_back_scrape_citations.py
grossir: It would be nice to ingest a single case dictionary up in line 59, but we do the ingestion of all the cases for a site object at a time, that's why I overwrite the `site.cases` with the filtered list

https://github.com/freelawproject/courtlistener/blob/f7337f80d5f9182dfe044ba12bd14689bd24ed1e/cl/scrapers/management/commands/cl_scrape_opinions.py#L279-L284

I could abstract it out and put most of the loop behavior inside a function `ingest_a_case`, but since it is a bigger refactor I didn't want to touch it in this PR

mlissner: Yeah, let's factor it out. It feels off to land this with the duplicative scraping being done, I think.

mlissner (commented):
Changes so far look good. Just need that last refactor.

grossir:
@mlissner can you review this again? I updated the PR with the requested changes

cl/scrapers/management/commands/cl_back_scrape_citations.py
mlissner, line 1: Should we throw an error around line 89? That indicates that we scraped what we thought was a citation, but then couldn't parse it, right?

mlissner (dismissed):
A few thoughts for you. Looks like a nice refactor though.

cl/scrapers/management/commands/cl_back_scrape_citations.py
grossir: The `make_citation` call in line 88 will log an error if no citation was found, so it's covered there
https://github.com/freelawproject/courtlistener/blob/fb892400cab5e2b7300c8fe1aea6bd2687b836f8/cl/scrapers/management/commands/cl_scrape_opinions.py#L50-L65

cl/scrapers/management/commands/cl_scrape_opinions.py
```diff
@@ -213,6 +219,7 @@ def save_everything(
 
 class Command(VerboseCommand):
     help = "Runs the Juriscraper toolkit against one or many jurisdictions."
+    object_type = "opinions"  # for logging purposes
```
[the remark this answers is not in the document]
grossir: I changed the name to `scrape_target_descr`, what do you think? 

I also thought about using the file name `_scrape_opinions`, `scrape_oral_arguments` to build the name, but I think a class variable is good and simple enough

mlissner:
Very nice. 

I've updated to HEAD and set for auto-merge. Thank you!

PR #4309, opened 2024-08-14
feat(proxy): Removes logic to handle legacy cached session data

### LINKED ISSUE #4298: Clean up old fetch proxy shim code
In https://github.com/freelawproject/courtlistener/issues/4192#issuecomment-2284523030 we discussed shim code that needs to be removed in our proxy code. 

There may be more, but here are examples @erosendo identified so far:

https://github.com/freelawproject/courtlistener/blob/57162841fd5aeda54bab59ee0b8439652fcda6e6/cl/recap/tasks.py#L1658-L1660

https://github.com/freelawproject/courtlistener/blob/57162841fd5aeda54bab59ee0b8439652fcda6e6/cl/recap/tasks.py#L2416-L2421

https://github.com/freelawproject/courtlistener/blob/57162841fd5aeda54bab59ee0b8439652fcda6e6/cl/lib/pacer_session.py#L165-L169

ERosendo:
This PR removes redundant session code introduced in #4192. This logic was specifically added to handle session data previously stored in the cache. With the recent changes to session data, this code is no longer necessary and can be safely removed.

This PR fixes #4298

### DIFF
--- a/cl/corpus_importer/tasks.py
+++ b/cl/corpus_importer/tasks.py
@@ -1011,12 +1011,9 @@ def get_pacer_case_id_and_title(
     )
 
     if not session_data and user_pk:
-        cookies_from_cache = get_pacer_cookie_from_cache(user_pk)
-        session_data = (
-            cookies_from_cache
-            if isinstance(cookies_from_cache, SessionData)
-            else SessionData(cookies_from_cache)
-        )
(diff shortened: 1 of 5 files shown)


mlissner (approved):

mlissner:
Thanks. Merged!

PR #4325, opened 2024-08-20
Add columbia ordering to management command

quevon24:
This PR adds the functionality to update_opinions_order command to fill ordering_key field for columbia opinions.

It will search for clusters that include columbia in their source (Z, ZL, ZU, ZLU, etc), clusters with more than one opinion, and whose opinion have a file assigned in the local_path field.

The command to run it:

`docker exec -it cl-django python /opt/courtlistener/manage.py update_opinions_order --action sort-columbia --xml-dir /opt/courtlistener/cl/assets/columbia`

We need to place the xml files in a directory and then pass the path to the command, which will allow us to read the xml assigned to the opinion, this is essential because we rely on the original xml to infer the correct order because some opinions were created in the wrong order due the old columbia importer.

To test it you can clone this clusters:

`docker exec -it cl-django python /opt/courtlistener/manage.py clone_from_cl --type search.OpinionCluster --id 9079419 4226317 8279828 8237637 8041148 4233235 4223647 8041161 8040857 8040920`

And paste this files: [test_xml.zip](https://github.com/user-attachments/files/16697543/test_xml.zip) in your local environment in this location: /opt/courtlistener/cl/assets/columbia 

And then run the command exactly as i put it above.


cl/corpus_importer/management/commands/update_opinions_order.py
```diff
@@ -1,14 +1,23 @@
...
 from django.db.models import Count
 
 from cl.lib.command_utils import VerboseCommand, logger
 from cl.search.models import SOURCES, Opinion, OpinionCluster
 
+VALID_COLUMBIA_SOURCES = [
+    key
+    for key in dict(SOURCES.NAMES).keys()
+    if SOURCES.COLUMBIA_ARCHIVE in key
+]
```
flooie, line 17: Are we using this?

flooie:
Thanks @quevon24 this looks good to me -

Issue #4326, opened 2024-08-21
Load data script syntax error + server side COPY command used


hopperj:
The script generated by the data export to load the data uses a quote character defined as \'`\', which is missing an escape on the '`' character itself. This breaks the command parsing and causes a syntax error.

Additionally, the `COPY` command is a server side only call, and should be replaced with `\COPY ` so users can run the script as intended.

PR #4327, opened 2024-08-21
Issue #4326

hopperj:
Fixing quote character not being escaped and replacing server side COPY with client side version


hopperj:
The second commit is to revert changing COPY->\COPY on the command exporting the data, as that hasn't caused an issue, and while using \COPY shouldn't either, best not to modify something that is working as intended. This does NOT revert the COPY->\COPY change built into the import script.

mlissner (approved):
Thanks. LGTM. Appreciate the little tweaks.

Issue #4140, opened 2024-06-26
Add button to download docket as CSV


mlissner:
A user asked for API access today so they could make CSVs of dockets, and I successfully pivoted them to building this mini-feaure into CL itself. So! 

I think the idea is to add a button to the docket page that allows it to be downloaded as a CSV. The logical place, I think, to add such a button would be somewhere in here: 

![image](https://github.com/freelawproject/courtlistener/assets/236970/c014045f-f35f-4215-a846-5498abffc34d)

Maybe just below the "Next" button at the top of the table, floated right?

When the button is clicked, the idea is to grab all the items from that docket, make them into a CSV, and provide that to the user. 

I think the needed fields would be:

 - entry.id
 - entry.entry_number
 - entry.date_filed
 - entry.time_filed
 - entry.pacer_sequence_number
 - entry.recap_sequence_number
 - entry.description
 - doc.id
 - doc.type (normalized to human-readable form)
 - doc.description
 - doc.acms_document_guid
 - doc.date_upload
 - doc.document_number
 - doc.attachment number
 - doc.pacer_doc_id
 - doc.is_free_on_pacer
 - doc.is_available
 - doc.is_sealed
 - doc.sha1
 - doc.page_count
 - doc.file_size
 - doc.filepath_local (this should be a full link to storage.courtlistener.com
 - doc.filepath_ia
 - doc.ocr_status (normalized to human-readable form)

Questions and suggested answers:
 - Q: What order does the CSV come in?
     - A: Same as the docket. Probably should use the same code.
 - Q: What do we name the file?
     - A: `$slugifiedCaseName.$courtId.$courtlistenerID.$generation-date.csv`. For example: `national-veterans-legal-services-program-v-united-states.dcd.4214664.2024-06-26.csv`
 - Q: If you filter to only some docket entries, does it make a CSV of only those entries or of all of them?
     - A: I think not. It might be useful, but it's probably not worth our time. Just trying to think about how this would affect the file name above is already annoying me.
 - Q: Is it better to send the items to the user in an email, so that we can do it as a background process or to just make the user wait while we do the work? 
     - A: Maybe a hybrid? If it's fewer than 1,000 entries, just do it. If it's more than that, do it as an email? This adds a lot of complexity, so maybe we skip the email thing at first, and only add it if we detect performance problems.
 - Q: Do we include docket metadata in the CSV?
[...]

PR #4334, opened 2024-08-21
Issue 4140 add download button de

RomaDF:


RomaDF:
Addressing issue: https://github.com/freelawproject/courtlistener/issues/4140
General comments/assumptions:
- I have added the class name to the columns in the csv. This is because we had repeated column names like id and description
- For the file name, the issues says: $slugifiedCaseName.$courtId.$courtlistenerID.$generation-date.csv. For example: national-veterans-legal-services-program-v-united-states.dcd.4214664.2024-06-26.csv
For the $courtlistenerID I understood docket id based on the example
- When creating the csv I have added the options quotechar'"' and quoting=csv,QUOTE_ALL. This means that every field in the csv will be converted to char and enclosed in ".
[screen_shot_.tar.gz](https://github.com/user-attachments/files/16696574/screen_shot_.tar.gz)

cl/opinion_page/urls.py
```diff
@@ -54,6 +55,11 @@
     path(
         "docket/<int:pk>/<blank-slug:slug>/", view_docket, name="view_docket"  # type: ignore[arg-type]
     ),
+    path(
+        "docket/download/<int:docket_id>/",
```
mlissner, line 59: I think, for better or worse, we have a pattern of doing this instead:

```suggestion
        "docket/<int:docket_id>/download/",
```

cl/search/models.py
mlissner, line 1: I think this might make more sense in `model_helpers.py`?

cl/opinion_page/views.py
```diff
@@ -563,6 +583,28 @@ async def make_thumb_if_needed(
...
+) -> HttpResponse:
+    """Download csv file containing list of DocketEntry for specific Docket"""
+
+    docket, _ = await core_docket_data(request, docket_id)
+    de_list = await fetch_docket_entries(docket)
+    court_id = docket.court_id
+    case_name = docket.slug
+
+    date_str = datetime.datetime.now().strftime("%Y-%m-%d")
+    filename = f"{case_name}.{court_id}.{docket_id}_{date_str}.csv"
```
mlissner, line 597: I think we wanted these to be dot-separated, so:

```suggestion
    filename = f"{case_name}.{court_id}.{docket_id}.{date_str}.csv"
```

cl/opinion_page/views.py
```diff
@@ -563,6 +583,28 @@ async def make_thumb_if_needed(
...
+
+    # TODO check if for large files we'll cache or send file by email
+    csv_content = await sync_to_async(generate_docket_entries_csv_data)(
+        de_list
+    )
```
mlissner, line 602: Our biggest issue here will be putting the file into memory before serving it. I wonder if we could stream it to the user as we generate it. I *think* all the pieces are there for this *but* if we do this *and* want to cache to disk, we'll need to stream to S3 and to the user. Now we're into the fun stuff...

Maybe later. :)

cl/opinion_page/views.py
```diff
@@ -336,19 +340,34 @@ async def redirect_docket_recap(
     )
 
 
+async def fetch_docket_entries(docket):
+    """Fetch docket entries asociated to docket
+
+    param request: current HttpRequest.
```
mlissner, line 346: This looks out of date.

cl/opinion_page/views.py
```diff
@@ -1,7 +1,10 @@
+import csv
 import datetime
+import logging
 from collections import OrderedDict, defaultdict
 from http import HTTPStatus
-from typing import Any, Dict, Union
+from io import StringIO
+from typing import Any, Dict, List, Union
```
mlissner, line 7: Did you wind up using `List` or `StringIO` in this file?

cl/search/models.py
```diff
@@ -1135,8 +1135,51 @@ class Meta:
         proxy = True
 
 
+# @pghistory.track(AfterUpdateOrDeleteSnapshot(), obj_field=None)
```
mlissner, line 1138: We'll want to remove this comment, right?

mlissner (changes requested):
Thanks Roma! This is great. I have only done a quick review, but here are my thoughts:

1. In #4140, I said the button location should be...

    > Maybe just below the "Next" button at the top of the table, floated right?

    That wasn't very clear. What I meant was that it should go in the actual table of entries, more like this:

    image

2. There's a comment above about signing our CLA. Can you do that please, so we can merge? 

3. We'll want ratelimiting on this before we send it live. Maybe 1/m? We have a decorator for this.

4. I made a comment about caching vs. streaming. Fun stuff, but it can certainly wait until we try this and see what happens. Maybe it'll be fine.

5. I haven't looked at your HTML or tests, I'll let @ERosendo dig into those. 

----

But big picture, this looks great, thank you so much for putting in this effort.

But despite my review (and probably more coming from Eduardo), please don't feel ANY need to continue working on this. It can easily wait, and you've done the task at hand, which was to show us your skills. 

I think it's great and I'll be in touch by email with next steps! 

Thank you!

mlissner:
Oh, one last thing. An example of the CSV will be helpful before we merge to eyeball it and make sure it looks about right — but this can wait!

cl/search/models.py
RomaDF: I have moved the Mixin to cl/lib/model_helpers.py. 
Is this what you meant?

cl/opinion_page/views.py
RomaDF: Yes! fixed

cl/opinion_page/views.py
RomaDF: Doc updated

cl/opinion_page/views.py
RomaDF: StringIO but import should not be in this file any longer, I have deleted them.

cl/search/models.py
RomaDF: No, this is a Mixin not a model, if  I understood correctly the doc this decorator should be use in models. I have deleted the comment and I have moved this to model_helpers

cl/search/models.py
mlissner: Yep, exactly, thanks.

ERosendo:
I've been reviewing the code changes in this PR and here are some thoughts:

> What I meant was that it should go in the actual table of entries.

I believe this proposal makes sense. We can relocate the code for the button from the `de_filter.html` template to the `de_list.html` template. Here are some approaches I've explored for integrating the button into our responsive design:

Screen Recording 2024-08-23 at 4 51 20 PM

Screen Recording 2024-08-23 at 4 46 01 PM

Adding 'csv' near the download icon offers a clear visual cue about the file format, but I'm uncertain about its responsiveness, @mlissner what do you think?

> An example of the CSV will be helpful before we merge to eyeball it and make sure it looks about right

@mlissner Here's an example of the CSV file:

[clarence-t-foos.ohnb.4826703.2024-08-23.csv](https://github.com/user-attachments/files/16733823/clarence-t-foos.ohnb.4826703.2024-08-23.csv)

I reviewed the file and noticed that the link in the`filepath_local` field is not a full URL to the storage location.

cl/search/models.py
```diff
@@ -1736,6 +1760,47 @@ def as_search_dict(self, docket_metadata=None):
...
+        ]
+        if get_column_name:
+            columns = [self.add_class_name(col) for col in columns]
+        return columns
+
+    def _get_readable_document_type(self, *args, **kwargs):
+        return self.get_document_type_display()
+
+    def _get_readable_ocr_status(self, *args, **kwargs):
+        return self.get_ocr_status_display()
```
mlissner, line 1791: Per Eduardo's comment, let's do one more of these for filepath_local. Notionally:

```
    def _get_full_filepath_local(self, *args, **kwargs):
        if self.filepath_local:
            return f"https://storage.courtlistener.com/{self.filepath_local}"
        return ""
```

mlissner:
I like the second gif you posted, Eduardo, with the button. Looks great to me. 

The CSV looks pretty good too. Lots of fields in there that'll confuse people, but the important ones are there too, and we like to provide lots of info when we can, so I think it's good.

cl/search/models.py
RomaDF: done

RomaDF: [Uploading united-states-v-cadden.mad.4275782.2024-08-25.csv…]()
Explample with change in filepath if any

RomaDF:
I have changed the button to de_list. It's not as good as Eduardo's gift, though

cl/lib/model_helpers.py
```diff
@@ -562,3 +562,46 @@ def linkify_orig_docket_number(agency: str, og_docket_number: str) -> str:
...
+    def get_column_function(self) -> Dict[str, Callable[[str], str]]:
+        """Get dict of attrs: function to apply on field value if it needs
+        to be pre-processed before being add to csv
+
+        returns: dict -- > {attr1: function}"""
+        raise NotImplementedError(
+            "Subclass must implement get_column_fuction method"
+        )
+
+    def to_csv_row(self) -> List[str]:
```
ERosendo, line 589: From Python 3.9 ([PEP 585](https://www.python.org/dev/peps/pep-0585/)) onwards `dict`, `list` and various other classes are now generic types. Using these rather than their typing counterpart is now preferred. From Python 3.9 you can now just do:

```suggestion
    def to_csv_row(self) -> list[str]:
```

cl/lib/model_helpers.py
```diff
@@ -1,7 +1,7 @@
 import contextlib
 import os
 import re
-from typing import Optional
+from typing import Callable, Dict, List, Optional
```
ERosendo, line 4: Importing `Dict` and `List` from typing is redundant. We can use built-in types (`dict` and `list`):

```suggestion
from typing import Callable, Optional
```

cl/opinion_page/views.py
```diff
@@ -361,11 +377,12 @@ async def view_docket(
         if cd.get("filed_before"):
             de_list = de_list.filter(date_filed__lte=cd["filed_before"])
         if cd.get("order_by") == DocketEntryFilterForm.DESCENDING:
-            sort_order_asc = False
```
ERosendo, line 364: Removing the `sort_order_asc` variable from this line without providing a suitable alternative will negatively impact the functionality of the filter bar in the `de_filter.html` template. We use this variable to set the active state of the radio buttons, ensuring that the filter bar accurately reflects the current sorting order. 

Screenshot 2024-08-26 at 5 15 28 PM

If removed, the 'asc' button will remain active even when the entries are sorted in descending order, leading to a confusing user experience.

cl/opinion_page/tests.py
```diff
@@ -1523,3 +1533,145 @@ async def test_block_cluster_and_docket_via_ajax_view(self) -> None:
...
+        DocketEntry.objects.all().delete()
+        RECAPDocument.objects.all().delete()
+        User.objects.all().delete()
+
+    def test_fetch_docket_entries(self) -> None:
+        """Verify that fetch entries function returns right docket_entries"""
+        res = asyncio.run(fetch_docket_entries(self.mocked_docket))
+        self.assertEqual(len(res), len(self.mocked_docket_entries))
+        self.assertIn(self.mocked_docket_entries[0], res)
+        self.assertNotIn(self.mocked_extra_docket_entries[0], res)
```
ERosendo, line 1637: We can use asynchronous alternatives and refactor this test to avoid importing `asyncio`. 

```suggestion
    async def test_fetch_docket_entries(self) -> None:
        """Verify that fetch entries function returns right docket_entries"""
        res = await fetch_docket_entries(self.mocked_docket)
        self.assertEqual(await res.acount(), len(self.mocked_docket_entries))
        self.assertTrue(await res.acontains(self.mocked_docket_entries[0]))
        self.assertFalse(
            await res.acontains(self.mocked_extra_docket_entries[0])
        )
```

cl/opinion_page/tests.py
```diff
@@ -1523,3 +1533,145 @@ async def test_block_cluster_and_docket_via_ajax_view(self) -> None:
...
+                            "has_alert": mock_user_has_alert.return_value,
+                            "timezone": "EST",
+                            "private": True,
+                        },
+                    )
+
+                    response = async_to_sync(download_docket_entries_csv)(
+                        self.request, self.mocked_docket.id
+                    )
+                    self.assertEqual(response["Content-Type"], "text/csv")
```
ERosendo, line 1677: While mocking is a great testing strategy, nested context managers can sometimes hinder code clarity. In this instance, using the `@mock.patch` decorator provides a more concise and readable approach to achieving the same goals.

```python
    @mock.patch("cl.opinion_page.utils.user_has_alert")
    @mock.patch("cl.opinion_page.utils.core_docket_data")
    @mock.patch("cl.opinion_page.utils.generate_docket_entries_csv_data")
    def test_view_download_docket_entries_csv(
        self,
        mock_download_function,
        mock_core_docket_data,
        mock_user_has_alert,
    ) -> None:
        """Test download_docket_entries_csv returns csv content"""
        mock_download_function.return_value = (
            '"col1","col2","col3"\r\n"value1","value2","value3"'
        )
        mock_user_has_alert.return_value = False
        mock_core_docket_data.return_value = (
            self.mocked_docket,
            {
                "docket": self.mocked_docket,
                "title": "title",
                "note_form": "note_form",
                "has_alert": mock_user_has_alert.return_value,
                "timezone": "EST",
                "private": True,
            },
        )

        response = async_to_sync(download_docket_entries_csv)(
            self.request, self.mocked_docket.id
        )
        self.assertEqual(response["Content-Type"], "text/csv")
```

ERosendo:
@RomaDF Thanks for your contributions to the pull request! I've added code to enhance the responsive design of the "export CSV" button. 👍 

@mlissner  I've added comments to enhance the code of the new tests, and refine type annotations and docstrings.

Before merging the code, it's important to address the point raised in this comment: 

https://github.com/freelawproject/courtlistener/pull/4334#discussion_r1731839531

cl/opinion_page/views.py
RomaDF: Sorry about this, it was not intentioanl. Fixed

cl/opinion_page/views.py
```diff
@@ -1,4 +1,6 @@
+import csv
 import datetime
+import logging
```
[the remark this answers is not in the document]
RomaDF: done

cl/opinion_page/views.py
```diff
@@ -336,19 +339,32 @@ async def redirect_docket_recap(
     )
 
 
+async def fetch_docket_entries(docket):
+    """Fetch docket entries asociated to docket
+
+    param docket: docket.id to get related docket_entries.
+    returns: DocketEntry list.
```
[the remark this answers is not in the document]
RomaDF: done

cl/opinion_page/tests.py
RomaDF: It makes sense. Done as suggested

ERosendo (approved):

PR #4364, opened 2024-08-28
Improve PDF file download for pacer free documents command

quevon24:
This PR introduce these changes:

- We need to add this cron job to run he daily scrape

```shell
manage.py scrape_pacer_free_opinions --action=do-everything --date-start `date -d "10 days ago" +"%Y-%m-%d"` --date-end `date +"%Y-%m-%d"`
``` 

- Implementation of a new way to rotate between courts to download the PDFs to not hit each court server with a lot of downloads. It will download one item for each court until it finishes all the pending documents to download, in the previous code we downloaded everything from each court. It will go through rows like this:

| court_id | pk  | row_number |
|----------|-----|------------|
| 1        | 10  | 1          |  
| 2        | 11  | 1          | 
| 3        | 13  | 1          |  
| 1        | 12  | 2          |  
| 2        | 14  | 2          | 
| 3        | 16  | 2          |  
| 1        | 15  | 3          |  

instead of this:
| court_id | pk  | row_number |
|----------|-----|------------|
| 1        | 10  | 1          |
| 1        | 12  | 2          |
| 1        | 15  | 3          |
| 2        | 11  | 1          |
| 2        | 14  | 2          |
| 3        | 13  | 1          |
| 3        | 16  | 2          |

- Alternate courts on a weekly basis to generate the documents report when running sweep to avoid hitting each court server with the entire date range

### DIFF
--- a/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
+++ b/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
@@ -16,6 +16,9 @@ from juriscraper.lib.string_utils import CaseNameTweaker
 from requests import RequestException
 from urllib3.exceptions import ReadTimeoutError
 
+from cl.corpus_importer.management.commands.bulk_iquery_project import (
+    CycleChecker,
+)
 from cl.corpus_importer.tasks import (
     delete_pacer_row,
     get_and_process_free_pdf,
@@ -217,12 +220,11 @@ def get_and_save_free_document_reports(
         # point for the sweep
         dates = make_date_range_tuples(date_start, date_end, gap=7)
 
-    if not dates:
-        # We are not using a custom date range, it is the daily cron
-        for pacer_court_id in pacer_court_ids:
-            court_failed = False
+    for pacer_court_id in pacer_court_ids:
+        court_failed = False
+        if not dates:
             # We don't pass the dates in the command, so we generate the range based
-            # on each court using last day queried until today to build the date range
+            # on each court
             date_end = datetime.date.today()
             date_start = get_last_complete_date(pacer_court_id)
             if not date_start:
@@ -233,43 +235,23 @@ def get_and_save_free_document_reports(
                 continue
             dates = make_date_range_tuples(date_start, date_end, gap=7)
 
-            # Iterate through the gap in dates either short or long
-            for _start, _end in dates:
-                exc = fetch_doc_report(
-                    pacer_court_id, _start, _end  # type: ignore
-                )
-                if exc:
-                    # Something happened with the queried date range, abort process for
-                    # that court
-                    court_failed = True
-                    break
-
-                # Wait 1s between queries to try to avoid a possible throttling/blocking
-                # from the court
-                time.sleep(1)
-
-            if court_failed:
-                continue
-    else:
-        # Custom date range, alternate courts on a weekly basis to generate report
-        # when running sweep based on specified date range
+        # Iterate through the gap in dates either short or long
         for _start, _end in dates:
-            court_failed = False
-            for pacer_court_id in pacer_court_ids:
-                exc = fetch_doc_report(
-                    pacer_court_id, _start, _end  # type: ignore
-                )
-                if exc:
-                    # Something happened with the queried date range, abort process for
-                    # that court
-                    court_failed = True
-                    break
-                # Wait 1s between queries to try to avoid a possible throttling/blocking
-                # from the court
-                time.sleep(1)
+            exc = fetch_doc_report(
+                pacer_court_id, _start, _end  # type: ignore
+            )
+            if exc:
+                # Something happened with the queried date range, abort process for
+                # that court
+                court_failed = True
+                break
 
-            if court_failed:
-                continue
+            # Wait 1s between queries to try to avoid a possible throttling/blocking
+            # from the court
+            time.sleep(1)
+
+        if court_failed:
+            continue
 
 
 def get_pdfs(
@@ -278,7 +260,6 @@ def get_pdfs(
     date_end: datetime.date,
     index: bool,
     queue: str,
-    pdf_days_ago: int,
 ) -> None:
     """Get PDFs for the results of the Free Document Report queries.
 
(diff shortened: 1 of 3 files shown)


quevon24:
> I only skimmed, but looks like the right direction. Let me know when it's ready for review! 

Yes, I’m only thinking if I forgot something before request a review 😅

quevon24:
It is ready @mlissner, i added another change

cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
```diff
@@ -231,23 +233,43 @@ def get_and_save_free_document_reports(
...
+                    pacer_court_id, _start, _end  # type: ignore
+                )
+                if exc:
+                    # Something happened with the queried date range, abort process for
+                    # that court
+                    court_failed = True
+                    break
+                # Wait 1s between queries to try to avoid a possible throttling/blocking
+                # from the court
+                time.sleep(1)
```
mlissner, line 269: OK, so the way this works is that if the user provides dates we iterate their courts for each date range. But if the user doesn't provide dates, we iterate the dates for each court, right? 

Three thoughts:
1. I don't understand why we need different logic for whether not dates were provided. 
2. If we do need different logic because one is better, why not apply it to both branches of the `if` statement?
3. I think this part was working fine before, and it has 1s waits between each iteration, so why do we need any changes here at all? The big concern is that we'd hammer a server with tons of requests. Yes, it's better to iterate courts for dates instead of dates for courts, but I think one request at a time with a 1s delay was fine? Or were there errors?

cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
```diff
@@ -273,24 +296,52 @@ def get_pdfs(
     courts
     :param index: true if we should index as we process the data or do it later
     :param queue: the queue name
+    :param pdf_days_ago: specify the number of days ago from which to download PDFs
```
mlissner, line 299: Sorry, I didn't realize we already had date_start and date_end, but it's a little weird to have two different ways of passing the date range to the function. Why don't we just use these parameters in our cron job and then we don't need pdf_days_ago?

cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
```diff
@@ -273,24 +296,52 @@ def get_pdfs(
...
+            days=pdf_days_ago
+        )
+        date_end = datetime.date.today()
         base_filter &= Q(date_filed__gte=date_start, date_filed__lte=date_end)
 
-    rows = PACERFreeDocumentRow.objects.filter(base_filter).only("pk")
+    # Filter rows based on the base_filter, then annotate each row with a row_number
+    # within each partition defined by 'court_id', ordering the rows by 'pk' in
+    # ascending order. Finally, order the results by 'row_number' and 'court_id' to
+    # download one item for each court until it finishes
```
mlissner, line 325: Nice comment!

cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
mlissner, line 1: Down around line 349, you're doing a good thing of enqueuing in a better order, but we still load up the entire thing in Celery simultaneously, which can be a problem. Could you include some code like this to add a day of 1s after each cycle: 
```
        if cc.check_if_cycled(d.court_id):
            print(
                f"Finished iteration {iterations_completed} on docket with id "
                f"{d.pk}. Sleeping {options['iteration_delay']} seconds."
            )
            time.sleep(options["iteration_delay"])
```

(That's from the iquery script.)

mlissner (changes requested):
A few thoughts for you, Kevin. Thank you!

cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
quevon24: you are right, we can just do this:

```shell
manage.py scrape_pacer_free_opinions --action=do-everything --date-start `date -d "2 days ago" +"%Y-%m-%d"` --date-end `date +"%Y-%m-%d"`
```

cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
quevon24: 1. When days are not provided, it is the daily cron and applied the end date as today and the start as today minus 10 days, but below I already commented how the cron would be so as not to have to do that

2. In one part of the process, the end date is obtained based on a table in the database. This date is the last day that the court was consulted. In this way, we do not lose anything, since the command will make the necessary requests until the court documents are updated to today. It is good to keep this like this because if for some reason there is a problem with the court server and we do not get anything, the next day we will get everything from the last day that everything worked correctly. For the sweep, we cannot use this table because it keeps telling us if we are up to date or not. That is why there are two "if" statements with different logic, one for the daily cron and another for the sweep.

3. I think the previous process was correct for the daily cron since there are few requests, for the full sweep many requests will be made per court (from 1950 until this month), that's why I thought that for the sweep it would be better to iterate weekly for each court instead of processing the entire court, if you think 1 second is enough I can revert it back to before without problem, how common is it that we get blocked for making too many requests?

cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
```diff
@@ -16,6 +16,9 @@
 from requests import RequestException
 from urllib3.exceptions import ReadTimeoutError
 
+from cl.corpus_importer.management.commands.bulk_iquery_project import (
+    CycleChecker,
+)
```
mlissner, line 21: Let's move this to a better place, where we have utilities of some kind. I think we're going to want this feature in more and more of our scripts.

mlissner (changes requested):
OK, one last thing, then I think we're good. Thanks Kevin!

cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
mlissner: I'd say keep it simple for now?

mlissner (approved):
Great. Thank you for sticking with it!

PR #4375, opened 2024-08-29
fix(scrapers.DupChecker): ensure raising SingleDuplicateError

grossir:
A bug was introduced in the previous [modification](https://github.com/freelawproject/courtlistener/pull/4303) of the DupChecker, where SingleDuplicateError was not raised when the item was duplicated if we were not in a full_crawl


mlissner:
Seems like the kind of code that'd benefit from some tests, but good to have it fixed.

grossir:
There are tests, I refactored them when I changed the DupChecker, I must have done it incorrectly, I will make another PR to fix them

PR #4388, opened 2024-08-30
Update harvard ordering command

quevon24:
I added a small change to update the way we get the harvard queryset, which has over ~350k objects.

The original approach could cause memory issues because we get the ~350k clusters and use prefetch_related to get their opinions, if we still wanted to use prefetch_related, we will have to use an iterator with chunk_size param to avoid getting everything.

The new approach is similar to the one bill implemented for columbia ordering, we get the cluster_ids of the clusters that we need to update their opinions, and then use the id to get the opinions and order them.


cl/corpus_importer/management/commands/update_opinions_order.py
```diff
@@ -46,13 +46,16 @@ def sort_harvard_opinions(options: dict) -> None:
...
         opinion_order = 1
         any_update = False
         with transaction.atomic():
+            opinions = Opinion.objects.filter(cluster_id=cluster_id).order_by(
+                "id"
+            )
             # We need to make sure they are ordered by id
-            for cluster_op in cluster.sub_opinions.all().order_by("id"):
+            for cluster_op in opinions:
                 if cluster_op.type == Opinion.COMBINED:
```
flooie, line 59: can we add a log here -

cl/corpus_importer/management/commands/update_opinions_order.py
```diff
@@ -46,13 +46,16 @@ def sort_harvard_opinions(options: dict) -> None:
...
 
     completed = 0
-    for cluster in harvard_clusters:
-        logger.info(f"Processing cluster id: {cluster}")
+    for cluster_id in harvard_clusters:
+        logger.info(f"Processing cluster id: {cluster_id}")
         opinion_order = 1
         any_update = False
         with transaction.atomic():
+            opinions = Opinion.objects.filter(cluster_id=cluster_id).order_by(
```
flooie, line 54: I think this may have a bug here

cl/corpus_importer/management/commands/update_opinions_order.py
```diff
@@ -169,8 +172,6 @@ def sort_columbia_opinions(options: dict) -> None:
     if limit:
         clusters = clusters[:limit]
 
-    print(clusters)
```
flooie, line 172: nice catch

PR #4393, opened 2024-08-31
Minor edits to citation API docs

anseljh:


mlissner:
Thanks. Documenting citations is always some of the trickiest to get right. 

There was a failing test here, but it's surely unrelated, so I went ahead and merged.

Thank you!

PR #4419, opened 2024-09-05
Display opinions in correct order in frontend

quevon24:
This change display opinions in the correct order in the frontend using the ordering_key field.

This order only affects opinions from Harvard or Columbia as these are the only sources with split opinions. For the other sources we keep the order by type.

I think the best way to do this was in the view instead of in the template because i think it is more readable in the view using python instead of using template tags.

It  will remove the combined opinion if we have multiple opinions.

I'll leave this PR as a draft while we wait for the ordering command to finish with the harvard data.

### DIFF
--- a/cl/opinion_page/views.py
+++ b/cl/opinion_page/views.py
@@ -83,10 +83,12 @@ from cl.recap.models import FjcIntegratedDatabase
 from cl.search.documents import OpinionClusterDocument
 from cl.search.models import (
     SEARCH_TYPES,
+    SOURCES,
     Citation,
     Court,
     Docket,
     DocketEntry,
+    Opinion,
     OpinionCluster,
     Parenthetical,
     RECAPDocument,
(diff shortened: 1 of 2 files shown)


cl/opinion_page/views.py
```diff
@@ -872,13 +874,31 @@ async def view_opinion(request: HttpRequest, pk: int, _: str) -> HttpResponse:
     )
     await authorities_context.post_init()
 
+    opinions = cluster.sub_opinions.all()
+
+    # Verify if the cluster has any of the two possible sources with multiple opinions
+    if (
+        SOURCES.HARVARD_CASELAW in cluster.source
+        or SOURCES.COLUMBIA_ARCHIVE in cluster.source
+    ):
```
mlissner, line 883: I wonder if this PR can wait until we have ordering across all case law, and, if so, if we can avoid this branch entirely. 

I'd much rather tell our API users to simply order by ordering_key, than to say, "Well, if it's this source or that source, then order this way, otherwise...."

quevon24: At the moment we had only thought about the ordering for Harvard and Columbia but we can add a new action in the command to populate the ordering_key field for all other clusters and automatically set the value in that field every time a new opinion is created. This way we can now order all opinions by ordering_key and get rid of the special cases.

With something like this in the save method:

```python
if self.pk is None and self.ordering_key is None:
	# Add order in new opinions with no defined order value
	last_position = Opinion.objects.filter(cluster=self.cluster).aggregate(models.Max("ordering_key"))["ordering_key__max"]
	self.ordering_key = (last_position or 0) + 1
```

mlissner: Yeah, I think we'll be bummed if we add this sorting key but don't do it well enough to have a simple: "Here's how you sort" comment in the documentation.

quevon24:
Maybe we should close this PR, since @flooie is working on redesigning opinions page

Issue #4376, opened 2024-08-29
Delete same hash opinion duplicates


grossir:
The PR that introduced the bug was [merged](https://github.com/freelawproject/courtlistener/pull/4303) on August 21, 2024; so we have scraped opinions duplicated with the same hash since that date until a fix is merged (WIP [here](https://github.com/freelawproject/courtlistener/pull/4375))

For example, all of these opinions have the same hash [1](https://www.courtlistener.com/api/rest/v3/opinions/10564413/), [2](https://www.courtlistener.com/api/rest/v3/opinions/10563539/), [3](https://www.courtlistener.com/api/rest/v3/opinions/10563970/)

We will have to delete the duplicated opinions, the clusters, and related objects like citations

mlissner:
Wah wah.

Why don't we have a unique constraint on that hash? Seems like the database could have prevented this...

grossir:
`\d+ search_opinion` on the docker compose DB returns
```
Indexes:
    "search_opinion_pkey" PRIMARY KEY, btree (id)
    "search_opinion_author_id_69e3caa8" btree (author_id)
    "search_opinion_cluster_id_09bd537a" btree (cluster_id)
    "search_opinion_date_created_76a4ddf9" btree (date_created)
    "search_opinion_date_modified_524fb7ff" btree (date_modified)
    "search_opinion_download_url_8428ad91" btree (download_url)
    "search_opinion_download_url_8428ad91_like" btree (download_url varchar_pattern_ops)
    "search_opinion_extracted_by_ocr_122ced11" btree (extracted_by_ocr)
    "search_opinion_local_path_8c124953" btree (local_path)
    "search_opinion_local_path_8c124953_like" btree (local_path varchar_pattern_ops)
    "search_opinion_sha1_62196033" btree (sha1)
    "search_opinion_sha1_62196033_like" btree (sha1 varchar_pattern_ops)
    "unique_opinion_ordering_key" UNIQUE CONSTRAINT, btree (cluster_id, ordering_key)
```

> search_opinion_sha1_62196033" btree (sha1)

So, it is indeed not unique (it should have "UNIQUE CONSTRAINT"). 
The index must be dropped and then build again, there is no way to add the UNIQUE constraint via [ALTER INDEX](https://www.postgresql.org/docs/16/sql-alterindex.html)
But the duplicates must be corrected before re-creating it.

mlissner:
[...]

PR #4425, opened 2024-09-06
tests(scrapers): refactor tests for DupChecker.press_on

grossir:
Aims to prevent further duplicates as seen on #4376, due to changes introduced in #4303

- Refactor tests for DupChecker.press_on method: replaces fixtures,  loops and if clauses by explicit test objects and explicit press_on calls for each scenario


mlissner (approved):
LGTM. Thanks for the refactor.

mlissner:
Set for auto-merge.

PR #4433, opened 2024-09-09
feat(search): Adds progress marker to search results

### LINKED ISSUE #4384: Add a spinner or progress marker of some kind to the search results
It'd be fun to do something fancy, but we probably don't have time to get it right. 

Nevertheless, ideas include:

 - Animate our new CL logo somehow to show a weighing of the scales or something.
 - Show messages to the users to educate them about the system or about FLP or both. This seems hard because users wouldn't have enough time to read longer things, so we'd be inclined to make search *slower*. That's no good.
 - Remove the search results and just show a loading document gif like lots of other sites use these days (see Kayak, for example).
 - Show a horizontal line across the top like Github does
 - Remove the research results and show a medium-size spinner like we do in financial disclosures when switching tabs
 - Just gray out the search buttons and put spinners in them.

Probably we should just do the last thing, or maybe the last two things, but it'd be fun to do more. 

Back in https://github.com/freelawproject/courtlistener/issues/4209#issuecomment-2285130213, we decided the spinner could wait until HTMX, but I agree that we should do it.

ERosendo:
This PR fixes #4384 by implementing a visual progress indicator to provide feedback during search requests. While the original issue outlined a `HTMX`-based solution, this implementation uses existing code and a more straightforward JavaScript event-driven approach due to conflicts with HTMX's history navigation functionality.

Key Changes:

- Updates the `submitSearchForm` to create a progress bar element and disable submit buttons upon form submission. 

- Adds an event listener to detect page transitions (the browser starts loading a new page). Upon detection, the progress bar's style is updated to reflect the search request's completion status.

Here's a gif showing the progress bar in action: 

Screen Recording 2024-09-10 at 10 09 09 AM

### DIFF
--- a/cl/assets/static-global/css/override.css
+++ b/cl/assets/static-global/css/override.css
@@ -1709,3 +1709,17 @@ rect.series-segment {
 .htmx-request.htmx-hidden-indicator {
     display:inline;
 }
+
+.turbo-progress-bar {
+  position: fixed;
+  display: block;
+  top: 0;
+  left: 0;
+  height: 3px;
+  background: #B53C2C;
+  z-index: 2147483647;
+  transition:
+    width 300ms ease-out,
+    opacity 150ms 150ms ease-in;
+  transform: translate3d(0, 0, 0);
+}
--- a/cl/assets/static-global/js/progress-bar.js
+++ b/cl/assets/static-global/js/progress-bar.js
@@ -1,0 +1,38 @@
+let trickleInterval = null;
+
+function updateProgressBar() {
+  let progressElement = document.getElementById('progress-bar');
+  let oldValue = 0;
+  if ('value' in progressElement.dataset) {
+    oldValue = parseFloat(progressElement.dataset.value);
+  }
+  let newValue = oldValue + Math.random() / 10;
+  progressElement.style.width = `${10 + newValue * 85}%`;
+  progressElement.dataset.value = newValue;
+}
+
+function installProgressBar() {
+  let progressElement = document.createElement('div');
+  progressElement.id = 'progress-bar';
+  progressElement.classList.add('turbo-progress-bar');
+  progressElement.style.width = '0';
+  progressElement.style.opacity = '1';
+  document.body.prepend(progressElement);
+  trickleInterval = window.setInterval(updateProgressBar, 300);
+}
+
+document.onvisibilitychange = () => {
+  if (document.visibilityState !== 'hidden') return;
+
+  let progressElement = document.getElementById('progress-bar');
+  if (progressElement == null) return;
+
+  window.clearInterval(trickleInterval);
+  progressElement.style.width = '100%';
+  progressElement.style.opacity = 0;
+
+  setTimeout(() => {
+    let progressElement = document.getElementById('progress-bar');
+    progressElement.remove();
+  }, 450);
+};
--- a/cl/assets/static-global/js/progress-bar.js
+++ b/cl/assets/static-global/js/progress-bar.js
@@ -7,7 +7,8 @@ function updateProgressBar() {
     oldValue = parseFloat(progressElement.dataset.value);
   }
   let newValue = oldValue + Math.random() / 10;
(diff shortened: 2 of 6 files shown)


cl/assets/static-global/js/progress-bar.js
```diff
@@ -0,0 +1,47 @@
...
+function disableAllSubmitButtons() {
+  // Get all submit buttons on the page
+  const submitButtons = document.querySelectorAll('input[type="submit"],button[type="submit"]');
+  // Disable each element
+  submitButtons.forEach((button) => {
+    button.disabled = true;
+  });
+}
+
+document.onvisibilitychange = () => {
```
mlissner, line 33: Can you help me understand why you're doing it this way?

mlissner (changes requested):
Looks pretty good and seems to work pretty well. A couple thoughts:

1. The Github progress bar stops around 80% across if things are taking forever. This one just slowly goes all the way across to the end. So if the back end is slow, the progress bar will look complete even as you're still waiting. I think it's better to stop partway if progress is stalled.

2. From a UX perspective, you won't see this bar unless you're scrolled to the top. It should always be visible regardless of scroll position.

I think that's it. I swear I had a third thing though. Hopefully it won't come to me in half an hour when you've mostly fixed the above!

cl/assets/static-global/css/override.css
```diff
@@ -1709,3 +1709,17 @@ rect.series-segment {
...
 }
+
+.turbo-progress-bar {
+  position: fixed;
+  display: block;
+  top: 0;
+  left: 0;
+  height: 3px;
+  background: #B53C2C;
+  z-index: 2147483647;
```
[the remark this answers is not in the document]
ERosendo: 😄 I'm sorry to disappoint you, but no, it's not. 

To ensure the progress bar always appears on top of other HTML elements, we need to set its `z-index` to a very high value. The maximum value for z-index is `2147483647`, which is the largest signed 32-bit integer. By setting the progress bar's z-index to this value, it will always be the topmost element on the page, regardless of the z-index values of other elements.

cl/assets/static-global/js/progress-bar.js
ERosendo: Without `HTMX` events to track request completion, we needed a way to update the progress bar(hide the bar) as the browser receives the server response and begins loading the next page. My initial exploration led me to the `beforeunload` event, which fires right before the user leaves the page. While this event seemed promising, the [MDN Web Docs](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#usage_notes) revealed its unreliability, particularly on mobile platforms. They recommend the [visibilitychange event](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event) as a more dependable signal.

mlissner: Interesting. I would have expected it to not need anything? Like, the whole page is loading, so when the new one comes, the bar gets nuked. I guess this is slightly faster because we get the signal a bit sooner? 

I was also curious how this is impacted by people like me that click between tabs a lot while waiting for things to load, but I wasn't able to get a slow query to see if there was an impact of some kind.

ERosendo:
Thanks for your feedback, @mlissner. I've addressed the issue with the progress bar for long-running requests. It will now stop at 85% to indicate that the process is still ongoing.

> From a UX perspective, you won't see this bar unless you're scrolled to the top. It should always be visible regardless of scroll position.

I've tested the progress bar in Chrome and Firefox and it's always visible. Here are GIFs demonstrating its behavior. 

- Firefox: 

Screen Recording 2024-09-10 at 3 02 27 PM


- Chrome

Screen Recording 2024-09-10 at 2 59 31 PM

mlissner (approved):

mlissner:
Great, here we go. I'll see if I can repro the issue with not seeing it when scrolled once it's in prod. I get the feeling I won't be able to!

PR #4439, opened 2024-09-10
fix(progress bar): Removes visibility change listener

ERosendo:
This PR addresses inconsistent progress bar behavior caused by visibility change listener. The progress bar now remains visible even when switching tabs or minimizing the browser, providing a more consistent user experience.

Screen Recording 2024-09-10 at 4 44 54 PM


mlissner (approved):

mlissner:
Progress! :)

PR #4442, opened 2024-09-11
Generate crosswalk and import cap pdf

jtmst:
# Import Harvard Case Law Access Project (CAP) PDFs to CourtListener

## Issues Addressed:
- [Import CAP PDFs to CL](https://github.com/orgs/freelawproject/projects/54/views/1?pane=issue&itemId=72804379)
- [Generate crosswalk between CAP and CL](https://github.com/orgs/freelawproject/projects/54/views/1?pane=issue&itemId=72804762)

## Changes Implemented:
1. Created a management command to generate crosswalk files between CAP and CL data.
2. Developed a management command to import CAP PDFs to CL using the generated crosswalk.
3. Implemented S3 storage integration for PDF storage (configurable for local storage in development).
4. Added error handling and logging for better debugging and monitoring.
5. Added settings for CAP R2 env variables
6. Added tests for both commands

### Prerequisites:
- Ensure you have the necessary environment variables set for R2 and S3 access.
- For local testing, configure the storage to use local file system instead of S3.

### Steps:

1. Generate the crosswalk:
   ```
    docker exec -it cl-django python /opt/courtlistener/manage.py generate_cap_crosswalk
   ```
   This command will create crosswalk files in `cl/search/crosswalks/`.

2. Import CAP PDFs:
   ```
    docker exec -it cl-django python /opt/courtlistener/manage.py import_harvard_pdfs
   ```
   This command will use the generated crosswalk to fetch and store PDFs.

3. To test with specific parameters:
   ```
    docker exec -it cl-django python /opt/courtlistener/manage.py generate_cap_crosswalk --reporter "A.2d" --volume 100

   ```

### Verification:
- Check the `cl/search/crosswalks/` directory for generated crosswalk files.
- Verify that PDFs are stored either in S3 or local storage (based on configuration).
- Examine the logs for any errors or warnings during the process.

### CAP Crosswalk File (Sample Data):
Crosswalk File Sample

### Imported PDFs (Local Storage):
Imported PDFs

### DIFF
--- a/cl/search/tests/test_generate_cap_crosswalk.py
+++ b/cl/search/tests/test_generate_cap_crosswalk.py
@@ -60,7 +60,10 @@ class TestGenerateCapCrosswalk(SimpleTestCase):
         mock_find_matching_case.return_value = mock_opinion_cluster
 
         # Call the command
-        call_command("generate_cap_crosswalk")
+        call_command(
+            "generate_cap_crosswalk",
+            output_dir="/opt/courtlistener/cl/search/crosswalks",
+        )
 
         # Check if crosswalk file was created
         expected_file_path = (
--- a/cl/settings/misc.py
+++ b/cl/settings/misc.py
@@ -56,3 +56,10 @@ FTM_LAST_UPDATED = env("FTM_LAST_UPDATED", default=date.today())
 
 # Pay and Pray quota
 ALLOWED_PRAYER_COUNT = env.int("ALLOWED_PRAYER_COUNT", default=5)
+
+
+# CAP
+CAP_R2_ENDPOINT_URL = env("CAP_R2_ENDPOINT_URL", default="")
+CAP_R2_ACCESS_KEY_ID = env("CAP_R2_ACCESS_KEY_ID", default="")
+CAP_R2_SECRET_ACCESS_KEY = env("CAP_R2_SECRET_ACCESS_KEY", default="")
+CAP_R2_BUCKET_NAME = env("CAP_R2_BUCKET_NAME", default="cap-static")


FRodriguez18 (commented):
Looks good! Didn't approve it because I think John proposed some changes, but this looks good to me 👍

cl/search/tests/test_generate_cap_crosswalk.py
```diff
@@ -0,0 +1,99 @@
...
+        mock_fetch_cases_metadata.assert_called_once_with("U.S.", "1")
+        mock_find_matching_case.assert_called_once()
+
+
+if __name__ == "__main__":
+    import unittest
+
+    unittest.main()
+
+    unittest.main()
```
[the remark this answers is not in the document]
jtmst: It was not, missed that on cleanup, fixed!

mlissner:
@flooie I just put this on your backlog to prioritize. The background is that we want to have the Harvard PDFs imported into CL and we want to do so regularly. Grab me when you have a sec, and I can give the details before you review or assign to somebody else to review.

cl/search/management/commands/generate_cap_crosswalk.py
```diff
@@ -0,0 +1,283 @@
...
+            return sorted(volume_directories)
+        except Exception as e:
+            logger.error(
+                f"Error fetching volume directories for {reporter_slug}: {str(e)}"
+            )
+            return []
+
+    def find_matching_case(self, case_meta):
+        try:
+            citation = case_meta["citations"][0]["cite"].split()
```
quevon24, line 171: It would be great if you could check the citation with get_citations() from eyecite and use the data from the returned object to filter the OpinionCluster objects, here is an example: https://github.com/freelawproject/courtlistener/blob/main/cl/corpus_importer/management/commands/harvard_opinions.py#L636

cl/search/management/commands/generate_cap_crosswalk.py
```diff
@@ -0,0 +1,283 @@
...
+                )
+
+            logger.info(
+                f"Processed {self.total_cases_processed} cases for {reporter_name}, found {self.total_matches_found} matches"
+            )
+
+    def fetch_volumes_for_reporter(self, reporter_slug):
+        # Fetch volume directories for a given reporter from R2
+        try:
+            response = self.s3_client.list_objects_v2(
```
quevon24, line 153: list_objects_v2 has a limit of 1000 objects, is there a possibility to have more than 1000?

cl/search/management/commands/generate_cap_crosswalk.py
```diff
@@ -0,0 +1,283 @@
...
+    def generate_complete_crosswalk(self):
+        reporters = self.fetch_reporters_metadata()
+
+        if self.single_reporter:
+            reporters = [
+                r for r in reporters if r["short_name"] == self.single_reporter
+            ]
+
+        for i, reporter in enumerate(
+            tqdm(reporters, desc="Processing reporters")
```
quevon24, line 91: A cool way to know that your command did not freeze 👌

cl/search/management/commands/import_harvard_pdfs.py
```diff
@@ -0,0 +1,220 @@
...
+from storages.backends.s3boto3 import S3Boto3Storage
+from tqdm import tqdm
+
+from cl.lib.storage import IncrementingAWSMediaStorage
+from cl.search.models import OpinionCluster
+
+logger = logging.getLogger(__name__)
+
+
+class HarvardPDFStorage(S3Boto3Storage):
```
quevon24, line 21: Maybe this could be in cl/lib/storage.py, what do you think @flooie?

cl/search/tests/test_generate_cap_crosswalk.py
```diff
@@ -0,0 +1,97 @@
...
+            crosswalk_data = json.load(f)
+            self.assertEqual(len(crosswalk_data), 1)
+            self.assertEqual(crosswalk_data[0]["cap_case_id"], 1)
+            self.assertEqual(crosswalk_data[0]["cl_cluster_id"], 100)
+            self.assertEqual(
+                crosswalk_data[0]["cap_path"], "/U.S./1/cases/test_case.json"
+            )
+
+        # Clean up the created file
+        os.remove(expected_file_path)
```
quevon24, line 85: You could also delete the file using tearDownClass method

cl/search/tests/test_import_harvard_pdfs.py
```diff
@@ -0,0 +1,188 @@
...
+        settings.CAP_R2_ACCESS_KEY_ID = "test-access-key"
+        settings.CAP_R2_SECRET_ACCESS_KEY = "test-secret-key"
+        settings.CAP_R2_BUCKET_NAME = "test-bucket"
+        self.court = Court.objects.get(pk="scotus")
+
+        # Create test dockets and clusters
+        self.dockets = []
+        self.clusters = []
+        for i in range(2):
+            docket = Docket.objects.create(
```
quevon24, line 45: Here you could optionally have used DocketFactory and OpinionClusterFactory. Although as it is it is also valid.

quevon24 (commented):
In general, it is necessary to add typing to the functions in addition to updating the docstrings of the functions so that they are in accordance with the format used in courtlistener.

cl/search/management/commands/generate_cap_crosswalk.py
jtmst: From my understanding reporter volumes are limited at 999 before they are rolled into a new series, so this should not be an issue

jtmst: On second thought, I cant say that with certainty there is never 1000+.   Good suggestion, will add pagination to be safe

cl/search/management/commands/generate_cap_crosswalk.py
jtmst: I've added this, but I'll note this could affect performance when run against all the data.  Run speed doesn't seem to be a huge concern at the moment, but something to note for later if the run time gets to get out of control in testing.

jtmst: Agreed that this will help accuracy though, thanks for the suggestion

cl/search/management/commands/import_harvard_pdfs.py
flooie: Yeah - I think we should move anything storage related together

jtmst: Makes sense, moved that to storage.py

cl/search/management/commands/generate_cap_crosswalk.py
```diff
@@ -0,0 +1,353 @@
...
+            logger.error(
+                f"Error fetching volume directories for {reporter_slug}: {str(e)}"
+            )
+
+        return sorted(volume_directories)
+
+    def find_matching_case(
+        self, case_meta: Dict[str, Any]
+    ) -> Optional[OpinionCluster]:
+        """Find a matching case in CourtListener database."""
```
quevon24, line 209: The docstrings for this function are missing

quevon24 (commented):
Everything looks good, you've addressed the comments I left, one function is still missing the docstrings, and a comment I left regarding the find_matching_case() function is still pending.

After solving what I mentioned before, I think it will be ready.

cl/search/management/commands/generate_cap_crosswalk.py
```diff
@@ -0,0 +1,283 @@
...
+
+            match_count = matches.count()
+
+            if match_count == 1:
+                matched_case = matches.first()
+                logger.debug(
+                    f"Single match found: {matched_case.case_name} (CL ID: {matched_case.id})"
+                )
+                return matched_case
+            elif match_count > 1:
```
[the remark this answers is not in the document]
jtmst: In this case I suppose that would be the result, but I was under the impression that when this occurs the citation would be differentiated by page number as "100 A.2d 36a" or "100 A.2d 36b".  Is this not the case? 

If they were the exact same citation starting on the same page then there would be multiple matches and a logged warning with the citations. 

If this is a more common occurrence than I thought then I'll need to find a way to be more specific, how would CL differentiate between the two in this case, by docket number?

cl/search/management/commands/generate_cap_crosswalk.py
jtmst: added

cl/search/management/commands/generate_cap_crosswalk.py
```diff
@@ -0,0 +1,357 @@
...
+                        logger.warning(
+                            f"Multiple matches ({match_count}) found for citation: {citation}"
+                        )
+                        return (
+                            matches.first()
+                        )  # Return the first match, but log a warning
+                    else:
+                        logger.debug(
+                            f"No match found for citation: {citation}"
+                        )
```
flooie, line 247: I am concerned about this part of the code.  

Why would we not use the Harvard CAP identifier to match up opinions.  We kept the Harvard XML file path field in our db.  Wouldn't it be easier (assuming you have that matched to the PDFs)  to use the specific identifier to the identifier found in the harvard file path?

mlissner: We also talked in our meeting today about how the same citation could map to two different scans. Imagine if page 31 of some book has two cases on it. At the top of the page, it has `Mike v. Jessie`, which is half a page long. Around the middle of the page, `Frank v. Howard` begins, but then goes on for another five pages.

In that case, both cases will have the same citation, but the scan for the first case will be one page long, while the scan for the second will be six pages long.

jtmst: I've updated this to instead query the filepath field for a matching path constructed from the case data and cap identifier.  Using this approach we shouldnt need to worry about cases with matching citations/

quevon24 (approved):
Everything looks very good, using the cap id greatly simplified the match, I have no more comments

mlissner:
I think we're just waiting for @flooie's final review here, but he's out sick today. Hopefully we'll get this merged shortly. Sorry for the delay.

flooie (approved):
This looks good to me-  Thanks for taking the time to rewrite the matching code.

cl/settings/third_party/r2.py
mlissner, line 1: I think I'd be confused if i looked in r2.py and only found CAP settings. Could we move these settings into misc.py instead?

mlissner (changes requested):
Sorry to chime in a bit late. I came in here to see if there were any DB migrations before merging, but I noticed a couple of last minute things I think we should tweak. 

@flooie or @quevon24, do you think we could make these changes so Josh can consider his part of this done?

cl/search/management/commands/generate_cap_crosswalk.py
```diff
@@ -0,0 +1,349 @@
...
+        if options["verbose"]:
+            logger.setLevel(logging.DEBUG)
+
+        self.dry_run = options["dry_run"]
+        self.single_reporter = options["reporter"]
+        self.single_volume = options["volume"]
+        self.output_dir = options["output_dir"] or os.path.join(
+            os.path.dirname(os.path.dirname(os.path.dirname(__file__))),
+            "crosswalks",
+        )
```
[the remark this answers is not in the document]
quevon24: i made the --output-dir argument required and removed the default directory

cl/settings/third_party/r2.py
quevon24: I moved the settings to misc.py

cl/search/crosswalks/example_crosswalk.json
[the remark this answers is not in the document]
quevon24: I couldn't find any use for the file other than documentation purposes, so I deleted it and added it as a comment in the generate_cap_crosswalk command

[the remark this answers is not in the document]
flooie: Thanks @quevon24 and @mlissner

flooie:
@mlissner Thanks for your suggestions.  The failing test passed on a re-run and appears to be related to an error in the meeting I missed.

mlissner (approved):

quevon24:
> get



> Woot! We're merged. @quevon24, do you want to file an issue to get this deployed and work with Ramiro on landing the changes?

Yes, I'll get to it, I think we would only need the values ​​of the environment variables necessary for the command, or are they already set?

mlissner:
They're not, I don't think. I'll get them to Ramiro.

PR #4447, opened 2024-09-12
feat(templates): Include additional information in Docket Alarms and Docket Notes

elisa-a-v:
This PR adds a column for the docket number both in docket alerts and in docket notes. It also adds a column for the docket's date filed in the alarm table, and a column for the docket's court in the docket notes.

The mobile version of all tables is currently awkward and hard to navigate, but I would suggest replacing the `<table>` and use flexbox instead, which I'm not sure is worth it at the moment.

## Docket Alerts Desktop
image
## Docket Alerts Mobile
alerts

## Docket Alarms Desktop
image
## Docket Alarms Mobile
notes


ERosendo (approved):
@elisa-a-v Thanks for the PR! I reviewed the changes and they look great.

@mlissner I've been experimenting with the new information in the table and noticed that we're currently using the `short_name` from the court model. While this works, I found that sometimes the `short_name` can contain a few words, which might lead to cluttered templates due to the limited table space.

As an alternative, I think we can consider using the value of the `citation_string` field. This field provides a cleaner and more concise representation. 

Here's a visual comparison:

Using `short_name`:

image

Using `citation_string`:

image

Let me know what you think! Happy to discuss this further.

mlissner:
Makes sense to me, thanks Eduardo. Elisa, how are things coming with the more responsive version? (Feel free to email too.)

elisa-a-v:
I'm finishing up, it looks like this:

#### Docket alerts:
docker alerts
#### Docket notes:
docket notes
#### Opinion notes:
opinion notes

I'm wondering in the case of search alerts if we shouldn't switch the first two columns so that the collapsed list is a bit more descriptive? now it would look like this:
Screenshot from 2024-09-17 18-57-35
but maybe this would be better?
Screenshot from 2024-09-17 18-58-02

I created a separate PR for the responsive version, is that ok?

mlissner:
Great! The separate PR is fine. Just let me know when it's ready to go and we'll give it a look.

mlissner:
Thanks Elisa. I put this and it's sibling PR on Eduardo's backlog for review.

ERosendo (approved):
Looks good. Thanks @elisa-a-v. :shipit:

PR #4457, opened 2024-09-14
4456 fix connection pool problems

mlissner:
Fixes: #4456 by removing connection pools. 

I did some research prior to this, but couldn't find anybody with similar problems that was using gunicorn, so I figured I better land the fix and we can continue researching (or simply not bother) on Monday.


mlissner:
Auto-merge is set.

Issue #4256, opened 2024-07-30
`cl.recap.mergers.find_docket_object` sometimes matches dockets when it shouldn't


grossir:
An example:

`Docket 68295573 already has a case_name Van Camp v. Van Camp, different than new value State v. Snyder`

The [docket](https://www.courtlistener.com/api/rest/v3/dockets/68295573/) in Courtlistener, that would have been overwritten has docket number "1 CA-CV 23-0297-FC", case name Van Camp v. Van Camp

The docket number for the Snyder case is "1 CA-CR 23-0297", which is a different case

So, the "docket_number_core" with value "230297" matches, but it shouldn't

This is a single example for Arizona, but on Sentry there are more records.

There is [another](https://www.courtlistener.com/api/rest/v3/dockets/63344833/) example where the mismatch doesn't have a straightforward solution:

- The oral argument with case name "In re: NEWMAN" has docket number 21-1228
![image](https://github.com/user-attachments/assets/2ba01d06-5e39-4d65-9d9f-86d31272e7b8)

- The [opinion](https://cdn.ca9.uscourts.gov/datastore/opinions/2024/07/30/21-1228.pdf) with case name "Edgar G. C. v. Garland" has the same docket number

There are some cases when it is a correct match, but the case name or other data point is slightly different: [ca3](https://www.courtlistener.com/api/rest/v3/dockets/63591559/)

------------------

The offending logic is in this function

https://github.com/freelawproject/courtlistener/blob/723b7ec84101b18fa2f0aa0dcb7ef7788dc74361/cl/recap/mergers.py#L84-L169

---- 

Sentry Issue: [COURTLISTENER-7XG](https://freelawproject.sentry.io/issues/5657914069/?referrer=github_integration)

```
Docket 68295573 already has a case_name Van Camp v. Van Camp, different than new value State v. Snyder
```

mlissner:
So:
 - probably we shouldn't rely on `docket_number_core` for state court docket numbers we don't understand very well.
 - I don't understand the CA9 example. Their docket numbers should be consistent. Maybe a typo on the court website? 
 - CA3, I guess it's reasonable to go with the latest value, maybe?

grossir:
I downloaded the details for 1236 events / logs from Sentry, which map to 486 unique dockets. Then, I manually inspected each court group, and found that some are indeed mixing up dockets, and some others are matching the correct docket, but bring updated values which may not be better than the old values

### Incorrect docket match

[...]

PR #4462, opened 2024-09-16
fix(scrapers.utils.get_existing_docket): return None when docket_number is an empty string

grossir:
Related to #4256

Doing a Docket lookup when the docket number is empty will match false positives


grossir:
So far only one case has surfaced where matching by empty docket number happens, in `nmctapp` 
https://www.courtlistener.com/api/rest/v3/dockets/67925513

This is also a pointer to improve the `extract_from_text` method for that scraper

flooie (approved):

PR #4470, opened 2024-09-17
4230 capture user queries

legaltextai:
i 've made a change in search/models, views, ran migrations, but somehow it generated changes to 558(!) files


mlissner:
Looks like you added your entire venv. Definitely recommend creating a new branch without these changes. It's not enough to just create another commit removing the files because we won't want to commit them and ship them around as part of the repo forever.

Issue #4230, opened 2024-07-22
Save and expose user search history


mlissner:
This ticket is to begin figuring out how to keep and expose user search queries. There are a few good reasons to do this:

1. From the user's perspective, it's useful to be able to repeat a query, pick up where you left off, etc. 

2. From our perspective, it's useful to be able to see the queries people are doing so we can see how the site is being used.

The big tradeoff for this is privacy:
 - We don't want to build profiles for people.
 - Whenever we store information about people, we increase the damage caused if we're hacked or have a data loss bug, and we raise the importance of keeping things private.
 - Searches can be quite private when aggregated. In 2006, [AOL search logs were revealed](https://en.m.wikipedia.org/wiki/AOL_search_log_release), and it wasn't great.

From a user perspective, this is pretty simple, I think:

 - Expose past searches in the drop down when they query things that are similar.
 - Put a search history tab in their profile
 - Have tools for managing search history

A few thoughts about history management:
 - We could make it very easy to disable history logging when you make a query.
 - We could make it very easy to delete history records for a period of time or completely.
 - We could make it easy to delete a single search history record from the typeahead dropdown or from the history tab.
 - We should make it possible to disable history logging altogether (or maybe we make it the default).

And some general thoughts: 
 - When accounts are deleted, the history must be nuked, not just anonymized.
 - If we're working on integrating this into the typeahead, we should probably think holistically about it — what other kinds of things do we want in there aside from previous queries?

Finally, before we do this, we should do a survey of how other websites do this. Google, Kayak, Lexis, West, and then use the best ideas we can find. I'm placing this on the volunteer backlog in case somebody wants to go analyze other search systems.

mlissner:
In https://github.com/freelawproject/free.law/issues/203, we're evaluating the privacy ramifications of this change. 

I think the above notes from July are pretty good. The one other thing I'd add is that we want to start logging performance of each query too, so that we can evaluate changes in aggregate. We'll want to do that in the database, and in https://github.com/freelawproject/courtlistener/issues/4360 we can get that data into Prometheus/Grafana, so we can see it.

For sequencing this work:

1. Our immediate need is performance. Let's figure out some basic models and start logging that. We can associate that with our users without any issue.

    **Question**: How do we log performance and queries for anonymous users?
    **Answer**: The simplest is that we make the FK to the user table nullable and just keep it null when the user is anonymous. 

1. We need to research what else is out there, as described above. What features exist that we might want?

[...]

PR #4471, opened 2024-09-17
fix(search): implements changes to search to capture user queries

legaltextai:
Makes changes to models, views, adds migrations

 Fixes: #4230


cl/search/views.py
```diff
@@ -514,41 +519,59 @@ def show_results(request: HttpRequest) -> HttpResponse:
...
+                            render_dict.update(do_search(request.GET.copy()))
+                    elif search_type == SEARCH_TYPES.RECAP_DOCUMENT:
+                        render_dict.update(do_es_search(request.GET.copy()))
+                    else:
+                        render_dict.update(do_search(request.GET.copy()))
+
+                    # Create and save the SearchQuery object
+                    SearchQuery.objects.create(
+                        get_params=request.GET.urlencode(),
+                        query_time_ms=render_dict.get("query_time", 0),
```
mlissner, line 566: I think I'd set this to null if we don't have the value, or if the called function always returns the value, I'd do:

```suggestion
                        query_time_ms=render_dict["query_time"],
```

By doing that, you'd enforce the calling function to always deliver on its promise.

cl/search/views.py
```diff
@@ -514,41 +519,59 @@ def show_results(request: HttpRequest) -> HttpResponse:
...
+                        hit_cache=render_dict.get("hit_cache", False),
```
mlissner, line 567: Similarly, defauling to False here seems a bit weird if we know the value will always be in `render_dict`.

cl/search/views.py
```diff
@@ -514,41 +519,59 @@ def show_results(request: HttpRequest) -> HttpResponse:
                 if not is_bot(request):
                     async_to_sync(tally_stat)("search.results")
 
+                    # Perform the search
+                    search_type = request.GET.get("type", SEARCH_TYPES.OPINION)
```
mlissner, line 523: Is there a reason you tweaked the case statement into an if/else, and moved it up here?

cl/search/models.py
```diff
@@ -49,6 +49,17 @@
...
 
+class SearchQuery(models.Model):
+    date_created = models.DateTimeField(auto_now_add=True, db_index=True)
+    date_modified = models.DateTimeField(auto_now=True, db_index=True)
+    get_params = models.CharField(max_length=255)
+    query_time_ms = models.IntegerField()
+    hit_cache = models.BooleanField()
+
+    def str(self):
+        return f"Query: {self.get_params} at {self.date_created}"
```
mlissner, line 60: Getting models exactly right is important because DB migrations are annoying, so I have a few comments here:

1. We can use `AbstractDateTimeModel` to automatically add the date-created and date_modified fields.
2. We'll want full descriptions of each field, as in other models.
3. In postgresql, you don't get a performance boost by doing max_length, and I can promise you that some queries will be really long. I'd remove that so that long queries don't crash things.
4. We'll want to link these queries to users, so we'll need a FK to them (there are lots of examples).
5. The documentation isn't (yet) clear about this, annoyingly, but `db_index` is considered the old way of adding an index. These days, they should go into the `Meta` class.

mlissner (changes requested):
Just a quick review here, but I did find a few things to address. I think you might do well to sit down with Gianfranco and pair program these, but I leave that up to you guys. 

Looks pretty good though!

PR #4474, opened 2024-09-19
fix(search): creates a model to capture user queries

legaltextai:
changes made to models.py and views.py to save user queries 
 
 instead of amending the previous branch, i created a new one. hope it's ok

 Fixes: #4230


legaltextai:
it's not clear what changes to the code caused the test failure

mlissner:
@grossir, since you've been involved in this one already, can you do a review and help bring it across the finish line, please? I think we just need some tweaks to get the tests passing again, but @legaltextai wants to get back into his AI wheelhouse a bit.

PR #4476, opened 2024-09-19
Add harvard scans to UI

jtmst:
Issue: https://github.com/orgs/freelawproject/projects/54/views/1?pane=issue&itemId=72804703

Adds the option to download the scanned split PDF from CAP when it is available. 

NOTE:  The PDFs that this link surfaces have not been imported.  

Screenshot: 
image

### DIFF
--- a/cl/opinion_page/templates/opinion.html
+++ b/cl/opinion_page/templates/opinion.html
@@ -378,7 +378,7 @@
                     {% if cluster.filepath_pdf_harvard %}
                         <li role="separator" class="divider"></li>
                         <li>
-                            <a href="{{ cluster.filepath_pdf_harvard }}" rel="nofollow">
+                            <a href="{{ cluster.filepath_pdf_harvard.url }}" rel="nofollow">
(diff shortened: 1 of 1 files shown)


mlissner:
Thanks Josh. @flooie, can you fold this into your big redesign PR and change it to a split button like we have for RECAP? 

image

We currently have a few links:

 - The court's website
 - Our copy of that PDF
 - The scan from Harvard

Assuming we have all three for a case, I think the main button could just say:

> Download Original

(That would link to the Harvard PDF.)

And the dropdown could have a link to our backup that says:

> Slip Opinion

An HR, and then:

> Link to Court

---

Some variations:

1. If we don't have the Harvard copy, then we promote the links, so that the button says "Download Slip Opinion" and the dropdown says "Link to Court".

2. If we only have the Harvard scan, then no dropdown, just the button.

Will this work OK? The one thing I don't like is that we don't mention Harvard anywhere, but I think it's crammed to put that as the default button text, unfortunately.

PR #4491, opened 2024-09-24
fix(users): Prevent text Injection in Email Query Parameter

### LINKED ISSUE #4466: Vulnerability Report: Text Injection in Registration
A security vulnerability has been identified in the registration success functionality. This vulnerability could allow attackers to inject malicious content into the registration success page, potentially leading to phishing attacks or malware downloads.

Here's how it works:

The registration success page relies on information within the URL to dynamically render its content. Specifically, it uses parameters like "next" and "email" to personalize the message. However, if these parameters aren't properly sanitized before being displayed, attackers can exploit this weakness.

An attacker can craft a malicious URL that includes the "next" and "email" parameters containing deceptive text and a link to a harmful website. Here's an example:

https://www.courtlistener.com/register/success/?next=//evil.com&email=visit%20https://evil.com/malware.exe%20to%20win%20$100%20giftcard

Here, the attacker is trying to inject a phishing link disguised as a congratulatory message about winning a gift card. This could trick a user into clicking the malicious link in the "email" parameter, potentially leading to malware download from the link.

ERosendo:
This PR fixes #4466 by adding stricter validation to the email query parameter to mitigate text injection risks. If an invalid email format is detected, a `SuspiciousOperation` exception is raised and logged at the Error level.

### DIFF
--- a/cl/users/tests.py
+++ b/cl/users/tests.py
@@ -173,6 +173,45 @@ class UserTest(LiveServerTestCase):
                         % next_param,
                     )
 
+    async def test_prevent_text_injection_in_success_registration(self):
+        """Can we handle text injection attacks?"""
+        evil_text = "visit https://evil.com/malware.exe to win $100 giftcard"
+        url_params = [
+            # A safe redirect and email
+            (reverse("faq"), "test@free.law", False),
+            # Text injection attack
+            (reverse("faq"), evil_text, True),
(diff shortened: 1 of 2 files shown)


mlissner (changes requested):
Simple enough, I suppose, thank you.

mlissner (approved):

Issue #4378, opened 2024-08-29
Ordering Opinions (API)


flooie:
Update API to include ordering for opinions

flooie:
Ordering key already is represented into the API.  I believe it may have been automatic.

mlissner:
A couple things:
 - If we have nested opinions, can/should they be ordered properly?
 - Can people sort by the new ordering key? 

I'm just trying to think of things we'd want if we were consuming the API, basically?

flooie:
> If we have nested opinions, can/should they be ordered properly?

- Yes they can, 
- Yes they should be ordered properly.

> Can people sort by the new ordering key?

- I don't think we should allow people to sort by ordering key - I think order only matters to allow people to read the opinions in order and I dont think it matters otherwise.  

The real problem I think is the confusion generated by grouping apples with oranges.  I'd rather we just return the sub-divided opinions and not return the or group the combined opinions where both exist.  But I know people want more data.

Issue #2499, opened 2023-01-27
Opinions presented out of order


flooie:
We have another fantastically fun bug to fix.  

[Daubert-v-Western-Meat-Co](https://www.courtlistener.com/opinion/3306830/Daubert-v-Western-Meat-Co/)

Unfortunately, the order presented by us is not the order it is presented in the book.  We clearly have an order that we present - but the addendums dont actually come at the end.  

In the following view the actual order should be as far as I can tell ....

------

**Lead Opinion** | Dissenting Opinion | **Concurrence** | Addendum Listed Last | **Addendum Listed First**

----
<img width="754" alt="Screenshot 2023-01-27 at 11 03 30 AM" src="https://user-images.githubusercontent.com/6464529/215132585-dafa50e6-2bf1-4222-9f58-e83cb55ca382.png">

flooie:
This feels like a rather bad bug - but im not sure how to fix it.

flooie:
I was hopeful that they were added to the system in the correct order --- but 

'https://www.courtlistener.com/api/rest/v3/opinions/3306610/',  - the lead opinion
'https://www.courtlistener.com/api/rest/v3/opinions/3306611/',  - the final addendum
'https://www.courtlistener.com/api/rest/v3/opinions/3306612/', - the first addendum 'https://www.courtlistener.com/api/rest/v3/opinions/3306613/', - the dissent
'https://www.courtlistener.com/api/rest/v3/opinions/3306614/ - the concurrence 

I wonder why the order was imported in this way, or if this is the correct order in a different book.

flooie:
There are two issues presented here.  one is order - the other is whether something counts as an opinion or not.  I'm going to let this issue stay with the first issue and the one addressed as order.  

The order issue was created by defaulting to order by type during the Columbia import.  

We need to now create an order in the opinion model.

mlissner:
Seems like this *might* be all good now? Want to review and respond, @flooie?

PR #4496, opened 2024-09-25
feat(opinion): Add ordering key

flooie:
Add ordering key to Opinion class meta field for sub opinions


cl/search/models.py
```diff
@@ -3472,6 +3472,7 @@ class Meta:
                 name="unique_opinion_ordering_key",
             )
         ]
+        ordering = ["ordering_key"]
```
mlissner, line 3475: I don't think I'd approach this this way. The problem with doing this is that it would apply the ordering everywhere, which imposes a performance penalty to queries where we don't care about the ordering. I think this is one of those django features that are considered an anti-pattern these days. 

Can you refactor to apply the ordering only where we need and want it?

quevon24:
This is the list of changes:

- Remove default ordering in Opinion model
- Remove unnecessary migrations
- Set opinions ordering in OpinionClusterViewSet (opinions displayed in correct order in API)
- Display opinions in the correct order in the old opinions template (remove ordering by type)

cl/opinion_page/views.py
```diff
@@ -796,7 +797,15 @@ async def view_opinion_old(
     unbound form.
     """
     # Look up the court, cluster, title and note information
-    cluster: OpinionCluster = await aget_object_or_404(OpinionCluster, pk=pk)
+    cluster: OpinionCluster = await aget_object_or_404(
+        OpinionCluster.objects.prefetch_related(
+            Prefetch(
+                "sub_opinions",
+                queryset=Opinion.objects.order_by("ordering_key"),
```
grossir, line 804: This will work; however I am confused by the general usage of `view_opinion_old`

The ordering is applied for "view_opinion_old", which is called only if `ui_flag_for_o` is not active.
I couldn't find any documentation for `ui_flag_for_o`, but from looking into the waffle admin and seeing it is only active for staff and opt-in users

I think we also want the ordering to work *when* that flag is active?

In general, I think some more comments would help understand what's going on 


-----------

Also, I wanted to test this using some clusters that are out of order without this PR, but couldn't find one. Many opinions with order have their ids in order with the ordering key. Do you have one handy? @flooie

grossir: This PR is related to 

https://github.com/freelawproject/courtlistener/issues/4378

actually found examples here:

https://github.com/freelawproject/courtlistener/issues/2499

grossir: So, the ordering for the new UI is done here; so that's not a problem
https://github.com/freelawproject/courtlistener/blob/077192564c5487dc35713afd8094740edb604a62/cl/search/models.py#L2878-L2889

An example of a opinion where the old UI sorts wrong (it sorts by type), and the new UI does it good (by ordering key):
https://www.courtlistener.com/opinion/3472329/a/

By cloning this cluster, and seeing the order we can see that this now has the desired order

`docker exec -it cl-django python /opt/courtlistener/manage.py clone_from_cl --type search.OpinionCluster --id 3472329`

image

The API response also has the desired order
image

----------------------

So, this is working as expected; my only comment would be to add some comments for the meaning of the flag; and for any differences between the old and the new, including how the ordering is done

grossir:
While testing this I noticed a related ordering problems, in the search results page; we may want to turn this into a separate issue
image

cl/opinion_page/views.py
grossir, line 804: About the case when there are opinions with both NULL and valid `ordering_key` values, it will not show the "Combined" or scraped opinion first
image

However, if you change
```python
from django.db.models import F
...
queryset=Opinion.objects.order_by(F("ordering_key").asc(nulls_first=True))
```

it will use the desired order
image

flooie: thanks @grossir

Issue #4467, opened 2024-09-17
Some single dockets are spread through multiple docket objects, need merging


grossir:
Some dockets are spread through multiple docket objects, which need merging. Have seen it in `mont` and `texapp`. Seems to happen when docket number core exists

- texapp [example](https://www.courtlistener.com/?q=court_id%3Atexapp&type=o&order_by=score%20desc&stat_Published=on&stat_Unpublished=on&stat_Unknown=on&docket_number=13-20-00065-CR)
- mont [example](https://www.courtlistener.com/?q=court_id%3Amont&type=o&order_by=score%20desc&stat_Non-Precedential=on&docket_number=DA%2023-0044)
- haw [example](https://www.courtlistener.com/?q=court_id%3Ahaw&type=o&order_by=dateFiled%20desc&stat_Published=on&stat_Unpublished=on&docket_number=SCMF-12-0000538)
- delsuperct [example](https://www.courtlistener.com/?q=court_id%3Adelsuperct&type=o&order_by=dateFiled%20desc&stat_Published=on&stat_Unpublished=on&docket_number=1201018188)

Related to https://github.com/freelawproject/courtlistener/issues/4256, it is the inverse: some dockets were not matched when they should have

---------------

Sentry Issue: [COURTLISTENER-85X](https://freelawproject.sentry.io/issues/5848914189/?referrer=github_integration)

```
texapp: more than 1 docket match for docket number '04-22-00669-CR'
```

grossir:
This is a script to merge the dockets in this issue , which should be a single docket. To test it, check the commented lines on how to use `clone_from_cl`, which would require the branch that allows cloning clusters together with dockets to be merged

```python
# Example query to get the docket ids
# https://www.courtlistener.com/api/rest/v3/dockets/?court__id=texapp&docket_number=14-22-00805-CV

# texapp example
# manage.py clone_from_cl --type search.Docket --id 63138128 63138127 61628798 61628797 --add-clusters
# texapp example 2
# manage.py clone_from_cl --type search.Docket --id 66610135 65750654 --add-clusters

# mont example
# manage.py clone_from_cl --type search.Docket --id 66906891 66769920 66683333 66556449 65640268 65392234 64686343 63558008  63359967 63273348 62979968 --add-clusters

# haw
# manage.py clone_from_cl --type search.Docket --id 66815418 66636686 65416047 --add-clusters

# del
[...]

PR #4497, opened 2024-09-26
feat(clone_from_cl): optionally add clusters when a docket is cloned

grossir:
For use in #4467

Adds `--add-clusters` optional flag to be used when a docket is cloned


cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -350,6 +352,7 @@ def clone_docket(
     docket_ids: list,
     add_docket_entries: bool,
     add_audio_files: bool,
+    add_clusters: bool,
```
quevon24, line 355: Everything looks, you could just add the missing parameters to the docstring: add_audio_files and add_clusters

quevon24 (approved):
Just a small comment regarding docstrings, but it is ready 👍

cl/scrapers/management/commands/clone_from_cl.py
grossir: Thanks @quevon24  I have updated the docstrings and set for automerge

PR #4499, opened 2024-09-26
Add XML Harvard to html citation generation 

flooie:
Add XML Harvard to html_with_citations generation

Also 
Some XML has xml encoding which is an issue for lxml used in eye cite.
This is stripped out in some cases and a test is added


cl/citations/tests.py
```diff
@@ -267,6 +267,48 @@ def test_make_html_from_html(self) -> None:
...
+        ]
+
+        # fmt: on
+        for s, expected_html in test_pairs:
+            with self.subTest(
+                f"Testing html to html conversion for {s}...",
+                s=s,
+                expected_html=expected_html,
+            ):
+                opinion = Opinion(html=s)
```
grossir, line 293: Shouldn't the key be `xml_harvard`
Like 
`opinion = Opinion(xml_harvard=s)`

cl/citations/annotate_citations.py
```diff
@@ -15,10 +16,18 @@ def get_and_clean_opinion_text(document: Opinion | RECAPDocument) -> None:
 
     :param document: The Opinion or RECAPDocument whose text should be parsed
     """
-    for attr in ["html_anon_2020", "html_columbia", "html_lawbox", "html"]:
+    for attr in [
```
grossir, line 19: Are these in priority order? It will pick up the first one available and skip the others. 

I think this should have a comment to make it clear

grossir (approved):
This looks good, I left a couple minor comments @flooie

cl/citations/tests.py
flooie: yes I suppose thats true - it gets there in the end though. ill chagne that.

cl/citations/annotate_citations.py
flooie: its hard to say a correct order - honestly, it mostly matters for harvard being above columbia - otherwise the rest is pretty much the same

PR #4502, opened 2024-09-27
fix(search): Set cache control header to the advanced view

ERosendo:
Key changes:

- Adds  the `@never_cache` decorator to set cache control header to the `advanced` view( which is used to render the following pages: **recap**, **oral arguments**, **financial disclosure**, and **judge** search) so it's always loaded fresh from the server. 

- Fixes a few typos in the migration guide.


mlissner (approved):

mlissner:
Thank you!

Issue #4507, opened 2024-09-30
Do front end and API for Pray and Pay


mlissner:
An earlier message has a nice prototype of how the front end could work: 

> I'm thinking through the front end, my first thought was to have something like this where the prayer emoji is available in rows of documents that need to be bought, right next to the "Buy on PACER" button. If someone clicks it without being logged in, they'll be directed to the sign in page. If they've exhausted their daily quota, the emoji will be grayed out with CSS.
> <img width="943" alt="image" src="https://github.com/user-attachments/assets/e5e13c72-5b8f-4d1b-a7d6-fccf832355c8">
> 
> Thoughts?

_Originally posted by @v-anne in https://github.com/freelawproject/courtlistener/issues/4486#issuecomment-2379736233_

-----

To do this, we'll need two things:

1. An API for creating and deleting prayers
2. The actual HTML tweaks

The API should probably use the Django Rest Framework. I think a good example for this would be `DocketAlertViewSet`, which is documented here: 

https://www.courtlistener.com/help/api/rest/alerts/#dockets

I don't think we need documentation for P&P. 

For the front end, I think the picture above is good, but two thoughts off the bat:

1. I think the icon needs a few states:
    - Unclicked with or without a count of other users
    - Clicked with or without a count of other users

2. I like how unobtrusive the icon is now, but it doesn't teach people what it is. Maybe a mouseover popup? Explaining or leading to a help page? 

Another question is how people find the leaderboard. Ideally it'd somehow be tied to the button, I think, but I'm not sure how. Maybe it could also live in the mouseover popup.

We usually use HTMX for doing updates like these, if that's feasible.

Finally, with UI stuff, it's really nice and fairly easy to use django-waffle to launch it for just a few people at a time. I'd definitely recommend that for a feature this big.

v-anne:
@mlissner, right now I'm thinking of doing the API first and allowing tech-savvy people to test the internals of the feature (does the leaderboard show the documents, etc.).

Good points on the front end--I hadn't thought it through that deeply until seeing your post. I will take a bit to digest it before offering a new design.

One thing that I've been thinking of, however, is whether it is beneficial to limit document requests to people with accounts. Do you have statistics for what fraction of the CourtListener DAU count signs in with an account? If we're limiting the feature to a small fraction of the user base, that seems counterproductive.

mlissner:
[...]

PR #4519, opened 2024-10-01
4507 pray and pay api

v-anne:
This is the PR for #4507, adding the API and frontend for the pray-and-pay project. It follows up on #4386 and #4498.


cl/favorites/tests.py
```diff
@@ -924,7 +924,7 @@ async def test_prayers_integration(self) -> None:
             email_text_content,
         )
         self.assertIn(
-            f"You requested it on {prayer_1.date_created.strftime("%b %d, %Y")}",
+            f"You requested it on {prayer_1.date_created.strftime("%b %-d, %Y")}",
```
mlissner, line 927: We saw this test failing too. I think you should delete this commit if you can, and rebase on main. That should get you the fix.

v-anne:
@albertisfu @mlissner I think a rough version of the API endpoint is done. Apologies for any errors, I mostly followed `DocketAlertViewSet` and made some changes when I thought appropriate.

mlissner:
Very cool. Thanks @v-anne. I'll put it on Alberto's backlog for review.

cl/favorites/views.py
```diff
@@ -190,3 +193,19 @@
...
+    request: HttpRequest, recap_document: RECAPDocument
+) -> HttpResponse:
+    user = request.user
+
+    # Call the create_prayer async function
+    new_prayer = await create_prayer(user, recap_document)
+
+    # Redirect back to the referring page after creating the prayer
+    return HttpResponseRedirect(
+        request.META.get("HTTP_REFERER", reverse("view_docket"))
```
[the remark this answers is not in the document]
v-anne: I am not wedded to this function, in fact I think it needs significant changes. I just pushed what I had to prevent the codebase from breaking, so I can see if the other tests are passing.

mlissner:
Nice progress. This is up next on Alberto's backlog for him to review.

mlissner:
Cool! Want to share some screenshots?

v-anne:
This is a random case I pulled from RECAP using the API:

image
image

At this point I've exhausted my quota so every request button is grayed out. You can see that the request button is not displayed on rows that have documents that are already available.

Here is the list of requested documents, which now includes a link to PACER to purchase them:

image

v-anne:
At this point, I think I've done as much as I can on this feature. I'll wait until Alberto has a chance to look at it before making any changes that are necessary.

mlissner:
Cool. I've got UI comments, but I'll hold them until Alberto has had a chance to chime in. This is looking great.

albertisfu:
@v-anne @mlissner  I've review the PR and applied the following tweaks:

- Added the Buy on PACER modal to the `top_prayers` template. It is required to display when users click the "Buy on PACER" button and do not have the RECAP extension installed.
- The "Pray" button looks good on the Docket page. However, I noticed it required two additional queries for every RECAPDocument in the Docket, which could be quite slow, especially for large cases. To improve this, I changed the approach to retrieve all the prayer counts and existing prayers in bulk for the RECAPDocuments on the page. This way, it only requires two additional queries per page. The counts and existing prayers are then appended to the RDs on the page, so the frontend approach to show the button remains unchanged.
- Added tests for the new bulk query methods: `get_prayer_counts_in_bulk` and `get_existing_prayers_in_bulk`.
- Passing `prayer_eligible` a single time within the context, as this variable only depends on the user. We can check it once and pass it to the template.

**Regarding the API:**

I understand the objective of the API is to allow users to create, list, and delete their own Prayers, correct?

Based on this, I added tests and adjusted the serializer validations to achieve the following:

- Create new prayers.
- Avoid duplicate prayers for the same document.
- List user prayers in the "Waiting" status.
- Delete user prayers in the "Waiting" status.
- Limit users' prayer creation based on the `prayer_eligible` method.
- Disallow PUT and PATCH methods, so users cannot change the status or RECAP document of their prayers.

**Regarding the frontend:**

I read that there is still work in progress regarding the button's asynchronous action and the display of tooltips.
Is this work considered part of this PR? Otherwise, should we place the frontend buttons behind a waffle flag, so that if this is merged, the buttons are not visible to the public?

Thanks @v-anne

v-anne:
@albertisfu, thanks for your great work on this.

My responses to your questions:

> I understand the objective of the API is to allow users to create, list, and delete their own Prayers, correct?
Yes, this was the intended use of the API. However, I think we didn't want to give users the ability to delete their own Prayers, because the quota resets daily. Even an accidental document request is relatively harmless as a result. I'm option to Mike's thoughts as to whether my view is wrong.

> I read that there is still work in progress regarding the button's asynchronous action and the display of tooltips.
Is this work considered part of this PR? Otherwise, should we place the frontend buttons behind a waffle flag, so that if this is merged, the buttons are not visible to the public?
Initially, I thought that the work on the buttons would take more time, but I was able to mostly finish it. I understand Mike has some feedback for UI, so maybe we should use a waffle flag to hide them momentarily.

cl/favorites/api_views.py
```diff
@@ -70,6 +70,8 @@ class PrayerViewSet(LoggingMixin, ModelViewSet):
     ordering = "-date_created"
     # Additional cursor ordering fields
     cursor_ordering_fields = ["date_created"]
+    # Only allow these methods. Restricting PUT and PATCH.
+    http_method_names = ["get", "post", "delete", "head", "options"]
```
v-anne, line 74: Is the current consensus to allow users to delete their requests?

cl/opinion_page/templates/includes/de_list.html
```diff
@@ -171,6 +171,26 @@
...
+                          {% if request.user.is_authenticated %}
+                          <a href="{% url 'create_prayer' rd.id %}"
+                            class="btn btn-default btn-xs prayer-button {% if rd.prayer_exists or not prayer_eligible %}disabled{% endif %}"
+                            data-document-id="{{ rd.id }}"
+                            {% if rd.prayer_exists %}
+                              title="You have already requested this document."
+                            {% elif not prayer_eligible %}
+                              title="You have reached your daily request limit."
+                            {% endif %}
+                            style="margin-left: {% if rd.page_count %}8px{% else %}44px{% endif %};">
```
v-anne, line 183: I am not a fan of hardcoding the amount of pixels to leave between the two buttons. Open to suggestions for improving it.

cl/opinion_page/templates/includes/de_list.html
```diff
@@ -171,6 +171,26 @@
...
+                            {{rd.prayer_count}} 🙏
+                          </a>
+                        {% else %}
+                          <a href="{% url 'sign-in' %}?next={{ request.path|urlencode }}"
```
v-anne, line 187: This directs users to the sign-in page instead of using the modal that the alerts feature employs, because I wasn't quite sure how to incorporate the modal. It works for a MVP, but obviously the modal would be better.

cl/favorites/views.py
```diff
@@ -190,3 +193,17 @@ async def open_prayers(request: HttpRequest) -> HttpResponse:
...
+async def create_prayer_view(
+    request: HttpRequest, recap_document: int
+) -> HttpResponse:
+    user = request.user
+    recap_document = await RECAPDocument.objects.aget(id=recap_document)
+
+    # Call the create_prayer async function
+    new_prayer = await create_prayer(user, recap_document)
+
+    return HttpResponse("It worked.")
```
v-anne, line 209: This seems problematic. The intention was to leave users on the same docket page instead of taking them to a new one that says "It worked."

v-anne:
@albertisfu -- one more thing about the API. I admit I only made a cursory search and did not find any results, but is it possible for users to actually know what the id of a particular RECAPDocument is? Otherwise, it seems very difficult to work directly with the API to request the purchase of a particular document.

For instance, I envision one use case of the API being a user being able to request new documents on a docket (after they're informed of them via webhook).

albertisfu:
> Yes, this was the intended use of the API. However, I think we didn't want to give users the ability to delete their own Prayers, because the quota resets daily. Even an accidental document request is relatively harmless as a result. I'm option to Mike's thoughts as to whether my view is wrong.

Got it. Yeah, an accidental document request is harmless. So giving users the option to delete their waiting prayers can be omitted if Mike agrees.

> Initially, I thought that the work on the buttons would take more time, but I was able to mostly finish it. I understand Mike has some feedback for UI, so maybe we should use a waffle flag to hide them momentarily.

Great. @mlissner let me know if we intend to merge this one and leave the remaining UI work for a different PR so I can set a waffle flag here.

> one more thing about the API. I admit I only made a cursory search and did not find any results, but is it possible for users to actually know what the id of a particular RECAPDocument is? Otherwise, it seems very difficult to work directly with the API to request the purchase of a particular document.
> 
> For instance, I envision one use case of the API being a user being able to request new documents on a docket (after they're informed of them via webhook).

Yes, it's possible for users to get the ID of a particular document. They can use the docket-entries API for checking entries in a docket, e.g:

https://www.courtlistener.com/api/rest/v4/docket-entries/?docket=60115540
And then use the `docket_entry` ID to request their RECAPDocuments:
https://www.courtlistener.com/api/rest/v4/recap-documents/?docket_entry=404738237
This is where users can see the RECAPDocument IDs.

Also, for instance, the Docket Alert Webhook shows both the `docket_entry` `ids` for a case and the `recap_document` IDs directly within the payload.

{
```
   "payload":{
      "results":[
         {
            "id":2208776613,
            "docket": 4214664,
            "date_filed":"2022-10-11",
            "description":"MOTION for Settlement Preliminary Approval by ALLIANCE FOR JUSTICE, NATIONAL CONSUMER LAW CENTER, NATIONAL VETERANS LEGAL SERVICES PROGRAM. (Attachments: # 1 Text of Proposed Order Proposed Order)(Gupta, Deepak) (Entered: 10/11/2022)",
            "date_created":"2022-10-11T14:21:40.855097-07:00",
            "entry_number":140,
            "date_modified":"2022-10-18T13:03:18.466039-07:00",
            "recap_documents":[
               {
                  "id":214672412,
                  "sha1":"3875437ed27ed37a221fa72c6b3925d0e9e733be",
                  "file_size":330183,
                  "is_sealed":null,
                  "thumbnail":"recap-thumbnails/gov.uscourts.dcd.178502/214672412.thumb.1068.png",
                  "ocr_status":2,
                  "page_count":28,
                  "date_upload":"2022-10-11T14:25:17.193565-07:00",
                  "description":"Motion for Settlement",
                  "filepath_ia":"https://archive.org/download/gov.uscourts.dcd.178502/gov.uscourts.dcd.178502.140.0.pdf",
                  "absolute_url":"/docket/4214664/140/national-veterans-legal-services-program-v-united-states/",
                  "date_created":"2022-10-11T14:21:40.859277-07:00",
                  "is_available":true,
                  "pacer_doc_id":"04509519006",
                  "date_modified":"2022-10-18T13:03:18.474889-07:00",
                  "document_type":1,
                  "filepath_local":"recap/gov.uscourts.dcd.178502/gov.uscourts.dcd.178502.140.0.pdf",
                  "document_number":"140",
                  "is_free_on_pacer":null,
                  "thumbnail_status":1,
                  "attachment_number":null,
                  "ia_upload_failure_count":null
               },
               {
                  "id":214673087,
                  "sha1":"",
                  "file_size":113664,
                  "is_sealed":null,
                  "thumbnail":null,
                  "ocr_status":null,
                  "page_count":6,
                  "date_upload":null,
                  "description":"Text of Proposed Order Proposed Order",
                  "filepath_ia":"",
                  "absolute_url":"/docket/4214664/140/1/national-veterans-legal-services-program-v-united-states/",
                  "date_created":"2022-10-11T14:23:38.689428-07:00",
                  "is_available":false,
                  "pacer_doc_id":"04509519007",
                  "date_modified":"2022-10-12T06:25:06.670192-07:00",
                  "document_type":2,
                  "filepath_local":null,
                  "document_number":"140",
                  "is_free_on_pacer":null,
                  "thumbnail_status":0,
                  "attachment_number":1,
                  "ia_upload_failure_count":null
               }
            ],
            "pacer_sequence_number":584,
            "recap_sequence_number":"2022-10-11.001"
         }
      ]
   },
   "webhook":{
      "version":1,
      "event_type":1,
      "date_created":"2022-10-11T14:21:40.855097-07:00",
      "deprecation_date":null
   }
}
```

cl/opinion_page/templates/includes/de_list.html
albertisfu: Yeah, this seems hard. I think this would require some custom CSS where you wrap both buttons within a div using a class with `display: table;`. Then, on each button, use `display: table-cell;` and adjust the buttons' widths to sizes that make them look aligned.

Do you want to try it? 

btw custom css can be added in: `override.css`

cl/opinion_page/templates/includes/de_list.html
albertisfu: I've incorporated the modal to request users to sign in if they're logged out.

cl/favorites/views.py
albertisfu: Yeah this would require an asynchronous request. We usually use HTMX for this kind of request, but an alternative can be done with a simple AJAX request.

For HTMX, it'll require the button to be rendered as a form in the frontend and use a submit method. You can see an example of this in `webhooks-form-create.html`.

In this case, it redirects to the create method of `webhooks-list`, which is handled by this method:
https://github.com/freelawproject/courtlistener/blob/main/cl/users/api_views.py#L60

This is an API method, but it can also be a regular view like the one you created. The difference is that the method should respond with something like:

```
return Response(
    status=HTTPStatus.CREATED,
    headers={"HX-Trigger": "webhooksListChanged"},
)
```

This returns an `HX-Trigger`, in this case called "webhooksListChanged", which is the signal to update something in the DOM. In this case, it can be the count for the specific Pray button that can also update the button status in case the new request made the user ineligible to request more prayers.

It could also be a good idea to add a spinner while sending the request to the server. This is also used in the webhook template referenced above.

Here are more details about HTMX:
https://htmx.org/docs/

Let me know if this is helpful.

cl/opinion_page/templates/includes/de_list.html
mlissner: We're redoing all of our HTML very soon, I'm frightened to say, so the normal reasons for separating out things like this might not apply. I've never allowed inline CSS before, but maybe this is fine?

cl/favorites/views.py
mlissner: I haven't looked at the JS here at all, but I'll reiterate that I love HTMX and you can learn it in about an hour, probably. 

> It could also be a good idea to add a spinner while sending the request to the server. This is also used in the webhook template referenced above.

I, um, think you can do a weird thing here. Increment as soon as clicked, then decrement if it fails. From the user's perspective, 99% of the time it just works. The 1% where the server fails or whatever, the user sees the prayer decrement, and presses the button again. No spinner needed, instant feedback for the user.

mlissner:
> So giving users the option to delete their waiting prayers can be omitted if Mike agrees.

I actually have an opinion on this one. On stackoverflow you can't undo your votes and it regularly bugs me. If I were using the prayer system and made a mistake or found something I wanted more, I'd want to be able to undo my prayer.

>> so maybe we should use a waffle flag to hide them momentarily.

> Great. @mlissner let me know if we intend to merge this one and leave the remaining UI work for a different PR so I can set a waffle flag here.

I'm also just a really big fan of waffles, because they'll let us get at least one public feedback loop before launching to everybody, so I'd say yes, let's add a waffle for the front end. I'd prefer to land the API in one PR and front end in another, but it seems hard at this point, so it's all good.

> For instance, I envision one use case of the API being a user being able to request new documents on a docket (after they're informed of them via webhook).

This is cool, but be careful, if we envision people doing this we'll need to document the darned API! I sort of don't care, honestly. If one of our API or webhook users wants documents badly, but isn't willing to pay for them, eh, whatever, they can figure out the API without docs if they want, and that's fine.

Thank you both, this is feeling very close. 

@albertisfu, one question for you. Do you think a separate review of this would be helpful? You've done a lot of work that @v-anne has reviewed, but that other FLP staff has not. What do you think?

cl/favorites/api_views.py
mlissner: Yes, see details below.

v-anne:
> I actually have an opinion on this one. On stackoverflow you can't undo your votes and it regularly bugs me. If I were using the prayer system and made a mistake or found something I wanted more, I'd want to be able to undo my prayer.

Done, I made this change.

> I'm also just a really big fan of waffles, because they'll let us get at least one public feedback loop before launching to everybody, so I'd say yes, let's add a waffle for the front end. I'd prefer to land the API in one PR and front end in another, but it seems hard at this point, so it's all good.

Added waffle flag "pray-and-pay" to the button.

I will make the HTMX change and then I think this will be ready for some final reviews.

albertisfu:
> Added waffle flag "pray-and-pay" to the button.
> 
> I will make the HTMX change and then I think this will be ready for some final reviews.

Thanks! I checked the waffle and worked well. Just tweaked the indentation.

Now that we have the waffle in place. I think the remaining fronted work can be done in a different PR if you want this to be merged first.

> @albertisfu, one question for you. Do you think a separate review of this would be helpful? You've done a lot of work that @v-anne has reviewed, but that other FLP staff has not. What do you think?

@mlissner Yeah, I think it'd be helpful if someone else review the work we've done here.

v-anne:
Sure, why not merge what we have done and leave the button for a final PR?

mlissner:
This all sounds great. I'll ask Eduardo to give it a review and we can merge once we've integrated any changes he identifies. Great work y'all!

v-anne:
By the way, @mlissner, what changes did you want to be made to the UI?

mlissner:
I think I'm holding my thoughts until I play with it a bit. Probably will have more constructive ones that way.

v-anne:
Ok, this is good to go from my end. Awaiting @ERosendo's review.

ERosendo (changes requested):
Thanks for the PR, @v-anne. The code looks good, however I have a few minor suggestions:

- I noticed a function named `get_lifetime_prayer_stats` that doesn't seem to be used within this PR. Could you please clarify if this function is intended to be part of this pull request or if it's a placeholder for a future feature?

- Before merging this PR, we should address the feedback regarding HTMX. Currently, after clicking the prayer button, the user is redirected to a page that simply says "It worked." This isn't the ideal user experience.  

@mlissner, I'm happy to help implement the HTMX code. It looks pretty easy to add. Let me know if you'd like me to take the lead on this. I think I can finish it in under an hour.

v-anne:
@ERosendo, thanks for your feedback. I responded to the suggestions accordingly.

As to your other comments:

> I noticed a function named get_lifetime_prayer_stats that doesn't seem to be used within this PR. Could you please clarify if this function is intended to be part of this pull request or if it's a placeholder for a future feature?

It is a placeholder for the future. I hoped to show how many documents users received as part of this feature, but it's not necessary to be used until the feature has been in place for some time.

> Before merging this PR, we should address the feedback regarding HTMX. Currently, after clicking the prayer button, the user is redirected to a page that simply says "It worked." This isn't the ideal user experience.

I'd definitely appreciate your help with the HTMX, but I think the current consensus was that the waffle flag would prevent anyone from seeing the button until this is fixed. So it can be done in either this PR or a future one.

ERosendo:
> It is a placeholder for the future

@v-anne Got it!. Let's also remove another unused import in the `views.py` file within the favorites module. We're importing this helper method but we're not using it.

> I think the current consensus was that the waffle flag would prevent anyone from seeing the button until this is fixed. So it can be done in either this PR or a future one.

I recommend refactoring the button in this PR to centralize the discussion on design and user experience. Additionally, the refactor should address these user experience issues:

- The button currently lacks a visual indicator to help users quickly identify previously requested documents. To enhance clarity, we should apply a distinct style. We can use the `prayer_exists` property to update the template. 

- The button will become disabled once the daily prayer limit is reached, preventing users from adding more prayers. However, users should still be able to remove existing prayers if they make a mistake or discover another document they wanted more.

Also, we should explore another method for setting the left margin of the button, potentially avoiding inline styles.

v-anne:
> I recommend refactoring the button in this PR to centralize the discussion on design and user experience. Additionally, the refactor should address these user experience issues:
The button currently lacks a visual indicator to help users quickly identify previously requested documents. To enhance clarity, we should apply a distinct style. We can use the `prayer_exists` property to update the template.

I agree, this is a good thing to change. I'm not sure about how to model this on the frontend. What do you recommend?

> The button will become disabled once the daily prayer limit is reached, preventing users from adding more prayers. However, users should still be able to remove existing prayers if they make a mistake or discover another document they wanted more.

Users should be able to remove their prayers now, as the `delete_prayer_view` function was created with that intention.

> Also, we should explore another method for setting the left margin of the button, potentially avoiding inline styles.

Mike said he was planning to revamp the entire site's HTML in the near future, so he was ok with leaving it hardcoded for now. However, we can change it if you prefer.

ERosendo:
> Users should be able to remove their prayers now, as the `delete_prayer_view` function was created with that intention.

@v-anne I understand that we can use delete_prayer_view to remove a prayer, but I've noticed that after creating the fifth prayer of the day, all buttons become disabled. This means there's no way to undo the last prayer or any previous ones. Am I missing something?

Please refer to this GIF for a demonstration: 

Screen Recording 2024-10-14 at 5 57 18 PM


After reaching the daily prayer limit, I'm unable to undo any of my prayers because the buttons are disabled. I'll address this issue in the refactor

v-anne:
Sorry about that @ERosendo. I didn't catch that issue in my local testing.

I suspect it was caused by the following line in `de_list.html`:

```html
class="btn btn-default btn-xs prayer-button {% if not prayer_eligible %}disabled{% endif %}"
```

ERosendo:
> Sorry about that @ERosendo. I didn't catch that issue in my local testing.

Don't worry @v-anne 

@mlissner check out this GIF showing the button's functionality with htmx. This refactor addresses the issues highlighted in https://github.com/freelawproject/courtlistener/pull/4519#issuecomment-2412262071. 

Screen Recording 2024-10-15 at 12 18 17 AM

Let me know your thoughts.

v-anne:
I like your changes to it. It's easier to see what documents a user has requested. The only suggestion I have is to more clearly disable the button (perhaps by making it unclickable) for documents that users haven't previously requested once their quota is exhausted.

ERosendo:
> Also, do the different states of the button have corresponding mouseover text?

Yes @v-anne. Here are screenshots: 

Screenshot 2024-10-15 at 4 59 28 PM

Screenshot 2024-10-15 at 4 59 59 PM

v-anne:
Just want to note a bug I found when testing locally: when trying to toggle a button that is currently untoggled, the mouseover reads "You have reached your daily request limit" even if I haven't exhausted the quota.

v-anne:
One last thing -- I think it would be helpful to have an explanation on `top_prayers.html` that explains to users how all of this works. I'm open to adding that in a subsequent PR with any API documentation (if needed), assuming the feature will gradually be rolled out to users.

mlissner:
I finally caught up here. I haven't looked at the code, but if everybody's happy and it's behind a waffle, let's merge. @ERosendo or @albertisfu, one of you want to go for it? 

I think this feature needs a lot of explanation to users before we launch it, both on the top-prayers page, and on the docket pages themselves, but let's get this merged and then we can play with that with some feedback from others.

ERosendo:
@mlissner @v-anne Here's a little enhancement to the button: 

Screen Recording 2024-10-15 at 6 44 45 PM


@mlissner I created the flag using the admin panel. I'll add one last commit and then merge this PR

ERosendo (approved):

mlissner:
I revered this PR in #4582 to fix #4581. No idea what was going on, but reverting fixed it. Let's see if we can figure that out, get a fix, and then do another PR.

Carter4242:
I'm like 75% sure when I refreshed the page a new button appeared on that page. Either Tags or Add Notes. I think there was only one drop down before so I'm betting it was Tags, but I'm not sure about that. Might be a clue to what's happening

v-anne:
That's disappointing, but we'll figure it out.

Mike, do you know if this PR also created issues with the RECAP extension? On PACER I no longer see the button to save a document to CourtListener.

ERosendo:
Sorry @v-anne :( .

 @mlissner I found the issue. In commit 09a9dd7, I added HTMX by updating the `footer-scripts` section of one of the templates we use on the docket page. However, I overlooked that this template extends from another HTML file that already includes scripts. As a result, my code overwrote the scripts section of the parent template, causing the issue described in ticket #4581.

To avoid this conflict, we should simply include the `pay_and_pray.js` file directly in the parent template, which is the file named `docket_tabs.html`.

mlissner:
I thought it would be something like that, thanks. 

I don't think this would also affect the extension, but we could open a new issue if needed for that.

PR #4520, opened 2024-10-01
feat(scrapers.update_from_text): new command

grossir:
Helps solve: https://github.com/freelawproject/juriscraper/issues/858

- New command to re-run Site.extract_from_text over downloaded opinions
- Able to filter by Docket.court_id ,  OpinionCluster.date_filed, OpinionCluster.precedential_status
- Updates tasks.update_from_document_text to return information for logging purposes
- Updates test_opinion_scraper to get a Site.extract_from_text method


cl/scrapers/management/commands/update_from_text.py
```diff
@@ -0,0 +1,159 @@
+from datetime import datetime
+
+from django.db import transaction
+
+from cl.lib.command_utils import VerboseCommand, logger
+from cl.scrapers.tasks import update_document_from_text
+from cl.search.models import PRECEDENTIAL_STATUS, Opinion, OpinionCluster
+
+
+def update_from_text(
```
flooie, line 10: can we make this - or the file a little more descriptive....

cl/scrapers/management/commands/update_from_text.py
```diff
@@ -0,0 +1,163 @@
...
+    """
+    with transaction.atomic():
+        changes = update_document_from_text(opinion, juriscraper_module)
+        if not changes:
+            logger.info("Did not get any metadata for opinion %s", opinion.id)
+            return
+
+        logger.info("Processing opinion %s", opinion.id)
+
+        # Check if changes exist before saving, to prevent unecessary DB queries
```
flooie, line 34: typo in `unnecessary`

cl/scrapers/management/commands/update_from_text.py
```diff
@@ -0,0 +1,163 @@
...
+            default="",
+            help=r"""A filter value in %Y/%m/%d format.
+            OpinionCluster.date_filed will have to be greater or equal""",
+        )
+        parser.add_argument(
+            "date-filed-lte",
+            default="",
+            help=r"""A filter value in %Y/%m/%d format.
+            OpinionCluster.date_filed will have to be less or equal""",
+        )
```
flooie, line 113: I tried to run this locally using opinion-ids and juriscraper-module and could not get it to run.  I had to comment out the date arguments.

flooie: Also - I havent mentioned this before but I dislike the format `%Y/%m/%d`. 

how do you feel about `%Y-%m-%d` - maybe its my American bias here.

cl/scrapers/management/commands/update_from_text.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Recommended use is to run over a sample of the target time period
+    and check if updates over Docket, OpinionCluster, Opinion and
+    Citation are as expected
+    """
+    stats = {}  # assigned at the end of a command run, for testing
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--juriscraper-module",
```
flooie, line 86: Perhaps this isnt possible - but do you think we could get away from the juriscraper module and just use court id?

cl/scrapers/management/commands/update_from_text.py
```diff
@@ -0,0 +1,163 @@
...
+        qs = OpinionCluster.objects.filter(**query).prefetch_related(
+            "sub_opinions"
+        )
+        for cluster in qs:
+            opinions = cluster.sub_opinions.all()
+            for op in opinions:
+                rerun_extract_from_text(op, juriscraper_module, stats)
+
+        logger.info("Modified objects counts: %s", stats)
+        self.stats = stats
```
flooie, line 163: maybe its just me but I prefer to have a clean handle method that doesnt contain large portions of code to run and passes that off.  do you think we could move this out of the handle into its own method.

cl/scrapers/management/commands/update_from_text.py
```diff
@@ -0,0 +1,163 @@
...
+            default="",
+            choices=[value for value, name in PRECEDENTIAL_STATUS.NAMES],
+            help="""A value of OpinionCluster.precedential_status. To be
+            used for filtering the Opinions to be processed
+            """,
+        )
+
+    def handle(self, *args, **options):
+        super().handle(*args, **options)
+        juriscraper_module = options["juriscraper_module"]
```
flooie, line 125: I think something like this 
```
        if not options['opinion_ids']:
            if not options['date_filed_gte'] or not options['date_filed_lte']:
                raise CommandError(
                    "Either --opinion-ids must be specified, or both --date-filed-gte and --date-filed-lte are required.")
```

after setting the date fields to optionals would be good

cl/scrapers/management/commands/update_from_text.py
flooie: maybe something like 
```
import pkgutil
import importlib
def find_juriscraper_module_from_court_id(court_id: str):
    # Import the base package
    package = importlib.import_module('juriscraper.opinions.united_states')
    for _, modname, ispkg in pkgutil.walk_packages(path=package.__path__,
                                                   prefix=package.__name__ + '.'):
        if modname.endswith(court_id):
            return modname
    return ""

court_id = "wis"
juriscraper_module = find_juriscraper_module_from_court_id(court_id)
```

flooie: I could be wrong but that would make this a lot easier to process - but perhaps there is a reason we always strictly follow module paths.

cl/scrapers/management/commands/update_from_text.py
```diff
@@ -0,0 +1,163 @@
...
+        if options["cluster_status"]:
+            query["precedential_status"] = options["cluster_status"]
+
+        qs = OpinionCluster.objects.filter(**query).prefetch_related(
+            "sub_opinions"
+        )
+        for cluster in qs:
+            opinions = cluster.sub_opinions.all()
+            for op in opinions:
+                rerun_extract_from_text(op, juriscraper_module, stats)
```
flooie, line 160: It's possible to have opinions that are merged with Harvard.  If thats the case this could bring back empty `plain_text` and `html` fields.  which will crash

flooie: See https://freelawproject.sentry.io/issues/6004622476/?project=5620212&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&stream_index=0
for an example I crashed locally.

cl/scrapers/tasks.py
```diff
@@ -54,12 +54,13 @@ def update_document_from_text(
...
     :param juriscraper_module: full module to get Site object
-    :return: None
+    :return: the extracted data dictionary
     """
     court = opinion.cluster.docket.court.pk
     site = get_scraper_object_by_name(court, juriscraper_module)
     if site is None:
-        return
+        logger.debug("No site found %s", juriscraper_module)
+        return {}
```
flooie, line 63: Perhaps we should just validate our modules so we dont end up here?

cl/scrapers/management/commands/update_from_text.py
```diff
@@ -0,0 +1,163 @@
...
+            )
+
+        court_id = juriscraper_module.split(".")[-1].split("_")[0]
+        gte_date = datetime.strptime(options["date_filed_gte"], "%Y/%m/%d")
+        lte_date = datetime.strptime(options["date_filed_lte"], "%Y/%m/%d")
+        query = {
+            "docket__court_id": court_id,
+            "date_filed__gte": gte_date,
+            "date_filed__lte": lte_date,
+        }
```
flooie, line 149: I would rather do this as validation when we start ... maybe something like this 

    def parse_date(self, date_string):
        """Parse a date string in the format YYYY/MM/DD."""
        try:
            return datetime.strptime(date_string, "%Y/%m/%d")
        except ValueError:
            from argparse import ArgumentTypeError
            raise ArgumentTypeError(f"Invalid date format: '{date_string}'. Expected format: YYYY/MM/DD.")

and then we can just assign the type 

        parser.add_argument(
            "--date-filed-gte",
            default="",
            type=self.parse_date,
            help=r"""A filter value in %Y/%m/%d format.
            OpinionCluster.date_filed will have to be greater or equal""",
        )

I think this both validates nicely and makes the code easier to read maybe.

cl/scrapers/tasks.py
```diff
@@ -70,14 +71,17 @@ def update_document_from_text(
             opinion.cluster.__dict__.update(data)
         elif model_name == "Citation":
             data["cluster_id"] = opinion.cluster_id
-            ModelClass.objects.get_or_create(**data)
+            _, citation_created = ModelClass.objects.get_or_create(**data)
+            metadata_dict["Citation"]["created"] = citation_created
```
flooie, line 75: Is it possible that the update will fail on OpinionCluster or Docket or is citation the only one where it could fail to create something.

flooie: Also - if we run this code - and we have a mistake in our regex pattern 
we could create citations with no known reporter - I wonder if we should validate that our citations are actually citations.

cl/scrapers/management/commands/update_from_text.py
grossir: I think we could support both passing a court_id or a full module; and the court_id argument should error if more than 1 court is found; so the user is forced to use the full module in that case

I think the full module is useful to differentiate between oral arguments and opinion scrapers for the same court, which would be ambiguous if only the court_id was passed. Given that this command should work on opinions only, that ambiguity doesn't count.

On the other hand some court_ids remain ambiguous, even in the context of opinions only. For example `lactapp_1` and `lactapp_5` scrape totally different sites; but their court_id is the same.

cl/scrapers/tasks.py
grossir: I just added a `logger.debug`, but that `if site is None` is there from before this PR.  I think trying to track all the cases where that condition may happen is out of the scope of this PR

cl/scrapers/management/commands/update_from_text.py
grossir: I think a Command's  `handle` should have the code to manipulate the input arguments, otherwise it would be just boilerplate taking up space. I actually tried to abstract it out, and the handle would look like this, which I don't think is good:
```python
     super().handle(*args, **options)
     some_other_function(options)
```

cl/scrapers/tasks.py
grossir: You are right we should validate `extract_from_text`  output before ingesting. 

We have 3 options for this:

1. change all juriscraper `Site.extract_from_text` to return citation strings to be parsed using `eyecite.find_citations`. This would unify the ingestion process between `extract_from_text` and the scrapers, but demands a lot of changes on juriscraper

2. validate the returned Citation objects, just checking that the reporter exists in `reporters-db` and that the reporter matches the Citation type returned

3. implement (2.) and also enable ingesting citation strings from `extract_from_text`, so we can start migrating the scrapers to (1.)


I have implemented (2.) for this PR

cl/lib/command_utils.py
```diff
@@ -22,6 +24,40 @@ def handle(self, *args, **options):
...
+            help=(
+                "The court(s) to scrape and extract. One of: "
+                "1. a python module or package import from the Juriscraper library, e.g."
+                "'juriscraper.opinions.united_states.federal_appellate.ca1' "
+                "or simply 'juriscraper.opinions' to do all opinions."
+                ""
+                "2. a court_id, to be used to lookup for a full module path"
+                "An error will be raised if the `court_id` matches more than "
+                "one module path. In that case, use the full path"
+            ),
```
flooie, line 57: Can I just clarify one thing.  This will allow us to keep our current scripts running using the original method right?

cl/lib/juriscraper_utils.py
```diff
@@ -42,3 +46,45 @@ def get_scraper_object_by_name(court_id: str, juriscraper_module: str = ""):
...
+
+
+def get_module_by_court_id(court_id: str, module_type: str):
+    """Given a `court_id` return a juriscraper module path
+
+    Some court_ids match multiple scraper files. These will force the user
+    to use the full module path. For example, "lactapp_1" and "lactapp_5"
+    match the same `court_id`, but scrape totally different sites, and
+    their Site objects are expected to have different `extract_from_text`
+    behavior
```
flooie, line 58: this is missing a return type I think.

flooie:
@grossir just a few comments.  This looks good and close.

cl/lib/command_utils.py
grossir: Yes, it's the same if you input a full module string; but it will also work if you send just a court id

cl/scrapers/management/commands/update_from_text.py
```diff
@@ -0,0 +1,200 @@
...
+            "date_filed__gte": options["date_filed_gte"],
+            "date_filed__lte": options["date_filed_lte"],
+        }
+
+        if options["cluster_status"]:
+            query["precedential_status"] = options["cluster_status"]
+
+        qs = OpinionCluster.objects.filter(**query).prefetch_related(
+            "sub_opinions"
+        )
```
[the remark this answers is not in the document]
grossir: You are referring to `OpinionCluster.source` containing a "C" for a Court Website source, right? 
So it will add this to the query: `{"source__contains": 'C'}`, because the cluster may have been updated

 I think it's a good idea, but I don't see an index for that field, so maybe it will slow down things; we will tell on prod I guess

cl/lib/command_utils.py
grossir: I just added tests for this; and retest it manually on my env; it works OK

cl/scrapers/management/commands/update_from_text.py
```diff
@@ -0,0 +1,201 @@
...
+            raise ValueError(
+                "Both `date-filed-gte` and `date-filed-lte` arguments should have values"
+            )
+
+        court_id = juriscraper_module.split(".")[-1].split("_")[0]
+        query = {
+            "docket__court_id": court_id,
+            "date_filed__gte": options["date_filed_gte"],
+            "date_filed__lte": options["date_filed_lte"],
+            "source__contains": "C",
```
flooie, line 172: lets not use "C" - it doesnt mean anything.  

`SOURCES.COURT_WEBSITE`

flooie (commented):
@grossir - can you make that one small tweak and otherwise looks good to me

grossir:
@flooie I updated the PR, please check again

flooie (approved):
Looks good to me

PR #4532, opened 2024-10-03
Add crosswalk_dir as argument in CAP import

quevon24:
I added the crosswalk directory argument for the CAP import command because it was not possible to specify it, i also made it required.

Related tests have been updated.

Usage example:
`docker exec -it cl-django python /opt/courtlistener/manage.py import_harvard_pdfs --crosswalk_dir /storage/tmp/crosswalk`

### DIFF
--- a/cl/search/management/commands/import_harvard_pdfs.py
+++ b/cl/search/management/commands/import_harvard_pdfs.py
@@ -44,6 +44,12 @@ class Command(BaseCommand):
         parser.add_argument(
             "--verbose", action="store_true", help="Increase output verbosity"
         )
+        parser.add_argument(
+            "--crosswalk-dir",
+            type=str,
+            help="Directory for reading crosswalk files",
+            required=True,
+        )
 
     def handle(self, *args: Any, **options: Any) -> None:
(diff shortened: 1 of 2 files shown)


cl/search/management/commands/import_harvard_pdfs.py
```diff
@@ -86,11 +100,9 @@ def process_crosswalks(
         logger.info(
             f"Processing crosswalks. Reporter: {specific_reporter}, Resume: {resume}"
         )
-        crosswalk_dir = os.path.join(
-            os.path.dirname(os.path.dirname(os.path.dirname(__file__))),
-            "crosswalks",
+        last_reporter_file = os.path.join(
+            self.crosswalk_dir, "last_completed_reporter.txt"
         )
-        last_reporter_file = "last_completed_reporter.txt"
```
flooie, line 106: why did this not work ?

quevon24: It will work, but will save the file in the current working directory instead of the crosswalk directory, i think all related files should be in the same directory

PR #4563, opened 2024-10-11
feat(cl_back_scrape_oral_arguments): add backscrape-start and backscrape-end optional args

grossir:
- Abstract into a function the backscraper `parser.add_argument` calls,
to prevent duplication
- Add dynamic backscraping support to oral argument scrapers
- Support days_interval optional argument, as expected in https://github.com/freelawproject/juriscraper/pull/1108


flooie:
Thanks

PR #4575, opened 2024-10-15
Update pacer free document command to ignore documents without case number

quevon24:
Small change to skip the document if it doesn't have a case number.
See issue #4547

When it is ready we can resume the sweep.


mlissner:
Got it, thanks Kevin! Let's get it cruising again!

PR #4577, opened 2024-10-15
feat(pghistory): enable Events admin for single object's history

grossir:
- Sets PGHISTORY_ADMIN_ALL_EVENTS to False
- Sets PGHISTORY_ADMIN_CLASS to a custom class that will only show data when filter by a single object, thus preventing unnecesary DB queries, like loading all events, or loading all events for an object type

### DIFF
--- a/cl/settings/django.py
+++ b/cl/settings/django.py
@@ -136,6 +136,7 @@ ROOT_URLCONF = "cl.urls"
 
 INSTALLED_APPS = [
     "daphne",
+    "pghistory.admin",
     "django.contrib.admin",
     "django.contrib.admindocs",
     "django.contrib.contenttypes",
@@ -189,6 +190,8 @@ if DEVELOPMENT:
 
 ASGI_APPLICATION = "cl.asgi.application"
 
+# Disable /admin/pghistory/events/ view that shows all the events without any filter
+PGHISTORY_ADMIN_ALL_EVENTS = False
 
 ################
 # Misc. Django #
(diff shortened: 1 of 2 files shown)


grossir:
pghistory admin list page has 3 entry points:
1. from the sidebar  that lists all admins for each app. Looks like: http://0.0.0.0:8000/admin/pghistory/events/
2. from an admin list page, for example, when checking the list of `search.Dockets`. Looks like: http://0.0.0.0:8000/admin/search/docket/
3. from a single object details page (for example, a single Docket)

- Entry point  (1) would load the whole Event dataset if one does not set `PGHISTORY_ADMIN_ALL_EVENTS` to False. However, one can still use a bunch of filters. For example, one could filter by all update actions across models, which would be a heavy DB hit and doesn't seem very useful.

- Entry point (2) returns all the events filtered by that model. Following our example, it would load all Docket events, allowing to click filters to reduce the queryset

With the proposed changes, the Events page would look like this when accesed through (1) and (2)
image



- The third entry point filters by an object ID, loading all events for that object, (in our example, all the changes for a Docket). *This is the targeted use case, to try and understand what happened with the changes of a single object.*

image

mlissner:
I think we just want to see events per object, right? A list of all events doesn't seem useful at all?

grossir:
Yes, this PR makes the pghistory events admin work only when filtering  by an object

quevon24 (approved):
Everything looks good, it is a nice way to avoid having to do a big query in main events  page and check the events only for specific objects, i'll set it to automerge

PR #4580, opened 2024-10-16
feat(favorites): frontend revisions to pray and pay project

v-anne:
This is the PR for making frontend changes to the pray and pay user interface.

Mike had some great comments in #4507, so we'll start with them and see what's needed once the low-hanging fruit is done.

Additions include:
- Allowing users to request a document from the `recap_document.html` page.
- Revamping UI of `top_prayers.html` to provide better UX.
- Creating `user_prayers.html` page to allow users to see their open and fulfilled prayers, and also summary statistics about how many requests they've had fulfilled. In a future PR, this will be made similar to tags to allow people (e.g., freelance journalists) to share documents they've requested.
- documentation for how pray and pay project works.
- minor grammar changes to email notifications.


v-anne:
@ERosendo this is the PR I'm going to build off of.

mlissner:
Seeing some commit messages around the user-prayer page. Reminds me that it doesn't *have* to be available in the profile section, just like tags aren't. The user prayers page could just be linked from the mouseover for each emoji, for example.

v-anne:
Yeah, tags are what I am aiming to emulate.

v-anne:
image
I'm having some issues getting the three buttons to display inline with each other. Ideally, the prayer button would be to the right of the other two.

v-anne:
image

Having same issue with the button on the modal. There's also a related issue where if I click the prayer button repeatedly, the count will increase despite having exhausted the quota.

ERosendo:
@mlissner sure! 👍

v-anne:
@ERosendo in addition to those two issues with the buttons, I've mostly finished a MVP of the `user_prayers.html` page. 

This is what it currently looks like. The status column is broken; I can't figure out why the backend isn't passing the information through to the template. I intend to add a button to allow users to delete requests and one for them to download available documents as well.

image

v-anne:
@mlissner @ERosendo I've finished up my end of this. I've highlighted some concerns of mine regarding bits that might merit closer scrutiny.

mlissner:
Can you post a few more screenshots for me, @v-anne?

v-anne:
image
image

ERosendo (changes requested):
@v-anne Here are my initial comments on the PR. I'll help you with the alignment issues in a bit.

cl/favorites/utils.py
```diff
@@ -211,19 +250,46 @@ async def get_user_prayer_history(user: User) -> tuple[int, float]:
     return count, total_cost
 
 
-async def get_lifetime_prayer_stats() -> tuple[int, int, float]:
+async def get_lifetime_prayer_stats(
+    status: int,
+) -> tuple[
+    int, int, float
+]:  # status can be only 1 (WAITING) or 2 (GRANTED) based on the Prayer model
```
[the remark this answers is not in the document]
mlissner: Seems like we need a rule or linter to this effect. It's really just too tempting to return a tuple like this when a dataclass is almost universally better. Hm.

mlissner:
@v-anne, I noticed this PR is paused. Are you waiting for feedback or something else?

v-anne:
Sorry @mlissner, I had stepped back because I thought @ERosendo was going to take a look at some of the issues I had raised with my code.

mlissner:
Got it. We're moving to a sprint/scrum model, but I'll get this on his new list. :)

v-anne:
Not related to these changes Mike, but I think there might be some issues with email. I'm not receiving alerts for every fulfilled request.

mlissner:
> Not related to these changes Mike, but I think there might be some issues with email. I'm not receiving alerts for every fulfilled request.

I checked your account and we've sent you 41 emails about pay-and-pray so far. If you want to debug a particular one, emails is probably the better way to go.

ERosendo:
Fixed alignment issue in `recap_document.html` and `redirect_to_pacer_modal.html` templates.


### recap_document.html:

image

----

image


----

image


### redirect_to_pacer_modal.html:

image

----

image

----

image

ERosendo (approved):

ERosendo:
@v-anne Thanks for the hard work in this PR. 

@mlissner In addition to the alignment corrections addressed in comment https://github.com/freelawproject/courtlistener/pull/4580#issuecomment-2523385303, I've refined the help page for the Pray and Pay project to include accurate screenshots and updated the user list of prayers to utilize HTMX for the delete button.

Here's the updated help page:

Screen Recording 2024-12-16 at 2 45 09 PM


Here's a GIF showing the new user prayer list.

Screen Recording 2024-12-16 at 2 53 22 PM

mlissner:
Seems OK to me, generally, thanks. Let's get one more set of eyes on this though, because there's a lot of code here. 

Thank you!

v-anne:
It looks pretty good to my eyes. Local testing seemed alright as well.

I'll defer to another reviewer to decide whether it's ready to merge. I think this will be a great feature in the years ahead!

cl/favorites/templates/user_prayers.html
```diff
@@ -0,0 +1,145 @@
...
+              <th>Case Name</th>
+              <th>Document Number</th>
+              <th>Document Description</th>
+              <th>Requested On</th>
+              <th>Status</th>
+              {% if is_page_owner %}<th>Action</th>{% endif %}
+            </tr>
+        </thead>
+        <tbody>
+            {% for prayer in prayers %}
```
albertisfu, line 52: I found it a bit confusing that we call this `prayers`, but they are actually `RECAPDocument` instances. Should we rename these variables?

cl/favorites/templates/user_prayers.html
```diff
@@ -0,0 +1,145 @@
...
+            <tr>
+              <td>{{ prayer.docket_entry.docket.court.citation_string }}</td>
+              {% with docket=prayer.docket_entry.docket %}
+              <td>
+                <a href="{% url "view_docket" docket.pk docket.slug %}">
+                  {{ prayer.docket_entry.docket|best_case_name|safe|v_wrapper }} ({{ prayer.docket_entry.docket.docket_number }})
```
albertisfu, line 58: We can simplify this line to:

```suggestion
                   {{ docket|best_case_name|safe|v_wrapper }} ({{ docket.docket_number }})
```

cl/favorites/templates/user_prayers.html
```diff
@@ -0,0 +1,145 @@
...
+                </a>
+              </td>
+              <td>
+                <a href="{% url "view_docket" docket.pk docket.slug %}#entry-{{ prayer.docket_entry.entry_number }}">
+                  {{ prayer.document_number }}
+                </a>
+              </td>
+              {% endwith %}
+              <td>{{ prayer.description }}</td>
+              <td>{{ prayer.date_created|date:"M j, Y"  }}</td>
```
albertisfu, line 68: This is displaying the `date_created` of the RECAPDocument. To display the `date_created` of the `Prayer`, we should use:

`{{ prayer.prayer_date_created|date:"M j, Y"  }}`

cl/favorites/templates/user_prayers.html
```diff
@@ -0,0 +1,145 @@
...
+              <td>{% if prayer.prayer_status == 1 %} Pending {% elif prayer.prayer_status == 2 %}
+                <div class="btn-group hidden-xs col-sm-4 col-md-3 hidden-print flex">
+                  {% if rd.filepath_local %}
+                    <a href="{{ rd.filepath_local.url }}"
+                        rel="nofollow"
+                        role="button"
+                        class="btn btn-primary btn-xs"
+                            {% if rd.date_upload %}
+                        title="Uploaded {{ rd.date_upload|timezone:timezone }}"
```
albertisfu, line 77: This line raises a `VariableDoesNotExist` error when loading the page with a `GRANTED` document because the `timezone` is not defined.

If we are going to display `date_upload` in UTC, we could use the custom filter: `datetime_in_utc`:

`{{ prayer.date_upload|datetime_in_utc }}`

This issue is also present in `prayer_email.html`

which could be the reason some emails are being missed, as seen in sentry: [COURTLISTENER-8FC](https://freelawproject.sentry.io/issues/6026700048/)

cl/favorites/utils.py
```diff
@@ -117,11 +124,16 @@ async def get_top_prayers() -> list[RECAPDocument]:
             "attachment_number",
             "pacer_doc_id",
             "page_count",
+            "is_free_on_pacer",
             "description",
+            "docket_entry__entry_number",
             "docket_entry__docket_id",
             "docket_entry__docket__slug",
+            "docket_entry__docket__case_name",
```
albertisfu, line 132: Since `best_case_name` can also use `case_name_full` and `case_name_short`. We should add them here to avoid additional queries.

cl/favorites/utils.py
```diff
@@ -144,7 +156,47 @@ async def get_top_prayers() -> list[RECAPDocument]:
...
+            "docket_entry__docket__court",
+        )
+        .only(
+            "pk",
+            "document_type",
+            "document_number",
+            "attachment_number",
+            "pacer_doc_id",
+            "page_count",
+            "is_free_on_pacer",
```
albertisfu, line 179: Similarly to `get_top_prayers` we should include `case_name_full` and `case_name_short`.

Additionally, some other fields are being used in the template:
```
date_upload
filepath_local
filepath_ia
```

cl/favorites/utils.py
```diff
@@ -200,30 +252,76 @@ def send_prayer_emails(instance: RECAPDocument) -> None:
...
 async def get_user_prayer_history(user: User) -> tuple[int, float]:
-    filtered_list = Prayer.objects.filter(user=user, status=Prayer.GRANTED)
+    filtered_list = Prayer.objects.filter(
+        user=user, status=Prayer.GRANTED
+    ).select_related("recap_document")
 
     count = await filtered_list.acount()
 
     total_cost = 0
     async for prayer in filtered_list:
```
albertisfu, line 262: I'm concerned about the performance of this `total_cost` computation if the user has thousands of prayers.

Would it be possible to use a similar approach, relying on a database query like the one you used in `get_lifetime_prayer_stats` to compute the cost?

cl/favorites/utils.py
```diff
@@ -200,30 +252,76 @@ def send_prayer_emails(instance: RECAPDocument) -> None:
...
+            price=Case(
+                When(recap_document__is_free_on_pacer=True, then=Value(0.0)),
+                When(recap_document__page_count__gt=30, then=Value(3.0)),
+                When(
+                    recap_document__page_count__gt=0,
+                    recap_document__page_count__lte=30,
+                    then=F("recap_document__page_count") * 0.10,
+                ),
+                default=Value(0.0),
+            )
```
albertisfu, line 308: I think it is possible to simplify this annotation a bit by relying on `Least`:

```suggestion
            price=Case(
                When(recap_document__is_free_on_pacer=True, then=Value(0.0)),
                When(
                    recap_document__page_count__gt=0,
                    then=Least(
                        Value(3.0),
                        F("recap_document__page_count") * Value(0.10)
                    ),
                ),
                default=Value(0.0),
            )
```

I would also suggest adding a test for this method since it is a bit complex. This will help guarantee its functionality and ease future maintainability.

cl/favorites/utils.py
```diff
@@ -200,30 +252,76 @@ def send_prayer_emails(instance: RECAPDocument) -> None:
...
+
+    total_cost = total_cost["price__sum"]
+
+    data = {
+        "count": prayer_count,
+        "num_distinct_purchases": distinct_prayers,
+        "total_cost": f"{total_cost:,.2f}",
+    }
+    one_day = 60 * 60 * 24
+    await cache.aset(cache_key, data, one_day)
```
albertisfu, line 321: We are caching the `data` here but it seems we are not using the previously cached values if they exist to avoid running the query again

cl/favorites/views.py
```diff
@@ -255,6 +269,38 @@ async def delete_prayer_view(
...
+    request: HttpRequest, username: str
+) -> HttpResponse:
+    requested_user = await aget_object_or_404(User, username=username)
+    is_page_owner = await request.auser() == requested_user
+
+    # this is a temporary restriction for the MVP. The intention is to eventually treat like tags.
+    if not is_page_owner:
+        return redirect("top_prayers")
+
+    prayers = await get_user_prayers(requested_user)
```
albertisfu, line 290: Similar to a previous comment, I got a bit confused since this method returns `RECAPDocuments`. Should we rename it to something like `rd_prayers`?

cl/opinion_page/views.py
```diff
@@ -723,6 +743,7 @@ async def view_recap_document(
             "timezone": COURT_TIMEZONES.get(d.court_id, "US/Eastern"),
             "redirect_to_pacer_modal": redirect_to_pacer_modal,
             "authorities": await rd.cited_opinions.aexists(),
+            "prayer_eligible": prayer_is_eligible,
```
albertisfu, line 746: It seems this context variable is not being used in any template. It's also present in `view_docket`.

albertisfu (changes requested):
This looks great. I really like how this feature is taking shape!

I've left some comments and suggestions above.

Additionally, I think it would be a good idea to add pagination to `user_prayers_view`, considering that some users may have thousands of prayers, which can impact performance. Or not sure if you're already considering this but it's not within the scope of this PR.

cl/favorites/templates/user_prayers.html
ERosendo: You're right! Thanks for catching this.

cl/favorites/templates/user_prayers.html
ERosendo: Given that we're iterating over documents with the highest prayer count, I think we can name the variable `rd_with_prayers`. What are your thoughts?

cl/favorites/templates/user_prayers.html
ERosendo: You're correct!

cl/favorites/templates/user_prayers.html
ERosendo: Great catch. Thanks!

cl/favorites/templates/user_prayers.html
albertisfu: yeah I agree!

cl/favorites/utils.py
ERosendo: alright!

cl/favorites/views.py
ERosendo: I apologize, I unintentionally removed this in commit [5fcb609](https://github.com/freelawproject/courtlistener/pull/4580/commits/5fcb609666bdb8e215391f60850c19c1ca1d91f4)

cl/favorites/utils.py
ERosendo: I apologize, I unintentionally removed this in commit [5fcb609](https://github.com/freelawproject/courtlistener/pull/4580/commits/5fcb609666bdb8e215391f60850c19c1ca1d91f4)

cl/favorites/views.py
ERosendo: I agree. I'll rename this variable

cl/opinion_page/views.py
ERosendo: You're correct! Thanks for bringing this to my attention.

cl/favorites/utils.py
ERosendo: > Would it be possible to use a similar approach, relying on a database query like the one you used in get_lifetime_prayer_stats to compute the cost?

Yes, that's possible. I'll update this helper function accordingly.

ERosendo:
@albertisfu Thanks for your feedback. I've implemented all your suggestions.

cl/favorites/templates/prayer_email.html
```diff
@@ -44,7 +45,7 @@ <h3 class="alt bottom" style="font-size: 1.5em; font-weight: normal; line-height
           <td style="text-align: center">{{ docket_entry.entry_number }}{% if rd.attachment_number  %}-{{ rd.attachment_number }}{% endif %}</td>
           <td>
             {% if docket_entry.datetime_filed %}
-              <span title="{{ docket_entry.datetime_filed|timezone:timezone}}">{{ docket_entry.datetime_filed|timezone:timezone|date:"M j, Y" }}</span>
+              <span title="{{ docket_entry.datetime_filed|datetime_in_utc}}">{{ docket_entry.datetime_filed|datetime_in_utc|date:"M j, Y" }}</span>
```
albertisfu, line 48: In this line: `{{ docket_entry.datetime_filed|datetime_in_utc|date:"M j, Y" }}`, an empty value is displayed.

I believe the issue is that the `datetime_in_utc` filter handles the date format internally, so applying an additional format causes it to fail. If we want to display a different format here, we might need to use a different approach, such as modifying `datetime_in_utc` to accept an optional custom format.

Same for the .txt version

cl/favorites/utils.py
```diff
@@ -200,30 +259,99 @@ def send_prayer_emails(instance: RECAPDocument) -> None:
...
+                    then=Least(
+                        Value(3.0),
+                        F("recap_document__page_count") * Value(0.10),
+                    ),
+                ),
+                default=Value(0.0),
+            )
+        )
+        .aaggregate(Sum("price", default=0.0))
+    )
```
albertisfu, line 285: I think we can avoid the duplicate code in this method and in `get_lifetime_prayer_stats` by encapsulating the shared logic into a separate method, something like:

```
async def compute_cost(queryset):
    return await (
        queryset.values("recap_document")
        .distinct()
        .annotate(
            price=Case(
                When(recap_document__is_free_on_pacer=True, then=Value(0.0)),
                When(
                    recap_document__page_count__gt=0,
                    then=Least(
                        Value(3.0),
                        F("recap_document__page_count") * Value(0.10),
                    ),
                ),
                default=Value(0.0),
            )
        )
        .aaggregate(Sum("price", default=0.0))
    )
```

And then just using:
`cost =  await compute_cost(filtered_list)`

OR:

`total_cost = await compute_cost(prayer_by_status.select_related("recap_document"))
`

cl/favorites/utils.py
```diff
@@ -200,30 +259,99 @@ def send_prayer_emails(instance: RECAPDocument) -> None:
...
+
+    total_cost = total_cost["price__sum"]
+
+    data = {
+        "count": prayer_count,
+        "num_distinct_purchases": distinct_prayers,
+        "total_cost": f"{total_cost:,.2f}",
```
albertisfu, line 348: Can we use the same key names here as in `PrayerStats`?

I believe if we do something like:

```suggestion
        "prayer_count": prayer_count,
        "distinct_count": distinct_prayers,
        "total_cost": f"{total_cost:,.2f}",
```

Then in both returns we can do:
`return PrayerStats(**data)`

albertisfu (commented):

Just left a couple of additional comments and suggestions.

cl/favorites/templates/prayer_email.html
ERosendo: we can pass `timezone` as a template variable and use this custom format, right?

cl/favorites/utils.py
ERosendo: You're correct! I'll create this new helper function.

cl/favorites/utils.py
ERosendo: You're correct! Thanks for bringing this to my attention.

cl/favorites/templates/prayer_email.html
albertisfu: yeah, that's also a valid approach!

ERosendo:
@albertisfu Ready for another review 😃

albertisfu (approved):
Thanks! this now looks good to me :shipit:

mlissner:
Right on. Into the maw it goes!! Thank you all for the teamwork polishing and refining this one. I'm looking forward to giving it a try.

PR #4586, opened 2024-10-17
fix(docs): correct quoting in examples with `jq`

cweider:
The example commands in the search API documentation (v4 and v3) were missing a necessary closing single quote.


mlissner (approved):

mlissner:
Thanks. Set for auto merge!

PR #4594, opened 2024-10-18
Update cap crosswalk command

quevon24:
The command has been updated to improve execution time, now it uses an exact match for filepath_json_harvard field to be able to use the filepath_json_harvard index (I need to check that it exists).

Originally the command used icontains to match cap id to a cluster but the problem is that the index is not being used which slows down the command, more details can be seen in the issue: https://github.com/freelawproject/infrastructure/issues/170

I also removed some duplicate log messages


mlissner (approved):

quevon24:
> LGTM, but I can't merge b/c it's still a draft.

Yes, I noticed that in this command we did not have a parameter to start from a specific reporter, I added it, it is now ready.

mlissner (approved):
Cool, nice one. Thank you!

PR #4604, opened 2024-10-21
chore(docker): improve usage of `CL_BASE_DIR`

cweider:
`docker-compose.yml` was going about specifying paths relative to the base directory in a handful of ways. Some were just convoluted (see `cl-postgresql`) while others were incorrect (see `cl-es`). With this patch, the `courtlistener` directory is consistently referred to with `${CL_BASE_DIR:-../../}`.


mlissner:
Nice, thank you.

PR #4611, opened 2024-10-23
Fix event admin page for pghistory

quevon24:
A quick fix for pghistory admin page to load events from a single object.


grossir (commented):
@quevon24 I was checking this on my local env and the changes in your commit are still doing the giant UNION ALL query / CTE. That CTE is described as being a scaling bottleneck by pghistory's [own](https://django-pghistory.readthedocs.io/en/3.5.0/performance/#the-events-proxy-model) docs, and I think is the cause of the admin pages timing out. You can look at it by activating the SQL debug logging in Django

The CTE is described [here](https://django-pghistory.readthedocs.io/en/stable/aggregating_events/#how-does-it-work) on the docs. Following the docs, I tried the `.tracks()` method and it makes the CTE dissapear. They say that using `objects.filter()` triggers the CTE

I hope you don't mind I pushed directly into the branch, I was testing the debugging and I think it should work now. Can you check  this proposed code ?

quevon24:
> @quevon24 I was checking this on my local env and the changes in your commit are still doing the giant UNION ALL query / CTE. That CTE is described as being a scaling bottleneck by pghistory's [own](https://django-pghistory.readthedocs.io/en/3.5.0/performance/#the-events-proxy-model) docs, and I think is the cause of the admin pages timing out. You can look at it by activating the SQL debug logging in Django
> 
> The CTE is described [here](https://django-pghistory.readthedocs.io/en/stable/aggregating_events/#how-does-it-work) on the docs. Following the docs, I tried the `.tracks()` method and it makes the CTE dissapear. They say that using `objects.filter()` triggers the CTE
> 
> I hope you don't mind I pushed directly into the branch, I was testing the debugging and I think it should work now. Can you check this proposed code ?

Great, a better and cleaner way to do it.

PR #4613, opened 2024-10-23
fix(make_dev_data): Prevent factories from uploading to AWS

ERosendo:
This PR addresses the dependency of the `make_dev_data` command on AWS credentials for creating dummy database entries.

The `docket` and `audio` factories were creating fake files using the `FileField` class, which triggered unnecessary **S3** uploads. To address this, the `FileField` is removed from both factories and [post-generation hooks](https://factoryboy.readthedocs.io/en/stable/reference.html#post-generation-hooks) are implemented. These hooks attach a stub file to the fake object, ensuring that factories still return objects with files **when requested** and allowing developers to use familiar declarations like `filepath__data` for testing purposes.

This refinement maintains compatibility and flexibility for factory usage


cl/audio/factories.py
```diff
@@ -15,8 +15,53 @@ class Meta:
...
     sha1 = Faker("sha1")
     download_url = Faker("url")
-    local_path_mp3 = FileField(upload_to="/tmp/audio")
-    local_path_original_file = FileField(upload_to="/tmp/audio/")
+
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
+        """Creates an instance of the model class without indexing."""
+        obj = model_class(*args, **kwargs)
+        # explicitly sets `index=False` to prevent it from being indexed in SOLR.
```
mlissner, line 23: Can you add instructions here for what we should do later when Solr is removed? E.g.: 

```suggestion
        # explicitly sets `index=False` to prevent it from being indexed in SOLR.
        # Once Solr is removed, we can just remove this method completely. (IS THIS TRUE?)
```

mlissner (approved):
Seems OK, but I'll let somebody else dig in and decide for sure. Perhaps @cweider is open to taking a look?

cl/audio/factories.py
ERosendo: > Once Solr is removed, we can just remove this method completely. (IS THIS TRUE?)

Correct. Once SOLR is removed, we'll need to update [save method](https://github.com/freelawproject/courtlistener/blob/a0aa1716f157905913473e98b4ef226f7f205ea1/cl/audio/models.py#L199-L218) and factory for this model. The index argument is exclusively used by SOLR for indexing purposes and will no longer be necessary.

mlissner: Great. Can we tweak the comment to that effect, please?

ERosendo: Sure!

cweider:
Yup, checks out over here on my non-AWS entitled instance.

PR #4618, opened 2024-10-24
refactor(workflows): isolate courtlistener build from test phase

cweider:
This isolation trims ~90s in billable GHA time and, by way of isolating the build, makes build errors easy to seperate from test errors.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,6 +11,39 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


mlissner:
This looks OK to me in theory. Waiting for the build/test scripts to finish, but set for automerge. Thank you!

mlissner:
A couple of the build commands weren't finishing, I think because they don't exist anymore, so I merged without them, and I tweaked our required build steps to the new one. I think we should be good going forward. Thank you again!

Issue #4313, opened 2024-08-15
Feature request: associate CAP IDs with clusters and allow querying `clusters` on that basis


beidelson:
With the impending deprecation of CAP's APIs, I'm looking at integrating CourtListener as a substitute data source for [Case Viewer](https://www.caseviewer.app/) (cf. #4312). It would help in this transition if the CL clusters that incorporate CAP's data also came with a field for the legacy CAP ID number (and if this field were available for querying the `clusters` endpoint). 

For one thing, Case Viewer stores user history and bookmarks, and these are implemented internally with URLs like `caseviewer://?capID=1905144`. With a `capID` field in the CL data, it would be easy to map these over; without it, the CAP deprecation will break them. Also, having matching ID numbers available from both CAP and CL would make it much easier to test the parsing of the same case from each source during the transition period.

In case it's helpful with this, I noticed that many clusters do already have their CAP IDs embedded in the `filepath_json_harvard` field—e.g., [Palsgraf](https://www.courtlistener.com/api/rest/v4/clusters/3602780/) has this:
```
"filepath_json_harvard": "/storage/harvard_corpus/law.free.cap.ny.248/339.1905144.json"
```
The CAP ID is `1905144`, the last bit of the filename there. Maybe that makes this a very simple transformation, although I don't know if the `filepath_json_harvard` field is uniformly present in the CAP-associated clusters (and in any case it's not possible to query the `clusters` endpoint on this basis as-is).

Thanks for considering.

mlissner:
Thanks Ben. @flooie what are your thoughts on the feasibility of this one? I'm +1 on doing it, but I don't know how hard it would be.

flooie:
Well. The simplest thing would be making the filepath json field usable in the api no?  That seems pretty simple to me.

mlissner:
Yeah, that'd be OK, but kind of clunky. It's not hard just to add the ID and do it right, I'd say.

PR #4284, opened 2024-08-07
Update opinioncluster model for CAP PDFs

jtmst:
Fixes: #4258

We need to update the OpnionCluster model to accommodate bringing it data for CAP PDFs.

This change adds a field:
  - filepath_pdf_harvard

It also modifies the make_upload_path in model_helpers to accommodate file paths for the OpinionCluster class.

PR #4622, opened 2024-10-25
feat(harvard_merger): add `harvard_id` field to `OpinionCluster`

cweider:
Harvard's Caselaw Access Project has been sunset. For projects which have existing references to CAP cases, there's a need to identify a CAP case's corresponding CL opinion cluster.

An indexed `harvard_id` column is added to `OpinionCluster`. The field is also added to the `fields` of `OpinionClusterFilter`.

For migration, this patch builds on work done in #4284 and #4442 and extends `import_harvard_pdfs` to populate the `harvard_id` column using CAP crosswalk file.

Fixes: #4313


cl/search/migrations/0037_add_harvard_id_to_opinioncluster.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN;
+ALTER TABLE "search_opinioncluster" ADD COLUMN "harvard_id" varchar DEFAULT '0' NOT NULL;
+ALTER TABLE "search_opinioncluster" ALTER COLUMN "harvard_id" DROP DEFAULT;
```
cweider, line 3: it would seem that this implicates #1106, but from the issue and the [migration page](https://github.com/freelawproject/courtlistener/wiki/Database-migrations) there’s not really any particular action to take? or it’s on a case-by-case basis?

Dunno - I welcome feedback on this.

cl/search/models.py
```diff
@@ -2731,6 +2731,11 @@ class OpinionCluster(AbstractDateTimeModel):
         storage=IncrementingAWSMediaStorage(),
         blank=True,
     )
+    harvard_id = models.CharField(
```
cweider, line 2734: Not convinced that `harvard_id` is the best name (`cap_id` is a leading alternative). Meh, this keeps it in-line with its `filepath_*_harvard` neighbors, above.

cl/search/tests/test_import_harvard_pdfs.py
```diff
@@ -119,3 +119,68 @@ def test_import_harvard_pdfs(
...
+        m = mock_open(read_data=json.dumps(crosswalk_data))
+
+        # Mock crosswalk_dir
+        crosswalk_dir = "/mocked_path/crosswalk_dir"
+
+        # Verify crosswalk_dir exists
+        self.assertTrue(
+            os.path.exists(crosswalk_dir),
+            f"Crosswalk directory does not exist: {crosswalk_dir}",
+        )
```
cweider, line 175: Everything above is copy-paste from `test_import_harvard_pdfs` above. I don’t love it, but “good enough”? Nudge me if it needs to be better.

cl/search/management/commands/import_harvard_pdfs.py
```diff
@@ -51,6 +51,13 @@ def add_arguments(self, parser):
             help="Directory for reading crosswalk files",
             required=True,
         )
+        parser.add_argument(
+            "--job",
+            type=str,
+            choices=["import_pdf"],
+            default="import_pdf",
+            help="",
+        )
```
cweider, line 60: I see other things running through the crosswalk file to do work (#4614). I’ve no sound opinion for what the “correct” approach is. `jobs` here is driven by my interest in not copy/pasting (and then owning) the moderate complexity surrounding crosswalk processing. @jtmst, I’m happy to take any direction that you can give.

cl/search/migrations/0037_add_harvard_id_to_opinioncluster.sql
mlissner: You're referring to the new column with a default? Happily that's fine these days.

cl/search/models.py
mlissner: Seems right to me.

cl/search/models.py
```diff
@@ -2731,6 +2731,11 @@ class OpinionCluster(AbstractDateTimeModel):
...
+        help_text="The ID of the item in the Caselaw Access Project (Harvard)",
+        default=0,
+        db_index=True,
```
mlissner, line 2737: `db_index` is sort of soft-deprecated these days. Those in the know say not to use it, and to prefer using the `indexes` meta variable. If you do that, it's worth seeing if you can also remove the `varchar_pattern_ops` index that you can see in the SQL, since we don't need it.

cl/search/models.py
```diff
@@ -2731,6 +2731,11 @@ class OpinionCluster(AbstractDateTimeModel):
         storage=IncrementingAWSMediaStorage(),
         blank=True,
     )
+    harvard_id = models.CharField(
+        help_text="The ID of the item in the Caselaw Access Project (Harvard)",
+        default=0,
```
mlissner, line 2736: I'd expect null if we don't have the ID?

mlissner (changes requested):
I just did a quick drive-by and didn't review your code in the management command, nor the tests, but here are a few bits of feedback. Thanks for doing this.

cl/search/models.py
cweider: 🤔  Well!, this explains oddity of needing to use a string in my test – it should be a `IntegerField`!

cl/search/models.py
cweider: SGTM. Very unfair to the DB to be using string for the numeric identifier – indexing will much improved by the switch to integer.

I see other bits of `OpinionCluster` using `db_index`; I can see switching those over could be a worthwhile, albeit separate, bit of housekeeping. Let me know.

cl/search/management/commands/import_harvard_pdfs.py
```diff
@@ -206,36 +200,61 @@ def parse_cap_path(self, cap_path: str) -> Tuple[str, str, str]:
...
+                            self.store_pdf_in_cl(cluster, pdf_content)
+                    else:
+                        logger.info(
+                            f"Dry run: Would fetch PDF from {pdf_path}"
+                        )
+
+                    self.processed_pdfs.add(pdf_path)
+
+                except OpinionCluster.DoesNotExist:
+                    logger.info(f"Cluster not found for id: {cl_cluster_id}")
```
cweider, line 244: Tempted to rely on the caller’s exception handler to do this logging ¯\_(ツ)_/¯

cl/search/models.py
mlissner: It's sort of on my list of things to do. I think the trick is to just add new indexes and then delete the old ones, but you could also get clever and figure out the names of the old indexes so that the actual DB migration isn't needed. And then, of course, when you get to this sort of a place, it sort of also feels like maybe it's just fine to leave alone.

quevon24:
@cweider The import_harvard_pdfs command has been updated, you may have conflicts in the PR

cweider:
Closing. Conflict is substantial and there’s no immediate need for this functionality.

mlissner:
Shoot, this seemed like an important feature, though I admit nobody has asked for it recently. @flooie, what do you think? Should we revive this or just let it go? Maybe it'd be good to at least have an open ticket to do the work, even just as a place for folks to say "me too."

PR #4628, opened 2024-10-26
fix(favorites): casting page_count from str to int to ensure pray-and-pay emails work

v-anne:
This is following up on the discussion in #4589 and the earlier fix in #4592, which was not comprehensive and missed another instance in the codebase of `page_count` being temporarily stored as a `str` instead of an `int`. It fixes this by casting `page_count` before the rd is created.

### DIFF
--- a/cl/disclosures/tasks.py
+++ b/cl/disclosures/tasks.py
@@ -330,11 +330,11 @@ def save_and_upload_disclosure(
     if len(disclosure) > 0:
         return disclosure[0]
 
-    page_count = async_to_sync(microservice)(
+    page_count = int(async_to_sync(microservice)(
         service="page-count",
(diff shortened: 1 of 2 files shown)


v-anne:
After reviewing @ERosendo's latest comment in the issue, I searched in the codebase for other occurrences of page_count being imported as a str, and found a total of three. The previous PR and this one address two of them, and the last one is for the financial disclosures tool, which is not affected by the pray-and-pay feature. Nevertheless, would it be beneficial to change the code in that spot to cast page_count to an int before it is added to the db? It might be better than trying to figure this out again in the future.

File is `cl/disclosures/tasks.py`.

mlissner (approved):
LGTM, thank you!

Issue #4625, opened 2024-10-25
Link to RECAP docket from oral argument audio pages


v-anne:
Is it possible to link to RECAP dockets from the oral argument audio pages? I think the easiest way to do this is to link from the docket number. Since there are sometimes multiple versions of the same PACER docket, it might be most practical to link to a search query based on the identified court and the case number.

This would take the code from the "appealed from" line on docket pages and adapt it to the audio context.
https://www.courtlistener.com/audio/78490/farhane-v-united-states/
https://www.courtlistener.com/docket/61402639/farhane-v-united-states/
![Image](https://github.com/user-attachments/assets/e895aa94-3b68-41bd-a1d4-fbff54bdae8e)

v-anne:
I think the following would need to change in `oral_argument.html`:

Replace this:

`<span class="meta-data-value select-all">{{ af.docket.docket_number }}</span>`

with this:

```html
<a
href="/?type=r&amp;docket_number={{ af.docket.docket_number }}&amp;court={{ af.docket.appeal_from.pk }}"
rel="nofollow"
data-toggle="tooltip"
data-placement="right"
title="Search for this docket number in the RECAP Archive.">{{ af.docket.docket_number }}</a>
```

Not entirely sure whether the existing function to render the view returns `af.docket.appeal_from.pk` already.

PR #4629, opened 2024-10-26
feat(audio): link to RECAP search from audio pages

v-anne:
This PR completes #4625, by allowing users to link from an audio page to the corresponding docket, similar to how existing appellate dockets allow users to see the lower courts the cases originated from.  

I wasn't sure if the site typically includes `class="meta-data-value select-all"` in `<a>` tags as well as plaintext, but if it does, I can add the class to the link as well.


mlissner:
Sure, why not? Thanks for the tweak!

PR #4644, opened 2024-11-02
refactor(docker): perform Webpack asset compilation eagerly

cweider:
While it is simpler to perform asset compilation with Webpack after the contents of the `cl` directory are copied, doing this comes at a cost: the layer generated by Webpack is invalidated any time that the contents of `cl` change. This happens much more frequently than the compiled `react.js` and its dependencies change.


cweider:
Premature – more rearranging is necessary for the work to not get overwritten by the copy.

PR #4647, opened 2024-11-04
Command to update case names

quevon24:
A command to update the case names using the metadata from datasets. This will update all possible names, not just those from Resource or a source combined with Resource.

You can specify the delay to between updates to avoid issues with redis (updating the case names will trigger indexing)
`docker exec -it cl-django python /opt/courtlistener/manage.py update_resource_casenames --filepath /opt/courtlistener/cl/assets/media/federal_3d.csv --delay 0.1`

You perform a dry run to verify that everything is fine
`docker exec -it cl-django python /opt/courtlistener/manage.py update_resource_casenames --filepath /opt/courtlistener/cl/assets/media/federal_3d.csv --dry-run`

You can control the chunk size when reading the csv to avoid memory issues:
`docker exec -it cl-django python /opt/courtlistener/manage.py update_resource_casenames --filepath /opt/courtlistener/cl/assets/media/federal_3d.csv --chunk-size 100000`


cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -111,7 +111,7 @@ def parse_date(date_str: str) -> date | None:
     return None
 
 
-def make_citations(cite1, cite2) -> list[FullCaseCitation]:
+def make_citations(cite1: str, cite2: str) -> list[FullCaseCitation]:
```
flooie, line 114: instead of returning full case citation - why not bring back the dictionary of volume page and reporter here so you dont need to remove dates etc later

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -126,6 +126,88 @@ def make_citations(cite1, cite2) -> list[FullCaseCitation]:
...
+    # Try to match row using both citations
+    for citation in valid_citations:
+
+        possible_matches = Citation.objects.filter(
+            **citation.groups,
+            cluster__docket__docket_number__contains=csv_docket_num,
+            cluster__date_filed=csv_date_filed,
+        )
+        if not possible_matches:
+            # Match not found with citation, docket number and date filed
```
flooie, line 154: Log?

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -187,102 +278,61 @@ def process_csv(
...
-                            f"Row index: {index} - Match found: {single_match} - New case name: {title} - Used citation: {citation.corrected_citation()} - To add: {citation_to_add[0] if citation_to_add else 'Invalid'}"
-                        )
 
-                    # Exit remaining_citations loop
-                    break
+                    if cluster_updated:
+                        total_clusters_updated = +1
+
+                    if docket_updated:
+                        total_dockets_updated = +1
```
flooie, line 307: drop the docket updated stuff- please

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -126,6 +126,88 @@ def make_citations(cite1, cite2) -> list[FullCaseCitation]:
...
+        matched_cluster.save()
+        logger.info(f"Case name updated for cluster id: {matched_cluster.id}")
+        cluster_case_name_updated = True
+
+    if not matched_cluster.docket.case_name or len(csv_case_name) < len(
+        matched_cluster.docket.case_name
+    ):
+        # Save case name in docket when we don't have it or when the case name in csv is smaller than the current case name
+        matched_cluster.docket.case_name = csv_case_name
+        matched_cluster.docket.save()
```
flooie, line 202: drop all this docket name updating.  we should just update the case name for the cluster.

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
quevon24: i did it this way because to use the add_citations_to_cluster function it needs a list of strings, and if we return the citations as a dict, we need to do something like: f"{cite.get("volume")} {cite.get("reporter")} {cite.get("volume")}" to get the full citation but we can get the same result using the FullCaseCitation corrected_citation()  method.

I also realized that the reporter returned in the dictionary is not the corrected reporter, for example: if you pass "276 Cal.Rptr. 272" to get_citations function, .groups will return {'volume': '276', 'reporter': 'Cal.Rptr.', 'page': '272'}

In the db we have that citation stored with "Cal. Rptr.", not "Cal.Rptr.", when you try to search for a Citation object using the matched reporter from the dict but not the corrected reporter it will not return any results. I tested the original approach using the dict and couldn´t match a lot of the rows, but changinq the queries to use the corrected_reporter() method from FullCaseCitation fixed that.

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
quevon24: The log is after the find_matches function, wait for it to finish checking the two citations.

If there are no matches with the two citations then log the message.

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,377 @@
...
+                valid_citations[1].groups.get("volume")
+            ):
+                # Volume number is not an integer e.g. 2001-1 Trade Cases P 73,218
+                logger.warning(
+                    f"Row index: {index} - Citation parsing failed."
+                )
+                continue
+
+            # Check if already have both citations from row
+            c = Citation.objects.filter(
```
flooie, line 262: better variable names please

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,377 @@
...
+
+            if len(valid_citations) < 2:
+                # Skipping row without two citations
+                continue
+
+            if not NUMBER_PATTERN.match(
+                valid_citations[0].groups.get("volume")
+            ) or not NUMBER_PATTERN.match(
+                valid_citations[1].groups.get("volume")
+            ):
```
flooie, line 254: I'm asking why not put this code in the parsing of valid citations.  instead of here.

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,377 @@
...
+                continue
+
+            # Check if already have both citations from row
+            c = Citation.objects.filter(
+                **get_citation_filter(valid_citations[0])
+            ).values_list("cluster", flat=True)
+            d = Citation.objects.filter(
+                **get_citation_filter(valid_citations[1])
+            ).values_list("cluster", flat=True)
+
```
flooie, line 268: why did you get rid of the docket number query and the date

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,377 @@
...
+    2012-635 (La.App. 3 Cir. 12/5/12) also includes date_filed when it is parsed
+
+    :param citation:
+    :return: dict with volume, reporter and page
+    """
+    return {
+        "volume": citation.groups["volume"],
+        "reporter": citation.corrected_reporter(),
+        "page": citation.groups["page"],
+    }
```
flooie, line 216: this should be in the make citation or extract citations -

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,360 @@
...
+        return True
+
+    # otherwise check if a match occurs on both sides of the V
+    v_index = csv_case_name.lower().index("v.")
+    hit_indices = [csv_case_name.lower().find(hit) for hit in overlap]
+
+    return min(hit_indices) < v_index < max(hit_indices)
+
+
+def parse_date(date_str: str) -> date | None:
```
flooie, line 92: shouldn't this be optional date ?

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,360 @@
...
+    for citation in valid_citations:
+        citation_query = Q(**citation) & Q(
+            cluster__docket__docket_number__contains=docket_number,
+            cluster__date_filed=date_filed,
+        )
+        citation_queries |= citation_query
+    possible_matches = Citation.objects.filter(
+        citation_queries
+    ).select_related("cluster")
+
```
flooie, line 172: if there are no valid citations - what happens?  Will this just return a query of all citations?  that would be bad right?

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,360 @@
...
+                total_dockets_updated = +1
+
+            # Add any of the citations if possible
+            add_citations_to_cluster(
+                [
+                    f"{cite.get('volume')} {cite.get('reporter')} {cite.get('page')}"
+                    for cite in valid_citations
+                ],
+                matches[0].cluster_id,
+            )
```
flooie, line 311: reporter or corrected_reporter ?

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,360 @@
...
+            "--dry-run",
+            action="store_true",
+            help="Simulate the update process without making changes",
+        )
+        parser.add_argument(
+            "--chunk-size",
+            type=int,
+            default=100000,
+            help="The number of rows to read at a time",
+        )
```
flooie, line 347: I thought we talked about just removing chunk size?

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,360 @@
...
+    "in",
+    "inc",
+    "re",
+    "st",
+    "ex",
+    "rel",
+    "v",
+    "vs",
+    "for",
+    "a",
```
flooie, line 41: I think we dont need "a" since we remove all single words from the comparison?

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,360 @@
...
+                )
+                case_name_match = check_case_names_match(
+                    west_case_name, cl_case_name
+                )
+                if case_name_match:
+                    matches.append(match)
+
+            if not possible_matches:
+                logger.info(f"Row index: {index} - No matches found.")
+                continue
```
flooie, line 277: if not possible matches appears twice

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,360 @@
...
+    """Tokenizes case name and removes single-character words except for letters with periods.
+
+    :param case_name: case name to tokenize
+    :return: list of words
+    """
+    return [
+        word.lower()
+        for word in WORD_PATTERN.findall(case_name)
+        if len(word) > 1
+    ]
```
flooie, line 57: I think I would simplify the code 

def tokenize_name(case_name: str) -> set[str]:
    """Tokenizes case name and removes single-character words except for letters with periods.

    :param case_name: case name to tokenize
    :return: list of words
    """

    return (
        set(
            [
                word.lower()
                for word in WORD_PATTERN.findall(case_name)
                if len(word) > 1
            ]
        )
        - FALSE_POSITIVES
    )
by changing this function to something like this.  

instead of doing all that set-ing and comparing later might as well just have a function that returns a nice set 


that should let the overap check to be nice and neat 

`overlap = tokenize_name(csv_case_name) & tokenize_name(cl_case_name)`

flooie (changes requested):
Getting closer but a few things jump out.  thanks @quevon24

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,360 @@
...
+            ) = row
+
+            clean_docket_num = docket.strip('="').strip('"')
+            if not clean_docket_num:
+                logger.info(f"Row index: {index} - No docket number found.")
+                continue
+
+            date_filed = parse_date(date_str)
+            if not date_filed:
+                logger.info(f"Row index: {index} - No valid date found.")
```
flooie, line 242: I think logging the errant date value would be valuable so it doenst have to be checked right?

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
quevon24: it is suggested to use | for union types instead of Optional when working with python 3.10+ (PEP 604)

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
quevon24: reporter, because we generate a list of dicts with volume, reporter(we get it with corrected_reporter()) and page in parse_citations()

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
quevon24: yes, that is right, we only keep words bigger than 1 character in tokenize_case_name()

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
quevon24: duplicated log message removed

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,360 @@
...
+            matches = []
+            for match in possible_matches:
+                cl_case_name = (
+                    match.cluster.case_name_full
+                    if match.cluster.case_name_full
+                    else match.cluster.case_name
+                )
+                case_name_match = check_case_names_match(
+                    west_case_name, cl_case_name
+                )
```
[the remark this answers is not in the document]
quevon24: The regex needs some tweaking, it almost works fine, i found some cases where there are spaces between abbreviations:

`"In re P. I.","Court of Appeal, First District, Division 3, California.","January 20, 1989","207 Cal.App.3d 316","254 Cal.Rptr. 774","A041221","254"` https://www.courtlistener.com/opinion/2168480/in-re-pi/

`"In re A. M.","Court of Appeal, First District, Division 2, California.","December 01, 1989","216 Cal.App.3d 319","264 Cal.Rptr. 666","A042237","264"`  https://www.courtlistener.com/opinion/2175113/in-re-am/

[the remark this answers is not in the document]
quevon24: what do you think about this:

```
def combine_initials(case_name: str) -> str:
    """Combine initials in case captions

    :param case_name: the case caption
    :return: the cleaned case caption
    """
    pattern = r"\b[A-Z](?:[A-Z\.]|\s)*[A-Z]\b\."

    return re.sub(
        pattern,
        lambda match: match.group(0).replace(" ", "").replace(".", ""),
        case_name,
    )
```

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,368 @@
...
+def combine_initials(case_name: str) -> str:
+    """Combine initials in case captions
+
+    :param case_name: the case caption
+    :return: the cleaned case caption
+    """
+
+    pattern = r"((?:[A-Z]\.?\s?){2,})(\s|$)"
+
+    return re.sub(pattern, lambda m: m.group(0).replace(".", ""), case_name)
```
flooie, line 220: this pattern doesn't appear to removal initials

flooie: perhaps 

```
initials_pattern = re.compile(r"([A-Z]{1}\.?\s?){2,}(\s|$)")

    match = initials_pattern.search(case_name)  # Search for the initials pattern
    if match:
        initials = match.group()
        compressed_initials = re.sub(r"(?!\s$)[\s\.]", "", initials)
        case_name = case_name.replace(initials, compressed_initials)  
```

this would work?  I tested it on 

case_names = [
    "M. X. Smith",
    "M.X. Smith",
    "M X Smith",
    "M. X. J. Smith",
    "M X J Smith",
]
and it worked as expected .

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -0,0 +1,368 @@
...
+            total_dockets_updated = +1
+
+        # Add any of the citations if possible
+        add_citations_to_cluster(
+            [
+                f"{cite.get('volume')} {cite.get('reporter')} {cite.get('page')}"
+                for cite in valid_citations
+            ],
+            matches[0].cluster_id,
+        )
```
flooie, line 326: id love to reuse this - but I dont think we should - its not as robust and causes use to re-create the citation to parse it again and do a bunch of eyecite.  I think it would be smarter to identify the type when we validate teh citations and just create the citation here.  


something like ... 


```

            matched_cluster = matches[0].cluster

            if dry_run:
                # Dry run, don't save anything
                continue

            # Update case names

            with transaction.atomic():
                cluster_updated, docket_updated = update_matched_case_name(
                matched_cluster, west_case_name
                )

                for citation in valid_citations:
                    if Citation.obects.filter(reporter=citation.get("reporter"), cluster=matched_cluster.id).exists():
                        logger.warning("Can not save mismatched citation")
                        raise("issue with reporter already here... if its not the same one")
                    citation['cluster_id'] = matched_cluster.id
                    Citation.objects.get_or_create(**citation)
```

I think.  

This means the parse citation would look something like this 

```

def parse_citations(citation_strings: list[str]) -> list[dict]:
    """Validate citations with Eyecite.

    :param citation_strings: List of citation strings to validate.
    :return: List of validated citation dictionaries with volume, reporter, and page.
    """
    validated_citations = []

    for cite_str in citation_strings:
        # Get citations from the string
        found_cites = get_citations(cite_str, tokenizer=HYPERSCAN_TOKENIZER)
        if len(found_cites) != 1:
            continue

        citation = found_cites[0]

        if isinstance(citation, FullCaseCitation):
            volume = citation.groups.get("volume")

            # Validate the volume
            if not volume or not volume.isdigit():
                continue

            if not citation[0].corrected_reporter():
                reporter_type = Citation.STATE
            else:
                cite_type_str = citation[0].all_editions[0].reporter.cite_type
                reporter_type = map_reporter_db_cite_type(cite_type_str)

            validated_citations.append(
                {
                    "volume": citation.groups["volume"],
                    "reporter": citation.corrected_reporter(),
                    "page": citation.groups["page"],
                    "type": reporter_type
                }
            )

    return validated_citations
```

I took this from the other function - but im not sure why we need the odd if citation reporter use Citation.State ... ?

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
quevon24: we could have abbreviations in both parties, but that way it will only apply to the first match, we should do something like:

```
def initials_func(case_name):
    initials_pattern = re.compile(r"([A-Z]{1}\.?\s?){2,}(\s|$)")

    matches = initials_pattern.finditer(case_name)  # Search for the initials pattern
    if not matches:
        return case_name
    for match in matches: # replace all abbreviations in the case name
        initials = match.group()
        compressed_initials = re.sub(r"(?!\s$)[\s\.]", "", initials)
        case_name = case_name.replace(initials, compressed_initials)
    return case_name

```
Because if not it fails like this:

```
D.L.M. v. T.J.S. -> DLM v. T.J.S.
R. L. C. R. v. L. Z. S. -> RLCR v. L. Z. S.
J. B. v. C. E. -> JB v. C. E.
```

After applying what I mentioned above, the regex only fails in this specific case:

`In re Gregory C. KAPORDELIS -> In re Gregory CKAPORDELIS
`

Also, I'm not sure if "M X J" in M ​​X J Smith should be considered an abbreviation since they are not together or have periods at the end of each abbreviated word.

I'm checking how to improve the regex, the other changes are ready. I also added a new date format to the list of valid formats.

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
quevon24: I have already checked it and I think this part is not necessary:

```
if not citation[0].corrected_reporter():
                reporter_type = Citation.STATE
```

if i remember correctly, we reused that from some old code, i guess we did it that way maybe because reporters-db wasn't complete as it is now.

I also tested if that condition occurs with any of the current citations we have on CL, but got not result, we never got around to that, i think it is safe to remove that part

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
flooie: how about something like this 

```
def initials_func(case_name):
	initials_pattern = re.compile(r"(\b[A-Z]{1}\.?\s?){2,}(\s|$)")

	matches = initials_pattern.finditer(case_name)
	if matches:
		for match in matches:
			initials = match.group()
			compressed_initials = re.sub(r"(?!\s$)[\s\.]", "", initials)
			case_name = case_name.replace(initials, compressed_initials)
	return case_name
```

quevon24:
The new changes are ready @flooie

flooie (approved):

mlissner:
There it is!

PR #4649, opened 2024-11-05
add support for last_updated field in crosswalk generation

jtmst:
Adds support for the updated_after param that checks CAP data for a last_updated field.  This will allow us to generate crosswalk files consisting of cases that have been updated after a given date.  Dates should be in either format: YYYY-MM-DD or YYYY-MM-DDTHH:MM:SS+00:00.

Changes:
- Adds updated_after param to the generate_cap_crosswalk command 
- Adds logic for filtering based on the given date
- Unit tests added

### DIFF
--- a/cl/search/management/commands/generate_cap_crosswalk.py
+++ b/cl/search/management/commands/generate_cap_crosswalk.py
@@ -1,11 +1,11 @@
 import json
 import logging
 import os
-from typing import Any, Dict, List, Optional
 from datetime import datetime
-import pytz
+from typing import Any, Dict, List, Optional
 
 import boto3
+import pytz
 from botocore.exceptions import ClientError
 from django.conf import settings
 from django.core.management.base import BaseCommand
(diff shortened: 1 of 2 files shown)


jtmst:
Marking as draft to potentially make some changes for performance.

cl/search/management/commands/generate_cap_crosswalk.py
```diff
@@ -74,6 +81,25 @@ def handle(self, *args: Any, **options: Any) -> None:
...
         self.start_from_reporter = options["start_from_reporter"]
+        self.updated_after = None
+        if options["updated_after"]:
+            try:
+                self.updated_after = datetime.fromisoformat(
+                    options["updated_after"]
+                )
+            except ValueError:
+                try:
+                    self.updated_after = datetime.strptime(
```
quevon24, line 92: maybe you can follow the same style and do this:

`self.updated_after = options["updated_after"]`

I think you can do this in one line, if something fails it will raise the ValueError exception:

`self.updated_after = datetime.strptime(updated_after, "%Y-%m-%d").replace(tzinfo=pytz.UTC)
`

quevon24 (approved):
Everything is fine, I just left two small optional comments

Issue #3950, opened 2024-04-05
Create a scraper status page


grossir:
Let's track here our ideas and requirements for a "scraper status page". Sometimes scrapers fail (for many reasons) and such a page would be useful to notice more errors that won't surface on Sentry

### Background
In #3949 we deleted `cl/scrapers/management/commands/cl_report_scrape_status.py` which depended on the `ErrorLog` model and created  a useful HTML report, that was sent via email and also to a slack channel

However, it seems nobody was really monitoring that, and both the content and the access to the report could be improved

### Ideas
...

mlissner:
We've never used materialized views and I haven't dug into them in a long time. What would we get out of that?

grossir:
A materialized view basically stores the results of a query as a named "table"
- saves computation time if you just want to check the stored results
- makes it easy to re-run the query and refresh the values  `REFRESH MATERIALIZED VIEW {view name}` 

So it's just a way of naming / storing a commonly used query, usually for analytic purposes

Then, we could create an admin page for it to access it without having DB access / having to rerun the query on a console. (we could do it for a plain query too, I think, but that would probably hit the DB each time something is done with the admin)

mlissner:
I think I'd suggest doing it the simple way first, and thinking about solutions like materialized views only if we need the performance they'd give. We could do it that way, right?

grossir:
While testing a implementation, I run into a problem: the query takes 3 minutes to execute, which would cause a timeout when loading the admin page. That would be another reason to use a materialized view. I tried some variations of the query for speed ( **a.** taking the most recently modified docket for each court; **b.** taking the most created docket for each court), but they raise false positives, which would kill the usefulness of a status page

```sql
EXPLAIN ANALYZE  (    
SELECT 
        court_id, 
        max(so.date_created) as latest_creation_date,
        now() - max(so.date_created) as time_since
    FROM 
        (
            SELECT id, court_id
            FROM search_docket 
            WHERE court_id IN (
                SELECT id 
                FROM search_court
                /* 
                    Only check courts with scrapers in use 
                */
                WHERE has_opinion_scraper AND in_use
            )
        ) sd 
    INNER JOIN 
        (SELECT id, docket_id FROM search_opinioncluster) soc ON soc.docket_id = sd.id
    INNER JOIN 
        search_opinion so ON so.cluster_id = soc.id
    GROUP BY 
        sd.court_id
    HAVING 
        /* 
            Only return results for courts with no updates in a week 
[...]

PR #4662, opened 2024-11-06
feat(scrapers.admin): create materialized view and admin page

grossir:
Concept for https://github.com/freelawproject/courtlistener/issues/3950

For this to work, the materialized view must be created directly on the DB by running the "query" attribute

For my local database, it looks like this
image

But, if if connected to the live DB, it would have these rows:
image


grossir:
To do this kind of queries we have 3 options
- run the query in a console, or hardcode it in a python string and run it from a postgres interface
- give the query a name, but don't store the results. That's a [VIEW](https://www.postgresql.org/docs/current/sql-createview.html). You could then do `SELECT * FROM view_name` and it will re-execute the query.
- give the query a name, and store the results. That's a MATERIALIZED VIEW

In the case of this query, I don't think a new index would help. I think the slowness comes from joining big tables. Some query optimization may help reduce it more. The raw version takes 10 minutes, prefiltering the docket table by courts that have scrapers takes it down to 3 minutes

**Why bother materializing it?**
- admin timeout
If we want to use the admin to look at the results of this query, we risk timing out, as happened with Pghistory admin. I am not sure what the server timeout is, but from Pghistory I think it's about 60 seconds. Having seen that the query takes around 3 minutes to complete, I would expect a timeout. 
- waiting (in case it didn't time out)
Even if it didn't timed out, waiting for 3 minutes each time would make for bad user experience. 
We could try to cache the query, too, for a couple days at least. That would alleviate some problems. I expect a few rows (less than 50) to be returned each time
- django admin hacking
to take advantage of the django admin HTML templates and ease of navigation, I am using the View (it would work on plain views, which are basically queries, and materialized views) as the concrete table for the django model.


**How to refresh it**
I expect to check this table a couple times a week. So we could refresh this every 2 or 3 days. We could use a plain cron job, the command is simple `REFRESH MATERIALIZED VIEW <name>` 

Or, we could implement the refresh logic as an async call when accessing the admin page, and send a django admin message (those banners when editing an object) to the user advising to refresh the page in a few minutes

mlissner:
Hm, in the past we had a script that would send an email with this kind of information, and it ran on a cron job, so I guess this is basically the same thing, but instead of storing the results of the script in an email, this stores the results in the materialized view. Seems fine, I suppose. :)

grossir:
To test this on a local environment
- clone some clusters using clone from CL
```
docker exec -it cl-django python /opt/courtlistener/manage.py clone_from_cl --type search.OpinionCluster --id 10034763 
```
- override the opinion's `date_created` and refresh the view
```
update search_opinion set date_created = '2024-11-01';
refresh materialized view scrapers_mv_latest_opinion ;
```
- checkout the admin page
```
http://0.0.0.0:8000/admin/scrapers/mvlatestopinion/
```

cl/scrapers/migrations/0004_create_mv_latest_opinion_customers.sql
mlissner, line 1: We wouldn't need this for the replicas.

mlissner (commented):
One comment so far. We also need an admin task to update the view, right? Want to include that?

grossir:
I added the command to refresh the view. Can you review this again @mlissner ?

mlissner (approved):
Let's give this a try.

mlissner:
I just created https://github.com/freelawproject/infrastructure/issues/210 to make the cronjobs, and assigned it to you. If you tell me the schedule, I can get it launched.

PR #4663, opened 2024-11-06
Cap import documentation

jtmst:
Adds docs detailing the usage of generate_cap_crosswalk and update_cap_cases for importing CAP data.


mlissner (changes requested):
This is great, Josh. We usually put our docs in the wiki, so I copied it over there: 

https://github.com/freelawproject/courtlistener/wiki/Importing-Harvard-Data

I'm going to close this one, but do you think you could do a tiny PR linking to that page in the `--help` flag for the commands?

PR #4696, opened 2024-11-15
fix(env): Change CL_API_KEY to CL_API_TOKEN

elisa-a-v:
When I tried to clone data from CL using the `clone_from_cl` command I got an error, because the command was expecting `CL_API_TOKEN` but the `.env.example` had `CL_API_KEY`. This PR fixes it.


mlissner:
Thank you! A dozen people dealt with this, fixed it locally, and didn't do a PR. We gotta fix these things to keep ourselves sane!

PR #4700, opened 2024-11-18
feat(api): Enable ordering docket entries by recap_sequence_number and entry_number

### LINKED ISSUE #4061: Add default website docket entry ordering to the API
From `cl.search.models.DocketEntry.Meta`, docket entries are sorted by: 

```
        ordering = ("recap_sequence_number", "entry_number")
```

But that ordering is not available in the API. Oof! Let's fix that for our customers and then document it in our new documentation page. 

I'm not sure how to sort by two fields like this in DRF, but I assume it's not too difficult.

elisa-a-v:
Resolves #4061


mlissner (approved):
LGTM!

PR #4701, opened 2024-11-18
fix(users): Enforce validation on first_name for increase security

### LINKED ISSUE #4687: Check vuln from Vaibhav

ERosendo:
This PR addresses a security vulnerability in the registration form. The `first_name` field was previously susceptible to Hyperlink Injection attacks. By allowing arbitrary input, malicious users could inject malicious links into the welcome email, potentially redirecting users to phishing sites or distributing malware.

References:

https://hackerone.com/reports/843421
https://hackerone.com/reports/158554
https://hackerone.com/reports/164833

fixes #4687


cl/users/forms.py
```diff
@@ -188,6 +190,16 @@
             )
         return email
 
+    def clean_first_name(self):
+        first_name = self.cleaned_data.get("first_name")
+        if not re.match(
+            r"""[^!"#$%&()*+,./:;<=>?@[\]_{|}~]+$""", first_name, re.IGNORECASE
```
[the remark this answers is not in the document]
ERosendo: The alert suggests limiting the input string length as a potential mitigation for this issue. While our clean method is run after validation, it's important to note that our form's structure and server-side validation limit the maximum input length to 150 characters.

As discussed in https://github.com/freelawproject/bigcases2/issues/598#issuecomment-2374634378, even if an attacker attempts to manipulate the form using JavaScript, our server-side validation will still enforce the 150-character limit.

Therefore, while limiting input length is a valid security measure, our existing validation and server-side controls already provide sufficient protection against this specific issue.

cl/users/forms.py
```diff
@@ -188,6 +190,16 @@ def clean_email(self):
...
+        ):
```
mlissner, line 197: Boy, I'm rusty on regexes these days, but why not do:

```suggestion
        import string
        if re.match(f"[{string.punctuation}]+$", first_name):
```

That flips the logic, uses `string.punctuation` (which I think is fine?) and removes `re.IGNORECASE`, because I can't see how it was relevant?

mlissner (approved):
One thought for you, but I might be wrong. 

Thank you. :)

cl/users/forms.py
ERosendo: @mlissner You're absolutely right! The re.IGNORECASE flag isn't necessary here, and your suggestion is great.

Before we implement this change, I'd like to clarify one thing:

The current regex allows names with apostrophes (') and hyphens/dashes (-). If we use `string.punctuation` to validate names, it will flag these characters as invalid. Is that ok?

ERosendo: 👍 :shipit:

mlissner: Wait, before we ship, we can flip the regex, remove the `not`, and remove the `IGNORECASE`, right?

ERosendo: Thank you for pointing that out. I've incorporated the refactor into the latest commit

mlissner:
Nice. Set for auto-merge!

PR #4717, opened 2024-11-20
feat(document_page): Add link to document in docket sheet

### LINKED ISSUE #3308: Make "document #X" link back to the docket sheet from the document page
I was just looking at the [document page here](https://www.courtlistener.com/docket/6247722/12/7/oscilloscope-pictures-inc-v-monbo/), and I was on exhibit F. I wanted to see the rest of the exhibits, but doing so was hard. The only way to do it now is to go back to the docket and find document 12. 

Instead, the headings should make this easy. There are two easy opportunities:

image

First, on attachment pages, we can make the words "Document #12" into a link. Second, in the left hand column under the "Back to Docket" link, we could put a "Back to Document 12" link.

Both or either of these would help. I think I'm partial to doing it in the heading, but I'm open to seeing the sidebar too.

(All of these links would go to the docket sheet at that item. In this case: https://www.courtlistener.com/docket/6247722/oscilloscope-pictures-inc-v-monbo/#entry-12 )

elisa-a-v:
Resolves #3308 

link to document

### DIFF
--- a/cl/opinion_page/templates/includes/rd_metadata_headers.html
+++ b/cl/opinion_page/templates/includes/rd_metadata_headers.html
@@ -3,7 +3,10 @@
   {% if rd.description %}{{ rd.description }}
     <span class="gray">&mdash;</span>
   {% endif %}
-  Document #{{ rd.document_number }}{% if rd.document_type == rd.ATTACHMENT %}, Attachment #{{ rd.attachment_number }}{% endif %}
+  <a href="{{ rd.docket_entry.docket.get_absolute_url }}#entry-{{ rd.document_number }}">
+    Document #{{ rd.document_number }}
+  </a>
+  {% if rd.document_type == rd.ATTACHMENT %}, Attachment #{{ rd.attachment_number }}{% endif %}
 </h3>
 <h4>{{ rd.docket_entry.docket.court }}</h4>
 <p class="bottom">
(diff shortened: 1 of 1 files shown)


cl/opinion_page/templates/includes/rd_metadata_headers.html
```diff
@@ -3,7 +3,10 @@ <h3>
   {% if rd.description %}{{ rd.description }}
     <span class="gray">&mdash;</span>
   {% endif %}
-  Document #{{ rd.document_number }}{% if rd.document_type == rd.ATTACHMENT %}, Attachment #{{ rd.attachment_number }}{% endif %}
+  <a href="{{ rd.docket_entry.docket.get_absolute_url }}#entry-{{ rd.document_number }}">
+    Document #{{ rd.document_number }}
+  </a>
```
mlissner, line 8:
```suggestion
    Document <a href="{{ rd.docket_entry.docket.get_absolute_url }}#entry-{{ rd.document_number }}">#{{ rd.document_number }}
  </a>
```

mlissner (approved):
I made a little tweak because I thought the link attracted too much attention. Hopefully this will look better. 

There's still a bug on paginated dockets like this one:

https://www.courtlistener.com/docket/4214664/national-veterans-legal-services-program-v-united-states/?page=2

In that case if you click on document 101 from page 2, it'll drop the pagination and take you to `#entry-101` on page 1, which doesn't exist. 

This also affects the "Back to Docket" button in the sidebar....something to fix, I guess.

PR #4731, opened 2024-11-25
fix(pacer_free_opinions): add ReadError to get_and_process_free_pdf

quevon24:
- Add ReadError to get_and_process_free_pdf decorator to retry for that exception
- Reduced the sleep time between court cycles when downloading pdfs from 2 to 1 second, we don't need to wait that long, in get_and_save_free_document_reports() we only wait 1 second to avoid possible  throttling/blocks from the court when we were collecting the documents metadata.


flooie (approved):
Looks fine to me.

PR #4744, opened 2024-11-27
feat(api): Enable filtering courts by parent court id

### LINKED ISSUE #4711: Add parent_court as a filter to the Court API
We need to be able to query the courts API for the courts that are children of a particular court. For example, I should be able to do this:

https://www.courtlistener.com/api/rest/v4/courts/?parent_court=nysupct

That doesn't work now, but it should.

This didn't make it into the current Sprint, unfortunately, so I'm going to put it on the unassigned zone and if we have time, perhaps we can do it.

elisa-a-v:
Resolves #4711

This PR also adds more tests for court filters, and it makes the assert message more descriptive in `FilteringCountTestCase.assertCountInResults` so it now includes the request url that prompted the response, for example:

```
AssertionError: 0 != 1 : Expected 1, but got 0 in /api/rest/v4/courts/?start_date__year=2021

r.data was: OrderedDict({'count': 0, 'next': None, 'previous': None, 'results': []})
```

### DIFF
--- a/cl/search/filters.py
+++ b/cl/search/filters.py
@@ -28,6 +28,10 @@ class CourtFilter(NoEmptyFilterSet):
         "cl.search.filters.DocketFilter", queryset=Docket.objects.all()
     )
     jurisdiction = filters.MultipleChoiceFilter(choices=Court.JURISDICTIONS)
+    parent_court = filters.CharFilter(
+        field_name="parent_court__id",
+        lookup_expr="exact",
+    )
 
     class Meta:
         model = Court
--- a/cl/api/tests.py
+++ b/cl/api/tests.py
@@ -1,5 +1,5 @@
 import json
-from datetime import date, timedelta
+from datetime import date, datetime, timedelta, timezone
 from http import HTTPStatus
 from typing import Any, Dict
 from unittest import mock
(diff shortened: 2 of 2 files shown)


mlissner:
To Eduardo for review. Thank you!

cl/search/filters.py
```diff
@@ -28,6 +28,10 @@ class CourtFilter(NoEmptyFilterSet):
         "cl.search.filters.DocketFilter", queryset=Docket.objects.all()
     )
     jurisdiction = filters.MultipleChoiceFilter(choices=Court.JURISDICTIONS)
+    parent_court = filters.CharFilter(
+        field_name="parent_court__id",
+        lookup_expr="exact",
```
ERosendo, line 33: We could simplify the filter addition by omitting the `lookup_expr` argument. The Filter class's [constructor](https://github.com/carltongibson/django-filter/blob/2494df96c6387a9fa411fcb00b696b15dfd9216b/django_filters/filters.py#L75-L103) defaults to using `settings.DEFAULT_LOOKUP_EXPR`, which is typically "exact". Unless we modify the default lookup expression in our settings module, the implicit behavior will suffice.

ERosendo (approved):
The code looks great! Thanks for the new test class, @elisa-a-v.

Just one small comment about the `Filter` class. while code modifications are not required for this particular PR, it could be useful for future filtering-related changes

@mlissner :shipit:

PR #4765, opened 2024-12-02
feat(opinions.html): Ensure Case Name displays

flooie:
Fixes Missing Case Names when Case Name Full is the only name

We missed this before - and should ensure we display a case name
or a case name full in the caption if its the only thing available.


mlissner (changes requested):
I think you can accomplish this by using the `best_case_name` template filter, no?

flooie:
Thats a great idea thank you @mlissner

mlissner (approved):

Issue #4762, opened 2024-11-30
Harvard PDFs


flooie:
Harvard PDFs need to be added for cases without scraped opinions.  Currently they do not display by default.

flooie:
this is finished with the latest PR #4772

PR #4772, opened 2024-12-03
fix(opinions.html): Enable Harvard SCAN tab

flooie:
Fix addresses #4762 

We enabled the scans to be downloaded but not viewed for the new harvard data.
This PR simply checks for scrapes and PDF scans

### DIFF
--- a/cl/opinion_page/templates/opinions.html
+++ b/cl/opinion_page/templates/opinions.html
@@ -290,55 +290,31 @@
 
             <ul class="nav nav-tabs hidden-print">
                 <li role="presentation" {% if tab == "opinions" or tab == "" %} class="active" {% endif %}>
-                    {% if tab == "opinions" %}
-                        <a href="#" data-toggle="tab">Opinion</a>
-                    {% else %}
-                        <a href="{% url 'view_case' cluster.pk cluster.slug %}">Opinion</a>
-                    {% endif %}
+                    <a href="{% url 'view_case' cluster.pk cluster.slug %}">Opinion</a>
                 </li>
                 {% if authorities_count > 0 %}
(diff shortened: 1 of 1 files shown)


flooie:
@mlissner the failing docket test appeared to be a connection issue with getting the docket image - not something related to this very small pr - im updating it to give it a chance to rerun

cl/opinion_page/templates/opinions.html
```diff
@@ -332,10 +332,10 @@ <h4 class="case-court">{{ cluster.docket.court }}</h4>
                         {% endif %}
                     </li>
                 {% endif %}
-                {% if has_downloads and "pdf" in pdf_path %}
+                {% if has_downloads and "pdf" in pdf_path or cluster.filepath_pdf_harvard %}
                     <li role="presentation" {% if tab == "pdf" %} class="active" {% endif %}>
                         {% if tab == "pdf" %}
```
mlissner, line 337: Is this if statement just here to make it so the tab isn't a link when it's the active tab? I don't think that's really necessary and I think we can simplify by removing this if statement here (and presumably elsewhere)?

mlissner (approved):

cl/opinion_page/templates/opinions.html
flooie: its also to avoid displaying the PDF tab if the download is an RTF or HTML file or some other form that we are not currently prepared to display.

mlissner: Oh sorry, I meant line 337.

flooie: yep

mlissner (approved):

PR #4776, opened 2024-12-03
Add Journal to Citation model type

flooie:
Add Journal as type of citation in the model

We added journals to eyecite, but it turns out many of the parallel citations are are in fact
journals. This distinction can be found in eyecite so it is worth bringing that distinction over
to CL. 

This change is needed to resolve https://github.com/freelawproject/courtlistener/issues/4775.

The alternative was to map journals to specialty reporters - but that did not seem correct.


cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
```diff
@@ -140,11 +140,16 @@ def parse_citations(citation_strings: list[str]) -> list[dict]:
         found_cites = get_citations(cite_str, tokenizer=HYPERSCAN_TOKENIZER)
         if not found_cites:
             continue
-
+        if len(found_cites) == 0:
+            logger.info("Unable to parse %s", cite_str)
+            continue
```
mlissner, line 145: Isn't this the same as line 141?

mlissner (changes requested):
One comment so far. It'd be nice to see a test here too, please, then maybe somebody from the case law team can review?

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
flooie: yes sorry let me clean this up .

flooie: No, I take that back. That’s not what I meant to write. The issue here is that we are correctly finding one citation when we pass in a single string citation, which is expected. However, the problem arises because the function identifies two editions for this citation: one journal and one reporter. We don’t want to arbitrarily choose the first one

quevon24 (approved):
everything looks good

flooie:
@mlissner looks like we are good to go - but there is a noop migration so I wouldn't merge this without your approval.

cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
mlissner: OK, but if `found_cites` is `[]`, line 141 will catch that and continue. I can't see how line 144 will ever get run.

mlissner:
I was also hoping for some tests on this one?

mlissner (approved):

Mtlgroupmarcel (changes requested):
- [ ]

PR #4777, opened 2024-12-03
fix(opinion_page.views): Update sponsorship timeframe

### LINKED ISSUE #3637: Stop showing vLex sposnsorship banner appropriately
Our contract with vlex requires that we put a banner on every opinion that we add during the contract, and then keep those banners up for three years from the date the banner first goes up. Our contract ends on January 31.

Right now, the code reads:

```python
    # Identify opinions updated/added in partnership with v|lex for 3 years
    sponsored = False
    if (
        cluster.date_created.date() > datetime.datetime(2022, 6, 1).date()
        and cluster.filepath_json_harvard
    ):
        sponsored = True
```

So this should be changed to only do items created between 2022/6/1 and 2024/01/31, and should be modified to do nothing for any case that was created more than three years ago.

Something like:

```
if creation-date more than three years ago:
    sponsored = False
elif  2022-6-1 < creation-date < 2024-01-31:
    sponsored = True
```

I think that's right!

flooie:
Fixes #3637 

Fixes an overly broad sponsorship bug

### DIFF
--- a/cl/opinion_page/views.py
+++ b/cl/opinion_page/views.py
@@ -888,12 +888,14 @@ async def view_opinion_old(
     )[:3]
 
     # Identify opinions updated/added in partnership with v|lex for 3 years
-    sponsored = False
-    if (
-        cluster.date_created.date() > datetime.datetime(2022, 6, 1).date()
-        and cluster.filepath_json_harvard
-    ):
-        sponsored = True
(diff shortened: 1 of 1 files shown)


mlissner (approved):
LGTM, thanks.

Issue #4726, opened 2024-11-21
Stop sending content to Solr and rip out indexing pipeline


mlissner:
Whenever we get new content, we send it to Solr so it's searchable. Once Solr is live, we can stop doing this and save ourselves a lot of trouble. Let's rip out all of this code, including the tasks, the stuff in the model, the templates, etc. It's clean up time.

PR #4791, opened 2024-12-05
feat(search): Rips out SOLR indexing pipeline

ERosendo:
This PR addresses #4726 

This PR includes the following changes: 

- Removes `scorched` dependency.

- Updates the settings module to clean up SOLR env vars.

- Removes Solr health check

- Removes custom `save` and `as_search_dict` methods from models.

- Refactors AudioFactory class to stop using the `index` argument in hooks.

- Updates admin classes to remove custom methods that were previously used to index data to Solr.

- Removes Solr-specific admin utility functions.

- Removes the normalize_search_dicts helper.

- Renames `solr_list` to `extract_solr_field_values` to better reflect its purpose.

- Removes the following test classes: `EmptySolrTestCase`, `SolrTestCase` and `IndexedSolrTestCase`

- Deletes the following commands:
   - `add_parallel_citations.py`
   - `import_columbia.py`
   - `cl_get_ftm_ids.py`
   - `cl_update_index.py`


cl/citations/tasks.py
```diff
@@ -214,7 +206,7 @@ def store_opinion_citations_and_update_parentheticals(
                 )
 
     # Finally, commit these changes to the database in a single
-    # transcation block. Trigger a single Solr update as well, if
+    # transcation block. Trigger a single update as well, if
     # required.
```
mlissner, line 210: This whole step is removed, so we can remove the comment about it.

```suggestion
    # transcation block. 
```

cl/search/models.py
```diff
@@ -3075,155 +2891,26 @@ def ordered_opinions(self):
...
-        except NoReverseMatch:
-            raise InvalidDocumentError(
-                "Unable to save to index due to missing absolute_url "
-                "(court_id: %s, item.pk: %s). Might the court have in_use set "
-                "to False?" % (self.docket.court_id, self.pk)
-            )
-
-        # Opinion
-        search_list = []
-        text_template = loader.get_template("indexes/opinion_text.txt")
```
mlissner, line 3199: I think everything in the cl.search.templates.indexes directory can be removed, right?

cl/stats/views.py
```diff
@@ -17,15 +16,13 @@ def health_check(request: HttpRequest) -> JsonResponse:
     """Check if we can connect to various services."""
     is_redis_up = check_redis()
     is_postgresql_up = check_postgresql()
-    is_solr_up = check_solr()
```
mlissner, line 20: Let's add an Elastic health check here, no?

mlissner (approved):
Nothing major. A couple more things we can delete, one feature we should upgrade, and I think we're there. Pretty easy to review, really.

cl/search/models.py
ERosendo: That's correct. I'll remove the directory

cl/citations/tasks.py
ERosendo: alright!

cl/stats/views.py
ERosendo: Sure! That's a great idea.

ERosendo:
@mlissner I've implemented the changes you suggested.

mlissner (approved):
AWESOME! 💀

Issue #4308, opened 2024-08-14
Optionally save response content and headers, to debug failing scrapers


grossir:
Some sites behave differently when scraped from the server or from local environments. It would be useful to have an optional feature to collect extra information about the failing request. For a start, the returned HTML or JSON and the response headers. To further reduce the friction, we would store these documents on a bucket accessible to the developer. 

The command should look something like:
`./manage.py cl_scrape_opinions --courts juriscraper.opinions.united_states.state.minn --verbosity 3 --save-responses`

To actually implement it we can just access the `site.request["response"]` object. Then, store what we need on AWS?

The most recent example is `minn`, which doesn't fail anymore due to SSL errors or a forbidden status response, but shows no results when ran on the server, in contrast to a local run.

@flooie

mlissner:
This sounds great, thanks @grossir. I do wonder if just storing everything is the way to go though:

 - S3 Glacier Instant Retrieval pricing is $0.004 per GB/month. Even if we had 100GB of data, that'd be $4/month. 
 - We can set up [S3 Lifecycle Rules](https://docs.aws.amazon.com/AmazonS3/latest/userguide/object-lifecycle-mgmt.html) to delete content when it's a few months old. 
 - Lifecycle transitions cost 2¢/1000 items. 

So I think for pennies we could store it all and you'd always have it for debugging.

grossir:
Some cost estimations confirm this would be cheap

- Number of scrapers in Juriscraper: 250
- Scrape frequency: every hour, every day
- Size of data stored: 30MB hourly   /  0.75 GB daily    /  22 GB per 30 days
- Number of objects stored: 250 per hour /  6000 per day / 180 000 per month
- Glacier Instant Retrieval minimum storage duration: 90 days

We could set a lifecycle rule for deleting the objects after the 90 days (minimum storage duration). If that's the case, we would be paying:

- 66 GB (3 months worth of data) of storage per month * 0.004 per GB = 0.26 USD
- 180 000 PUT per month * 0.02 per 1000 requests = 3.6 USD
- a few GET / LIST / etc * $0.01	per 1000 requests = ?
- a few retrievals * 0.003 per GB retrieved = ?
- 180k deletes per month, seem to be free

So, putting the objects in the bucket seems to be the most expensive part; we could actually keep them for more time for little relative cost.

--------------------------------------

[...]

PR #4794, opened 2024-12-05
feat(scrapers): save raw scraped responses on S3 

grossir:
Solves https://github.com/freelawproject/courtlistener/issues/4308

- adds a  function to upload response content and headers to S3
- implements a new storage class in the private bucket, using Glacier Instant Retrieval
- changes the scraper commands to pass the new function to always store scraped responses
- pending changing the lifecycle rules to make the stored objects expire


cl/scrapers/utils.py
```diff
@@ -446,3 +449,49 @@ def scraped_citation_object_is_valid(citation_object: dict) -> bool:
...
+    :return None
+    """
+
+    storage = S3GlacierInstantRetrievalStorage()
+    response = site.request["response"]
+    headers = dict(response.headers)
+
+    scraper_id = site.court_id.split(".")[-1]
+    scrape_type = site.court_id.split(".")[1]  # opinions or oral args
+    now_str = datetime.now().strftime("%Y_%m/%d/%H_%M_%S")
```
mlissner, line 471: I think I'd suggest: 

```suggestion
    now_str = datetime.now().strftime("%Y/%m/%d/%H_%M_%S")
```

That'd give you daily separation, which seems generally good.

cl/scrapers/utils.py
```diff
@@ -446,3 +449,49 @@ def scraped_citation_object_is_valid(citation_object: dict) -> bool:
...
+        content = json.dumps(content, indent=4)
+    elif isinstance(content, dict):
+        content["response_headers"] = headers
+        content = json.dumps(content, indent=4)
+    else:
+        # Appending the response headers to the content would save a
+        # PUT request and an extra download step, but it may cause
+        # encoding and upload conflicts, so we save them apart
+        json_string = json.dumps(dict(response.headers), indent=4)
+        storage.save(f"{base_name}_headers.json", ContentFile(json_string))
```
mlissner, line 494: If you just have a rule of always uploading two files, you could really simplify this, right? Just send response.content in file number 1 and response.headers in file number 2?

Feels a bit tidier in a way too, since you won't be adding headers to a scraped JSON object?

mlissner (changes requested):
Looks generally good. Nice job making the saving routine pluggable by others. Just one suggestion to simplify and another to tweak the download location.

Thanks!

cl/scrapers/utils.py
```diff
@@ -446,3 +449,37 @@ def scraped_citation_object_is_valid(citation_object: dict) -> bool:
...
+    base_name = f"responses/{scrape_type}/{scraper_id}/{now_str}"
+
+    headers_json = json.dumps(dict(response.headers), indent=4)
+    storage.save(f"{base_name}_headers.json", ContentFile(headers_json))
+
+    try:
+        # both tests for and parses JSON content
+        content = json.loads(response.content)
+        extension = "json"
+    except:
```
mlissner, line 480: Definitely need the exception type here, but otherwise this looks great and ready to merge to me.

mlissner (approved):

cl/scrapers/utils.py
grossir: @mlissner I added the exception types, I think it should be ready to merge

mlissner:
Merged! Can you create an infrastructure task for setting up S3 however needed, please?

Issue #4788, opened 2024-12-04
Case caption on new case page needs to be an H1 tag


mlissner:
Using a div for this is bad for accessibility, bad for printing, and bad for SEO. This should be an H1 (but should look the same).

Issue #4786, opened 2024-12-04
Nofollow and noindex the new PDF page on new case law page


mlissner:
I don't think we want Google wandering onto our PDF page, so let's add a `nofollow` to the link, and add `noindex` to the page. 

I suspect doing the same for the other tabs is also probably wise, but we can punt thinking too hard about that for now.

Issue #4785, opened 2024-12-04
HTML `<title>` tags need tweaks on new case pages


mlissner:
The various title tags need to be tweaked to say what's what. I currently have three of the tabs open, and I see this:

![Image](https://github.com/user-attachments/assets/d563ea97-0c2f-4466-adb0-1fff154eeba3)

They should say:

 - Authorities for ...
 - Citations to ...
 - Similar cases to ...
 - Summaries of ...
 - Download PDF for ...

And drop the citations from the title tag (they're too long).

PR #4795, opened 2024-12-06
Small Front End UI Tweaks

flooie:
PR addresses three small front end cleanups.

#4785 Tweak tabbed text for each tab
#4786 No follow and no index for PDF case law page
#4788 Case caption converted to H1 tag


mlissner (approved):
Nice!

Issue #4741, opened 2024-11-26
Add tailwind to dev and prod environments for CourtListener


mlissner:
We use tailwind with Django in bots.law, so we should be able to just copy that over from the freelawproject/big_cases2 repo so that it works here too.

mlissner:
Use tailwind v4!

ashkonf:
PR: https://github.com/freelawproject/courtlistener/pull/4800

References:
* https://django-tailwind.readthedocs.io/en/latest/installation.html
* https://tailwindcss.com/blog/tailwindcss-v4-alpha#using-the-cli
* https://github.com/search?q=repo%3Afreelawproject%2Fbigcases2%20tailwind&type=code

Issue #4738, opened 2024-11-26
Analyze using Vue vs. Django-Components


mlissner:
Figure out:
 - How do we do basic components using these libraries and what are the pros/cons
 - What's the basic approach to getting them into our dev env?
 - How would we deploy them?
 - Etc.

Optionally: Can add a simple component to /components/ for each, if that makes sense.

elisa-a-v:
[Here](https://medium.com/swlh/using-vue-js-alongside-django-template-45cbe17b3646)'s a very easy way to gradually start integrating Vue into our codebase; the only downside is you can't use [Single File Components](https://vuejs.org/guide/scaling-up/sfc) (SFC) when you include Vue via CDN cause they require a build step, and SFCs are extremely useful and one of Vue's key features, so I'd definitely look for an alternative that supports them.

One option would be to create a SPA from scratch. I think the hardest part would be creating the API, because if we use Vue the frontend would be fairly easy to develop, but idk most of the logic is already there and I think DRF makes it really easy too so it might not be that bad.

I **love** [Quasar](https://quasar.dev/) for this, it's a very underrated meta framework that works on top of Vue3 and provides several customizable components out of the box, besides [taking care of many things for you](https://quasar.dev/introduction-to-quasar). I believe it is HIGHLY underrated, and it provides several modes (SPA being one of them) that we could use which would even make it possible to eventually launch native mobile apps with the same codebase.

One of Quasar's modes is SSR, which I haven't used but from what I've seen that sounds like a viable option for this. I had a [chat with ChatGPT](https://chatgpt.com/share/674f58e2-a478-800f-8400-35041f83d6c8) about it if you want to read it, bottom line is:

> Rebuilding your frontend using Quasar's SSR mode can offer significant benefits, especially regarding SEO and initial load performance. However, it comes with increased complexity and potential costs. If your project's success heavily relies on search engine visibility and delivering content quickly to users, SSR is a strong candidate.
>
> On the other hand, if you prefer a simpler development process, lower server costs, and your application doesn't depend heavily on SEO (e.g., an internal tool or a user-authenticated app), an SPA might be more appropriate.

There are other alternatives I haven't used for SSR using Vue3, like Nuxt 3, so that's something we might want to consider too.

mlissner:
Thanks Elisa, this is really helpful to get the conversation going. We're going to be figuring out our goals as we go a bit, unfortunately, but I'll highlight a few that I think help move this conversation forward:

[...]

Issue #4736, opened 2024-11-26
Create blank components page


mlissner:
Set up a view with a basic template that loads at `/components/`. For now it's blank. In a future issue, we'll begin populating it with either vue, django-components, or something else.

ashkonf:
PR: https://github.com/freelawproject/courtlistener/pull/4800

leohentschker:
Hi there! I took a less ambitious whack at this item with a smaller pr [here](https://github.com/freelawproject/courtlistener/pull/4852), as I would love to start incrementally adding in tailwind support.

If this looks good I'll plan on opening up a second pr [here](https://github.com/leohentschker/courtlistener/pull/3) that should add core Tailwind support.

mlissner:
Leo, @ERosendo will be reviewing @ashkonf's PR and yours soon. You're officially scheduled for review. :)

elisa-a-v:
@ERosendo and I think we should probably just close this issue and have @leohentschker open a new PR with his implementation of tailwind support to address #4741, would that make sense to you @mlissner ? Do we really want a blank POC in the repo? might just be enough to test the tailwind support locally then go on with #4815, #4739, and #4740 , unless I'm missing something.

mlissner:
Sure, sounds fine! Thanks.

PR #4800, opened 2024-12-08
Installed tailwind, added components page, created toy components with django-components and embedded Vue

ashkonf:
Screenshot 2024-12-08 at 6 42 53 PM

* https://github.com/freelawproject/courtlistener/issues/4736
* https://github.com/freelawproject/courtlistener/issues/4741
* https://github.com/freelawproject/courtlistener/issues/4738


mlissner:
@ashkonf, can you please sign the CLA per the instructions above?

cl/assets/templates/base.html
```diff
@@ -54,6 +54,10 @@
 
   <link href="{% static "css/font-awesome.css" %}" rel="stylesheet">
 
+  <link href="{% static "css/app.css" %}" rel="stylesheet">
```
mlissner, line 57: I'm finding this a bit weird. We link to app.css, which imports tailwindcss. Gotta be a better way. I think Bots.law just has:

```
  {% tailwind_preload_css %}
  {% tailwind_css %}
```

Honestly not sure what the preload vs. regular do, but definitely simpler/cleaner?

cl/simple_pages/templates/components.html
```diff
@@ -0,0 +1,31 @@
+{% extends "base.html" %}
+{% load humanize %}
```
mlissner, line 2: I don't think you need `humanize`, do you?

cl/simple_pages/templates/components.html
```diff
@@ -0,0 +1,31 @@
...
+{% load component_tags %}
+{% load static %}
+
+{% block title %}Components – CourtListener.com{% endblock %}
```
mlissner, line 6: Nit:

```suggestion
{% block title %}Components Library – CourtListener.com{% endblock %}
```

cl/simple_pages/templates/components.html
```diff
@@ -0,0 +1,31 @@
...
+{% block sidebar %}{% endblock %}
+
+{% block content %}
+  <div class="col-xs-12">
+    <h1 class="">Components</h1>
```
mlissner, line 11:
```suggestion
    <h1 class="">Components Library</h1>
```

cl/simple_pages/templates/components.html
mlissner, line 1: Let's add a paragraph to this page explaining what it's for and why it exists.

mlissner (changes requested):
OK, this PR has a lot of bits to it to make things work, and most of it is so complicated/fragile I can barely figure it out. I think @ERosendo might have some better intuition, since he set up tailwind on Bots.law, but I did make a number of comments for things I noticed. 

I think we should approach this PR in two steps:

1. Focus on cleaning up the various configurations, settings, installations, etc, so they're polished and optimized.

    As part of this, I think we'll want to include the reloader in our docker-compose.yaml file, similar to how big cases does it.

3. Dial in the implementation of the two components, so we can see the parts that really matter.

@ashkonf, can you work on the comments I've made so far, and @ERosendo, can you review after that, please?

mlissner:
I'm setting this for Eduardo's next sprint, but Ashkon, if you can address my suggestions before then, that'd be great.

cl/settings/django.py
```diff
@@ -160,6 +172,10 @@
     "django_elasticsearch_dsl",
     "pghistory",
     "pgtrigger",
+    "tailwind",
+    "theme",
```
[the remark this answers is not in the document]
ERosendo: > In Big Cases, we put this in the web app, but even that feels off. Maybe it belongs in the cl.assets directory?

I attempted to configure the assets directory as the target for Tailwind's asset generation, but encountered the following error: `cl.assets is not in INSTALLED_APPS`. This suggests that the directory specified for the Tailwind command must be registered within the `INSTALLED_APPS` setting.

To resolve this, we can either leverage an existing module such as `simple_pages` or explicitly include the dotted path to the asset folder in the `INSTALLED_APPS` list.

cl/assets/templates/base.html
ERosendo: > Honestly not sure what the preload vs. regular do

According to the documentation, the {% tailwind_css %} tag generates a stylesheet link for the app that has all the tailwind assets. This is the only step required to include Tailwind CSS on a page. The preload tag generates a preload directive for the stylesheet, which improves loading performance in production.

cl/settings/django.py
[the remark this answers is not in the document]
mlissner: No harm in adding it to INSTALLED_APPS, right?

PR #4807, opened 2024-12-10
Remove unnecessary log message in lookup_and_save()

quevon24:
We got this in sentry because it is logged as an error: `Free opinion scraper: Got multiple results while attempting save.`

The detail is that later in the code a solution is applied for this, so i removed the unnecessary log message in lookup_and_save()


flooie (approved):
Looks removed to me

PR #4811, opened 2024-12-11
Remove solr dock from tests, docker, and remaining strays

mlissner:
This addresses part of #4749, but the wiki remains.

I went ahead and did this because it was just too tempting to nuke it from orbit, and I wanted to do a grep for "Solr" across the entire codebase and purge whatever was left (not much!).

Changes:
 - Remove solr from tests and docker compose
 - Tweak a few variable names and strings to remove solr references
 - A handful of odds and ends I had laying around

After this PR, solr does not appear in CL at all anymore. Not once.


elisa-a-v (commented):
Thank you getting this done! It works fine, but I can still find "solr" referenced in a couple places besides the migration guide and the API docs:
- Docstring for `get_docket_by_pacer_case_id` states that the method returns "A dict indicating if we need to update Solr".
- Docstring for `get_object_list` simply states "Perform the Solr work".
- The help text for the Docket's field `date_last_index` says "The last moment that the item was indexed in Solr." 
- The help text for the `query_time_ms` field in the SearchQuery model mentions solr in the help text.
- The SearchQuery model still has Solr as one of the choices for the `engine` field.

For the first three, would it be safe to replace "solr" by "elasticsearch"? or is there something else that needs to be removed/updated?

I guess we're keeping the last two due to the data that's already in the DB, but I think it might be a good idea to add some notice that this choice has been deprecated, maybe in the help text, the display text for the SOLR engine choice, or simply a comment.

Other than that, LGTM! I think we could merge this PR and address these when we update the wiki.

mlissner:
Great, thanks Elisa. I just updated the strings and left the model stuff. I don't think it's really worth a migration. Merging!

PR #4825, opened 2024-12-16
fix(api) adds v4 usage to api tracking

### LINKED ISSUE #4787: API Recent Usage Page doesn't include v4 usage
You can see your recent API usage here: https://www.courtlistener.com/profile/api-usage/

But it doesn't seem to show v4 usage. Hopefully just a small tweak to make it always work?

leohentschker:
Resolves #4787 by checking usage over a set of versions (`v3` and `v4`) instead of just `v3` in the `invert_user_logs` function.

**Note**: I don't actually know that Redis interface is the same across v3 and v4! Definitely needs eyes from someone more familiar. Also, this is my first contribution to the project, would love help on any and all conventions to confirm I'm contributing in a way that is helpful.


cl/api/utils.py
```diff
@@ -736,7 +736,10 @@ def invert_user_logs(
 
     dates = make_date_str_list(start, end)
     for d in dates:
-        pipe.zrange(f"api:v3.user.d:{d}.counts", 0, -1, withscores=True)
+        for version in ["v3", "v4"]:
+            pipe.zrange(
+                f"api:{version}.user.d:{d}.counts", 0, -1, withscores=True
+            )
```
mlissner, line 742: I haven't read the code *super* carefully, but I'd be pretty surprised if this works. Before your change `results` variable would contain a list of results, one per day. And the `dates` variable would contain a list of dates, one per day. From there you can `zip` them together, and you're off to the races. 

But when you do the above, you wind up with a list of counts, with 2 values per day, if somebody has used both v3 and v4 on a given day. 

I suspect this is only working because you didn't test that? I think the solution is to add v3 and v4 together.

mlissner (changes requested):
Thanks for the handy PR. I appreciate that you figured out where to work on the code and that you included a test. But I think you've got a bug, which I highlighted for you. 

Let me know what you think, and thank you again!

cl/api/utils.py
leohentschker: Ah! My mock is super wrong then, appreciate the review. I'll read the docs on zrange and update.

cl/api/utils.py
```diff
@@ -732,28 +732,43 @@ def invert_user_logs(
...
 
     dates = make_date_str_list(start, end)
-    for d in dates:
-        pipe.zrange(f"api:v3.user.d:{d}.counts", 0, -1, withscores=True)
-    results = pipe.execute()
+
+    versions = ["v3", "v4"]
+
+    for version in versions:
+        pipe = r.pipeline()
```
mlissner, line 741: I don't see how this could work. You establish a new pipe for each version, so this would always throw away the pipe (and it's results) for v3.

mlissner (changes requested):

cl/api/utils.py
leohentschker: Ah you're correct, I'm sorry for the silly mistake there.

leohentschker:
@mlissner -- I've updated this but I'm worried I'm wasting your time here with reviews! Any way that I can do some more testing on my own to make sure this is working as anticipated? I'm running API calls locally but I have yet to see them get tracked in the local redis store.

mlissner:
Yeah, I don't think it's quite right, but I'll assign it back to one of our devs who should be able to tweak and land this. (It was on her backlog anyway before you grabbed it!)

Thank you!

elisa-a-v (commented):
Hi @leohentschker, thank you so much for your work on this, your eagerness to help and persistence in improving the code is really appreciated! After reviewing the code I decided to go in a different direction, but I wanted to share some thoughts about approaching problems like this one, which might help with future contributions.

While you correctly identified code that needed changing, the modifications made it harder to follow the logic. Code clarity is crucial in such big projects like CL that have so many contributors, because it helps everyone understand the logic, and makes debugging way easier when issues arise.

Some ways you could improve the code clarity would be to:
- write self-documenting code using descriptive names, adding comments when the logic isn't immediately clear from the code itself.
- break down complex logic into clear, smaller steps.

Also, it's great you added unit tests, but how did you go about verifying this implementation besides them? it's important to use multiple methods to test your code, especially when you're testing complex, non-trivial logic--maybe this could have been an issue? Notice [these tests](https://github.com/freelawproject/courtlistener/blob/4787-include-v4-usage/cl/api/tests.py#L3304-L3457) fail with your implementation.

I hope you find this feedback helpful, and thanks again for your time and effort!

@mlissner I guess I should just go ahead and close this?

leohentschker:
Appreciate the feedback @elisa-a-v ! Starting on a redis issue here was probably the wrong call. I’ll close this one out

PR #4854, opened 2024-12-23
feat(authorities): link docket authorities to search results

### LINKED ISSUE #4134: Link docket authorities to search results
When you're looking at the [authorities](https://www.courtlistener.com/docket/63337698/authorities/united-states-v-twitter-inc/) page for a docket, it shows you all the case law that has been cited in the case so far:

image

That's cool, but it leaves a big question unanswered: Which documents are making the citations to those cases?

For example, in the screenshot above, we learn that _U.S. v. Arthur Young_ has been cited six times. But which documents do that?

Well, that case has an ID of `111119`, so we can actually check using the search engine with a [query](https://www.courtlistener.com/?q=docket_id%3A63337698%20cites%3A111119&type=r&order_by=score%20desc) like: 

    docket_id:63337698 cites:111119

That reveals:

image

Cool! 

So there is an easy way to fix this. Just make the word "references" into a link to the search results, so it looks like:

image

And the link goes to the search results. 

One gotcha: The citations are to opinions, not clusters, but the authorities page shows clusters. The way to solve this is to make the link use the opinion IDs, not the cluster IDs, and to be sure to include all of the sub-opinions for the cluster in an OR query. For example, cluster 123 might have sub-opinion 345 and 456. You might think that the query should be `docket:whatever cites:123`, but that'd be wrong. The correct query is `docket:whatever cites:(345 OR 456)`. We do this elsewhere too.

elisa-a-v:
Resolves #4134

authorities link

It took me a while to get the authorities to show up in the docket page, so I added a note in the `clone_from_cl` docstring hinting devs to use `find_citations_and_parantheticals_for_recap_documents` in the django shell to achieve this (thanks @ERosendo for the tip!)

### DIFF
--- a/cl/opinion_page/templates/includes/authorities_list.html
+++ b/cl/opinion_page/templates/includes/authorities_list.html
@@ -3,7 +3,14 @@
 <ul>
   {% for authority in authorities %}
   <li>
-    {{ authority.depth }} reference{{ authority.depth|pluralize }} to
+    {% if docket %}
+    <a href="/?type=r&q=docket_id%3A{{ docket.pk }}%20AND%20cites%3A({{ authority.cited_opinion.cluster.sub_opinions.all|OR_join }})">
+    {% endif %}
+      {{ authority.depth }} reference{{ authority.depth|pluralize }}
+    {% if docket %}
+    </a>
+    {% endif %}
+    to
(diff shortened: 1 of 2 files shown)


mlissner (approved):
Looks great, thanks.

PR #4881, opened 2025-01-02
feat(api): include v4 in recent API usage

### LINKED ISSUE #4787: API Recent Usage Page doesn't include v4 usage

elisa-a-v:
Closes #4787


mlissner (approved):
I didn't check the tests, but the approach looks good to me. Eduardo, I'm going to assign to you because you're helping Elisa get set up, but I don't have a handle on your workload or that of @albertisfu, so if you feel busy, can you guys please discuss and see who makes more sense this sprint? And I'm happy to break a tie if you both feel busy?

cl/api/utils.py
```diff
@@ -704,56 +705,58 @@ def invert_user_logs(
...
+        _count = int(_count)
+        out.setdefault(_user_id, OrderedDict())
+        out[_user_id].setdefault(_date, 0)
+        out[_user_id][_date] += _count
+        out[_user_id].setdefault("total", 0)
+        out[_user_id]["total"] += _count
+
+    for d, v3_data, v4_data in zip(dates, results[::2], results[1::2]):
+        for user_id, count in chain(v3_data, v4_data):
+            update_user_counts(user_id, count, d)
```
ERosendo, line 759: I think we can use the `itertools.batched` function in the outer loop to process the data in chunks. This function iterates over the list, yielding tuples of items with a maximum size of `n`.

Here's the documentation for reference: [itertools.batched](https://docs.python.org/3/library/itertools.html#itertools.batched)

```suggestion
    for d, api_usage in zip(dates, batched(results, 2)):
        for user_id, count in chain(*api_usage):
            update_user_counts(user_id, count, d)
```

This approach enhances future maintainability. If a new API version (e.g., v5) is introduced, we simply need to adjust the batch size instead of modifying the slicing expressions for previous versions (v3, v4) and adding a new slicing expression for v5.

ERosendo: @elisa-a-v looks great!

ERosendo (approved):

Issue #4845, opened 2024-12-20
Next round of Pray and Pay revisions


mlissner:
It's getting really close. Thanks everybody for the work getting it so far. It's cool already, but we need the UX to be good for it to work. I hope this doesn't feel like moving goalposts. We're close! 

Here are my thoughts for the next round:

1. **DONE:** From the leader board, clicking on documents that aren't on page 1 of the docket doesn't work. For example, this link fails because the docket page only goes to entry 159 (and the link is for number 396):

    https://www.courtlistener.com/docket/63290154/missouri-v-biden/#entry-396

    A simple solution is to take the users to the document page instead:

    https://www.courtlistener.com/docket/63290154/396/missouri-v-biden/

1. **DONE:** Speaking of those pages, I just noticed they don't have the icon. Darn.

1. **DONE:** We still need to teach people what the emojis mean and how to find the prayer leader board. One thought is that the first time somebody prays for something, we could pop up a modal congratulates them on their first prayer and then tells them about the leaderboard, how it all works, and points them to the help page.
 
    We'd need a reliable way to know if the prayer was somebody's first. Maybe the API for creating a prayer could efficiently return if it's the first one the user has? 

1. **DONE:** We're going to set up an email list that reminds people to buy things with their PACER accounts at the end of each quarter. We'll just need a button for it so people can go to the sign up page, and we'll want that button on our help page with an explanation there too.

1. The leader board page:

    - **DONE:** The `h1` heading on the leader board needs space below it. Maybe add `.v-offset-below-3` to it?
    - **DONE:** Maybe change the heading to: "Community's Most Wanted PACER Documents"
    - **DONE:** I think the big blue box can be removed, but its idea was good. Instead, let's do a sub-heading that says: "122 prayers granted totaling $153.30." Same idea; friendly approach.
    - **DONE:** With the blue box gone, we need a place for:
       - the new email list sign up button: "Remind Me to Help"
          - This can pop open a modal explaining about the $30/quarter that you get for free, explaining that we can email to remind people to contribute towards the end of each quarter. The modal should have a button the user can click to go to the sign up page.
       - For "Your Prayers"
       - And for "Learn More" -> Goes to the help page

    As for where the buttons should go, I think it should mimic our new opinion page. It's not yet released, but it looks like this across the top:

    ![Image](https://github.com/user-attachments/assets/0058a14b-da2f-4290-bde9-01ff6221c3c4)

[...]

PR #4902, opened 2025-01-08
4845: Allow users to see how many prayers they have remaining

v-anne:
This PR enables users to see how many document requests they have remaining as part of the pray-and-pay project. It's part of #4845.


mlissner:
Looks good in theory. Setting to auto-merge!

PR #4905, opened 2025-01-10
fix(templates): remove stitcher from CourtListener

### LINKED ISSUE #2948: Remove stitcher from CourtListener
This should be easy. Stitcher is going away as a company, so we should stop advertising it. Simple enough.

elisa-a-v:
Closes #2948 

Also removed all logos and mentions of specific podcast apps to make it future proof in case more of them stop existing.


cl/lib/context_processors.py
```diff
@@ -65,7 +65,7 @@ def inject_settings(request):
     % reverse("advanced_oa"),
     'CourtListener has an <a href="%s">API</a> so anybody can easily use our data.'
     % reverse("api_index"),
-    'Oral argument podcasts <a href="%s">are available in variety of apps</a> like iTunes, Stitcher Radio, and Google Music.'
+    'Oral argument podcasts <a href="%s">are available in variety of apps</a> like iTunes, Pocket Casts, and Google Music.'
```
elisa-a-v, line 68: I replaced it with a different one instead of just removing it because it looked better when it was three things being listed instead of just "iTunes and Google Music" but I can also remove it if we don't want to add Pocket Casts for some reason 😅

mlissner: You know, I hate to say it, but now that you've drawn my attention here, I'm realizing that Google Music doesn't exist anymore either! Maybe we should just say something like, "available wherever you get your podcasts" — that'd be future-proof too, and we can just remove all the logos?

mlissner (changes requested):

cl/lib/context_processors.py
elisa-a-v: Yeah, that makes sense! The only thing is that I personally think the logos make it look a bit nicer and more inviting, so I'd consider how this will look like after the redesign. If this code isn't going to be used anymore, then we might as well keep them for now :shrug:

mlissner: Sure, if you want. I think after the redesign it'll just be a narrow help page, like the others, so it might not be so bad without them, but sure, either way. :)

mlissner (approved):
Looks OK, but we should delete the static files too, please.

mlissner (approved):

mlissner:
Set for auto merge, thank you!

PR #4912, opened 2025-01-12
4845 pray and pay revisions pt. 3

v-anne:
This is part 3 of the revisions to pray-and-pay outlined in #4845.

Included:
- adding tab to get to user prayer page from the dropdown
- caching user prayer stats for 24 hours to avoid overloading db with repeated queries
- modifying function that retrieves user stats to use same class as `get_lifetime_prayer_stats `


cl/favorites/utils.py
```diff
@@ -301,14 +315,17 @@ async def get_user_prayer_history(user: User) -> tuple[int, float]:
...
 
-    return count, total_cost
+    # return count, total_cost
 
+    data = {
+        "prayer_count": count,
+        "distinct_count": "",
+        "total_cost": f"{total_cost:,.2f}",
+    }
+    one_day = 60 * 60 * 24
```
mlissner, line 325: I think I'd make this cache *much* smaller. Maybe 30s or a minute or two? People will want to see it change, so the cache will confuse them if they don't. I also am not sure we need the cache. I think these queries should be pretty fast?

cl/favorites/utils.py
```diff
@@ -301,14 +315,17 @@ async def get_user_prayer_history(user: User) -> tuple[int, float]:
 
     total_cost = await compute_prayer_total_cost(filtered_list)
 
-    return count, total_cost
+    # return count, total_cost
```
mlissner, line 318: We'll want to remove this line.

cl/favorites/templates/user_prayers.html
```diff
@@ -14,7 +14,7 @@
 {% block content %}
   <div class="col-xs-12">
     <h1 class="text-center v-offset-below-3">{% if is_page_owner %}Your PACER Document Prayers{% else %}PACER Document Requests for: {{ requested_user }}{% endif %}</h1>
-    {% if is_page_owner %}<h3 class="text-center"><b>{{ count|intcomma }}</b> prayers granted totaling <b>${{total_cost|floatformat:2 }}</b> (<b>{{ num_remaining }}</b> remaining today).</h3>{% endif %}
+    {% if is_page_owner %}<h3 class="text-center"><b>{{ user_history.prayer_count|intcomma }}</b> {{ user_history.prayer_count|pluralize:"prayer,prayers" }} granted totaling <b>${{ user_history.total_cost }}</b> (<b>{{ num_remaining }}</b> remaining today).</h3>{% endif %}
```
mlissner, line 17: Did we lose the floatformat here?

cl/favorites/tests.py
```diff
@@ -891,18 +891,18 @@ async def test_get_top_prayers_by_number_and_age(self) -> None:
 
     async def test_get_user_prayer_history(self) -> None:
         """Does the get_user_prayer_history method work properly?"""
-        # Prayers for user_2
-        await create_prayer(self.user_2, self.rd_4)
+        # # Prayers for user_2
+        # await create_prayer(self.user_2, self.rd_4)
```
mlissner, line 895: Little cleanup here.

mlissner (commented):
Not sure if this is ready, but I went ahead and did a quick review. Hopefully helpful. Thank you as always!

cl/favorites/templates/user_prayers.html
v-anne: It's no longer necessary because of this line in `get_user_prayer_history`:

```
    data = {
        "prayer_count": count,
        "distinct_count": "",
        "total_cost": f"{total_cost:,.2f}",
    }
```

cl/favorites/utils.py
v-anne: Changed it to a minute. My reasoning for caching these stats was similar to that for caching those on the leaderboard--it may be resource intensive to constantly compute statistics for any given user.

v-anne:
I'm continuing to break PRs into chunks and this should be ready to go.

mlissner (approved):
Great! Merging!

cl/assets/templates/base.html
```diff
@@ -139,14 +139,16 @@ <h1>You did not supply the "private" variable to your template.
                     <li><a href="{% url "profile_alerts" %}"
                            tabindex="203"><i class="fa fa-bell-o gray fa-fw"></i>&nbsp;Alerts</a></li>
                     <li><a href="{% url "profile_notes" %}"
-                           tabindex="205"><i class="fa fa-bookmark-o gray fa-fw"></i>&nbsp;Notes</a></li>
+                           tabindex="204"><i class="fa fa-bookmark-o gray fa-fw"></i>&nbsp;Notes</a></li>
                     <li><a href="{% url "tag_list" username=user.username %}"
                            tabindex="205"><i class="fa fa-tags gray fa-fw"></i>&nbsp;Tags</a></li>
+                    <li><a href="{% url 'user_prayers' user.username %}"
+                           tabindex="206"><i class="fa gray fa-fw"></i>&nbsp;Prayers</a></li><!--need to fix class-->
```
mlissner, line 146: Oops, just noticed this isn't behind the flag, and that it has a todo for the icon. It's probably fine at this point if the user prayers page is available to folks, but we should fix the icon quickly. Did you have a plan for this, @v-anne?

v-anne: I'll submit a hotfix to put it behind the flag for the time being.

v-anne: How about
'<i class="fa-solid fa-hands-praying"></i>' for the time being? It's not a perfect replica of the emoji but it seems to get the message across.

http://fontawesome.com/icons/categories/hands

mlissner: I kind of like this one: 

https://fontawesome.com/v5/icons/hand-holding-heart?f=classic&s=solid

But it's not supported by v4 of Font Awesome. Upgrading doesn't make sense since we're redoing the website anyway, so probably something else makes more sense, but I haven't found one I like. We could hide it for now, in any case.

v-anne: https://fontawesome.com/icons/hands-praying?f=classic&s=solid

I came across this one too, but it may have the same issue.

mlissner: It does. I think it also is kind if a bad icon!

ERosendo: > I kind of like this one:
> 
> https://fontawesome.com/v5/icons/hand-holding-heart?f=classic&s=solid
> 
> But it's not supported by v4 of Font Awesome. 

Hey @mlissner and @v-anne  sorry I'm a bit late to the conversation. I don't think an upgrade is necessary just for this one icon.

I checked the link and noticed there's an SVG version of the icon available. We can use that if we really like it.

Let me know what you think!

image

v-anne: I think that looks good!

Issue #4903, opened 2025-01-09
Use eyecite to validate citations incoming from juriscraper's `site.extract_from_text`


grossir:
We had an error with "PA Super" reporter, which is a variation of "Pa. Super."
Currently, this fails with a KeyError, but it will also fail due to the reporter type not matching

The object got from `extract_from_text`
```python
{
cluster_id: 10311260,
page: "3",
reporter: "PA Super",
type: 8,
volume: "2025"
}
```

https://github.com/freelawproject/courtlistener/blob/9f67ed6835422fe53d42a9c6c5206cf3bb9542b8/cl/scrapers/utils.py#L436-L449

We must imitate the behaviour of `eyecite.get_citations`, which handles both the variation and the type

```python
from eyecite import get_citations
In [5]: get_citations("2024 PA Super 3")[0].all_editions[0].reporter
Out[5]: Reporter(short_name='Pa. Super.', name='Pennsylvania Superior Court Neutral Citation', cite_type='neutral', source='reporters', is_scotus=False)

In [6]: get_citations("2024 PA Super 3")[0].all_editions[0].reporter.cite_type
Out[6]: 'neutral'
```

grossir:
After discussing this with @flooie , we decided to change juriscraper to return citation strings that can be parsed with `eyecite.get_citations` instead of returning already parsed out dictionaries suitable for the model. That way, both parsing and validation will be done in Courtlistener, using the same pipeline as used to process other citations incoming from scrapers

PR #4913, opened 2025-01-13
refactor(scrapers.tasks.update_from_text):  reuse make_citation in update_from_text

grossir:
Solves #4903

- Move make_citation from cl_scrape_opinions into scrapers.utils
- Move citation_is_duplicated from cl_back_scrape_citations into scrapers.utils
- Delete scraped_citation_object_is_valid, now we rely on eyecite, used by make_citation
- Refactor test site to account for changes


cl/scrapers/tasks.py
```diff
@@ -50,31 +49,32 @@ def update_document_from_text(
...
         elif model_name == "OpinionCluster":
             opinion.cluster.__dict__.update(data)
         elif model_name == "Citation":
-            data["cluster_id"] = opinion.cluster_id
-            if scraped_citation_object_is_valid(data):
-                _, citation_created = ModelClass.objects.get_or_create(**data)
-                metadata_dict["Citation"]["created"] = citation_created
+            citation = make_citation(data, opinion.cluster, court_id)
+            if not citation or citation_is_duplicated(citation, data):
+                continue
```
flooie, line 75: Would it make more sense to throw the error logging here - and not pass court id into make citation.

grossir: It's awkward to pass the court_id only for logging, but if we didn't do it inside this function we would have to make an outside `logger.error` call everytime `make_citation` returns None

grossir:
I fixed the tests; and added a call to `extract_doc_content` to `update_from_text`, for when the text extraction failed. I noticed we weren't covering this use case on the command, and it's actually a very important one.

Please give it another check @flooie , but notice that before we merge this, we would have to:
1. merge the https://github.com/freelawproject/juriscraper/pull/1298 in Juriscraper
2. publish a new juriscraper
3. update the poetry requirement in this PR to that new Juriscraper version

So that it doesn't break anything
I tested it in my local env with the juriscraper branch, and it's working!

```
# copy some PA Super cluster where citation extraction had failed; then

docker exec -it cl-django pip uninstall juriscraper
docker exec -it cl-django pip install https://github.com/freelawproject/juriscraper/archive/extract_from_text_citations.zip

docker exec -it cl-django python /opt/courtlistener/manage.py update_from_text --courts juriscraper.opinions.united_states.state.pasuperct --verbosity 3 --date-filed-gte=2024-12-01 --date-filed-lte=2026-01-01 --cluster-status=Published
```

poetry.lock
```diff
@@ -5690,6 +5937,6 @@ test = ["coverage[toml]", "zope.event", "zope.testing"]
 testing = ["coverage[toml]", "zope.event", "zope.testing"]
 
 [metadata]
-lock-version = "2.0"
+lock-version = "2.1"
```
flooie, line 5940: think this is causing the issue?

grossir: It seems there is no issue? 
Tests are passing and they imply building the docker images, so this poetry.lock seems to be working

flooie:
@grossir this looks great - and will be great to get these errors gone.

Issue #1178, opened 2020-02-10
Give aria attributes to citations


mlissner:
I just got off the phone with somebody at LII who has some expertise about making websites accessible. They pointed out that they put aria labels into their data, and made me realize we should probably put them in our citations. This is perhaps 4× more true now that we're adding links for supra, id, etc. The idea here is to avoid making these nearly unreadable by screenreaders. 

I'm going to send them an example of the HTML we've got, and then I'll get their suggestion in here for discussion.

mlissner:
Looks like we should add an `aria-label` attribute with the name of the case. That's not as easy as I'd hoped, but I guess it helps avoid a reader that just reads the word "Id." to you over and over again.

mlissner:
Here's the standard: https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html

It says:

> The intent of this Success Criterion is to help users understand the purpose of each link so they can decide whether they want to follow the link.

I asked on stackoverflow if there's a way to hide these links b/c I don't think they do much for users:

https://stackoverflow.com/questions/60174252/hide-link-from-screenreaders-but-show-text

One reply is to simply use `role="presentation"` seems like a really simple solution that isn't a terrible compromise.

mlissner:
More on this conversation here: https://github.com/w3c/aria/issues/1044#issuecomment-586087963

mlissner:
So stackoverflow votes me down and says to do `role=none`, which seems wrong. 

The w3c folks say to not do anything in particular (I think). 

It sounds like just doing "aria-label attribute with the name of the case" would be the easiest way forward (and something of a useful middle ground). 

If we go that way, truncating the case name to some number of words is probably wise since they can be really long.

grossir:
I have been reading ARIA related stuff to work on this issue. I will write some pointers I consider important

The `aria-label` attribute [supresses](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA8.html#:~:text=Per%20the%20Accessible%20Name%20and,link%20text%20by%20assistive%20technology.) the text content of the link
> Per the [Accessible Name and Description Computation](https://www.w3.org/TR/accname/) and the [Accessible Name and Description Computation](https://www.w3.org/TR/html-aam-1.0/#accessible-name-and-description-computation) in the HTML Accessibility API Mappings 1.0, the aria-label text will override the text supplied within the link. As such the text supplied will be used instead of the link text by assistive technology

[...]

PR #4932, opened 2025-01-15
feat(annotate_citations): use aria-attributes

grossir:
Solves #1178

Use aria-description on the anchor tag for a resolved citation. The description uses the case name, truncated if it is too long


mlissner (commented):
Nice. Seems easy enough! 

Did you see we have a function called best_case_name or get_best_case_name or similar? It might be a useful tool in this. 

Why do you need it to support dockets and opinions/clusters?

grossir:
> Why do you need it to support dockets and opinions/clusters?

From looking at `get_and_clean_opinion_text` I thought it would apply to RecapDocuments; but I see that it doesn't
https://github.com/freelawproject/courtlistener/blob/b9536c5bb516af4ca402123f5bd72685e38f264b/cl/citations/annotate_citations.py#L12

mlissner (approved):
Looks pretty good to me, but I don't know much about aria and I don't know this code super well. Onwards for a more careful review. Thanks Gianfranco!

mlissner:
Process-wise, I assigned this to you, Bill, for triage or to review, as you see fit. Gianfranco, I removed you as assignee, so that it will show up in Bill's sprint view. Bill, if you're not going to review you should assign it to somebody else AND make them reviewer.

cl/citations/annotate_citations.py
```diff
@@ -61,8 +63,12 @@ def generate_annotations(
                 "</span>",
             ]
         else:  # If successfully matched...
+            case_name = trunc(best_case_name(opinion.cluster), 60, "...")
```
flooie, line 66: Why TRUNCate this?

flooie: I mean I understand why we do it with the metadata field but an aria attribute is for accessibility and that user may want the full name that is being linked no?  for the description we are already on the page ...

grossir: > Why TRUNCate this?

In case it's too long; since we expect the `aria-*` to be useful specially for "Id", "Supra", "Ibid" where there is no further textual reference; but could be *too* repetitive, worst in case of long case names

>  I mean I understand why we do it with the metadata field but an aria attribute is for accessibility and that user may want the full name that is being linked no? for the description we are already on the page ...

The targeted case is a screen reader. Setting `aria-description` makes both the attribute value and the link content be read. You can check a longer comment about that [here](https://github.com/freelawproject/courtlistener/issues/1178#issuecomment-2587858711)

@flooie

mlissner:
Sorry, random thought: Do we need to worry about special characters in the aria-attribute and encode them?

flooie:
@mlissner I would think so. but how many random characters do you expect in our name strings?  Are you worried about characters messing up readers or messing up the HTML

mlissner:
@flooie, yes, I'm worried about a case with a name like, "A&M Univesity v. Foo" that would mess up our HTML.

flooie (approved):
Thanks @grossir

mlissner:
@flooie, @grossir, did we resolve this?

> I'm worried about a case with a name like, "A&M Univesity v. Foo" that would mess up our HTML.

flooie:
no - apologies

flooie:
@mlissner - in my limited testing - ampersands do not cause an issue with the HTML.

mlissner:
What's your testing, because I don't think you can put am ampersand into an HTML attribute without it being invalid. It's possible that it renders correctly, but that'd just be the browser fixing your error, and it's probably possible for worse errors to cause real problems (like what if there's a double quote in the case name, for example).

mlissner:
This says in html5 ampersands are kind of OK, but mostly not: https://stackoverflow.com/a/19442133/64911. But other characters like double quotes will definitely cause issues.

flooie:
I tested the ampersand and saw no issue extracting it back out - but Im sure you are right about quotes being an issue - ill make the change to escape theHTML

PR #4940, opened 2025-01-18
feat(redesign): introduce new_base.html with waffle flag

### LINKED ISSUE #4815: Create new `base.html` file and replace base.html for new design
I think all of our templates inherit from base.html with a tag like: 

```
{% extends "base.html" %}
```

We can't get very far on the redesign until base.html is replaced with a new template that uses tailwind and components. 

I think what we can do to make this transition is to add a waffle to all of the new pages.

Ideally, we could do something like:

```
{% load waffle_tags %}
{% flag "is_new_design" %}
   {% extends "new_base.html" %}
{% else %}
   {% extends "base.html" %}
{% endif %}
```

Unfortunately, that won't work because the `extends` tag has to go first, so what we need to do instead is move it to the view with something like:

```
def my_view(request):
    is_new_design = waffle.flag_is_active(request, "is_new_design")
    context = {"is_new_design": is_new_design}
    return TemplateResponse(request, "help/alert_help.html", context)
```

And then in your template, you can do:

```
{% extends is_new_view|yesno:"new_base.html,base.html" %}
```

From here: https://stackoverflow.com/a/7817990/64911

Later, when the new view is launched, we remove the flags, and we're good to go.

----

So to complete this ticket:

1. Create `new_base.html` as a clone of base.html, but using tailwind and components. As an MVP, it should just have blank spaces for the header, navbar, and footer.
2. Apply `new_base.html` to `/components` as an MVP.

elisa-a-v:
Closes #4815

This PR adds a middleware that checks a waffle flag (`use_new_design`) and changes the old template with the new one when a new template exists.

It also includes a copy of `base.html` without bootstrap, and a new template for the help index that extends from it, both of which need to be updated when tailwind support is ready; things do NOT look good!

To identify the new template we prepend "v2_" to the old template name, which means if the old `template_name` includes a dir, like `"help/index.html"`, the new template should be in `"v2_help/index.html"` and NOT in `"help/v2_index.html"`.

image

image


mlissner (commented):
This looks pretty good! We will need a way of handling tests though, so that they pass. Maybe an easy way to handle this is to make sure that the flag for the new design is off by default, so that the tests continue running on the old design instead of the new. Then, when we're closer to being done, we can flip the default and make sure they work on the new design. 

I think the rest looks good. This will be fun and weird.

mlissner:
Let's get tests passing before merging, but @erosendo, it might be good to have your review now anyway, just to see if you think the design is a good direction. 

I like it, I think, but this is a weird project with a lot of ways to do things, so more perspectives is probably smart.

elisa-a-v:
> Maybe an easy way to handle this is to make sure that the flag for the new design is off by default, so that the tests continue running on the old design instead of the new. Then, when we're closer to being done, we can flip the default and make sure they work on the new design.

The tests were indeed creating a flag that is on by default, as per waffle's config (`WAFFLE_CREATE_MISSING_FLAGS = True` and `WAFFLE_FLAG_DEFAULT = env("WAFFLE_FLAG_DEFAULT", default=True)`), but that shouldn't have been an issue for the missing templates as Django templates handles that (we're defaulting to the old template if the new one doesn't exist, even when the flag is on).

The issue with the tests was that we already had some templates that started with `new_` (`new_visualization.html`) so the tests were using the wrong templates. I changed the prefix to `v2_` instead and also added a check to see if the flag exists before asking waffle if it's active, this way we don't create a new one automatically and default to `False`.

elisa-a-v:
Oh no I broke the tests again 😮‍💨 I'll see what's going on

elisa-a-v:
The broken tests are those that are counting the queries executed (`ApiQueryCountTests`), and since we're manually checking the DB for the flag, we now have an additional query every time. I could either patch the test and leave a `TODO` to clean up after we've completely rolled out the new design, or remove the manual check and just live with the automatic flag creation, what do you think is best here @mlissner ?

ERosendo:
@elisa-a-v I really like the middleware you created! It seems like your approach addresses @mlissner's concerns about testing. However, I have a slight concern regarding the `_flag_is_active` method. the current implementation  introduces an additional database query on each page load, which may have a slight impact on overall system performance

While looking through the Django-waffle documentation, I came across some [decorators](https://waffle.readthedocs.io/en/stable/testing/automated.html#unit-tests) that might make testing easier. There's one called `override_flag` that I think could be useful. We could potentially use this to [decorate](https://github.com/jazzband/django-waffle/blob/f5fd6fc411372c196eaf59b383749ecabbee218f/waffle/tests/test_testutils.py#L308-L313) our testing classes from the [cases.py](https://github.com/freelawproject/courtlistener/blob/0368e5120407bc9ac2721fae05829058954ae733/cl/tests/cases.py) file (I believe we typically use these classes instead of Django's built-in testing classes directly).

I think exploring this alternative approach could be beneficial because it might also addresses Mike's concerns about the flag during testing, and it would save us an extra database query when loading pages. What do you think?

elisa-a-v:
Thanks @ERosendo !

You raise an important issue with the overhead introduced by the additional DB query, and now I see it makes no sense to keep that there. In practice, the only issue that the check is preventing is the automatic flag creation with a default `True` value. We shouldn't have an additional DB query in each and every request just for that!!

Let's revert the last two commits and make sure we create the flag before merging. I see I have enough permissions to do it, shall I create it right away just for superusers and staff? @mlissner

mlissner:
Checking the DB for a query like this is almost instant (the table is tiny and the OS will cache it), so I had accepted that it was going to be something we did. That aside though, I don't see how decorating the tests gets us out of doing the query. Whether or not the flag is true or false, we still have to check it, don't we?

Anyway, I guess my stance is I think whatever we do here is fine to get the tests passing and the waffle in place. If we can avoid the query, great, but I don't see how except to use a setting, and that seems annoying to deal with.

elisa-a-v:
> I don't see how decorating the tests gets us out of doing the query.

I don't think it does.

> if we can avoid the query, great, but I don't see how except to use a setting

Well, the query is only useful while the flag doesn't exist, because we want to prevent waffle from creating it with a default `True` value. After that, we don't really need it anymore, and the built-in waffle `flag_is_active` method is enough.

Note the tests were initially failing not because of a flag that was on, but because of the prefix used that clashed with the preexisting `new_visualization.html` template that was used as the new version of `visualization.html`—this caused the tests to fail because these are different templates used in different places, the former is not the replacement for the latter. Changing the prefix from `new_` to `v2_` fixed this issue (see [this commit](https://github.com/freelawproject/courtlistener/pull/4940/commits/7aa59be162c1de6aeb8fc84d50c1f952aa21e097) for which all checks were successful).

The "cleanest" solution code-wise, imo, is reverting the last two commits (the ones that add the DB query and the tests patch), that way we use the plain `flag_is_active` and the tests are fine. Only thing we need to be careful about is creating the flag before merging.

mlissner:
Sounds good to me. Thanks for explaining again.

elisa-a-v:
Cool. Do we have a preferred way to undo commits? I kinda like `git reset` better than `git revert` but I don't want to `push --force` without checking first.

mlissner:
Some around here have strong opinions. I'm happy with things that get it done. I'd probably reset and force, personally. :)

mlissner:
I'm going for it! Let's see how this goes and if needed, I can revert. Looks like the waffle is in place, so I think we should be good to go.

mlissner:
Total success: https://bsky.app/profile/michaeljaylissner.com/post/3lgmyvtejis2a

🚀

Issue #4920, opened 2025-01-14
Store NO_MATCH_RESOURCE citations


grossir:
We can store the unmatched citations in order to 
- know how many we are missing
- update the resolutions as we fill the gaps

We need to:
1. Create a model for storing them, under the `citations` app: maybe `UnmatchedCitation` ?
```python
class UnmatchedCitation(models.Model):
    citing_opinion_id = models.ForeignKey(Opinion, help_text="The opinion citing this citation")
    #  we currently don't match FullJournalCitation or FullLawCitation, may be interesting to track which are referenced more so as to decide which to implement first
    resource_type = models.TextField(help_text="The eyecite resource type")
    citation_string = models.TextField(help_text="The unparsed citation string in case it doesn't match the regular citation model below")
    
    # below is the same structure as the search.Citation model
    volume = models.SmallIntegerField(help_text="The volume of the reporter", null=True)
    reporter = models.TextField(
        help_text="The abbreviation for the reporter",
    )
    page = models.TextField(
        help_text=(
            "The 'page' of the citation in the reporter. Unfortunately, "
            "this is not an integer, but is a string-type because "
            "several jurisdictions do funny things with the so-called "
            "'page'. For example, we have seen Roman numerals in "
            "Nebraska, 13301-M in Connecticut, and 144M in Montana."
        ),
    )
    type = models.SmallIntegerField(
        help_text="The type of citation that this is.", choices=Citation.CITATION_TYPES
    )

```

2. Update ingestions processes so that when a new citation is ingested, it is looked up in the `UnmatchedCitation` citation table; and if it exists, all the Opinions citing it have their annotated HTML updated

mlissner:
That table looks about right to me. I think we'll want an index on some of the fields so that lookups in the table can be fast, and I guess we'll want a unique_together on the opinion and citation fields? `citing_opinion_id` can just be called `citing_opinion`.

I'm not sure that your second step is the right approach. Like, it might be better when adding citations just to flag the ones that are no longer Unmatched, and to have a second script that comes around to re-run the citation finder on those opinions. 

One thing I don't like about your approach is that if you add `n` citations that a single opinion lacked, then the citation finder will be run `n` times on that opinion. Like, if an opinion is missing four citations and you add all four in a batch, it'll get extracted four times, when once would have been enough.

A possibly-better way to do this is to add a column to the `UnmatchedCitation` table for `status`. It could have values like:

 - `unmatched` -> We don't have it in the Citation table
[...]

PR #4949, opened 2025-01-21
feat(citations): add UnmatchedCitation model and logic

grossir:
Solves #4920

- Add new model UnmatchedCitation on citations app
- refactor `cl.search.models.Citation` to create a BaseCitation abstract model to reuse on the UnmatchedCitation model
- updates `cl.citations.tasks.store_opinion_citations_and_update_parentheticals` to handle storing and updating unmatched citations
- updates `cl.search.signals` to update UnmatchedCitation status when a new Citation is saved
- updates citations command find_citations to delete all UnmatchedCitation when the `--all` argument is passed
- add tests
- add `update_unmatched_citations` command to trigger update for found citations


cl/citations/models.py
```diff
@@ -0,0 +1,96 @@
...
+            "The citing Opinion.html_with_citations update "
+            "failed because the citation is ambiguous",
+        ),
+        (FAILED, "The citing Opinion.html_with_citations update failed"),
+    )
+    citing_opinion: models.ForeignKey = models.ForeignKey(
+        Opinion,
+        help_text="The opinion citing this citation",
+        on_delete=models.CASCADE,
+        related_name="eyecite_citations",
```
mlissner, line 45: It's usually best to do the name of the model. I'd suggest:

```suggestion
        related_name="unmatched_citations",
```

cl/citations/models.py
```diff
@@ -0,0 +1,96 @@
...
+        help_text="The unparsed citation string in case it doesn't match the "
+        "regular citation model in BaseCitation"
+    )
+    court_id: models.TextField = models.TextField(
+        help_text="A court_id as identified by eyecite from the opinion's "
+        "context. May be useful to know where to find missing citations"
+    )
+    year: models.TextField = models.TextField(
+        help_text="A year identified by eyecite from the opinion's context"
+    )
```
mlissner, line 61: Why text instead of integer for this?

mlissner (commented):
I didn't read all the code, but I did a quick look through at the models (since I always have opinions about them!). One or two little comments, but I'll leave the rest to others. 

This seems well thought out and like a really nice improvement. It's going to be powerful to know which of our citations are the most needed.

cl/citations/models.py
grossir: You are right, this field can be of integer type; even of smallint type

The regex used to get `metadata.year` is `\d{4}`
https://github.com/freelawproject/eyecite/blob/0909c26841f9aff42f22d029a6c565d1a4dcde07/eyecite/regexes.py#L130-L139

flooie:
@quevon24 can you review this please?

cl/citations/tests.py
```diff
@@ -2449,3 +2458,93 @@ async def test_can_throttle_user_exceeding_citation_limit_by_big_margin(
...
+    cluster = None
+    opinion = None
+
+    @classmethod
+    def setUpClass(cls):
+        cls.cluster = OpinionClusterFactoryWithChildrenAndParents()
+        cls.opinion = cls.cluster.sub_opinions.first()
+        UnmatchedCitation.objects.all().delete()
+
+    def test_1st_creation(self) -> None:
```
quevon24, line 2489: Maybe split this big test into smaller tests, something like: test_save_unmatched_citations, test_signal_matching_citation and test_single_citation_resolved

quevon24 (commented):
All looks good, only a suggestion related to the tests

cl/citations/tests.py
grossir: Originally I did it that way; but then I was inheriting from `SimpleTestCase` and it would break other tests (not precisely sure why).
Then I changed to TransactionTestCase as parent class, but the signal wouldn't work unless they were all in the same function
You can check the changes in this commit
https://github.com/freelawproject/courtlistener/pull/4949/commits/4304ccb620d579ed270939ba50ff78a030b2e77c

quevon24: Perfect, then it's fine, after all the tests work without any problem. Everything else is fine.

mlissner:
@grossir, Chaco and I went to do these DB migrations today, but the tests are failing. Chaco is going to do the migrations anyway (they just add new tables), and we'll be able to merge this when the tests are passing.

grossir:
@mlissner this is fixed now; tests were failing from an eyecite update

PR #4955, opened 2025-01-23
feat(annotate_ciations): Add test for unsafe characters

flooie:
Aria description needs to be safely handled
Escape case names and added tests


grossir (approved):
This looks great. Let's merge it

PR #4956, opened 2025-01-23
4845 pray-and-pray revisions pt. 4

v-anne:
Making further revisions to the pray-and-pay feature, as outlined in #4845.

Changes include:
- Adding pagination to the community leaderboard
- Breaking the user prayers page into two tables, one of pending requests and one of granted ones
- Buttons on the leaderboard (help, your prayers, sign up for the email reminders)
- Button for mailing list on help page
- Modal that shows up when users make their first request, to explain how everything works.


v-anne:
@mlissner I'm getting tripped up by some tests that may be flaky. I didn't touch disclosures but their tests started failing now.

mlissner:
There's a flaky test for some reason. It'll get fixed shortly, you can ignore. If that's all there is holding this up, let me know and we'll review anyway.

v-anne:
This may be ready. I say "may" because I'm having issues rebuilding the docker containers after tailwind was added, so I can't actually see what the frontend looks like. That being said, no germane tests failed earlier, so I suspect nothing is seriously broken with these changes.

ERosendo:
> because I'm having issues rebuilding the docker containers after tailwind was added, so I can't actually see what the frontend looks like.

Hey @v-anne!, could you please describe the error messages or unexpected behavior you're encountering when trying to build the docker containers? Have you tried running `docker compose up -d --build`?

v-anne:
@ERosendo, there aren't any error messages per se, but the containers are hanging indefinitely at various stages.

For example, this was pending for almost 25 minutes:
> cl-postgres         | 2025-01-24 16:31:23.842 UTC [22] LOG:  checkpoint complete: wrote 3 buffers (0.0%); 0 WAL file(s) added, 0 removed, 0 recycled; write=0.003 s, sync=0.002 s, total=0.011 s; sync files=2, longest=0.001 s, average=0.001 s; distance=0 kB, estimate=0 kB

I suspect it may be because of my computer's age (5 years old), but I'm not completely sure.

mlissner:
That sort of sounds like it's just the last log line, and nothing had to be printed. Is it actually stuck?

v-anne:
> That sort of sounds like it's just the last log line, and nothing had to be printed. Is it actually stuck?

The script proceeded past that step but tripped up at other places. I quit the operation multiple times after it took 5+ hours without finishing.

A few weeks ago everything ran on this same computer in less than 20 minutes. Did something significant change?

mlissner:
Sorry @v-anne, I think this didn't get assigned to somebody properly, so it got a bit stuck. I just put Eduardo on it, so it's on his list now, but he has some high priority work that will have to come first. But now we're on track at least!

v-anne:
I ultimately got Docker to work, but it wasn't pretty. I deleted the codebase and the containers from my machine, cloned the repo again, and then set up the containers afresh.

v-anne:
Getting this strange error in relation to the pagination on the community leaderboard. Not quite sure how to fix it.

The buttons for the pagination show up fine, but when I try clicking them this shows up:

```
Environment:


Request Method: GET
Request URL: http://0.0.0.0:8000/prayers/top/?page=2

Django Version: 5.1.5
Python Version: 3.13.1
Installed Applications:
['daphne',
 'pghistory.admin',
 'django.contrib.admin',
 'django.contrib.admindocs',
 'django.contrib.contenttypes',
 'django.contrib.auth',
 'django.contrib.humanize',
 'django.contrib.messages',
 'django.contrib.sessions',
 'django.contrib.sites',
 'django.contrib.sitemaps',
 'django.contrib.staticfiles',
 'corsheaders',
 'hcaptcha',
 'markdown_deux',
 'mathfilters',
 'rest_framework',
 'rest_framework.authtoken',
 'django_filters',
 'storages',
 'waffle',
 'admin_cursor_paginator',
 'django_elasticsearch_dsl',
 'pghistory',
 'pgtrigger',
 'cl.alerts',
 'cl.audio',
 'cl.api',
 'cl.citations',
 'cl.corpus_importer',
 'cl.custom_filters',
 'cl.disclosures',
 'cl.donate',
 'cl.favorites',
 'cl.people_db',
 'cl.lasc',
 'cl.lib',
 'cl.opinion_page',
 'cl.recap',
 'cl.recap_rss',
 'cl.scrapers',
 'cl.search',
 'cl.simple_pages',
 'cl.stats',
 'cl.users',
 'cl.visualizations',
 'tailwind',
 'django_extensions',
 'django_browser_reload',
 'debug_toolbar']
Installed Middleware:
['django.contrib.sessions.middleware.SessionMiddleware',
 'django.contrib.auth.middleware.AuthenticationMiddleware',
 'django.middleware.security.SecurityMiddleware',
 'django_permissions_policy.PermissionsPolicyMiddleware',
 'django.middleware.clickjacking.XFrameOptionsMiddleware',
 'corsheaders.middleware.CorsMiddleware',
 'csp.middleware.CSPMiddleware',
 'django.middleware.common.CommonMiddleware',
 'django.middleware.csrf.CsrfViewMiddleware',
 'django_ratelimit.middleware.RatelimitMiddleware',
 'waffle.middleware.WaffleMiddleware',
 'django.contrib.messages.middleware.MessageMiddleware',
 'cl.lib.middleware.RobotsHeaderMiddleware',
 'cl.lib.middleware.IncrementalNewTemplateMiddleware',
 'pghistory.middleware.HistoryMiddleware',
 'django_browser_reload.middleware.BrowserReloadMiddleware',
 'debug_toolbar.middleware.DebugToolbarMiddleware']



  File "/usr/local/lib/python3.13/site-packages/django/core/handlers/exception.py", line 42, in inner
    response = await get_response(request)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/site-packages/django/core/handlers/base.py", line 271, in _get_response_async
    response = await middleware_method(request, response)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/courtlistener/cl/lib/middleware.py", line 86, in process_template_response
    old_template = response.template_name
                   ^^^^^^^^^^^^^^^^^^^^^^

Exception Type: AttributeError at /prayers/top/
Exception Value: 'TemplateResponse' object has no attribute 'template_name'
```

mlissner:
I opened a fresh issue for this. It's related to a new middleware we're using as part of the redesign, but I don't know why it'd cause you trouble. https://github.com/freelawproject/courtlistener/issues/5034

mlissner:
One thing I think I haven't seen is this item from the list:

> We still need to teach people what the emojis mean and how to find the prayer leader board. One thought is that the first time somebody prays for something, we could pop up a modal congratulates them on their first prayer and then tells them about the leaderboard, how it all works, and points them to the help page.

Did I miss that or are you thinking launch before that's done? I fear that doing so would put zillions of prayer hands around without explaining what they mean or providing a link to the top prayers page or help page.

mlissner:
OOOOh, exciting. We're still suffering the yoke of a client thing, but should be wrapping that up this week. Sorry!

mlissner:
Just using a cookie? Seems fine to me!

cl/favorites/templates/top_prayers.html
```diff
@@ -12,22 +12,29 @@
...
+        <i class="fa fa-heart-o"></i>&nbsp;Remind Me to Help
+      </a>
+    </li>
+    <li>
+      <a href={% if user.is_authenticated %}"{% url 'user_prayers' user.username %}"{% else %}"{% url 'sign-in' %}?next={% url 'top_prayers' %}"{% endif %}><div class="fa gray fa-fw">
+        {% include "includes/hand-holding-heart.svg" %}
+      </div> Your Prayers</a>
+    </li>
+    <li><a href="{% url "pray_and_pay_help" %}">Learn More</a></li>
+  </ul>
```
ERosendo, line 29: This section looks great overall, but I noticed a small alignment issue. To help illustrate this, Here's a GIF that highlights the problem.

Screen Recording 2025-02-18 at 10 10 46 AM

cl/favorites/templates/user_prayers.html
```diff
@@ -121,6 +121,85 @@ <h1 class="text-center v-offset-below-3">{% if is_page_owner %}Your PACER Docume
...
+                    {% endif %}
+                </div>
+                {% endif %}
+              </td>
+            </tr>
+            {% empty %}
+            <tr>
+                <td colspan="2">No document requests granted. Consider making one!</td>
+            </tr>
+            {% endfor %}
```
ERosendo, line 198: Let's hide this table until we have some granted requests to showcase.  It's not very useful without them.

Screenshot 2025-02-18 at 10 31 00 AM

cl/favorites/views.py
```diff
@@ -192,11 +193,25 @@ async def open_prayers(request: HttpRequest) -> HttpResponse:
...
+    def paginate_open_prayers(top_prayers, prayer_page):
+        paginator = Paginator(top_prayers, 25, orphans=10)
+        try:
+            return paginator.page(prayer_page)
+        except PageNotAnInteger:
+            return paginator.page(1)
+        except EmptyPage:
+            return paginator.page(paginator.num_pages)
+
+    paginated_entries = await paginate_open_prayers(top_prayers, page)
```
ERosendo, line 208: We need to update the cache decorator for this view. The current cache decorator doesn't take into account the query string parameters when caching the page. This is a problem because pagination uses the query string to pass the page number to the view. If we don't include the query string in the cache key, we'll end up returning the wrong page every time.

If we still want to use caching for this view, I recommend using the [cache_page decorator](https://docs.djangoproject.com/en/5.1/topics/cache/#the-per-view-cache).

cl/favorites/utils.py
```diff
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 from datetime import timedelta
+from typing import Optional
```
ERosendo, line 3: You don't need to import Optional for type hints in Python 3.10 and later. Python 3.10 introduced the union operator` | `, which you can read about in [PEP 604](https://www.python.org/dev/peps/pep-0604/).  For Python 3.10 and up, the preferred (and more concise) way to specify an optional argument is `Type | None`.  For example, use `str | None` or `list | None`.

cl/favorites/templates/top_prayers.html
ERosendo: We can likely resolve the alignment issue by wrapping the ul element in a div

```html
  <div class="col-xs-12">
    <ul class="nav navbar-nav navbar-right">
      {% include 'includes/remind_me_modal.html' %}
      <li class="hidden-xs">
        <a href="#" class="donate" data-toggle="modal" data-target="#remindMeModal">
          <i class="fa fa-heart-o"></i>&nbsp;Remind Me to Help
        </a>
      </li>
      <li>
        <a href={% if user.is_authenticated %}"{% url 'user_prayers' user.username %}"{% else %}"{% url 'sign-in' %}?next={% url 'top_prayers' %}"{% endif %}><div class="fa gray fa-fw">
          {% include "includes/hand-holding-heart.svg" %}
        </div> Your Prayers</a>
      </li>
      <li><a href="{% url "pray_and_pay_help" %}">Learn More</a></li>
    </ul>
  </div>
```

cl/favorites/templates/includes/remind_me_modal.html
ERosendo, line 1: I'm not convinced a popup is the best way to go here. I think tweaking the signup form wording in Neon  and using a simple button here would be a better approach. What do you think, @mlissner?

Here's a GIF showing the banner in action: 

Screen Recording 2025-02-18 at 4 46 09 PM

mlissner: I think this is OK, actually, but the button looks like it's about 3cm tall?

I like the modal because otherwise you get whisked away to donate.free.law, which is confusing.

ERosendo: > but the button looks like it's about 3cm tall?

@mlissner are you referring to the 'Remind me to help' button?

ERosendo:
Hey @v-anne, thanks for your work on this! 

I've added a couple of commits to refine the logic for showing the congratulatory popup after a user's first prayer.  Your original approach of including the popup HTML with every prayer button was adding a lot of unused elements to the page.  My update uses a helper function to check if the user has created prayers before, and only includes the html for the popup when it's their first time. This keeps the HTML much cleaner.

cl/favorites/templates/includes/remind_me_modal.html
ERosendo: I've made a small adjustment to the padding of the 'Remind me to help' button. 

Screen Recording 2025-02-18 at 7 04 25 PM

Let me know what you think

mlissner: yeah, that looks more normal, thanks.

mlissner:
> My update uses a helper function to check if the user has created prayers before, and only includes the html for the popup when it's their first time. This keeps the HTML much cleaner.

Does this add the modal to the page hundreds of times? If so, why not add it once? That should work fine, I think?

ERosendo:
> Does this add the modal to the page hundreds of times? 

The original implementation was adding the popup hundreds of times to the same page. The updated logic now adds the HTML only once on the user's initial request.

> why not add it once? That should work fine, I think?

Sounds good. I'll test it out.

ERosendo:
The popup display logic has been refactored. The popup is now a separate component that can be included in other templates using the `include` tag.  This component has been added to pages that may have pray buttons.  The existing cookie-based logic remains: the modal opens only if the corresponding cookie is not set, and subsequently sets the cookie to prevent further modal openings. This approach accomplishes two things:

1. The popup is added to the page only once, as a separate component, not tied to the button itself.
2. The database query previously used to determine popup visibility is no longer needed, as the display logic now relies on the cookie

ERosendo (approved):

cl/favorites/templates/includes/first_prayer_modal.html
mlissner, line 1: One other thing we might add here is a note about how many prayers each user can make each day, because otherwise they just start failing silently, right?

cl/assets/static-global/css/override.css
```diff
@@ -1837,3 +1837,7 @@ rect.series-segment {
 .gap-2 {
   gap: 0.5rem
 }
+
+.prayer-top-buttons > li > a {
+  padding: 10px 15px !important;
```
mlissner, line 1842: We usually try hard not to use `!important`, preferring instead to use specificity to win CSS wars. Does this fail without the `important` argument?

mlissner:
Two more little things from me and one about JS we can remove. Otherwise, looks good to me!

cl/assets/static-global/css/override.css
ERosendo: My apologies, I was aware of that 🥲.

I've just found a solution that avoids the use of !important.

cl/favorites/templates/includes/first_prayer_modal.html
ERosendo: This popup includes the daily request limit in its message:

image


>  because otherwise they just start failing silently, right?

Once a user hits the daily prayers limit, the button displays a tooltip explaining that they've reached the limit. This GIF illustrates the tooltip
 
Screen Recording 2025-02-18 at 10 09 01 PM

Let me know what you think.

cl/opinion_page/static/js/pray_and_pay.js
```diff
@@ -79,3 +79,17 @@ document.addEventListener('htmx:oobBeforeSwap', function (event) {
 document.addEventListener('htmx:oobAfterSwap', function (event) {
   $('[data-toggle="tooltip"]').tooltip();
 });
+
```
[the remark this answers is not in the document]
ERosendo: > Is it still necessary to have this JS block since the modal logic has been changed?

Yes, this logic is still needed. I removed the database query for first prayer validation, so we're now relying exclusively on the cookie.

mlissner:
Looks great to me! Shall we merge? Once merged, we can roll it out to a percentage of users, I think, but usually we just go for it and it's ok.

cl/favorites/templates/includes/first_prayer_modal.html
ERosendo: I apologize for the confusion. The screenshot was a mockup used to confirm the wording. I'll create a commit to update the modal.

ERosendo: commit [578a9da](https://github.com/freelawproject/courtlistener/pull/4956/commits/578a9da5dec1105c953548d2840a46112a867606) updates the message.

ERosendo:
I believe this is ready to be squashed and merged. :shipit:

mlissner:
Thanks for all the patience and hard work on this one. I'm looking forward to giving it a try!

PR #4957, opened 2025-01-23
chore(Dockerfile): Include Tailwind asset compilation in Docker build

ERosendo:
This PR fixes the deployment workflow by removing the step that generates Tailwind CSS assets within a temporary pod. Instead, Tailwind CSS asset generation is now integrated into the Docker build process.


mlissner (approved):
Seems OK. Merging.

Issue #4958, opened 2025-01-23
Flaky Selenium Tests in Disclosure module


ERosendo:
While working on PR #4931, Selenium tests for the disclosure module started failing unexpectedly. However, they passed when the action was rerun.

You can find the failing GitHub action run here:

https://github.com/freelawproject/courtlistener/actions/runs/12938299727/job/36088203841?pr=4931

We tried to use the screenshots generated at the end of the test run to help debug the issue. Unfortunately, the `DisclosureReactLoadTest` class overrides the tearDown method, which prevents screenshots from being captured for these specific tests.

elisa-a-v:
I found another one:
[This PR](https://github.com/freelawproject/courtlistener/pull/4940) had all checks successful until `main` was merged into it, then we started getting failed Selenium tests.
- [first failed job](https://github.com/freelawproject/courtlistener/actions/runs/12940358566/job/36094512448?pr=4940), 6 errors including the ones in the failed job mentioned by @ERosendo in the description for this issue.
- [second failed job](https://github.com/freelawproject/courtlistener/actions/runs/12940358566/job/36096537222?pr=4940), 2 errors, same ones.

[Third attempt](https://github.com/freelawproject/courtlistener/actions/runs/12940358566/job/36097017455) was finally successful.

ERosendo:
@mlissner I've been investigating the issue with our tests and noticed they started failing after we added the `cl-tailwind-reload` service to our `docker-compose` file. It seems the tests are failing because the Disclosure search component, a React component, is not loading properly. These react assets should be generated and updated by the `cl-webpack` service.

Here's a breakdown of both services for reference:

```
cl-webpack:
    container_name: cl-webpack
    working_dir: /opt/courtlistener/cl
    image: node:16
    volumes:
      - ${CL_BASE_DIR:-../../}:/opt/courtlistener
    env_file:
      - ../../.env.dev
    command: sh -c 'npm install && exec npx webpack --progress --watch --mode=development'
``` 

```
 cl-tailwind-reload:
    container_name: cl-tailwind-reload
    working_dir: /opt/courtlistener/cl
    image: node:16
    volumes:
      - ${CL_BASE_DIR:-../../}:/opt/courtlistener
    env_file:
      - ../../.env.dev
    command: sh -c 'npm install && exec npm run dev'
    tty: true
``` 

Both services run `npm install`, which might be causing conflicts because they share the same mounted volume. I suspect one installation might be overwriting files created by the other.

[...]

PR #4959, opened 2025-01-23
tests(disclosures): Tweaks the DisclosureReactLoadTest class

ERosendo:
This PR updates the `tearDown` method of the `DisclosureReactLoadTest` class to capture a screenshot at the end of each test run. This will help in visual regression testing and debugging.

 This addresses one of the issues found in #4958

### DIFF
--- a/docker/courtlistener/docker-compose.yml
+++ b/docker/courtlistener/docker-compose.yml
@@ -49,6 +49,12 @@ services:
       - ../../.env.dev
     working_dir: /opt/courtlistener/cl
     command: sh -c 'npm install && exec npx webpack --progress --watch --mode=development'
+    healthcheck:
+      test: sh -c 'ps aux | grep -v grep | grep webpack >/dev/null 2>&1 && echo 0 || echo 1'
+      interval: 5s
+      retries: 5
+      start_period: 10s
+      timeout: 10s
 
   # Task Server
   cl-celery:
@@ -84,7 +90,10 @@ services:
       - ${CL_BASE_DIR:-../../}:/opt/courtlistener
     env_file:
       - ../../.env.dev
-    command: sh -c 'npm install && exec npm run dev'
+    command: sh -c 'npm run dev'
+    depends_on:
+      cl-webpack:
+        condition: service_healthy
     tty: true
 
   cl-django:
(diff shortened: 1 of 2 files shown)


ERosendo:
@mlissner in commit [d58b3c8](https://github.com/freelawproject/courtlistener/pull/4959/commits/d58b3c81b953469a249f1574ded7531f655aa641).  I updated the `docker-compose.yml` file to ensure a specific startup order between the `cl-webpack` and `cl-tailwind-reload` services.

In the latest comment on #4958, I explained the reasoning and motivation behind this change

By default, Docker Compose doesn't guarantee that a container is completely 'ready' before starting others. It only checks if the container is up and running.

The documentation [here](https://docs.docker.com/compose/how-tos/startup-order/#control-startup) recommends using the `depends_on` and `condition` attributes to ensure a service is truly ready before starting another one.

Commit [d58b3c8](https://github.com/freelawproject/courtlistener/pull/4959/commits/d58b3c81b953469a249f1574ded7531f655aa641) introduces a health check for the `cl-webpack` service. This verifies that the `npm install` command has completed before starting the tailwind service. The health check specifically checks if the `npx webpack` command is already running within the service. 

The health check configuration includes:

- `start_period` : the grace period during which Compose will not consider the health check failures (set to 10 seconds). However, if the health check succeeds, then the container is marked healthy, and the grace period ends early.
- `interval` : checks are performed every 5 seconds.
- `timeout` : each health check command has a 10-second timeout.
- `retries` : 5 consecutive failures are allowed before the service is considered unhealthy.

docker/courtlistener/docker-compose.yml
```diff
@@ -49,6 +49,12 @@ services:
       - ../../.env.dev
     working_dir: /opt/courtlistener/cl
     command: sh -c 'npm install && exec npx webpack --progress --watch --mode=development'
+    healthcheck:
+      test: sh -c 'ps aux | grep -v grep | grep webpack >/dev/null 2>&1 && echo 0 || echo 1'
+      interval: 5s
+      retries: 5
+      start_period: 10s
+      timeout: 10s
```
mlissner, line 57: Seconds count with tests and with launching a compose file. 

Why not be more aggressive since the test uses almost no resources. Perhaps:

```suggestion
    healthcheck:
      test: sh -c 'ps aux | grep -v grep | grep webpack >/dev/null 2>&1 && echo 0 || echo 1'
      interval: 1s
      retries: 45
      start_period: 1s
      timeout: 1s
```

Yours gives it 10s to initially start, then retries up to five times every five seconds. That's 10 + 5 * 5 = 35 total seconds. Mine starts after 1s, then retries every 1s up to 45 times, for a total of 46s. I can't imagine why the test would take even 1s, so I reduced the timeout to almost nothing.

Thoughts?

mlissner (approved):
LGTM with one comment to consider. I leave to your judgment how to proceed, and please merge when you're satisfied. 

Thank you!

PR #4961, opened 2025-01-24
Fix bulk data script by using default quoting and escape character

quevon24:
This PR addresses parsing issues when importing opinions bulk data into PostgreSQL. Previously, backticks (`) were used as the quoting character, which was unnecessary and contributed to parsing errors. This has been removed, and the default quoting character (double quotes) is now used instead.

Additionally, an escape character has been set to handle cases where double quotes are embedded within column values. This ensures proper parsing and prevents errors caused by special characters, maintaining data integrity during imports.

With these changes, the bulk data should now import reliably without requiring custom configurations or workarounds.

The following are examples of the executed SQL commands to create the data dump and to load the data:

Command to create the dump:

```
psql --command "COPY search_opinion (
               id, date_created, date_modified, author_str, per_curiam, joined_by_str,
               type, sha1, page_count, download_url, local_path, plain_text, html,
               html_lawbox, html_columbia, html_anon_2020, xml_harvard,
               html_with_citations, extracted_by_ocr, author_id, cluster_id
           ) TO STDOUT WITH (FORMAT csv, ENCODING utf8, HEADER, ESCAPE '\\', FORCE_QUOTE *)" --host "db_host" --username "db_username" --dbname "db_name"
```

Command to load the data:

```
psql --command "\COPY public.search_opinion (
               id, date_created, date_modified, author_str, per_curiam, joined_by_str,
               type, sha1, page_count, download_url, local_path, plain_text, html,
               html_lawbox, html_columbia, html_anon_2020, xml_harvard,
               html_with_citations, extracted_by_ocr, author_id, cluster_id
           ) FROM '/path/opinions-2025-01-23.csv' WITH (FORMAT csv, ENCODING utf8, ESCAPE '\\', HEADER)" --host "db_host" --username "db_username" --dbname "db_name"
```


mlissner (commented):
Looks good. Do we need updates to the documentation/release/change notes, and do we need to tweak the import script as well?

mlissner:
Let's be sure to get this done before the next data generates.

quevon24:
@mlissner I updated the bulk data help page and the part of the script responsible for generating the import shell script.

mlissner (approved):
Great! Let's hope we've got it finally. :)

PR #4968, opened 2025-01-25
fix(authorities): clarify number of references in docket authorities

### LINKED ISSUE #4865: Clarify number of references on Authorities Page vs. Search Results
Noticed after doing #4134

On the “Authorities” page for a given docket, we display the references to opinions in the docket's documents. However, each "authority" in the list is not actually an opinion, but an element of a `Docket`'s `authorities`:

```python
    @property
    def authorities(self):
        """Returns a queryset that can be used for querying and caching
        authorities.
        """
        return OpinionsCitedByRECAPDocument.objects.filter(
            citing_document__docket_entry__docket_id=self.pk
        )
```

```python
class OpinionsCitedByRECAPDocument(models.Model):
    citing_document = models.ForeignKey(
        RECAPDocument, related_name="cited_opinions", on_delete=models.CASCADE
    )
    cited_opinion = models.ForeignKey(
        Opinion, related_name="citing_documents", on_delete=models.CASCADE
    )
    depth = models.IntegerField(
        help_text="The number of times the cited opinion was cited "
        "in the citing document",
        default=1,
    )
```

This means an opinion cited by multiple documents in the same docket will be listed multiple times, like the authorities in [this docket](https://www.courtlistener.com/docket/4214664/authorities/national-veterans-legal-services-program-v-united-states/). Notice `Norman v. United States, 429 F.3d 1081 (Fed. Cir. 2005)` being listed 14 times, where [all links point to the same search](https://www.courtlistener.com/?type=r&q=docket_id%3A4214664%20AND%20cites%3A(211579)) which yields 14 docket entries.
Image
Image

This also means the counts are confusing. Each "depth" represents only the number of times the opinion is cited in one citing document, not the total number of references across all documents in the docket. We could:
1. Make the counts less confusing by making the authorities list include the citing document and changing the text to something like `5 references to <CITED_OPINION> in <CITING_DOC>`, with a link to the citing doc.
2. Group authorities by opinion, then aggregate the depth. This way we would get the summary of all references to a given opinion in a given docket, instead of having the same opinion repeated several times. This does sound like it could be a lot more work, but potentially more informative if we could also include a sub-list of all citing docs with their links.

    So instead of:

    > - 5 references to Norman v. United States, 429 F.3d 1081 (Fed. Cir. 2005)
    > Court of Appeals for the Federal Circuit Nov. 18, 2005
    > - 5 references to Norman v. United States, 429 F.3d 1081 (Fed. Cir. 2005)
    > Court of Appeals for the Federal Circuit Nov. 18, 2005
    > - 2 references to Norman v. United States, 429 F.3d 1081 (Fed. Cir. 2005)
    > Court of Appeals for the Federal Circuit Nov. 18, 2005

    It would be:

    > - 12 references to Norman v. United States, 429 F.3d 1081 (Fed. Cir. 2005)
    > Court of Appeals for the Federal Circuit Nov. 18, 2005
    >   - 5 references in citing doc 1
    >   - 5 references in citing doc 2
    >   - 2 references in citing doc 3

3. Update the search results to display the depth of treatment when a cites query is made, so that each result shown says something like, "22 references to case XYZ".

Whatever the option, we should make sure that the opinions' authorities page doesn't get broken since `authorities_list.html` is used in both docket and opinion authorities, which are not `OpinionsCitedByRECAPDocument` but `OpinionsCited` so we don't always have the same attributes available.

elisa-a-v:
Fixes #4865 

This PR not only makes reference counts less ambiguous, it also fixes the actual number of opinions cited.

Note that before this PR, the count shown in the header ("The filings in this docket cite X opinions") was not actually the number of **opinions**, but rather the number of `OpinionsCitedByRECAPDocument` instances related to the docket. This is wrong because multiple docs could cite the same opinion, as is the case in the examples provided below (both have the exact same data), so each opinion can be counted multiple times. This is now fixed as we're counting actual Opinions.

<details>
<summary>Before this PR</summary>

</details>

<details>
<summary>After this PR</summary>

</details>


mlissner:
Looks good!

mlissner (changes requested):
That's a doozy of a query! I gather you and Eduardo worked on it? If so, I think I'm happy merging this as is, if you think another set of eyes would help, then let's have him review.

That aside, I do have one other tweak, please.

Looks good though! Nice enhancement.

elisa-a-v:
> That's a doozy of a query! I gather you and Eduardo worked on it?

Yes we did, he helped me get rid of some pesky duplicated queries. With the data I have for the docket in my screenshots, that page performs 30 DB queries before this PR, and only 14 after.

mlissner (approved):

mlissner:
INTO THE MAW. Thank you, this is a nice improvement that moves this feature from "neat" to "useful".

PR #5016, opened 2025-01-31
Link citations cannot be disambiguated

quevon24:
This PR addresses the issue of linking citations that cannot be disambiguated to /c/reporter/volume/page/ (e.g. "114 F.3d 1182" will be linked to/c/f3d/114/1182/) so the user can figure it out which open.

It uses a similar approach that is being used for NO_MATCH_RESOURCE but setting an impossible citation for the FullCaseCitation object (999999 U.S. 999999) to have a different hash than NO_MATCH_RESOURCE because the object is being used as key for citation resolution dictionary.


cl/citations/annotate_citations.py
```diff
@@ -62,6 +66,20 @@ def generate_annotations(
                 '<span class="citation no-link">',
                 "</span>",
             ]
+        elif opinion is MULTIPLE_MATCHES_RESOURCE:
+            # Multiple matches, can't disambiguate, link citation to citation
+            # lookup page
+            citation_groups = citations[0].groups
```
grossir, line 72: Probably best to use this, seems to handle a special case when page has a value but volume is None.

https://github.com/freelawproject/courtlistener/blob/55686ba567920b671e7ea3945615e8cc73946704/cl/opinion_page/views.py#L1532-L1541

grossir, line 72: There may be a bug here. When resolving citations to `MULTIPLE_MATCHES_RESOURCE`, all ambiguous citations will be grouped to that resource. But, they may have different ambiguous cites, so doing `citations[0]` may be incorrect. I suggest you add this situation into the test cases, too; to actually see that it can handle it

```python
{
MULTIPLE_MATCHES_RESOURCE: [
        FullCaseCitation(volume=100, reporter=US, page=100),
        FullCaseCitation(volume=300, reporter=US, page=100),
        FullCaseCitation(volume=22, reporter=WI, page=3),
]
}
```

grossir (changes requested):
This is looking good, but there may be a bug that needs addressing

cl/citations/annotate_citations.py
quevon24: Thank you, you were right, I already fixed it

quevon24: I implemented it, i just had to import it inside the function to avoid circular imports

quevon24: Also added an additional test case to verify that multiple citations with multiple results can be annotated

quevon24:
@grossir I addressed all your comments, but  i should probably get this PR done first: https://github.com/freelawproject/courtlistener/pull/5049

cl/citations/annotate_citations.py
```diff
@@ -59,39 +59,49 @@ def generate_annotations(
...
+            # Multiple matches, can't disambiguate
+            for c in citations:
+                # Annotate all citations can't be disambiguated to citation
+                # lookup page
+                citation_dict = {
+                    "reporter": c.groups.get("reporter"),
+                    "volume": c.groups.get("volume", None),
+                    "page": c.groups.get("page", None),
+                }
+                kwargs = make_citation_url_dict(**citation_dict)
```
grossir, line 83: You could simply do
`kwargs = make_citation_url_dict(**c.groups)`

since `c.groups` already looks like a dict

```python
In [1]: from eyecite import get_citations

In [2]: get_citations("Lorem ipsum, 114 F.3d 1182,")
Out[2]: [FullCaseCitation('114 F.3d 1182', groups={'volume': '114', 'reporter': 'F.3d', 'page': '1182'}, metadata=FullCaseCitation.Metadata(parenthetical=None, pin_cite=None, year=None, court=None, plaintiff=None, defendant=None, extra=None, resolved_case_name_short=None, resolved_case_name=None))]

In [3]: get_citations("Lorem ipsum, 114 F.3d 1182,")[0].groups
Out[3]: {'volume': '114', 'reporter': 'F.3d', 'page': '1182'}
```

grossir (approved):
This looks great Kevin, I think it's ready to merge. Just one style comment below

cl/citations/annotate_citations.py
quevon24: perfect, thanks, i just pushed that change

PR #5017, opened 2025-01-31
Update wl case name importer

quevon24:
@grossir almost finished running the case name importer command, He found a problem with some files because they include special characters, so they are not encoded with UTF-8 but with ISO-8859-1, it was corrected to be able to execute the command with those files that failed

He also suggested that I make the argument to pass the csv path accept multiple files at the same time, the argument to accept multiple files has already been updated.

I fixed some alerts that flake8 gave me related to the length of the lines.

Regarding the fact that the bankruptcy.csv file did not work, it is because the "B.R." reporter returns two possible reporters: Baltimore City Reports and Bankruptcy Reporter. This must be corrected in reporters-db to solve the problem for that last file and be able to finish executing the command.

### DIFF
--- a/cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
+++ b/cl/corpus_importer/management/commands/update_casenames_wl_dataset.py
@@ -52,7 +52,8 @@ DATE_FORMATS = (
 
 
 def tokenize_case_name(case_name: str) -> set[str]:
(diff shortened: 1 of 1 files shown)


flooie:
Did you update reports db?   And I assume this means using dates to filter it out?

PR #5033, opened 2025-02-03
fix(people_db): Add missing religion and position values

ERosendo:
This PR updates the `Person` model to include four new religion values and the `Position` model to include the "Special Trial Judge" type.


ERosendo:
This script normalizes the religion field in the person table by mapping religious denomination names to their corresponding internal values defined in the Person.RELIGIONS choices:


```python
from django.db.models import Q
from cl.people_db.models import Person

religion_mapping = {value: key for key, value in Person.RELIGIONS}
p_records = Person.objects.filter(
    Q(religion="Episcopalian")
    | Q(religion="Methodist")
    | Q(religion="Presbyterian")
    | Q(religion="Roman Catholic")
)

for p in p_records:
    p.religion = religion_mapping.get(p.religion)
    p.save()
```

mlissner (approved):
Looks good, merging, thank you!

PR #5036, opened 2025-02-04
feat(match_citations): extract_references_using_resolutions

grossir:
wip


flooie:
I think we need to resolve conflicts here @grossir

PR #5047, opened 2025-02-06
feat(citations): store_opinion_citations_and_update_parentheticals no…

grossir:
…w uses find_reference_citations_from_markup

Adds logic to use https://github.com/freelawproject/eyecite/pull/203

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -81,7 +81,7 @@ django-hCaptcha = "^0.2.0"
 reporters-db = "*"
 django-waffle = "^4.1.0"
 nameparser = "^1.1.3"
-eyecite = "*"
+eyecite = {git = "https://github.com/freelawproject/eyecite.git", branch = "198-find-name-only-reference-citations"}
 scipy = "^1.14.1"
 scikit-learn = "^1.5.2"
 numpy = "^2.1.2"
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -81,7 +81,7 @@ django-hCaptcha = "^0.2.0"
 reporters-db = "*"
 django-waffle = "^4.1.0"
 nameparser = "^1.1.3"
-eyecite = "*"
+eyecite = {git = "https://github.com/freelawproject/eyecite.git", branch = "198-find-name-only-reference-citations"}
```
flooie, line 84: I assume this just needs to be reverted and we need to release a new version of eyecite?

grossir: We need to release a new version of eyecite for this to work. But I also need to update the code to account for the changed interface

flooie (approved):
looks good to me

grossir:
Glad to announce our first couple stand alone references on a live [opinion](https://www.courtlistener.com/opinion/10335771/xx/)

image

grossir:
Some more from NY
image

image

mlissner:
These screenshots look a lot like magic, y'all.

PR #5049, opened 2025-02-06
Find and linkify pincites

quevon24:
This PR helps to annotate pin cites from FullCaseCitation objects:

For example:

-  `8 Wheat. 574` will be linked to `8 Wheat. 543 `(https://www.courtlistener.com/opinion/85404/johnson-grahams-lessee-v-mcintosh/) because it cointains page 574
- `334 U. S. 131, 144-145, n. 6 `We will have two links, one for `334 U. S. 131` linked to "https://www.courtlistener.com/opinion/104551/united-states-v-paramount-pictures-inc/" and `144-145, n. 6` linked to "https://www.courtlistener.com/opinion/104551/united-states-v-paramount-pictures-inc/#144"

It includes tests for the cases mentioned above


flooie:
is this ready?

quevon24:
> is this ready?

No, I'm just finishing adding some tests

cl/citations/tests.py
```diff
@@ -998,6 +1084,82 @@ def test_no_duplicate_parentheticals_from_parallel_cites(self) -> None:
...
+        pincite_1 = f"""<span class="citation pin-cite" data-id="{matched_opinion.pk}"><a href="/opinion/{self.citation7.cluster.pk}/united-states-v-paramount-pictures-inc/#144" aria-description="Citation for case: United States v. Paramount Pictures, Inc.">144-145, n. 6</a></span>"""
+        pincite_2 = f""" <span class="citation pin-cite" data-id="{matched_opinion.pk}"><a href="/opinion/{self.citation7.cluster.pk}/united-states-v-paramount-pictures-inc/#144" aria-description="Citation for case: United States v. Paramount Pictures, Inc.">144, n. 6</a></span> (1948).</p>"""
+
+        # Verify that a citation without pin cite matches
+        self.assertIn(main_citation, new_html)
+        # Verify that a citation with pin cite with single page number matches
+        self.assertIn(pincite_1, new_html)
+        # Verify that a citation with pin cite with page range matches
+        self.assertIn(pincite_2, new_html)
+
```
flooie, line 1162: can this be refactored and incorporated into 

class CitationTextTest(SimpleTestCase):
    def test_make_html_from_plain_text(self) -> None:
    

I think that would make this much easier for me to understand.

quevon24: I updated the tests

cl/citations/tests.py
```diff
@@ -336,6 +340,34 @@ def test_make_html_from_matched_citation_objects(self) -> None:
...
+            ("""A clearance is the period of time, usually
+                                stipulated in license contracts, which must elapse
+                                between runs of the same feature within a particular area
+                                or in specified theatres.” United States v. Paramount
+                                Pictures, Inc., 334 U. S. 131, 144, n. 6 (1948).""",
+             f"""<pre class="inline">A clearance is the period of time, usually
+                                stipulated in license contracts, which must elapse
+                                between runs of the same feature within a particular area
+                                or in specified theatres.” United States v. Paramount
+                                Pictures, Inc., </pre><span class="citation" data-id="MATCH_ID"><a href="MATCH_URL" {aria_description}>334 U. S. 131</a></span><pre class="inline">, </pre><span class="citation pin-cite" data-id="MATCH_ID"><a href="MATCH_URL#144" {aria_description}>144, n. 6</a></span><pre class="inline"> (1948).</pre>"""),
```
flooie, line 354: can you refactor this again please

something more like 

```
("""United States v. Paramount Pictures, Inc., 334 U. S. 131, 144, n. 6 (1948).""",
 f'<pre class="inline">United States v. Paramount Pictures, Inc., </pre>'
 f'<span class="citation" data-id="MATCH_ID"><a href="MATCH_URL" {aria_description}>334 U. S. 131</a></span>'
 '<pre class="inline">, </pre>'
 f'<span class="citation pin-cite" data-id="MATCH_ID"><a href="MATCH_URL#144" {aria_description}>144, n. 6</a></span>'
 '<pre class="inline"> (1948).</pre>'),
```

I dont think we need all the extra text at the start for the test and removing it makes it easier to see what is happening.  Also  if you could model the other tests where it's not wrapped in triple quotes that also makes it easier for me to see what is going on.

cl/citations/match_citations.py
```diff
@@ -55,12 +58,25 @@ def resolve_fullcase_citation(
...
     if type(full_citation) is FullCaseCitation:
         db_search_results: list[Hit]
+        # Look for matches using full citation
         db_search_results, _ = es_search_db_for_full_citation(full_citation)
+        if len(db_search_results) == 0:
+            # We didn't get an exact match on the volume/reporter/page.
+            # Perhaps it's a pin cite. Find closest citations filtering by
+            # volume and reporter and excluding self cites. e.g. 8 Wheat. 574
+            # points to 8 Wheat. 543
+            db_search_results, _ = es_search_db_for_partial_citation(
```
grossir, line 68: This is going to generate a secondary ES query for each citation that does not have a match. Isn't that going to be costy?

The input for this function is supossed to be a FullCaseCitation; but for this if clause we are assuming that the incoming parameter is actually a ShortCaseCitation that looks like a FullCaseCitation, in which case the "page" part of the citation is actually a pin cite. 

But don't we expect this to happen only on exceptional cases?

grossir (commented):
@quevon24  this was an interesting review. I think it would be easier if the PR was split in 2 parts

1. pincite annotation  (I think this part is easier to understand)
2. using pincites to resolve citations (maybe harder to understand since it involves ES)

The only interaction between both parts is the usage of `page_pin_cite` for annotation, which I expect to be a somewhat rare use case (for the usage currently given to the partial ES search)


### about pincite annotation

2 general comments:

1. I think part of this logic should be in eyecite. 

`calculate_pin_cite_absolute_span` and part of `get_pin_cite_annotation`. Maybe we need a Pincite object to live in  `Citation.Metadata.pincite = Pincite()` that encapsulates logic to do stuff with spans, etc. However, this may take more time, and we seem to be in a hurry, so it could be left for future refactoring

2. I think there should be a single link for citation+pincite, instead of a link for a citation and another for its pincite

### about opinion resolution using pincites
2  general comments:

1. As the code is now, I think trying a secondary lookup each time we don't find a full case citation may generate a lot of ES queries; and it may produce false positive matches (I am not so sure about this second part; I would like to see a counter example in the test cases)

2. I think a big use case is being ignored: when we actually resolve a full case citation on the first pass; and it has a pincite; we may be able to resolve to the precise opinion in the cluster; instead of to the first opinion in the cluster


-------
These is my review, but I leave @flooie and you to decide what to take and what to leave

cl/citations/tests.py
```diff
@@ -320,22 +327,37 @@ def test_make_html_from_matched_citation_objects(self) -> None:
...
             # Id. citation without page number ("Id. Something else")
             ('asdf, Id. Lorem ipsum dolor sit amet',
-             '<pre class="inline">asdf, </pre><span class="citation" data-id="'
+             '<pre class="inline">asdf, </pre><span class="citation id-citation" data-id="'
              f'MATCH_ID"><a href="MATCH_URL" {aria_description}>Id.</a></span><pre class="inline">'
              ' Lorem ipsum dolor sit amet</pre>'),
+
+            # Annotate citation but also pin cite
+            ('United States v. Paramount Pictures, Inc., 334 U. S. 131, 144, n. 6 (1948).',
+             f'<pre class="inline">United States v. Paramount Pictures, Inc., </pre><span class="citation" data-id="MATCH_ID"><a href="MATCH_URL" {aria_description}>334 U. S. 131</a></span><pre class="inline">, </pre><span class="citation pin-cite" data-id="MATCH_ID"><a href="MATCH_URL#144" {aria_description}>144, n. 6</a></span><pre class="inline"> (1948).</pre>'),
```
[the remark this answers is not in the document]
quevon24: Yes you are right, this should go in one link

cl/citations/match_citations.py
quevon24: I think this is not very common, but probably necessary if we want to link as much of the document as possible, currently the Citation Lookup Tool solves this strange type of pin cites, the bad thing is that the only way to verify it is by making a query(db or ES), the other option would be to send them directly to the citation lookup tool and have it take care of resolving the citation, for example send it to https://www.courtlistener.com/c/wheat/8/574/ although instead of annotating the citations without link with <span class="citation no-link"> now those would be linked to the citation lookup tool

cl/citations/match_citations_queries.py
```diff
@@ -130,6 +132,103 @@ def es_case_name_query(
...
+    # Create a temporal item and add it to the values list (cluster_id, page)
+    closest_opinion_clusters = [(0, full_citation.groups["page"])]
+
+    for result in results:
+        # Get the citations from OpinionDocument that matched the partial
+        # citation
+        valid_citations = [
+            get_citations(citation)[0]
+            for citation in result["citation"]
+            if partial_citation_str in citation and get_citations(citation)
```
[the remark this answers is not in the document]
quevon24: what do you think about using walrus operator to avoid calling get_citations twice?

```
valid_citations = [
    citation_result[0]
    for citation in result["citation"]
    if partial_citation_str in citation and (citation_result := get_citations(citation))
] 
```

cl/citations/annotate_citations.py
```diff
@@ -62,18 +194,60 @@ def generate_annotations(
...
+                    annotations.append([c.span()] + annotation)
+                else:
+                    # It is a FullCaseCitation, we have three possible cases
+                    # 1) A regular citation without pin cite
+                    # 2) A citation with pin cite in metadata
+                    # 3) A pin cite specified in page
+                    if hasattr(opinion, "page_pin_cite"):
+                        # Case 3: Get the pin cite if we used one to match it
+                        # with an opinion. This is for a special pin cite
+                        # like "8 Wheat. 574" which points to "8 Wheat. 543"
```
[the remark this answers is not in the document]
quevon24: I updated the comment

cl/citations/tests.py
```diff
@@ -320,22 +327,37 @@ def test_make_html_from_matched_citation_objects(self) -> None:
...
+
+            # Annotate citation but also pin cite with page range,
+            # we annotate using first page in pin cite
+            ('United States v. Paramount Pictures, Inc., 334 U. S. 131, 144-145, n. 6 (1948).',
+             f'<pre class="inline">United States v. Paramount Pictures, Inc., </pre><span class="citation" data-id="MATCH_ID"><a href="MATCH_URL" {aria_description}>334 U. S. 131</a></span><pre class="inline">, </pre><span class="citation pin-cite" data-id="MATCH_ID"><a href="MATCH_URL#144" {aria_description}>144-145, n. 6</a></span><pre class="inline"> (1948).</pre>'),
+
+            # Dummy text to annotate these citations types: FullCaseCitation,
+            # ShortCaseCitation, SupraCitation and IdCitation
+            ('In Strong v. Waterman, the court held that a trustee must act in good faith and in the best interest of the beneficiaries. 11 Paige 607. It further clarified that a trustee’s discretion remains subject to judicial review. 11 Paige at 611. This principle had been previously established. Supra, at 609. The court reaffirmed this rule. Id. at 609.',
```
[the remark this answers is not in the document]
quevon24: Do you mean that if you had after `Id at 609.` for example, something like `11 Paige 609` ?

something like this: `In Strong v. Waterman, the court held that a trustee must act in good faith and in the best interest of the beneficiaries. 11 Paige 607. It further clarified that a trustee’s discretion remains subject to judicial review. 11 Paige at 611. This principle had been previously established. Supra, at 609. The court reaffirmed this rule. Id. at 609. 11 Paige at 609.`  If this is the case i think it should work.

or do you mean that we have a text that only contains a short citation but no full case citation before to help us to guess to which case points the short citation?

cl/citations/annotate_citations.py
```diff
@@ -62,18 +194,60 @@ def generate_annotations(
...
+                    ).lower()
+                    if c.metadata.pin_cite:
+                        annotation = generate_annotation(
+                            opinion, class_name, c.metadata.pin_cite
+                        )
+                    else:
+                        # We can have these citations types without a pin cite
+                        annotation = generate_annotation(opinion, class_name)
+
+                    annotations.append([c.span()] + annotation)
```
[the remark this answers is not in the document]
quevon24: i rewrote that part

cl/citations/annotate_citations.py
```diff
@@ -62,18 +194,60 @@ def generate_annotations(
...
+                        annotations.append([c.span()] + annotation)
+
+                    if c.metadata.pin_cite:
+                        # Case 2: We could have a pin cite after the
+                        # FullCaseCitation, annotate the pin cite text
+                        # e.g. "144, n. 6" in "334 U. S. 131, 144, n. 6"
+                        pin_cite_annotation = get_pin_cite_annotation(
+                            c, plain_text, opinion
+                        )
+                        if pin_cite_annotation:
```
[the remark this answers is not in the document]
quevon24: i removed this part and refactored the code

quevon24:
@grossir Thanks for your review

I tried to address all of your comments as best as possible, i was annotating some pin cites incorrectly so i removed the get_pin_cite_annotation and calculate_pin_cite_absolute_span methods.

I tried to improve my comments to provide a better explanation, i left a few comments, i hope you can help me with them

cl/citations/match_citations.py
flooie: I agree with @grossir that this is too uncommon to be worth the second ES query.

quevon24: I have already removed it from the code

grossir:
@quevon24 can you update this branch to account for the conflicts?

quevon24:
> @quevon24 can you update this branch to account for the conflicts?

yes, i'll fix it

cl/citations/annotate_citations.py
```diff
@@ -85,18 +133,40 @@ def generate_annotations(
...
-                f'<span class="citation" data-id="{opinion.pk}">'
-                f'<a href="{opinion_url}"'
-                f' aria-description="Citation for case: {safe_case_name}"'
-                ">",
-                "</a></span>",
-            ]
-            # Annotate all matched citations for the resource
-            annotations.extend([[c.span()] + annotation for c in citations])
+            for c in citations:
+                # TODO add ReferenceCitation
```
grossir, line 137: I think you can add `ReferenceCitation` with the same logic as for `(SupraCitation, IdCitation, ShortCaseCitation)`
You should also add a test case for that citation model

cl/citations/annotate_citations.py
```diff
@@ -85,18 +133,40 @@ def generate_annotations(
...
+                # Handle FullCaseCitation cases
+                elif c.metadata.pin_cite:
+                    # Case 1: FullCaseCitation with pin cite
+                    # e.g. "334 U. S. 131, 144, n. 6"
+                    annotation = generate_annotation(
+                        opinion,
+                        extra_class="pin-cite",
+                        pin_cite=c.metadata.pin_cite,
+                    )
+                    annotations.append([c.full_span()] + annotation)
```
grossir, line 162: this line in conjunction with line 169 is adding the citation annotation twice, but with different spans. I don't know if that's desireable

Perhaps you could just do something like
```python
for c in citations:
       span = c.span()
       if isinstance(...):
       ...
       ....
        # Handle FullCaseCitation cases
        elif c.metadata.pin_cite:
                ...
                span = c.full_span()
        ....
        annotations.append([span] + annotation)
                
```
The conflict is not appearing in the actual annotation in the tests; it must be doing some filtering; but it's better to do it here, I think

cl/citations/annotate_citations.py
```diff
@@ -49,21 +50,68 @@ def get_and_clean_opinion_text(document: Opinion | RECAPDocument) -> None:
...
+        "</a></span>",
+    ]
+    return annotation
+
+
 def generate_annotations(
     citation_resolutions: Dict[
         MatchedResourceType, List[SupportedCitationType]
     ],
+    plain_text: str,
```
grossir, line 101: plain_text is no longer used, should be removed; also the :param comment in the docstring

cl/citations/tests.py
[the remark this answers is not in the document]
grossir: I meant tests for the elastic search querying, but that's outdated now.

cl/citations/match_citations_queries.py
grossir, line 1: maybe just rever the changes for this file too, since they are unused?

grossir (commented):
Without the elastic search parts this PR is more clear, I think it's mostly ready
The most important missing part is adding ReferenceCitations

Also, I am pointing some minor style improvements
Also, have you tested this as a webpage, and saw if the link fragment actually leads to the proper page in opinions where it can happen?

cl/citations/match_citations_queries.py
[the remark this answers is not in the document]
grossir: I think that would have been a valid solution too

grossir (approved):
@quevon24 I think this is ready for merging

PR #5059, opened 2025-02-11
Update dependencies february

quevon24:
This PR update about eight dependencies as reported by dependabot from February 1st to February 10th.

It also upgrades sentry release action in sentry-release.yml workflow


flooie:
Great. thanks Kevin

PR #5068, opened 2025-02-12
feat(templates): add time uploaded to document page

### LINKED ISSUE #5039: Add time to date upload field on front end
Now that I'm using the RECAP alerts, I'm constantly looking at documents and wondering how far behind I am from when the doc was available. Currently we show: 

Image

<sup>From [this page](https://www.courtlistener.com/docket/69613359/1/afl-cio-v-department-of-labor/)</sup>

Our `date_upload` field on the `AbstractPDF` class is actually a DateTimeField. We should show the time in the HTML! 

Note that the `date_filed` field just above it does some stuff with the timezone. We should make sure the two dates have the same time zone.

elisa-a-v:
Closes #5039 

image

### DIFF
--- a/cl/opinion_page/templates/includes/rd_metadata_headers.html
+++ b/cl/opinion_page/templates/includes/rd_metadata_headers.html
@@ -33,7 +33,7 @@
 <p class="bottom">
   <span class="meta-data-header">Uploaded:</span>
   <span class="meta-data-value">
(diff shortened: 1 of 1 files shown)


mlissner (approved):

mlissner:
Nice, thank you!

PR #5082, opened 2025-02-17
Hotfix to 4228: linkify-admin-citations

v-anne:
This is a minor change to #4228, linking to administrative agency dockets. It accounts for the Federal Register exceeding 100,000 pages last year by changing the regex and adding two new tests to confirm the function still works.


mlissner:
Not that I'm aware of, but we can take this over the finish line for you. Thanks for the fix (and your patience)!

mlissner:
We assigned this one to Alberto, so he'll take a look soon. Sorry about the flaky tests!

PR #5094, opened 2025-02-18
fix(citations): remove MULTIPLE_MATCHES_RESOURCE in recap_citations.py

quevon24:
I didn't see that the do_resolve_citations function was also used in this function for recap, i had to remove citations with multiple matches


mlissner:
Automerge set!

grossir (approved):
Let's merge this, it fixes the problem and has no side effects

mlissner:
Can you guys mark the Sentry issue resolved, please, so it sends alerts anew if it ever breaks again?

PR #5131, opened 2025-02-21
fix(citations): get_years_from_reporter

### LINKED ISSUE #5130: Bug in get_years_from_reporter causes citation matching to fail
```
def get_years_from_reporter(
    citation: FullCaseCitation,
) -> tuple[int, int]:
    """Given a citation object, try to look it its dates in the reporter DB"""
    start_year = 1750
    end_year = date.today().year

    edition_guess = citation.edition_guess
    if edition_guess:
        if hasattr(edition_guess.start, "year"):
            start_year = edition_guess.start.year
        if hasattr(edition_guess.end, "year"):
            start_year = edition_guess.end.year
    return start_year, end_year
```

start_year is assigned twice - once for start and end this causes all reporter years to be returned as the end year to present day.  

This is causing a cascading failure for matching citations

flooie:
Small typo assigned end year to start year
causes errors matching citations because
reporter years wont match reducing
elastic searches to zero

Fixes #5130

### DIFF
--- a/cl/citations/match_citations_queries.py
+++ b/cl/citations/match_citations_queries.py
@@ -172,7 +172,7 @@ def es_search_db_for_full_citation(
             full_citation.citing_opinion is not None
             and full_citation.citing_opinion.cluster.date_filed
         ):
-            end_year = min(
+            end_year = max(
                 end_year,
                 full_citation.citing_opinion.cluster.date_filed.year,
             )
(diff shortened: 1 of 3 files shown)


mlissner:
Fantastic. We should give an award for whomever finds/fixes the oldest bug each sprint.

flooie:
ok, the fix in get year causes tests to fail because when the opinion cited was generated in a factory it did not set the cluster date filed to match the reporters-db dates for yeates.  As such a date that works was added which fixes tests.  🤞

grossir (approved):

PR #5132, opened 2025-02-21
fix(authorities): restore authorities tab in parties page

### LINKED ISSUE #4992: Restore authorities tab in docket authorities page
After merging #4968 the authorities tab disappeared:

image

elisa-a-v:
Fixes #4992 

To display the tab we were relying on the context explicitly declaring `authorities` for every view that included the `docket_tabs.html`—instead we now get it directly from the Docket instance, so it should always be available for docket tabs.


mlissner (approved):

mlissner:
Nice one. Auto-merge set.

PR #5148, opened 2025-02-26
fix(match_citations): Update find cites

flooie:
Remove self links in backup method
Check if volume is digit to avoid bug
Use corrected reporter to lookup citations

### DIFF
--- a/cl/citations/match_citations.py
+++ b/cl/citations/match_citations.py
@@ -69,7 +69,7 @@ def resolve_fullcase_citation(
         if len(db_search_results) == 0:
             # If no citation is found use get_clusters_from_citation as a backup
             volume = full_citation.groups.get("volume", None)
-            reporter = full_citation.groups.get("reporter", None)
+            reporter = full_citation.corrected_reporter()
             page = full_citation.groups.get("page", None)
             if (
                 volume is not None
--- a/cl/citations/match_citations.py
+++ b/cl/citations/match_citations.py
@@ -73,12 +73,22 @@ def resolve_fullcase_citation(
             page = full_citation.groups.get("page", None)
             if (
                 volume is not None
+                and volume.isdigit() == True
                 and reporter is not None
                 and page is not None
             ):
                 clusters, _count = async_to_sync(
                     get_clusters_from_citation_str
                 )(volume=volume, reporter=reporter, page=page)
+
+                # exclude self links
+                clusters = [
+                    cluster
+                    for cluster in clusters
+                    if cluster.id != full_citation.citing_opinion.cluster.pk
+                ]
+                _count = len(clusters)
+
                 if _count == 1:
                     # return the first item by ordering key
                     return clusters[0].ordered_opinions.first()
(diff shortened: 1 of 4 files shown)


cl/citations/match_citations.py
```diff
@@ -73,12 +73,22 @@ def resolve_fullcase_citation(
             page = full_citation.groups.get("page", None)
             if (
                 volume is not None
+                and volume.isdigit() == True
```
grossir, line 76: no need for `== True`, since `volume.isdigit()` already is either True or False

cl/citations/match_citations.py
```diff
@@ -73,12 +73,22 @@ def resolve_fullcase_citation(
...
             ):
                 clusters, _count = async_to_sync(
                     get_clusters_from_citation_str
                 )(volume=volume, reporter=reporter, page=page)
+
+                # exclude self links
+                clusters = [
+                    cluster
+                    for cluster in clusters
+                    if cluster.id != full_citation.citing_opinion.cluster.pk
```
grossir, line 88: this function is used by `do_resolve_citations`, which is used in both `recap_citations` and `match_citations`. When it's called from  `recap_citations` this line `full_citation.citing_opinion.cluster.pk` will raise an AttributeError because `full_citation.citing_opinion` will be `None`

The good part is that you don't have to check for self citations in recap documents; since they are only looking for citations to plain old opinions, not to other recap docs

cl/citations/tasks.py
```diff
@@ -334,9 +334,10 @@ def store_unmatched_citations(
         # values in required fields
         groups = unmatched_citation.groups
         if (
-            not groups.get("reporter")
-            or not groups.get("volume")
-            or not groups.get("page")
+            not groups.get("reporter", None)
```
grossir, line 337: the default value for when "key" is not found in `dict.get("key")` is `None`, so no need to put it as explicit default

grossir (commented):
Looks generally good, just found a possible error when the changed function is called by `recap_citations`

Apart from that, on the PR summary it says
> Use corrected reporter to lookup citations
I don't see that change

cl/citations/match_citations.py
```diff
@@ -69,16 +69,26 @@ def resolve_fullcase_citation(
         if len(db_search_results) == 0:
             # If no citation is found use get_clusters_from_citation as a backup
             volume = full_citation.groups.get("volume", None)
-            reporter = full_citation.groups.get("reporter", None)
+            reporter = full_citation.corrected_reporter()
```
grossir, line 72: may also use `page = full_citation.corrected_page()` to help find the NY citations

cl/citations/match_citations.py
```diff
@@ -68,17 +68,31 @@ def resolve_fullcase_citation(
...
                 clusters, _count = async_to_sync(
                     get_clusters_from_citation_str
                 )(volume=volume, reporter=reporter, page=page)
+
+                # exclude self links
+                clusters = [
+                    cluster
+                    for cluster in clusters
+                    if (
+                        full_citation.citing_opinion is not None
```
grossir, line 89: doing this will just make the count 0 when getting into this step on `recap_citations`

Why not just do the self check when it's not a recap document?
```python
if not getattr(full_citation, 'citing_document', False):
      clusters = [ ... ]
      _count = len(clusters)
```

grossir (approved):
Looks good!

PR #5150, opened 2025-02-26
feat(admin): add UnmatchedCitation to admin

quevon24:
This is a basic implementation of UnmatchedCitation model to admin.

You can filter the items by type or status and you can search by volume, reporter or page.

image


flooie:
Ok, this is great thanks @quevon24 

A few things that I think would be great.

I feel like the fields are super long 

Screenshot 2025-03-04 at 10 47 00 AM

Maybe we should clean up that one particular drop down

Screenshot 2025-03-04 at 10 49 30 AM

I dont get anything useful from being able to see the type in the dropdown. I would prefer to be able to perhaps see the cluster and click a link to jump to the opinion that the unmatched citation is in.  So I guess if I could see the citing opinions so I can jump into it it would be more useful.

quevon24:
@flooie The changes are ready

List (display citing_opinion with link):
image


Edit (change how status is displayed in edit view):
image

flooie (approved):
Thanks @quevon24

PR #5157, opened 2025-02-27
feat(pages): make docket and case names one-click selectable

### LINKED ISSUE #4984: Make docket name and case name one-click selectable
When you click on a docket number, it uses some fancy CSS thing to automatically select the element for copy/pasting. It's neat! 

We should also do this for docket names and for case names. I don't know why we didn't! I think we just need to add a class to these and that should do it.

elisa-a-v:
Closes #4984 

I went ahead and added the class to the opinion citations too, I assumed that was something people might want to copy often.

Please let me know if I missed any other instances that need to be made one-click selectable!

### DIFF
--- a/cl/api/templates/rest-docs-v2.html
+++ b/cl/api/templates/rest-docs-v2.html
@@ -1,7 +1,6 @@
 {% extends "base.html" %}
 {% load static %}
 
-
 {% block title %}REST API, v2 &ndash; CourtListener.com{% endblock %}
 {% block description %}
     The first REST API for federal and state case law and still the best.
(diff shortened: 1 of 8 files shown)


elisa-a-v:
I looked around and those were the only ones I found that were not links, same with the PACER document page—both the docket name and number are links, and the citation has the class already. I don't think those need to be links though, since we already have a "Back to Docket" link in the sidebar and both link to the docket too, shall I change them so they're one-click selectable instead of links?

mlissner:
Oh, I hadn't thought of that. Actually, the links are more important. I use them regularly and people will be sad if they stop working. (I agree that the back to docket thing is **right there!**, but people complained in the past that it was too hard to get back to the docket, so we made the header a link too.)

Want to put that back and then merge?

elisa-a-v:
Oooh I completely understand! I reverted the commit and set to auto-merge.

PR #5160, opened 2025-02-28
fix(opinions): remove toc from sidebar

quevon24:
Small change that removes the toc from the sidebar of the new case law ui.

Before:

image

After:

image

A different tab:

image


flooie:
can you add teh screenshots for the other page where it still exists?

quevon24:
> can you add teh screenshots for the other page where it still exists?

which page?

flooie:
I meant of the page that still contains the TOC

mlissner:
Is this right? I don't see a TOC at all now? https://www.courtlistener.com/opinion/107252/miranda-v-arizona/

image

PR #5161, opened 2025-02-28
fix(help): update opinion and cluster endpoint documentation

quevon24:
The documentation has been updated to indicate that the citations field is part of the Cluster API, not the Opinion API. This change ensures users can correctly find citation data within the Cluster object.


mlissner:
Thanks for the fix!

PR #5165, opened 2025-02-28
fix(opinions): use default style for buttons in new case law ui

quevon24:
- use default style for buttons in new case law ui
- remove unused styles

image


flooie:
we should probably standardize these drop-downs 

its strange that I didnt do that before but can you make both the same in style 

Screenshot 2025-03-06 at 1 58 26 PM
Screenshot 2025-03-06 at 1 58 29 PM

quevon24:
> we should probably standardize these drop-downs
> 
> its strange that I didnt do that before but can you make both the same in style
> 
> Screenshot 2025-03-06 at 1 58 26 PM Screenshot 2025-03-06 at 1 58 29 PM

Which option of the two dropdown do you like more?

mlissner:
What standardization are you looking for, Bill? One is a button with a drop down, the other is a button drop down. Different things?

flooie:
@mlissner they do the same thing - the dropdown a button to click on.  If I'm the only one it bothers thats fine.  But I feel like they should both look the same.  I prefer the separator @quevon24  personally.

quevon24:
I think the main reason for having two different buttons is because on the first button with dropdown, if you click on "Get Citation Alerts" it will show a modal, but if you open the dropdown, you can see an option that takes you to the alerts help page.

In the second button, we can have multiple options to download (harvard pdf, a file from different source), so we cant perform an action if we had a button + dropdown, because user needs to select which files want to download

Issue #5133, opened 2025-02-22
Citation Matching always creates unmatched annotation for self cites


flooie:
`es_search_db_for_full_citation`

This function removes self citations, which is good but because it doesnt find itself we will create an unmatched citation for every opinion.  

Im not sure what the right approach is here, but creating an annotation citation no link - as well doesnt seem like the right choice to me. 

```
  must_not = []
  if full_citation.citing_opinion is not None:
      # Eliminate self-cites.
      must_not.append(Q("match", id=full_citation.citing_opinion.pk))
```

PR #5167, opened 2025-02-28
fix(store_unmatched_citations): do not store self cite as Unmatched

grossir:
Solves #5133

- compare the citing opinion's cluster citations with the unmatched citations, to exclude self cites.
- refactor `logger.error` calls for invalid full cites
- add tests

### DIFF
--- a/cl/citations/tasks.py
+++ b/cl/citations/tasks.py
@@ -325,6 +325,7 @@ def store_unmatched_citations(
     """
     unmatched_citations_to_store = []
     seen_citations = set()
+    citations_to_this_cluster = [str(c) for c in opinion.cluster.citations]
 
     for unmatched_citation in unmatched_citations:
         if not isinstance(unmatched_citation, FullCaseCitation):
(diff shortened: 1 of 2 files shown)


cl/citations/tasks.py
```diff
@@ -334,16 +337,21 @@ def store_unmatched_citations(
         # values in required fields
         groups = unmatched_citation.groups
         if (
-            not groups.get("reporter", None)
-            or not groups.get("volume", None)
-            or not groups.get("volume").isdigit()
-            or not groups.get("page", None)
+            not groups.get("reporter")
+            or not groups.get("volume")
+            or not groups.get("page")
```
flooie, line 342: why are we removing this is digit line?

grossir: to make the `logger.error` message proper; it's not null, it's not a integer like string. The `isdigit()` check is now  a few lines below

flooie (approved):

Issue #5111, opened 2025-02-20
Round III of Pray and Pay front end revisions


mlissner:
This ticket tries to tally up what's remaining for P&P that's in #4845. It also includes my latest suggestions for ways it could be even better when we launch:

 - [x] Pagination on various prayer pages

 - [x] Ensure mailing list goes out near end of each quarter automatically (I'll talk to @jeniferw)

 - [x] New: The three buttons on the leader board are  too far right. There should be a small margin on the right between the rightmost button and the edge of the white part. See the "Donate" button in the navbar for an example.

 - [x] Need some spacing on the leaderboard above the table so it's more balanced.

 - [x] The two tables in the "Your Prayers" page is confusing b/c it doesn't explain why there are two tables. We should use tabs, pills, or something else for this to separate and label the tables. At a minimum, we should put a heading above the second table.

 - [x] At certain widths, the right side of the prayers emoji on the docket page is unclickable. I narrowed it down to this CSS:

    ```
    .prayer-button[data-gap-size="large"] {
      margin-left: 44px;
    }
    ```

 - [x] The modals need a bit of work. Here's an example of creating a note that looks about right: 

    ![Image](https://github.com/user-attachments/assets/f5547c17-3d7e-4213-8052-ca71831d3eef)

    The differences:
     - The header is bigger
     - The action is in the lower right
     - There's no X button (but that's a mistake, the P&P modals are better here)

    So for example, this one could be improved with:

    ![Image](https://github.com/user-attachments/assets/cecb23e3-e8b5-4110-9a66-fd329b9a8872)

     - Bigger header
     - Change the text to say:

        > Every quarter, PACER waives the first $30 you spend. Sign up here and we'll remind you each quarter to use your allotment

     - Move the sign up button to the lower right, and maybe remove the close button that's there.

    And on this modal:

    ![Image](https://github.com/user-attachments/assets/386cfb84-eb7c-4c1c-a317-04df080c95eb)

     - Bigger header
     - Change text to say:

        > By clicking 🙏, you used 1 of your 5 daily prayers asking somebody to buy a document for you from PACER. Your prayer will now show up on our [leaderboard of most-wanted PACER documents](https://www.courtlistener.com/prayers/top/), and when the community buys it for you, we'll send you an email to let you know.

     - From there, we can put a Learn More link in the left and a Got It on the right.

I think that's it for now. Just the last little things, if you've still got patience for this, @v-anne!

mlissner:
[...]

PR #5186, opened 2025-03-03
feat(favorites): modifying pray-and-pay ranking algorithm

v-anne:
This PR follows up on a discussion related to the pray-and-pay ranking algorithm in #5111. The existing algorithm takes the geometric mean of the number of requests for a document and the average age of those requests. The way the code is currently written, the average age is always measured in seconds. That means for almost all documents, average age >>> number of requests, so a very old request will outweigh multiple recent requests for a different document.

To resolve this issue, the new algorithm first sorts by the number of requests for a document, and then breaks ties by sorting by number of views a particular docket (not document) has.

Edit:

I also added a quick fix to the link issue for PACER attachments identified in #5292.

### DIFF
--- a/cl/alerts/tests/tests.py
+++ b/cl/alerts/tests/tests.py
@@ -1643,9 +1643,12 @@ class SearchAlertsUtilsTest(SimpleTestCase):
                 },
             ],
         }
+        mock_date = now().replace(day=27, hour=5)
         for rate, test_cases in test_cases.items():
             for test_case in test_cases:
-                with self.subTest(rate=rate, test_case=test_case):
+                with self.subTest(
+                    rate=rate, test_case=test_case
+                ), time_machine.travel(mock_date, tick=False):
                     expected_date_start = date(
                         test_case["year"],
                         test_case["cut_off_month"],


alex:
FWIW, as I documented in https://github.com/freelawproject/courtlistener/issues/5266, I'm not sure the current algorithm works as intended.

Using the simpler "order by prayers" algorithm makes a ton of sense to me. (I think it'll also allow the to prayers page to include the prayer count for each document without confusing users).

mlissner:
That sounds good to me. It's on @ERosendo's backlog as a P3 (lowest priority), but the P3's tend to break up our P1's nicely.

ERosendo:
> I am trying to understand whether the function get_absolute_url has to be included in the "only" part of the ORM queries in get_top_prayers and get_user_prayers.

@v-anne We don't need to include get_absolute_url in the query. As a model property, it's calculated directly. We only need to ensure all fields required for its calculation are present in the query, which they are.

cl/favorites/templates/top_prayers.html
```diff
@@ -63,7 +63,7 @@ <h3 class="text-center v-offset-below-3"><b>{{ waiting_stats.prayer_count|intcom
                 </a>
               </td>
               <td>
-                <a href="{% url "view_recap_document" docket.pk prayer.docket_entry.entry_number docket.slug %}">
+                <a href="{{ prayer.get_absolute_url }}" rel="nofollow">
```
ERosendo, line 66: The document link fix is working correctly. However, the table becomes cluttered when multiple attachments from the same entry are listed consecutively, as shown in this screenshot:

Screenshot 2025-03-28 at 10 16 03 PM

To improve readability, I suggest adding the attachment number to each record when available. This would help distinguish individual attachments: 

Screenshot 2025-03-28 at 10 20 26 PM

Let me know what you think

ERosendo: Here's the code for the template: 

```
                <a href="{{ prayer.get_absolute_url }}" rel="nofollow">
                  {% if prayer.attachment_number %}
                    {{ prayer.document_number }}-{{ prayer.attachment_number }}
                  {% else%}
                    {{ prayer.document_number }}
                  {% endif %}
                </a>
```

ERosendo (approved):
This code looks generally good. I've added a few await statements to correct the failing tests. There's just one comment concerning the table. Once that's addressed, we can merge this PR.

ERosendo (approved):
LGTM :shipit:

ERosendo:
> Do you know why the tests failed earlier?


A different [PR](https://github.com/freelawproject/courtlistener/pull/5322) fixed the failing test; it wasn't related to your changes

PR #5187, opened 2025-03-03
feat(find_citations): Add no html with citations arg

flooie:
We need to re-run anything without html with citations and generate one for all the missing opinions

Also enable find citations to run even when no citations are found.


cl/citations/tasks.py
```diff
@@ -156,10 +156,6 @@ def store_opinion_citations_and_update_parentheticals(
         **get_markup_kwargs(opinion),
     )
 
-    # If no citations are found, then there is nothing else to do for now.
-    if not citations:
-        return
```
grossir, line 161: Wouldn't it be better to let it create the cited html, then save the opinion and exit?

```python

    # Resolve all those different citation objects to Opinion objects,
    # using a variety of heuristics.
    citation_resolutions: Dict[
        MatchedResourceType, List[SupportedCitationType]
    ] = do_resolve_citations(citations, opinion)

    # Generate the citing opinion's new HTML with inline citation links
    opinion.html_with_citations = create_cited_html(
        opinion, citation_resolutions
    )
    
    if not citations:
          opinion.save()
          return

```
Otherwise, a bunch of queries will be run against the DB, even when all should return none, like

`opinion.opinions_cited.all()`

```
update_unmatched_status = UnmatchedCitation.objects.filter(
        citing_opinion=opinion
    ).exists()
```

```
        opinion_clusters_to_update = OpinionCluster.objects.filter(
            sub_opinions__pk__in=opinion_ids_to_update
        )

```
etc

grossir (commented):
Looks good; I just think we should keep an early exit after html_with_citations has been created

Issue #5191, opened 2025-03-03
Handle DataError for volume smallint field when creating UnmatchedCitations


grossir:
We need to handle DataErrors raised from the following:

SmallInt has a max value of 32 767

We are seeing the following volumes from random numbers next to reporter-like tokens; or  artifacts from OCRing the documents"

- In opinion [10815309](https://www.courtlistener.com/opinion/10348721/nonemacher-k-v-nonemacher-s/): "**892301655** Va. 861", comes from "PACSES NO. 892301655 V. 861 EDA" which mixes a couple docket numbers and the "v" from plaintiff / apellant

- In opinion [10660768](https://www.courtlistener.com/opinion/10194176/edmar-financial-company-llc-v-currenex-inc/): "**60606** P. 312443" comes from "Chicago IL 60606 P 312443 3200" a zipcode and something else

- In opinion [10354499](https://www.courtlistener.com/opinion/9887889/the-avon-company-v-fareva-morton-grove-inc/): "**21842005** WL 2105592" where the WL cite should be just 2005, but the OCR has mixed them together

Sentry Issue: [COURTLISTENER-96D](https://freelawproject.sentry.io/issues/6318388481/?referrer=github_integration)

```
NumericValueOutOfRange: smallint out of range
  File "django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
  File "psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)

DataError: smallint out of range
(10 additional frame(s) were not displayed)
...
  File "cl/citations/tasks.py", line 130, in find_citations_and_parentheticals_for_opinion_by_pks
    store_opinion_citations_and_update_parentheticals(opinion)
  File "cl/citations/tasks.py", line 241, in store_opinion_citations_and_update_parentheticals
    store_unmatched_citations(unmatched_citations, opinion)
  File "cl/citations/tasks.py", line 377, in store_unmatched_citations
    UnmatchedCitation.objects.bulk_create(unmatched_citations_to_store)
```

PR #5192, opened 2025-03-04
fix(store_unmatched_citations): prevent DataError from bad volume value

grossir:
Solves #5191

Check that the volume value does not exceed the smallint postgres type max value of 32_767; this is not normal, and is usually an layout extraction artifact; a OCR issue; or a citation lookalike


flooie (approved):
Small Ints strike again.

PR #5197, opened 2025-03-04
fix(favorites): adjust cache and prayer price calculation

v-anne:
This PR has three quick fixes to issues raised in #5195.

1. Adjusting cache of prayer statistics to last for a minute instead of 24 hours. This is to account for the relatively few number of prayers pending right now.
2. Making the default price of a document $0.10 instead of $0.00 if we don't know its length, as every paid document costs at least $0.10.
3. Removes periods from headers on the prayer pages.

### DIFF
--- a/cl/citations/management/commands/find_citations.py
+++ b/cl/citations/management/commands/find_citations.py
@@ -73,6 +73,12 @@ class Command(VerboseCommand):
             help="Parse citations for all items",
         )
         parser.add_argument(
+            "--no-html-with-citations",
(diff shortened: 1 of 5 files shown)


mlissner (approved):
LGTM. Easier than I'd expect!

mlissner:
Thanks. I squashed and merged these. I think to avoid the mess of commits, I'd sync my fork with the upstream repo, and then make a new branch off main for the next task.

PR #5214, opened 2025-03-05
fix(favorites): Refines get_user_prayers to eliminate duplicates

### LINKED ISSUE #5211: Pending PACER requests duplicating
This is a strange bug. On the pending requests page of the pray-and-pay project, all requests for a document are shown if a particular user was one of the people who requested it.

Image

To summarize, other users requested both of the documents before me. I subsequently requested them today to evaluate how the ranking algorithm was performing (that is a separate issue), but when I went to the pending requests page, the page is also showing me information from when the original users made the first request for each document.

I don't know what is happening, as each user can only request a given document once (and you should be able to confirm in the db that I only requested each of these documents once).

The function `get_user_prayers()`, which is used to populate this table, also specifically filters for requests made by a particular user.

This seems like a fairly urgent bug to fix.

ERosendo:
This PR refines the `get_user_prayers` helper method to avoid including duplicate documents in the query. Unit tests have been added to prevent future regressions.

Fixes #5211

### DIFF
--- a/cl/favorites/tests.py
+++ b/cl/favorites/tests.py
@@ -29,6 +29,7 @@ from cl.favorites.utils import (
     get_prayer_counts_in_bulk,
     get_top_prayers,
     get_user_prayer_history,
+    get_user_prayers,
     prayer_eligible,
 )
 from cl.lib.test_helpers import AudioTestCase, SimpleUserDataMixin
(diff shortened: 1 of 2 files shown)


mlissner (approved):
Thanks!

PR #5229, opened 2025-03-07
fix(generate_annotations): fix `make_citation_url_dict` usage

grossir:
Solves #5158

Ensures `make_citation_url_dict` only gets the arguments it's ready to get


flooie (approved):
looks good to me

PR #5231, opened 2025-03-08
feat(pnp): filter out unavailable documents from prayers list

v-anne:
This PR is part of #5195.

Namely, it:
- Checks whether particular documents are available for purchase on PACER after someone requests them as part of the pray-and-pay project
- If those documents are unavailable, users are notified via email.
- Updates the page_count of the requested documents that are available for purchase
- Adds prayer buttons to the community leaderboard that show how many requests a particular document has.
- Updated documentation

In a future PR, I will modify the leaderboard ranking algorithm to take the document availability into consideration.


mlissner:
> To fix that, we could add the button to rows that satisfy if rd.is_sealed, on the understanding that this function will check if the document remains sealed, and if not, will allow the user to make the request.

My first thought is: Let's just not bother with sealed content. It adds a lot of complexity to the UX. My second thought is: People sometimes *really* want sealed content, so let them pray for it, and maybe someday they'll get an email and appreciate it. 

If we do the second thing, we could just let them pray for such things, then if things are sealed, they get demoted in the leaderboard, and that's that. If it's an important enough item, somebody will buy it someday and they'll get an email. Simple enough?

mlissner:
Sure, seems fine. Do we want to put a cache in front of this, so that if 100 people pray for the same doc in an hour, they can't trigger 100 checks to PACER to see if it's sealed?

ERosendo:
@v-anne Just checking in on the cache implementation from the suggestion in https://github.com/freelawproject/courtlistener/pull/5231#issuecomment-2711999048. Have you started working on it?

ERosendo:
@v-anne got it! 

I've taken a look at your code, and I was wondering about the document availability check for each new prayer. It seems like it would happen quite frequently.

I was thinking, perhaps we could explore a different  approach? we could consider implementing a daemon that periodically checks for new prayer availability. This daemon, similar to `cl_send_rt_percolator_alerts`, could execute your code every five minutes, thereby removing the availability checks from the `create_prayer` method.

This approach would result in prayers from sealed documents appearing briefly on the leaderboard. However, it would keep the `create_prayer` method simple by shifting availability checks to the daemon.

mlissner:
In general, I try to avoid daemons (they're harder to test, more likely to break, use more resources, etc). What do we get from doing that instead of just checking prayers as they're created?

ERosendo:
> What do we get from doing that instead of just checking prayers as they're created?

Using the daemon could potentially eliminate the need for caching

mlissner:
I guess we could eliminate the caching too, if we just check if there's already a prayer for a particular document, but we'd have to add a column to the prayers table to track if things are unavailable. I appreciate how having such a column would allow us to demote those prayers instead of removing them mysteriously.

I think that could work, but in any case, I don't think using a daemon really fixes the need for the cache because if people are slowly requesting a given doc every few minutes, we'd wind up checking it over and over, right? 

If 20 people pray for the same doc,

ERosendo:
> we'd have to add a column to the prayers table to track if things are unavailable. I appreciate how having such a column would allow us to demote those prayers instead of removing them mysteriously.

We can use the `is_sealed` field from the `RecapDocument` model to demote prayers (This PR implements logic to update the field in cases where the document is not accessible.)

> if people are slowly requesting a given doc every few minutes, we'd wind up checking it over and over, right?

I was thinking we could filter out documents that are already marked as 'sealed' (if we've done that in a previous run) and also those where we've already updated the page count.

mlissner:
That helps, but some items aren't sealed and also aren't available, like transcripts...that's why I was thinking another field might be needed for something that's just generally unavailable.

mlissner:
That sounds great to me, yes. My one thing I'd suggest, though nobody seems to like it, is for the unavailable docs to still appear on the leaderboard, just demoted to the bottom and with a label or something saying they were unavailable as of XYZ date.

I just worry that disappearing things confuses people, they'll go pray for it again, and the cycle will repeat.

mlissner:
Thanks for posting the model. You're learning our pain points! 

A couple thoughts:

1. I don't think you mean to use `auto_now_add`. In the docs it says:

    image

    So I think you want to use `default` instead.

2. I don't think you need a db_index on the `last_checked` field. Instead I think you want a multi-column index on both fields.

3. I think for related_name I'd do `prayer_availability`. I've found it's almost always best to just use the class name for this field, and that if that doesn't work, it usually means the class name should also change. 

Thank you!

ERosendo:
Hey @mlissner, I reviewed the model proposal and noticed @v-anne added the `@pghistory.track()` decorator. Do you think it's necessary to introduce a table for event tracking?

mlissner:
Yeah, we could live without it. I agree.

cl/favorites/models.py
```diff
@@ -200,3 +201,30 @@ class Meta:
...
+            models.UniqueConstraint(
+                fields=["recap_document"],
+                name="unique_recap_document",
+            ),
+        ]
+        indexes = [
+            models.Index(
+                fields=["recap_document", "last_checked"],
+            ),
+        ]
```
ERosendo, line 230: I don't think we really need that multicolumn index. The queries in this PR primarily use the recap document ID to access the table and we're not using the `last_checked` column for filtering.

However, if the goal with the multicolumn index was to avoid fetching data from both the index and the heap when the table gets big, then I think a covering index would be a better way to go. We could use the UniqueConstraint with the [include](https://docs.djangoproject.com/en/5.1/ref/models/constraints/#include)  argument to do that.  Here's my suggestion: 

```suggestion
        constraints = [
            models.UniqueConstraint(
                fields=["recap_document"],
                name="unique_recap_document",
                include=["last_checked"]
            ),
        ]
```

cl/favorites/models.py
```diff
@@ -200,3 +201,30 @@ class Meta:
...
+class PrayerAvailability(models.Model):
+    recap_document = models.ForeignKey(
+        RECAPDocument,
+        help_text="The document people prayed for.",
+        related_name="prayeravailability",
+        on_delete=models.CASCADE,
+    )
+    last_checked = models.DateTimeField(
+        help_text="The time when the availability of this document was last checked.",
+        default=timezone.now(),
```
ERosendo, line 216:
```suggestion
        auto_now=True,
```

cl/favorites/utils.py
```diff
@@ -51,7 +52,13 @@ async def create_prayer(
         new_prayer, created = await Prayer.objects.aget_or_create(
             user=user, recap_document=recap_document
         )
-        return new_prayer if created else None
+        if created:
+            if not fetch_prayer_availability.delay(recap_document.pk):
```
ERosendo, line 56: This wont work as expected. The delay method returns immediately upon invocation, without awaiting the completion of the associated task and the returned value is an [AsyncResult](http://docs.celeryproject.org/en/latest/reference/celery.result.html#celery.result.AsyncResult%20AsyncResult) instance.

ERosendo:
@v-anne Based  in https://github.com/freelawproject/courtlistener/pull/5231#issuecomment-2725728115, I thought this PR was to be addressed first. Is there a reason for the change in order?

cl/recap/tasks.py
```diff
@@ -3435,3 +3437,89 @@ def do_recap_document_fetch(epq: EmailProcessingQueue, user: User) -> None:
...
+@app.task(
+    bind=True,
+    autoretry_for=(RedisConnectionError, PacerLoginException),
+    max_retries=5,
+    interval_start=5,
+    interval_step=5,
+    ignore_result=True,
+)
+@transaction.atomic
+def fetch_prayer_availability(self, pk: int) -> bool:
```
ERosendo, line 3491: Let's refactor this method. We can change it to a regular function and remove the call to _run_availability_check. The function should simply return a boolean value, indicating whether or not we need to schedule a task to check document availability. Here's a suggestion for the signature:

```suggestion
def should_check_prayer_availability(self, pk: int) -> bool:
```

cl/recap/tasks.py
```diff
@@ -3435,3 +3437,89 @@ def do_recap_document_fetch(epq: EmailProcessingQueue, user: User) -> None:
         process_recap_email.si(epq.pk, user.pk),
         extract_recap_pdf.s(),
     ).apply_async()
+
+
+def _run_availability_check(rd: RECAPDocument) -> bool:
```
ERosendo, line 3442: this function should be implemented as a task. It allows for clear decoupling of the court page validation logic from the main request flow. Additionally, this refactoring creates a natural point for integrating email notification functionality (the prayer_unavailable method)

cl/favorites/utils.py
ERosendo, line 1: after implementing https://github.com/freelawproject/courtlistener/pull/5231#discussion_r2019605581 and https://github.com/freelawproject/courtlistener/pull/5231#discussion_r2019609256, we should refactor the `create_prayer` function to enhance readability. Here's my suggestion: 

```python
async def create_prayer(
    user: User, recap_document: RECAPDocument
) -> Prayer | None:
    is_user_eligible, _ = await prayer_eligible(user)
    if not is_user_eligible or recap_document.is_available:
        return None

    new_prayer, created = await Prayer.objects.aget_or_create(
        user=user, recap_document=recap_document
    )
    if not created:
        return None

    if should_check_prayer_availability(recap_document.pk):
        check_prayer_availability.delay(recap_document.pk)

    return new_prayer
```

ERosendo (changes requested):
I've added comments suggesting refactoring improvements to this PR. Please review them and let me know when it's ready for another review.

ERosendo:
> Because we are now checking for document availability in create_prayer, should we also update the API endpoint to incorporate the same logic?

@v-anne  Good catch!. I thought we were using the `create_prayer` method for the API view too, but I just double-checked and we're not. So, to keep things cleaner and avoid repeating ourselves, I think we should move the `should_check_prayer_availability` method and the task scheduling code into a `post_save` signal.

let me know what you think.

ERosendo:
> I am not familiar with the methodology of signals. Would doing that cover requests made via the API as well?

Yes, We can use a `post_save` signal on the `Prayer` model to make sure our validation happens automatically after saving. This way, we only need to write the validation code once, and it'll work for both the website and API, saving us from doing the same thing multiple times.

ERosendo:
> Can you point me to any existing examples in the codebase?

Certainly. Here's an example of one of our `post_save` signals:

https://github.com/freelawproject/courtlistener/blob/2b324d99be03ca3633538cb8761283aa217fcd77/cl/search/signals.py#L547-L578

cl/favorites/views.py
```diff
@@ -205,6 +207,21 @@ def paginate_open_prayers(top_prayers, prayer_page):
...
+        # Check prayer existence in bulk.
+        existing_prayers = await get_existing_prayers_in_bulk(
+            user, recap_documents
+        )
+
+    # Merge counts and existing prayer status to RECAPDocuments.
+    for rd in recap_documents:
+        rd.prayer_count = prayer_counts.get(rd.id, 0)
+        rd.prayer_exists = existing_prayers.get(rd.id, False)
+
```
ERosendo, line 224: This code has a few issues: 

1. We probably don't need to use the `list()` method. The `object_list` property is already a `queryset`, and `get_existing_prayers_in_bulk` should be able to handle that directly. If `mypy` raises a type error, I'd prefer to adjust the type hint instead of using `list()`.

2. The `get_prayer_counts_in_bulk` method might not be necessary anymore. In a previous [PR](https://github.com/freelawproject/courtlistener/pull/5186), the `prayer_count` was included in the `queryset` returned by the `get_top_prayers` method.

3. We need to make sure the existing_prayers variable has a default value. Otherwise, the code could run into an error when trying to access it before it's been set, especially when someone checks the board without logging in.

4. Since this code is part of an asynchronous view, we should probably use async for instead of a regular for loop when we're working with the records in the `recap_documents` variable.

Here's my suggestion: 

```suggestion
    recap_documents = paginated_entries.object_list

    user = await request.auser()
    existing_prayers = {}
    if user.is_authenticated:
        # Check prayer existence in bulk.
        existing_prayers = await get_existing_prayers_in_bulk(
            user, recap_documents
        )

    # Merge counts and existing prayer status to RECAPDocuments.
    async for rd in recap_documents:
        rd.prayer_exists = existing_prayers.get(rd.id, False)

```

cl/favorites/templates/top_prayers.html
ERosendo, line 1: Here's how the most wanted documents page would look with this change: 

image

Personally, I think it looks a bit better with the button on the right and a little smaller:

image

@mlissner I'd be interested to hear your thoughts on the top prayers table.

Also, the button isn't working because the `pray_and_pay.js` file isn't being loaded in the template.

ERosendo:
> Should I add tests for the signal?

@v-anne Yes, that would be great! if you encounter any challenges with the tests, just let me know. I'd be glad to help :)

ERosendo:
> can you point me to tests for other signals? I am not that familiar with them.

Sure! Here's an example of a test for a [post-save](https://github.com/freelawproject/courtlistener/blob/a16b0efbabc34b225047615f7fb331d0de69a267/cl/search/signals.py#L583-L598) signal:

https://github.com/freelawproject/courtlistener/blob/a16b0efbabc34b225047615f7fb331d0de69a267/cl/citations/tests.py#L2770-L2784

Generally, we want to test the result of the signal after the records are created. So, for your case, we'll want tests that verify: 

- New data is correctly added to the `PrayerAvailability` table.
- Emails are being sent to the appropriate users.
- The task is being scheduled when it should be.

ERosendo:
> Would you suggest I incorporate this into a function that already exists, like test_create_prayer?

That's a good thought!  However, to keep our tests focused and maintainable, I wouldn't add the validation logic into the existing `test_create_prayer` function. It's better to create new, dedicated test methods for this validation within the `RECAPPrayAndPay` test class. This will keep the `test_create_prayer` test focused on its original purpose and make the new validation tests easier to identify and manage.

> do I need to find real PACER documents that the server will check availability for? 

No, there's no need to find real PACER documents for these tests. We use mocks specifically to avoid making actual requests to PACER during testing.

I'm happy to help you get these tests set up now, and then you'll be all set to take the lead on future ones.

ERosendo:
@v-anne I've added tests for the new signal and task in this PR.

@mlissner this PR introduces a new model, so migration files are included. I've only added the `.sql` file for replicas since this PR doesn't involve creating trigger or event tables. Is that the correct approach, or should I also include the `_customer.sql` file even though it's currently identical to the file for replicas?

Also, could you please take a look at my comment here:  https://github.com/freelawproject/courtlistener/pull/5231#discussion_r2032181552? Your input would be appreciated.

cl/favorites/templates/top_prayers.html
mlissner: Let's go with the right, because that doesn't make the rows taller, and because I don't think there's enough context on this page to make a vote particularly meaningful. I don't mind if people *can* do it on this page, but I'm not positive that we want to *encourage* it.

mlissner:
I made a comment on the button placement question. @ERosendo can you help guide the DB questions, please?

mlissner:
OK, yeah, you migration files look good to me!

cl/favorites/templates/top_prayers.html
ERosendo: Thanks! I'll update the template. 👍

ERosendo:
@v-anne I've updated the template based on the feedback in https://github.com/freelawproject/courtlistener/pull/5231#discussion_r2032181552. Could you please highlight those comments?

cl/favorites/utils.py
```diff
@@ -363,3 +373,55 @@ async def get_lifetime_prayer_stats(
     await cache.aset(cache_key, data, one_minute)
 
     return PrayerStats(**data)
+
+
+def prayer_unavailable(instance: RECAPDocument, user_pk: int) -> None:
```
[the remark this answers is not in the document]
ERosendo: > Should this function take the primary key of the user, or the user object itself?

I think it's okay to just pass the `user_id`. We're only using it to filter the open prayers list. If the method needed to query the database again using the ID to get more user details, then I'd agree we should pass the whole object. But since we're just filtering, passing the ID makes sense.

ERosendo (approved):
@v-anne This PR is ready to go! No further comments from me.

ERosendo:
@v-anne that issue should be resolved once PR https://github.com/freelawproject/courtlistener/pull/5301 is merged

PR #5246, opened 2025-03-12
Fix/sitemap generation issues (command)

alextop:
The sitemaps_infinite app is based on the SQL `cursor pagination` approach. 

Cursor pagination allows to paginate huge database tables without using `OFFSET` SQL instruction, that makes the `SELECT` very slow or even impossible when the number of records is very big. 

On the contrary, cursor pagination works in the constant time, no matter what page is being selected.

P.S. Please do not forget to submit new sitemap index file to google.

All details are described in the README.md file.


cl/sitemaps_infinite/conf.py
ERosendo, line 1: I think we can remove this file. It seems redundant to have `FILES_PER_CALL` and `CELERY_TASK_REPETITION` defined here when we can just pull those values directly from our settings.

Also, I think we can clean things up by getting rid of `make_sitemaps_list` helper and the `SITEMAPS` constant.

Here's my idea for a refactor to remove this file altogether:

1. Since we're making a new Django app specifically for this sitemap generation approach, let's create a `urls.py` file in the `sitemaps_infinite` folder. We can move the `large-sitemap.xml` index view and the `large-sitemap` sections into this new file (`urls.py`).

2. You created a variable named `pregenerated_sitemaps`. Let's move that into this new `urls.py` file within the `sitemaps_infinite` folder. This variable already holds the same data as the output of `make_sitemaps_list`, so we can use it instead of the SITEMAPS constant.

    The `urls.py` in `sitemaps_infinite` would look like this:

```python

from django.contrib.sitemaps import views as sitemaps_views
from django.urls import path
from django.views.decorators.cache import cache_page

from cl.opinion_page.sitemap import DocketSitemap
from cl.search.models import SEARCH_TYPES
from cl.sitemap import cached_sitemap

pregenerated_sitemaps = {
    SEARCH_TYPES.RECAP: DocketSitemap,
}

urlpatterns = [
    path(
        "large-sitemap.xml",
        cache_page(60 * 60 * 24, cache="db_cache")(sitemaps_views.index),
        {
            "sitemaps": pregenerated_sitemaps,
            "sitemap_url_name": "sitemaps-pregenerated",
        },
    ),
    path(
        "large-sitemap-<str:section>.xml",
        cached_sitemap,
        {"sitemaps": pregenerated_sitemaps},
        name="sitemaps-pregenerated",
    ),
]
```

3. With the pregenerated sitemap URLs in their own file, we'll need to update the main urls.py like this:

```diff
...
from cl.opinion_page.sitemap import (
    BlockedDocketSitemap,
    BlockedOpinionSitemap,
-   DocketSitemap,
    OpinionSitemap,
)
...


sitemaps = {
    ...
}

- pregenerated_sitemaps = {
-     SEARCH_TYPES.RECAP: DocketSitemap,
- }


urlpatterns = [
    ...
    # Pregenerated sitemaps, shorter cache time for sitemap index
-   path(
-       "large-sitemap.xml",
-       cache_page(60 * 60 * 24, cache="db_cache")(sitemaps_views.index),
-       {
-           "sitemaps": pregenerated_sitemaps,
-           "sitemap_url_name": "sitemaps-pregenerated",
-       },
-   ),
-   path(
-       "large-sitemap-<str:section>.xml",
-       cached_sitemap,
-       {"sitemaps": pregenerated_sitemaps},
-       name="sitemaps-pregenerated",
-   ),
+  path("", include("cl.sitemaps_infinite.urls")),
...
]
```

4. Then, we'll need to replace any `from cl.sitemaps_infinite import conf` calls with `from django.conf import settings` or `from cl.sitemaps_infinite.urls import pregenerated_sitemaps`.

5. Finally, we can remove conf.py and utils.py from the sitemaps_infinite app.


Let me know what you think

cl/sitemaps_infinite/base_sitemap.py
```diff
@@ -0,0 +1,263 @@
...
+    @property
+    def cursor(self):
+        return self._cursor
+
+    @property
+    def has_next(self):
+        return self._has_next
+
+    @property
+    def ordering(self) -> Tuple[str]:
```
ERosendo, line 115:
```suggestion
    def ordering(self) -> tuple[str]:
```

cl/sitemaps_infinite/base_sitemap.py
```diff
@@ -0,0 +1,263 @@
+import logging
+from datetime import datetime
+from functools import cached_property
+from typing import Any, Tuple, Union
```
ERosendo, line 4:
```suggestion
from typing import Any, Union
```

cl/sitemaps_infinite/base_sitemap.py
```diff
@@ -0,0 +1,263 @@
...
+
+    This sitemap is intended to be pre-generated by a Command rather than generated on-the-fly,
+    as accessing it directly will raise an `InvalidPage` exception.
+
+    The `get_urls_by_cursor` method is used to generate the URLs,
+    taking an optional cursor parameter to fetch the next set of URLs.
+    """
+
+    # set url scheme to be compatible with the regular sitemap
+    protocol = settings.SITEMAPS_URL_SCHEME
```
ERosendo, line 85: I've looked at the [get_protocol](https://github.com/django/django/blob/0d92428d77fafff373e05dd5a6cdb62bd1dfbda0/django/contrib/sitemaps/__init__.py#L81C1-L83C52) method in the Sitemap class, and I believe we can simplify things. We don't need a new environment variable for the URL scheme. Instead, we can override the method in this class and leverage existing settings to provide the default. Here's a suggestion:

```suggestion
    def get_protocol(self, protocol=None):
        env_based_protocol = (
            "http"
            if any([settings.TESTING, settings.DEVELOPMENT])
            else "https"
        )
        return self.protocol or protocol or env_based_protocol
```

cl/sitemaps_infinite/sitemap_generator.py
```diff
@@ -0,0 +1,288 @@
...
+
+# Set up a task if repetition period is set in conf
+@transaction.atomic
+def generate_urls_chunk(force_regenerate: bool = False) -> None:
+    """Generates the next chunk of URLs for the sitemap. This function is responsible for managing the sitemap generation process, including caching, cursor handling, and limiting the number of files generated per call.
+
+    The function iterates over the configured sitemap sections, loading the appropriate `InfinitePaginatorSitemap` class for each section. It then generates the sitemap pages for each section, caching the results and updating the cursor data in Redis. The function stops generating new pages once the configured limit of files per call has been reached.
+
+    The function uses the `make_cache_key()` and `make_expiration_time()` helper functions to generate the cache keys and expiration times for the sitemap pages. It also handles cases where the cached cursor data does not match the current cursor, indicating that the sitemap data may have changed and needs to be regenerated.
+    """
```
ERosendo, line 52: Breaking docstring paragraphs into multiple lines significantly enhances readability

```suggestion
    """
    Generates the next chunk of URLs for the sitemap. This function is 
    responsible for managing the sitemap generation process, including caching,
    cursor handling, and limiting the number of files generated per call.

    The function iterates over the configured sitemap sections, loading the 
    appropriate `InfinitePaginatorSitemap` class for each section. It then 
    generates the sitemap pages for each section, caching the results and 
    updating the cursor data in Redis. The function stops generating new pages
    once the configured limit of files per call has been reached.

    The function uses the `make_cache_key()` and `make_expiration_time()` helper
    functions to generate the cache keys and expiration times for the sitemap 
    pages. It also handles cases where the cached cursor data does not match 
    the current cursor, indicating that the sitemap data may have changed and 
    needs to be regenerated.
    """
```

Additionally, could you clarify the intended behavior of the `force_regenerate` argument? Based on the method signature, I expected it to clear and recreate the cache from page one. However, it didn't seem to do that when I tried it.

ERosendo (changes requested):
@alextop Thanks for your contribution. The code generally looks good. I've added comments pointing out some areas for improvement before merging. We also need to address the conflicts with the main branch. I tried to resolve them, but it seems I don't have permission to push changes to your fork.

Please let me know when you've made the changes and are ready for a second review.

alextop:
@ERosendo 
Hi!
Can you please check the test results? I see that the test failed but it's not related to my changes at all...
Thanks!

mlissner:
@ERosendo, can you take this on? I suspect the issues are related to old code that'll get fixed if we fix the conflicts and rebase, but I think Alex is out of time to work on this unfortunately. Thank you!

alextop:
If it's only the poetry things I can fix it in 10 min, but yesterday the pipeline failed at the different test in the different module. That's why I asked for help.

alextop:
Aha, now everything is good!

cl/sitemaps_infinite/tasks.py
ERosendo, line 1: @alextop could you please remove this task and the associated file? As @mlissner confirmed in my previous [comment](https://github.com/freelawproject/courtlistener/pull/5246/files/2344a17d05a085b69f68887678b4dcca9017b936..584fe3abea11a17aabedd19a17ae0b9a23276588#r2015640323), we're using Celery's cron functionality.

ERosendo (changes requested):
@alextop changes look great!. However, I did notice that your poetry.lock file seems to have picked up a few more changes than I was expecting. Would you mind recreating it?

Here's how you can do it:

1. Go into the `cl-django` container and run:
```
poetry self update 2.1.1
```
2. Then run:
```
poetry lock
```

that should update the poetry.lock file to only include the changes for the new dependency this pull request is adding (`django-cursor-pagination`).

Thanks!

alextop:
Ok sure. Strange that I did exactly the same last time when I was resolving the conflics.
Anyway, have just done that again, pushing...

ERosendo:
Thanks @alextop, it looks much better now!

ERosendo (approved):
:shipit:

mlissner:
With this finally merged, I could use some help understanding the next steps. I guess we need a cronjob and we need to submit this sitemap to Google? Why isn't this sitemap listed in our root sitemap instead though?

Issue #4819, opened 2024-12-13
Update `Opinion` model to support versions


grossir:
From the parent issue

data model changes: there are many ways we could model the opinion "versions". I think that the simplest is adding a field that points to the "main" version
```python
class Opinion(models.Model):
    cluster = models.ForeignKey(
        OpinionCluster,
        help_text="The cluster that the opinion is a part of",
        related_name="sub_opinions",
        on_delete=models.CASCADE,
    )
    ....
    main_version = models.ForeignKey(self, null=True)
```

This has the advantage of being simple enough, and :

- You can find the "main" version via a simple query Opinion.objects.filter(main_version__isnull=True)
- Given the "main" version you can find all it's versions Opinion.objects.filter(main_version=main_version_id)
these queries should be covered by an index
- The Opinion model already has all the metadata fields for comparing versions: dates, the uploaded document copy, the original URL, the extracted text
all the versions would be grouped under the same OpinionCluster

grossir:
Setting a "Large" size because I intend to at least draft the 3 issues together

PR #5254, opened 2025-03-13
feat(search.models): add Opinion.main_version field

grossir:
Solves #4819

A field to track opinion versions and manage duplication due to different hashes / versions of the same document


mlissner:
It looks like this is duplicated in #5279. Should we just close it?

PR #5260, opened 2025-03-13
Populate new opinion tabs dynamically

quevon24:
This PR updates the new case law ui to populate the opinion tabs dinamically using htmx:

By default it will load all the tabs, the ones that requires ES wont display the count: Cited by and related cases
The authorities and summaries will show the count because those are calculated with: cluster.aauthority_count() and cluster.parentheticals.acount() and if the count is 0 by default it wont display the count (in the image we can see 2 in summaries because we already know the count and the count for authorities is 0, so it is not showing the count)

image

If the htmx request fails, it will continue to display the tabs, the tabs will only be enabled if the request is successful:

image

Disabling the new flag: `ui_flag_for_o_es` will hide the count for those tabs (the user wont know if we have them until going to the tab page) and will display the ES tabs, in the image below Related Cases is not displaying count and Cited by is being displayed because we dont know if we have data to show or not

image

For each tab we need to add a no results message  in case the user clicked the tab before htmx call finished loading the tabs

image

image

image

image


mlissner:
To be clear here, we should show the tabs no matter what. It's just that we won't know their counts at page load, and the counts will come in momentarily later. If the count is zero, we can disable the tab rather than remove it. If the person clicks the tab before the counts come in, we can show an error for that. But the tabs always show in this design. I think it works better than having tabs pop in as the page loads.

quevon24:
> To be clear here, we should show the tabs no matter what. It's just that we won't know their counts at page load, and the counts will come in momentarily later. If the count is zero, we can disable the tab rather than remove it. If the person clicks the tab before the counts come in, we can show an error for that. But the tabs always show in this design. I think it works better than having tabs pop in as the page loads.

I updated the PR message with new screenshots and with the things you mentioned in the issue

mlissner:
Yeah, looks good to me. I'd just move *all* the counts out of the tabs. That'll allow the page to load quickly (good for SEO), and it'll make more of the tabs work the same way. 

Thank you!

quevon24:
> Yeah, looks good to me. I'd just move *all* the counts out of the tabs. That'll allow the page to load quickly (good for SEO), and it'll make more of the tabs work the same way. 
> 
> Thank you! 

then just remove the count from tabs but still use them to show/hide them with htmx, right?

And only show the counts inside the tab content

mlissner:
Not quite, when the page loads:

 - Show all tabs and they're clickable
 - HTMX goes and gets counts for all tabs that make sense
 - JS puts the counts into the tab headers
 - If the count is zero, the tab says "Related Cases (0)" and is disabled, but still shown.
 - If the count is greater than zero, show the count, and leave the tab clickable

Help? 
If people click a tab before HTMX returns, take them to the correct page, where they'll either see the error and the zero items message, or they'll see the content they're looking for.

quevon24:
> Not quite, when the page loads:
> 
>  - Show all tabs and they're clickable
>  - HTMX goes and gets counts for all tabs that make sense
>  - JS puts the counts into the tab headers
>  - If the count is zero, the tab says "Related Cases (0)" and is disabled, but still shown.
>  - If the count is greater than zero, show the count, and leave the tab clickable
> 
> Does that help? 

In this case I use a view to re-render the tabs with htmx using a partial, would it be better if i just update the counter with js?

mlissner:
Yeah, it might be!

quevon24:
@mlissner 

When page loads show all the tabs with zero count and enabled:
image

When the htmx request finishes, refresh the tabs, disabled tag links will point to #:
image

If `ui_flag_for_o_es` flag is disabled, it will enable cited by and related cases tab but with 0, authorities and summaries still being calculated:

image

Also added count in heading of remaining tabs to maintain consistency across all tabs that display the count:

image

image

mlissner:
Looking good! I'd just not show a number upon first page load, then insert the number.

quevon24:
> Looking good! I'd just not show a number upon first page load, then insert the number.

Perfect, i just pushed a new change:

first page load, no counts in tabs:

image

mlissner:
LGTM!

cl/opinion_page/templates/includes/opinion_tabs.html
ERosendo, line 1: Hey @quevon24! I was reviewing the responsive design and noticed this template overflows a bit before it transitions to the mobile layout. Here's a GIF showing the problem: 

Screen Recording 2025-03-18 at 9 07 59 AM

cl/opinion_page/templates/includes/opinion_tabs_content.html
```diff
@@ -0,0 +1,398 @@
...
+             rel="nofollow"
+             class="btn btn-default"
+          >View Citing Opinions</a>
+        </p>
+    </div>
+
+{% elif tab == "related-cases" %}
+    {# Related Cases #}
+    <div class="tab-pane fade in active" id="related">
+        <h2 class="opinion-section-title jump-link">Related Cases ({{ related_clusters|length|intcomma }})</h2>
```
ERosendo, line 181: I noticed the count in the header doesn't match the counts on the tabs. It's a bit confusing. (Screenshot attached). is it possible to make them the same, or just take the count out of the header?

Screenshot 2025-03-18 at 11 56 58 AM

quevon24: Yes, no problem with that, I forgot to clear the ES cache when I took the screenshots.

I delete the cache with these sql statements for cited by and related cases:

```
DELETE FROM django_cache WHERE cache_key LIKE ':1:clusters-cited-es:%';
DELETE FROM django_cache WHERE cache_key LIKE ':1:cited-by-count-es:%';
```

ERosendo: This is likely not a cache issue. The related cases are retrieved using the `es_get_related_clusters_with_cache` helper, and the number of results is limited by the `RELATED_COUNT` setting. Consequently, the header count will consistently be 5 unless that setting is modified, as the header reflects the number of results returned from ES. 

https://github.com/freelawproject/courtlistener/blob/8bc5b2f324d3723bdcd49c297302a3a9d7c7ed25/cl/opinion_page/utils.py#L308-L310

quevon24: Thanks eduardo, I took a look again,  I created a lot of dummy data, that is why the ~300 is correct (I cleared the cache and it is slightly bigger).

I noticed that in the html I'm counting the results instead of using the count and as you mentioned it will always be limited to RELATED_COUNT setting, I'll change that so the tab number matches the count in the tab content.

mlissner: It might actually be better not to show a count inside the tab since we're going to do estimated counts in the tab itself. If the estimate has a 5% error rate, it might say 19k, when then the actual number might be 18,412 and that might confuse people. The actual number isn't really that interesting and we've got it in the tab. Skip it in the tab content?

ERosendo: Hey Kevin,  thanks for the dummy data script. I got everything populated, but I'm seeing some of the headers in the tabs have duplicate counts. Here's a few examples:

Screenshot 2025-03-18 at 4 46 03 PM


Screenshot 2025-03-18 at 4 49 03 PM

Removing the count from the content makes it much cleaner. Here's how it looks in production

image

mlissner: > For all tabs content?

Yes, I think that'll be best, and we can rephrase some of the headers to just remove their counts.

quevon24: Thanks for all your feedback

I fixed the responsive design



https://github.com/user-attachments/assets/4d473561-2d84-4307-8770-9e784183a282




Removed the counts and the extra text from the tab content:

image

image

image

image

mlissner: Looks good to me. Let's see it live, and we can see if we need more explainer, but this should help with performance and accuracy in the meantime. Thanks Kevin!

ERosendo (changes requested):
@quevon24 Thanks for the tweaks. I've verified that the responsive layout issue is fixed. Before merging, I have one remaining comment:

After pulling your latest changes, I noticed that the "Related cases" tab has been renamed, and the button to view the full list of cases is now located in the upper right corner of the tab's content, next to the header. However, in the "Cited by" tab, the button to load the full list is positioned at the bottom of the list. is this placement difference intentional? or should both buttons be consistently located in the same position, next to their respective headers?

Here are screenshots to illustrate the placement of the buttons:

### Similar Cases tab

Screenshot 2025-03-19 at 9 24 30 AM
----

### Cited By tab

Screenshot 2025-03-19 at 9 24 57 AM

quevon24:
> @quevon24 Thanks for the tweaks. I've verified that the responsive layout issue is fixed. Before merging, I have one remaining comment:
> 
> After pulling your latest changes, I noticed that the "Related cases" tab has been renamed, and the button to view the full list of cases is now located in the upper right corner of the tab's content, next to the header. However, in the "Cited by" tab, the button to load the full list is positioned at the bottom of the list. is this placement difference intentional? or should both buttons be consistently located in the same position, next to their respective headers?
> 
> Here are screenshots to illustrate the placement of the buttons:
> 
> ### Similar Cases tab
> Screenshot 2025-03-19 at 9 24 30 AM
> ### Cited By tab
> Screenshot 2025-03-19 at 9 24 57 AM

Yes, those changes are correct, are from this PR: https://github.com/freelawproject/courtlistener/pull/5224 which was merged yesterday

mlissner:
Sounds like we're good to go here??

ERosendo (approved):
> Sounds like we're good to go here??

Yes, this PR looks good. 🚀 

My apologies, @quevon24. I've just noticed that there's another PR specifically addressing the 'Cited by' tab.

Following our Slack discussion and @mlissner's suggestion, we can tweak the `es_get_related_clusters_with_cache` method to estimate the unique document count rather than calculating the exact count. This would improve performance. However, I recommend implementing that change in a separate PR to keep this one focused.

mlissner:
Thanks all. This will be a nice improvement.

PR #5262, opened 2025-03-14
feat(stats): Updates health check endpoint

### LINKED ISSUE #5259: Remove Elastic from health check endpoint, move it to its own
The /monitoring/health-check endpoint was failing yesterday because Elastic was down. When that healthcheck fails, pods get restarted and then nothing works. 

Let's remove Elastic from that health check. The site (and particularly the API) can be healthy even when Elastic isn't. 

We already have a separate healthcheck for Redis, so let's make one that's similar for Elastic, and we can monitor it separately.

ERosendo:
This PR fixes #5259 by moving the Elasticsearch health check logic from `/health-check` to a dedicated `/elasticsearch-status` endpoint.

### DIFF
--- a/cl/stats/urls.py
+++ b/cl/stats/urls.py
@@ -2,6 +2,7 @@ from django.urls import path
 
 from cl.stats.views import (
     celery_fail,
+    elasticsearch_status,
     health_check,
     redis_writes,
     replication_status,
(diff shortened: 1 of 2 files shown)


mlissner (approved):

mlissner:
Great. One solution landed. Thank you.

PR #5272, opened 2025-03-18
feat(templates): add note about data source and quality on docket filters

### LINKED ISSUE #4991: Add note to bottom of dockets explaining sources and update quality
I've been getting more inbound emails with confused questions about why various dockets aren't up to date. A simple thing we should have done years ago is to put a notice at the bottom of the docket entries that explains where the content comes from, encourages installation of RECAP (with link), provides a link to download the docket, links to the RECAP coverage page, and says that they're not necessarily up to date. Whew, that's three links, but it's all helpful, and we should see if we can cram it all in there. 

This would only need to go on cases where `date_terminated=None`.

elisa-a-v:
Closes #4991 

Adds a notice about data sources and update quality to docket pages in the filters card.


cl/opinion_page/templates/includes/de_filter.html
```diff
@@ -108,4 +108,48 @@
...
   </div>
+
+{% if docket.date_terminated is None %}
+  <p style="margin-top: 10px;">
+    <i class="fa fa-info-circle"></i> Docket entries on this page are sourced from PACER and may not be up to date. Learn more about our coverage <a href="{% url 'coverage_recap' %}">here</a>.
+  </p>
+  <p style="margin-bottom: 0;">
+    Notice an issue with the data displayed? Please <a href="{% url 'contact' %}">let us know</a>, and to further help improve public access to court records, consider <a href="https://free.law/recap/" class="alert-link">installing RECAP</a>—it's quick, easy, and free!
+  </p>
+{% endif %}
```
mlissner, line 119: Now that I'm seeing this, I think we should just show it on all pages, and maybe simplify it a bit to:

```suggestion
<p style="margin-top: 10px; margin-bottom: 0;">
  <i class="fa fa-info-circle"></i> Items on this page are sourced from <a href="https://www.pacer.gov">PACER</a> and <a href="https://free.law/recap/">RECAP</a> and may not be up to date. <a href="{% url 'coverage_recap' %}">Learn more about our coverage</a>.
</p>
```

That gets it on one line (I think), that we can just put on all the pages.

mlissner (changes requested):
Yeah, I don't love adding CSS to the page itself (and I actually wonder if our CORS policy even allows it), but if it works, that's fine by me since you're right, it's a short term thing, and it'll be your problem to deal with soon enough! 

But I have to say I don't love the new table and mobile sizes. The way they center the widgets instead of distributing them bugs me, particularly with this new note added. 

Can we revert that and either make no tweaks to the widgets or instead just make minimal ones? I don't want to spend time on this since we're dropping it soon, and I don't want to change it unless it's definitely better, since we'll be changing it a lot soon. The pagination buttons in this header aren't a huge deal if they're missing b/c they're at the bottom of the page anyway. They're a fairly recent addition to the filter bar, actually, so I feel OK with them being omitted sometimes.

Thank you!

mlissner (approved):

mlissner:
It's a tricky little spot to get right. It's eaten my time too. Merging!

PR #5274, opened 2025-03-19
feat(opinions.html): Add citation_spans property and update opinions template

flooie:
Add citation_spans property and update opinions template

- Added `citation_spans` property to wrap citations in `<span class="citation">` tags for better front-end handling.
- Updated `opinions.html` template to use `citation_ spans` instead of `citation_string` for improved styling and interaction.
- Added corresponding CSS styles to `opinions.css` to enhance citation display and enable user selection.

https://github.com/user-attachments/assets/0da3ed11-239b-423b-b991-cb33c3b9e888


mlissner (commented):
Thanks Bill and glad to see you're feeling better. Two quick things:

1. Why do the HTML in your view instead of in the template? 
2. Why add a CSS rule for this instead of using the `select-all` class we already have?

mlissner:
Yeah, generally it's best not to have HTML in our views, so let's do the ordering in the view and then HTML-wrap in the template.

For the CSS, yes, let's upgrade the override CSS so everything benefits.

Thanks!

flooie:
Thanks @mlissner back to you

PR #5278, opened 2025-03-19
Return estimate count for similar cases and cited by tabs

quevon24:
This PR fixes the problem that the tab count is different from the count shown in the advanced search.

Cardinality aggregation was added to have an approximate count and be close to what is seen in the advanced search.

Also take advantage of this PR to slightly refactor the view that updates the tabs and remove a print that was placed in the code.

### DIFF
--- a/cl/custom_filters/templatetags/extras.py
+++ b/cl/custom_filters/templatetags/extras.py
@@ -367,6 +367,7 @@ def build_docket_id_q_param(request_q: str, docket_id: str) -> str:
     return f"docket_id:{docket_id}"
 
 
+@register.filter
 def humanize_number(value):
     """Formats a number into a human-readable abbreviated form
 


ERosendo (approved):
@quevon24 The code looks good.

 I noticed there's another suggestion about number formatting in the issue, are you planning to address that suggestion from @mlissner in a separate PR?

quevon24:
> @quevon24 The code looks good.
> 
> I noticed there's another suggestion about number formatting in the issue, are you planning to address that suggestion from @mlissner in a separate PR?

thanks eduardo, which formatting issue? i missed that one

ERosendo:
@quevon24 Here's the suggestion: 

> We have some code to convert numbers like 20,203 into 20k, and things like that. I think any value over a couple hundred probably can be approximated?

quevon24:
> @quevon24 Here's the suggestion:
> 
> > We have some code to convert numbers like 20,203 into 20k, and things like that. I think any value over a couple hundred probably can be approximated?

Thanks, i'll address that in this PR, let you know when i'm done

quevon24:
The changes are ready @ERosendo 

Example humanizing counts:

image

For the image above:

```
authorities_count = 1050
cited_by_count = 329
summaries_count = 10987
related_cases_count = 10
```

i couldn't find the code to convert numbers in the template tags files, so i added a new template tag named `humanize_number`

cl/custom_filters/templatetags/extras.py
```diff
@@ -365,3 +365,54 @@ def build_docket_id_q_param(request_q: str, docket_id: str) -> str:
...
+            return f"{int(num)}M"
+        else:
+            return f"{num:.1f}M"
+    else:
+        # Format as billions
+        num = value / 1000000000
+        if num == int(num):
+            return f"{int(num)}B"
+        else:
+            return f"{num:.1f}B"
```
mlissner, line 418: Seeing this, I think you're right that we didn't have this code before.

This might be a good opportunity to use Python's weird underscore trick:

```suggestion
    if value < 1_000:
        # No formatting for numbers less than 1_000
        return str(value)
    elif value < 1_000_000:
        # Format as thousands
        num = round(value / 1_000)
        # Check if the number is a whole number
        if num == int(num):
            return f"{int(num)}K"
        else:
            return f"{num:.1f}K"
    elif value < 1_000_000_000:
        # Format as millions
        num = value / 1_000_000
        if num == int(num):
            return f"{int(num)}M"
        else:
            return f"{num:.1f}M"
    else:
        # Format as billions
        num = value / 1_000_000_000
        if num == int(num):
            return f"{int(num)}B"
        else:
            return f"{num:.1f}B"
```

quevon24: i didn't know that, thanks, it makes it more readable

mlissner: It's somewhat recent and a favorite of mine. I think I can't count zeros as well as other people or something because without the underscores I feel crazy!

cl/custom_filters/templatetags/extras.py
```diff
@@ -365,3 +365,54 @@ def build_docket_id_q_param(request_q: str, docket_id: str) -> str:
     if request_q:
         return f"({request_q}) AND docket_id:{docket_id}"
     return f"docket_id:{docket_id}"
+
+
+@register.filter
+def humanize_number(value):
```
ERosendo, line 371: @quevon24 I've found that the function isn't working as expected in certain scenarios. I've documented some test cases that reveal these issues, for example: 

```
>>> humanize_number(1050)
'1K' # Expected: '1.1K' (according to the docstring)

>>> humanize_number(999_999)
'1000K' # Expected: '1M'

>>> humanize_number(999_999_999)
'1000.0M' # Expected: '1B'
```

Here's my suggestion to refine the filter function: 

```python
@register.filter
def humanize_number(value):
    try:
        num = float(f"{value:.3g}")
    except (TypeError, ValueError):
        return value

    if num < 1_000:
        # No formatting for numbers less than 1_000
        return str(value)

    magnitude = 0
    while abs(num) >= 1000:
        magnitude += 1
        num /= 1000.0

    abbreviation = [
        "",
        "K",
        "M",
        "B",
    ][magnitude]
    rounded_num = floatformat(num, 1)
    return f"{rounded_num}{abbreviation}"

```

We also need to add some tests to ensure we don't introduce regressions later on.

quevon24: Thanks, i'll take a look at it and add some tests for the tag

quevon24:
@ERosendo I made a few changes, i liked your magnitude approach, i  made a few changes to solve the issue i commented to you related to convert 1.0K to 1K

Also added a test case

mlissner:
If you're done here, Kevin, should we assign back to Eduardo? I'm noticing the tests all failed for some reason.

quevon24:
> > I'm noticing the tests all failed for some reason.
> 
> @quevon24 removed the `@register.filter` decorator in [6816af5](https://github.com/freelawproject/courtlistener/pull/5278/commits/6816af5f19891ac7a472069e4b4e4c3a486a1d8e)

Yes, i didn't see that, i pushed a change to bring back that decorator

ERosendo (approved):
LGTM

mlissner:
Cool! Faster backend, nicer front end.

PR #5283, opened 2025-03-20
chore(tailwind): config tailwind with new branding

### LINKED ISSUE #5281: Add typography and colors from figma to tailwind config

elisa-a-v:
Closes #5281 

### Before:

image

### After:

image

### Figma

image


cl/simple_pages/templates/v2_help/index.html
```diff
@@ -8,11 +8,11 @@
 
 
 {% block content %}
-  <div class="hidden-xs col-sm-2 col-md-3"></div>
+  <div class="text-sm-semibold text-primary-600">COURTLISTENER</div>
```
mlissner, line 11: We can use CSS to do the uppercasing here. It's better because:

1. It can be copy/pasted and the user will get CourtListener instead of COURTLISTENER (neat). 

2. Developers can't turn it off once it's componentized. All eyebrows will be uppercase as enforced by the component.

mlissner (commented):
OK, this is exciting. A few thoughts:

1. Take a look at Big Cases' CSS file: https://github.com/freelawproject/bigcases2/blob/main/bc/web/static_src/src/styles.css

    I think there are a few lessons we can borrow from there, including fallback fonts (woff2 -> woff -> eot -> svg), swapping in the font, etc.

2. Did you consider componentizing this or do you think it's easier to do that in a second PR? What we want to avoid is winding up copy/pasting things like `text-sm-semibold text-primary-600` every time we need an "eyebrow" (that's unfortunately what the red uppercase header is called).

Thank you!

elisa-a-v:
Thanks for the feedback! This is still a small step but now it's starting to look a bit more like the new design, so I agree it's exciting!

> Take a look at Big Cases' CSS file

Thank you for pointing this out, I'll add the fallbacks and font swapping—anything else that comes to mind?

I also noticed the `@layer base` there—maybe we could style the HTML headers in that directive so instead of adding the `text-h1-bold` class to whatever HTML tag, we could simply use `<h1>`, but I don't think we can do the same with the rest of the typography classes, so it might be better to just keep all typography definitions in the same place and be consistent throughout, so I'm not sure. Maybe @leohentschker has some insight here?


> Did you consider componentizing this or do you think it's easier to do that in a second PR? What we want to avoid is winding up copy/pasting things like text-sm-semibold text-primary-600 every time we need an "eyebrow" (that's unfortunately what the red uppercase header is called).

Oh I was only focusing on styles here, specifically colors and typography, so I hadn't thought of that but yeah that makes perfect sense to me! I think we should do it when we get to the components for the help page though, so we keep this PR within scope.

mlissner:
I don't think anything else jumps out at me in the Big Cases file, no. 

I actually don't understand the @layer base stuff anymore (if I ever did). Maybe could you educate me on the different approaches with code examples?

elisa-a-v:
Well, what I've gathered is basically this:

- Tailwind uses layers to generate the styles in the correct order when compiling the CSS.
- There's `base`, `components` and `utilities` layers:
	- The base layer is the first one to be added to the compiled CSS, so everything else can override these. It's meant to be used to style **raw HTML** tags directly.
	- The components layer is meant to be used to declare **classes** that combine multiple CSS properties that belong together to style something like cards or buttons or other components. They appear after the base styles in the compiled CSS.
	- Then there's the utility layer, which is basically for **single-purpose utility classes**. These can have multiple CSS properties but they should all be closely related to each other (exactly like the typography classes I added, so now that I'm reading this, I think this is actually where those classes should be). These come last in the compiled CSS, so they can override the previous styles if needed.
- Technically we *could* use a single layer (or no layer at all?) and I think it would probably work, but this layering system is meant to keep things organized and to help developers understand the order in which the styles can be overridden—this can be very important to avoid surprises cause CSS overrides can be a pain to figure out sometimes when a style isn't taking effect.

A simple example could be something like this:
```css
@tailwind base;
@tailwind components;
@tailwind utilities;

@layer base {
  h1 {
    @apply text-4xl font-bold text-gray-800 text-shadow-sm;
  }
  div {
    @apply bg-white;
  }
}

@layer components {
  .btn {
    @apply px-4 py-2 font-bold bg-blue-500 hover:bg-blue-600 text-white;
  }
}

@layer utilities {
  .text-shadow-md {
    text-shadow: 0 2px 4px rgba(0,0,0,0.1);
  }
}
```

Which is the `input.css` file that Tailwind uses to compile the final CSS. Note that the `h1` style includes a `sm` shadow class, so if we do `<h1>Header!</h1>`, it'll have a small shadow, but if we add the `text-shadow-md` class like so: `<h1 class="text-shadow-md">Header!</h1>`, then we can be sure that the utility class will override the default `h1` small shadow with a slightly bigger one.

Same with the div: if we simply use `<div>something...</div>`, it'll have a white background, but if we add the `btn` class, it'll be overridden and have a blue background. If we added another utility class that sets the background color, then that one would override both previous ones.

Is that clear enough?

mlissner:
Thanks Leo, that makes sense. The problem, it would seem, is that layers make one more place where the tweaks to default HTML might be stored, so you're not sure if the thing is {bold/large/small/etc} because of component classes, layers, or something else?

elisa-a-v:
Thanks for your input, @leohentschker ! Although now your comment makes me think I may have misunderstood everything :sweat_smile: 

> However if you need to go beyond just the core elements, you run into an odd situation in which there is a new vector of change that impacts styles outside of what is in the classname.

I'm honestly not sure I fully understand the issue here, nor what your suggestion for this PR is. This might be my lack of experience with tailwind showing, but from what I've read, I thought this was exactly the type of problem that layering is trying to prevent. Core elements already have default styles anyway, so using the base layer overrides those, and it's like we have our own set of branded native HTML. Why would it then be an issue to have classes on top of that when we need to customize the look in specific situations?

With this in mind, when we talked about locking in the new branding styles, I thought we were talking about not just configuring tailwind to use our custom colors, but also adding the classes for the different typographies used in Figma. I chose classes instead of overriding the native HTML elements because I couldn't see a 1-1 mapping, so I thought it was best to be consistent and always use classes to style text. And from what I read, I thought adding classes in tailwind should always be done via layering anyway.

Could you please expand on your suggestions so I know how to move forward with this particular PR? Thanks!

elisa-a-v:
> layers make one more place where the tweaks to default HTML might be stored

Oh I was under the impression that this was the ONLY place we'd do any custom CSS at all: the `input.css` file with the layers and nowhere else. Isn't that tailwind's recommended approach? This would keep all our custom CSS in one place, with a clear hierarchy so we would know which styles take precedence.

mlissner:
> Oh I was under the impression that this was the ONLY place we'd do any custom CSS at all

Oh, I'm not sure, honestly, but it does seem important to keep such things in one place. I *think* in Big Cases we use override.css for that, maybe, but I dig the layers concept. I'd say let's land whatever approach you like best, unless @leohentschker has a moment to chime in, and then we can change it later if it makes us sad.

leohentschker:
I'm adding unnecessary confusion here! Post review of the above, I support moving forward with layers, and I think this is definitely a two-way door.

elisa-a-v:
I'd say this is now ready. I finally changed the approach with typographies, got rid of the classes and instead configured tailwind to use our sizes and widths. You can see an example of the change in use in [this commit](https://github.com/freelawproject/courtlistener/pull/5283/commits/908cbfe758bb22018a32d112dac71d21d557d74b), in the `index.html` file. I also changed the HTML structure there a little bit so it now resembles the figma design a bit more.

mlissner (approved):
Looks good. Let's land it, and if we arent' totally happy with it later, it's easy to change. 

Thank you!

PR #5287, opened 2025-03-20
Tweaks to summaries tab on new opinions UI

quevon24:
Update summaries tab o new opinions UI, display summary first, then date, court name, case name, court and "Show X summaries like this"
Add summaries explanation

image


mlissner:
I see this is not related to your change, but the groups shouldn't be indented. We *do* currently indent the members of the group though: 

image

Otherwise the screenshot looks good to me!

flooie:
Yeah - I agree with @mlissner  just need to remove the indent.

mlissner:
SHIP IT!!!

PR #5306, opened 2025-03-25
feat(django-cotton + alpine) Enabling Django Cotton components with Alpine

leohentschker:
**TLDR:** Enables adding custom components via django cotton + alpine
* Enables [alpine js](https://alpinejs.dev/) for lightweight, client-side, secure interactivity 
* Enables [django-cotton](https://django-cotton.com/) for re-usable, client-side components

### Preview
image


cl/assets/static-global/js/alpinejscsp@3.14.8.min.js
mlissner, line 1: Is there a non-minimized version that we can show when dev is True so that devs can debug, or is there a whatchamacallit file to make sense of the minified one?

poetry.lock
mlissner, line 1: Lots of funny changes in here. I suspect we may need to regenerate it with 2.1.x to make it match. (We should really switch to `uv`)

cl/simple_pages/templates/components.html
```diff
@@ -6,6 +6,8 @@
 {% tailwind_css %}
 {% tailwind_preload_css %}
 
+<script src="{% static 'js/alpinejscsp@3.14.8.min.js' %}" defer init nonce="{{ request.csp_nonce }}"></script>
```
mlissner, line 9: Hm, do we want this here or in new_base.html? I guess if we put it here we can just remember to include it only on the pages that need it; maybe that's better if we don't need it on a lot of pages.

cl/simple_pages/templates/cotton/h1.html
mlissner, line 1: I think you and Elisa have a different vision on how to style an H1, which is sort of great. I'm +1 on merging both and then playing and deciding as we go, so long as we remember to evaluate it very soon, before we married to both.

cl/assets/static-global/js/alpinejscsp@3.14.8.min.js
leohentschker: Yes! Will do.

poetry.lock
mlissner: @elisa-a-v, can you help with this, please?

elisa-a-v: Sure! Leo and I actually dicussed this earlier today, he will let me know if he gets stuck and I can take it from there :)

elisa-a-v (approved):
The non-minified version of Alpine has been added, and I created a new lock file using the correct poetry version.

I think this is ready to be merged!

mlissner:
Looks pretty good, but usually we'll have an if/else in our template to pick the minified or the regular. Want to add that and then merge?

mlissner (approved):

elisa-a-v:
> usually we'll have an if/else in our template to pick the minified or the regular. Want to add that and then merge?

Oh I didn't add it because we only had it in the `components.html` file, but now that I think about it, I'll move it to the new base template with the if/else, because I don't think we'll actually have any pages that don't use it at all 🤔 so I think we should have the script in the new base template _only_ (I think having it repeatedly could create conflicts with different inits, but I'm not sure), and if we ever need to get rid of it, we can wrap it in a DTL block then override it in the one page that doesn't need it.

Does that make sense?

mlissner:
I can't imagine needing to get rid of it. It's harmless since the user will have it cached, so I say, yeah, let's if/else the dev version and make it part of the default.

PR #5307, opened 2025-03-25
Fix broken mobile view

quevon24:
- Fix disabled pdf tab
- Fix behavior when clicking disabled tab (changed # to #tab-id, we need the link to keep the styles even if it is disabled)
- Display download pdf button next to tab header name (useful on iphone/ipad because embedded viewer doesn't work, user can download the pdf easily)
- Display "no pdf available" in case the url is accessed even when the tab is disabled

PDF tab fixed:
image

PDF tab responsive:
image

New download button:
image

Responsive:
image


mlissner:
Looks good visually. I'll leave it to others for full review!

flooie:
I dont know @mlissner - it looks pretty wonky in iOS and iPadOS 

We still have the squished image and it's confusing that you cant use the PDF.  I think we need to either use PDF.js to enable PDF embedding on the tab or drop the image/pdf on smaller devices.  Or... perhaps fix the thumbnail to be the correct size and make it a link on smaller devices to open teh PDF viewer.  

That former vs the latter I'm not sure.  

@quevon24 


For reference on iPadOS we retain the squished thumbnail 

Simulator Screenshot - iPad Pro 11-inch (M4) - 2025-03-31 at 12 35 17

Im not necessarily arguing for PDF.js ... it's a bit ... wonky in its own right but this doesnt seem to be the right solution.

mlissner:
Yeah, on mobile we can just show a friendly error and a download button. It's not like they get anything useful by seeing a thumbnail or even from pdf.js.

flooie:
are you grouping tablets into mobile?

mlissner:
Yeah, I think so. Neither really benefits from an embedded PDF viewer.

quevon24:
Now it will show a message on mobile devices with download button:

image

And if for some uknown reason the pdf viewer in your browser is disabled, it will show a different message with the download button:

image

flooie (approved):
Looks good to me

PR #5309, opened 2025-03-26
fix(alerts): Tweaks the SearchAlertSerializer's validate method

### LINKED ISSUE #5083: Validation error creating an alert via API
This should be a really easy one. I think we just need to make sure that the `query` parameter is required. This has only happened once, but it's worth fixing just to continuously make our APIs better.

----

Sentry Issue: [COURTLISTENER-94K](https://freelawproject.sentry.io/issues/6305979166/?referrer=github_integration)

```
AttributeError: 'NoneType' object has no attribute 'encode'
(8 additional frame(s) were not displayed)
...
  File "cl/alerts/api_serializers.py", line 30, in validate
    qd = QueryDict(attrs.get("query").encode(), mutable=True)
```

ERosendo:
This PR fixes #5083 by updating the validate method from the SearchAlertSerializer to allow patch requests.

### DIFF
--- a/cl/alerts/tests/tests.py
+++ b/cl/alerts/tests/tests.py
@@ -661,6 +661,74 @@ class AlertAPITests(APITestCase, ESIndexTestCase):
         search_alert = Alert.objects.all()
         self.assertEqual(await search_alert.acount(), 0)
 
+    async def test_can_validate_required_fields(self) -> None:
+        # Test creation with missing required fields:
+        response = await self.client.post(self.alert_path, {}, format="json")
+
+        self.assertEqual(response.status_code, HTTPStatus.BAD_REQUEST)
+        self.assertIn("This field is required.", response.json()["name"])
+        self.assertIn("This field is required.", response.json()["query"])
+        self.assertIn("This field is required.", response.json()["rate"])
+
+        # Create a valid alert for subsequent update testing:
+        alert_1 = await self.make_an_alert(
+            self.client,
+            alert_name="alert_1",
+            alert_query=f"q=testing_query&type={SEARCH_TYPES.OPINION}",
+        )
+        alert_1_data = alert_1.json()
+        # Try to update an alert using a PUT request with missing required
+        # fields:
+        alert_1_path_detail = reverse(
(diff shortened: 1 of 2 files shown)


cl/alerts/tests/tests.py
```diff
@@ -661,6 +661,74 @@ async def test_invalid_alert_type_fail(self) -> None:
...
+            self.client,
+            alert_name="alert_1",
+            alert_query=f"q=testing_query&type={SEARCH_TYPES.OPINION}",
+        )
+        alert_1_data = alert_1.json()
+        # Try to update an alert using a PUT request with missing required
+        # fields:
+        alert_1_path_detail = reverse(
+            "alert-detail",
+            kwargs={"pk": alert_1_data["id"], "version": "v3"},
```
mlissner, line 684: Hmmm, should this be v4 instead?
```suggestion
            kwargs={"pk": alert_1_data["id"], "version": "v4"},
```

cl/alerts/tests/tests.py
```diff
@@ -661,6 +661,74 @@ async def test_invalid_alert_type_fail(self) -> None:
...
+
+        # Verify initial alert state:
+        self.assertEqual(alert_1_data["alert_type"], SEARCH_TYPES.OPINION)
+        search_alert = Alert.objects.all()
+        self.assertEqual(await search_alert.acount(), 1)
+
+        # Construct the detail URL for the alert:
+        alert_1_path_detail = reverse(
+            "alert-detail",
+            kwargs={"pk": alert_1_data["id"], "version": "v3"},
```
mlissner, line 710: And here too: 

```suggestion
            kwargs={"pk": alert_1_data["id"], "version": "v4"},
```

If you see other v3's, seems like we should nuke 'em.

mlissner (approved):
Just a couple tweaks, I think, please, then merge, thank you!

cl/alerts/tests/tests.py
ERosendo: You're absolutely right. I'll tweak this right away.

PR #5313, opened 2025-03-26
Fix error when returning estimate count for similar cases tab

quevon24:
Implement early return for es_related_case_count function when the cluster doesn't have opinions.

In the code of the funtion, we use a function to build an ES "more like this" query which requires a list of sub opinions pks.

There are clusters without opinions, for example:

http://www.courtlistener.com/opinion/3561702/go/
http://www.courtlistener.com/opinion/3560425/go/
http://www.courtlistener.com/opinion/3895596/go/
http://www.courtlistener.com/opinion/3310252/go/

This doesn't affect the other function used to estimate cited by count because it doesnt use a "more like this" query


ERosendo (approved):
LGTM :shipit:

Issue #4724, opened 2024-11-21
Deprecate Supreme Court Citation Visualizations


mlissner:
More than ever, we need to focus on what we can do exceptionally well. I think it's time to deprecate [citation visualizations](https://www.courtlistener.com/visualizations/scotus-mapper/). 

Here's why:

1. We haven't upgraded them in years.
2. In 2024, there were 25 published visualizations created by 16 users, totaling 2675 views.
3. The champion of this project was @Colinstarger, but he hasn't created a viz for exactly two years as of today.
4. We haven't been adding the needed SCOTUS short names or citations to make these good for several years, with no complaints.
5. We're redesigning the site and it's one less thing to redesign if we drop it first.

Here's how:

7. I'm reaching out to Colin to make sure he's OK.

2. Contact the folks that created visualizations in the last two years to apologize and let them know they won't be able to create new ones.

3. Keep the existing visualizations. They're static, and there's no need to delete them. Some are embedded in other websites.

3. Remove the link from the Case Law drop down.

4. Redirect the viz homepage

5. Close all viz-related bugs.

6. Rip out the viz created code and dependencies.

7. Remove these pages:

     - https://www.courtlistener.com/visualizations/scotus-mapper/new/
     - https://www.courtlistener.com/visualizations/gallery/
     - https://www.courtlistener.com/visualizations/scotus-mapper/ --> This can redirect to an FAQ that provides a short explanation of them.

8. Remove them from the homepage.

9. Find other places they exist and adapt.

10. Add a little banner on the top of each visualization saying that the project is archived.

And I think that's it? What else?

mlissner:
OK, turns out there's still demand for this, so it's not a full deprecation, but:

 - We'll remove the front end code as above, with redirections. No more HTML related to visualizations, except perhaps an FAQ saying they're deprecated or something like that.
 - We'll maintain the visualization API, and make sure it works for creating new ones.
 - We'll maintain the embedding webpage, so that visualizations can be embedded.

The final result will be an API folks can use to create visualizations, some documentation for it, and the ability to embed them into other webpages.

jbgolds:
> We'll maintain the embedding webpage, so that visualizations can be embedded.

For clarity @mlissner , this is the page we want to still keep, right?  https://www.courtlistener.com/visualizations/scotus-mapper/4829/embded/?

Wondering as want to know which templates I can delete / keep.

mlissner:
No, it's this: 

[...]

PR #5334, opened 2025-04-01
Remove old opinions ui

quevon24:
This PR introduces these changes:

- Remove templates of old opinions ui
- Remove views of old opinions ui
- Update related tests to frontend opinions ui

We can delete the `ui_flag_for_o` flag in admin panel


quevon24:
@flooie This is almost ready, just one question. Are visualizations still used? there is no link to them in the new ui. e.g. https://www.courtlistener.com/opinion/107252/miranda-v-arizona/visualizations/

mlissner:
We're deprecating them, so this is fine: #4724

flooie:
@quevon24 things look and work the same, so I think this is good to go when @mlissner gives the ok.

cl/opinion_page/urls.py
```diff
@@ -35,11 +34,6 @@
         name="court_publish_page",
     ),
     # Opinion pages
-    path(
-        "opinion/<int:pk>/<blank-slug:slug>/visualizations/",
-        cluster_visualizations,  # type: ignore[arg-type]
-        name="cluster_visualizations",
-    ),
```
mlissner, line 42: I think this PR is fine, except I don't think we should remove URLs that used to work. Can we add a redirect for this in urls.py (with a date and comment, like the others)? The redirect can go to the API documentation for visualizations: 

https://www.courtlistener.com/help/api/rest/v4/visualizations/#deprecation-notice

And we should add a section at the bottom saying that they're deprecated. The section can say something like (feel free to make this better):

> Unfortunately, our system for visualizing Supreme Court networks has not gotten much traction among users, and is partially deprecated as of early 2025. 
> 
> If you are interested in creating, deleting, or updating visualizations, you can still do so through our APIs, but it is no longer possible to display visualizations on CourtListener.com itself. Moving forward, to support existing users, the only way to display visualizations is through their `embed` links. 
> 
> To embed a visualization on a website you control, use code like the following on your site:
> ```
> <iframe height="540" 
>         width="560" 
>         src="https://www.courtlistener.com/visualizations/scotus-mapper/YOUR_ID_HERE/embed/" 
>         frameborder="0" 
>         allowfullscreen>
> </iframe>
> ```
> 
> Just replace `YOUR_ID_HERE` with the ID of your visualization and it should work on your website.
> 
> We apologize for this deprecation and hope you understand that we cannot always maintain all the features and experiments we undertake.

Does that feel about right? I'd just like anybody that has this link in their bookmarks or whatever to understand what's happening.

mlissner (approved):
This looks good to me and I'm excited to get it merged (nothing better than deleting code!), but I think we need a smoother path whenever we remove a URL. Once that's in, let's merge!

cl/opinion_page/urls.py
quevon24: It sounds good, so that users don't see a 404 page

mlissner: Exactly, yeah.

quevon24: Redirection has been added and deprecation notice has been added to api help page.

Navitating to https://www.courtlistener.com/opinion/107252/miranda-v-arizona/visualizations/ will take you to https://www.courtlistener.com/help/api/rest/visualizations/#deprecation-notice

image

cl/opinion_page/urls.py
```diff
@@ -34,6 +35,11 @@
         name="court_publish_page",
     ),
     # Opinion pages
+    path(
+        "opinion/<int:pk>/<blank-slug:slug>/visualizations/",
+        visualizations_deprecation_redirect,
+        name="cluster_visualizations",
+    ),
```
mlissner, line 42: Check out cl/urls.py, where we usually put our redirections. I think you can use the approach there to avoid having a view too.

cl/api/templates/visualizations-api-docs-vlatest.html
```diff
@@ -75,6 +75,18 @@ <h2 id="editing">Editing and Deleting Visualizations</h2>
   <p>Similar commands can be used to update other non-readonly fields.</p>
   <p>To soft-delete a visualization, flip the <code>deleted</code> field to <code>True</code>. To hard-delete, send an HTTP <code>DELETE</code> request.
   </p>
+
+  <h2 id="deprecation-notice">DEPRECATION NOTICE</h2>
```
mlissner, line 79: I don't think we need to use caps lock, so:

```suggestion
  <h2 id="deprecation-notice">Deprecation Notice</h2>
```

mlissner (changes requested):
A couple little things, but looks good!

cl/opinion_page/urls.py
quevon24: Thanks, good suggestion, I'm so used to doing it the Django way that I didn't think about this 😅

cl/api/templates/visualizations-api-docs-vlatest.html
quevon24: The changes are ready 👍

mlissner (approved):
Let's gooo! Bye, bye old stuff!

quevon24:
> Let's gooo! Bye, bye old stuff!

Flag `ui_flag_for_o` deleted 👍

PR #5390, opened 2025-04-14
fix(SLUGIFIED_EDITIONS): now using collision aware slugify

### LINKED ISSUE #5359: Slugifying editions discards some reporters
This affects endpoints where `cl.citations.utils.SLUGIFIED_EDITIONS` is used
https://github.com/freelawproject/courtlistener/blob/7d48ec7cef614a724533fced58abd17f2c13bae9/cl/citations/utils.py#L14-L16

- `cl.opinion_page.views.citation_redirector` 
- `cl.citations.api_views`

repeated slug 'cit' 'CIT' 'C.I.T.'
repeated slug 'day' 'Day.' 'Day'
repeated slug 'la' 'La.' 'LA'
repeated slug 'mspb' 'MSPB' 'M.S.P.B.'
repeated slug 'me' 'Me.' 'ME'
repeated slug 'nd' 'ND' 'N.D.'
repeated slug 'nm' 'NM' 'N.M.'
repeated slug 'pa' 'Pa.' 'PA'
repeated slug 'sd' 'SD' 'S.D.'
repeated slug 'vt' 'Vt.' 'VT'

Found while [researching](https://github.com/freelawproject/juriscraper/issues/1150#issuecomment-2790896524) `Vt.` vs `VT` , which are actually different reporters

To reproduce:

```python
from reporters_db import EDITIONS, VARIATIONS_ONLY
from django.template.defaultfilters import slugify

In [13]: mapper = {}
    ...: for item in EDITIONS.keys():
    ...:     slug = str(slugify(item))
    ...:     if slug in mapper:
    ...:         print(f"repeated slug '{slug}' '{item}' '{mapper[slug]}'")
    ...:     mapper[slug] = item
```

--------------

### Solution

Fixing this will require updating the slugify function so that it keeps points `"."`? Perhaps we can do it only when there is a repeated key in the `SLUGIFIED_EDITIONS` dictionary to prevent introducing to much noise to the URLs

grossir:
Fixes #5359

Use a collision aware slugify function
To resolve the collision I am using the `REPORTERS[reporter][cite_type]` (which resolve to 'state', 'neutral', 'federal', etc; because other options had drawbacks
- using `reporter.replace(".", "-")` would work for "C.I.T." but not for "Vt.", because Django's `slugify` would delete the trailing hyphen
- using the   `REPORTERS[reporter][name]` may produce very big names . For example for `'M.S.P.B.'` it would be the long `Decisions of the United States Merit Systems Protection Board` slug


cl/opinion_page/tests.py
```diff
@@ -500,6 +499,23 @@ async def test_slugifying_reporters(self) -> None:
         )
         self.assertEqual(r.url, "/c/f2d/56/9/")
 
+    async def test_slugifying_reporters_collision(self) -> None:
```
quevon24, line 502: The test looks good, but now that you have identified the repeated slugs, maybe you could add them as test pairs like:
```

test_pairs = [("Vt.", "VT"), ("La.", "LA"), ("Me.", "ME")]

for r1, r2 in test_pairs:

	response1 = await self.async_client.get(
            reverse(
                "citation_redirector",
                kwargs={"reporter": r1},
            )
        )
    response2 = await self.async_client.get(
        reverse(
            "citation_redirector",
            kwargs={"reporter": r2},
        )
    )
    self.assertNotEqual(response1.url, response2.url)


```

quevon24 (commented):
What a good way to create slugs to differentiate the reporters 👍

Just a small suggestion related to the test

cl/opinion_page/tests.py
grossir: Thanks Kevin, I added these examples

PR #5398, opened 2025-04-15
fix(favorites): make sure granted prayers aren't marked as sealed

v-anne:
Minor modification to `send_prayer_emails` in `cl/favorites/utils.py`.

In `cl/favorites/tasks.py`, we already make sure to mark documents as not sealed if we discover they are available for purchase, but that function is only called when someone requests a document as part of the pray-and-pay project. Now, if someone goes directly to PACER to purchase a document and the RECAP extension uploads it to the repository, we confirm that it is not sealed and deletes any corresponding rows from the `PrayerAvailability` model.

I wasn't sure if this belonged here or in `handle_recap_doc_change` in `cl/search/signals.py`.

Edit: one question I have is whether it's possible for a particular RECAPDocument to be marked as `is_available = True` and `is_sealed = True`. Those two statuses seem contradictory.

### DIFF
--- a/cl/favorites/signals.py
+++ b/cl/favorites/signals.py
@@ -1,6 +1,6 @@
 from datetime import timedelta
 
-from django.db.models.signals import post_save
+from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
 from django.utils.timezone import now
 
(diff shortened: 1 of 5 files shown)


mlissner:
@v-anne, we established a new system in our latest sprint retro to prioritize volunteer work into either fast or slow lanes. Fast ones we allow into the current sprint, and slow ones get slated for future sprints. You can read more about this in the dev guide. See number 6, here: https://github.com/freelawproject/courtlistener/wiki/Getting-Started-Developing-CourtListener#guidelines-for-contributions

This one, since it's just an optimization, is a good candidate for the slow review, and I've slated it for our next sprint (known as the Eminem sprint; we're in Daft Punk now; they're alphabetical).

So look for a review "soon", but it may take a bit....

ERosendo:
> do you know the answer to this?

@v-anne I don't have the exact answer about how the `is_sealed` field is handled every time a new document gets ingested from all the various sources we have (_I do know that the `save_pacer_doc_from_pq` function – that's what we use when we get documents from recap.email – definitely updates that field_). However, in [d7baddd](https://github.com/freelawproject/courtlistener/pull/5398/commits/d7badddb013d203d88eb280f7c965c3f5d714d57).  I've implemented a signal that triggers upon the deletion of a `PrayerAvailability` record. This signal double-checks and, if necessary, corrects the `is_sealed` status of the associated `RecapDocument`, which helps keep our data consistent.

ERosendo:
> After this is merged, can someone run a quick script to delete RECAPDocuments from the PrayerAvailability model if is_available = True? I doubt many are affected at this point but it would still be a good thing to do.

@mlissner , when you get a chance, could you please run this script?  

```python
from cl.favorites.models import PrayerAvailability

PrayerAvailability.objects.filter(recap_document__is_available=True).delete()
```

mlissner:
Done: 

```
PrayerAvailability.objects.filter(recap_document__is_available=True).delete()
(23, {'favorites.PrayerAvailability': 23})
```

Issue #4998, opened 2025-01-30
Redesign all remaining help pages


mlissner:
We'll do one help page first, in https://github.com/freelawproject/courtlistener/issues/4740. In this issue, we'll do the rest of the help pages.

elisa-a-v:
The pages that have the same "Help page" structure are as follows:
- [ ] Docket and search alerts
- [ ] @recap.email help page
- [ ] Tags and notes help page
- [ ] Markdown help page
- [ ] Coverage help pages
    - [ ] RECAP
    - [ ] Case law
    - [ ] Financial disclosures
    - [ ] Oral arguments
- [ ] API help pages
- [ ] Contribute help page
- [ ] Terms and policies

Issue #4739, opened 2024-11-26
Identify and make components for basic help page


mlissner:
Once we have a component tool identified, we need to figure out which components are needed for a basic help page and add them to /components/.

elisa-a-v:
#### Help page components:
- [ ] nav menu (side menu in desktop, dropdown selector in mobile)

    <details>
    <summary>Nav menu desktop</summary>

    ![Image](https://github.com/user-attachments/assets/3dffe285-846f-4174-901b-d80782d1696b)
    </details>

- [ ] tabbed cards

    <details>
    <summary>Tabbed cards desktop</summary>

    ![Image](https://github.com/user-attachments/assets/aea21837-f4bc-4c94-8ac7-d59d167e3fd8)
    </details>

    <details>
    <summary>Tabbed cards mobile</summary>

    ![Image](https://github.com/user-attachments/assets/88d4b8ed-e44f-424d-9840-cc6648887dba)
    </details>

- [ ] code blocks (probably only needs tailwind styling for `<code>` tags)

    <details>
    <summary>code block</summary>

    ![Image](https://github.com/user-attachments/assets/85fafd0e-1976-4d5e-9a34-60826d5d52d3)
    </details>

- [ ] table

    <details>
    <summary>Table</summary>

    ![Image](https://github.com/user-attachments/assets/516a959d-a3ae-45be-8c25-24b08c2ac31b)
    </details>



- [ ] "Listen up!/Pro tip!" card (probably only needs a tailwind utility class)

    <details>
    <summary>"Listen up!/Pro tip!" card</summary>

    ![Image](https://github.com/user-attachments/assets/d4b3d701-bc9c-4a8f-a560-3722bfb3ef4b)
    </details>

PR #5400, opened 2025-04-15
feat(components): add tabbed cards and navigation menu components

elisa-a-v:
#4998
#4739

## Changes

- Add several new components:
    - Navigation menu
    - Tabbed cards
    - Code block
    - Two-column list
    - Eyebrow
    - Layout with navigation menu
- Add new template tag to inject SVG inline.
- Add new template tag to register component scripts at the end of the `<body>`.
- Update Alerts help page with new components.
- Update components page:
    - Extend from new base template.
    - Include and describe new components.

#### 1.1. Tabbed cards

The `title` is only used in ARIA attributes, it's meant to add more context to assisting technologies.

The code for this component was heavily influenced by django-cotton's [docs](https://django-cotton.com/docs/alpine-js).

<details>
<summary>Tabbed cards screenshots</summary>

#### Mobile

image

#### Desktop

image

</details>

###### Screenshot code

```html
<c-tabs title="My tabs">
  <c-tabs.tab name="First tab">
    First tab panel content
  </c-tabs.tab>
  <c-tabs.tab name="Second tab">
    Second tab panel content
  </c-tabs.tab>
  <c-tabs.tab name="Third tab">
    Third tab panel content
  </c-tabs.tab>
</c-tabs>
```

#### 1.2. Navigation menu

For highlighted items, the parent template needs to implement the `intersect.js` Alpine component, and add `x-intersect="show"` to the elements referenced.

This uses Alpine's intersect plugin, so the `x-intersect` directive can be used with [modifiers](https://alpinejs.dev/plugins/intersect#modifiers) for more control; see `x-intersect.margin.-100px="show"` in the new alerts help page, which adds a negative margin to the intersect handler, preventing the event to be fired immediately when the element enters the viewport.

It is recommended to reference the container for the entire section instead of the headings otherwise the dynamic highlighting will not work as expected.

The code for the mobile dropdown select was based on Tailwind Plus UI Blocks [Simple Custom Select Menu](https://tailwindcss.com/plus/ui-blocks/application-ui/forms/select-menus).

<details>
<summary>Navigation menu screenshot</summary>

#### Desktop

image

</details>

###### Screenshot code

```html
<c-navigation-menu
  :nav_items="[
    {'href': '#first', 'text': 'First item'},
    {'href': '#second', 'text': 'Second item'},
    {'href': '#with-children', 'text': 'With children', 'children': [
      {'href': '#first-child1', 'text': 'First child'},
      {'href': '#second-child1', 'text': 'Second child'},
      {'href': '#third-child1', 'text': 'Third child'},
    ]},
    {'href': '#another', 'text': 'Another item'},
    {'href': '#more-children', 'text': 'More children', 'children': [
      {'href': '#first-child2', 'text': 'First child'},
      {'href': '#second-child2', 'text': 'Second child'},
      {'href': '#third-child2', 'text': 'Third child'},
    ]},
    {'href': '#last', 'text': 'Last'},
  ]"
></c-navigation-menu>

<div id="first"><h2>First item</h2></div>
```

#### 2.1. Require script

To keep the template clean and avoid duplicated code and potential conflicts, we introduce a template tag that injects all required scripts at the end of the `<body>`. This is especially useful for reusable components that may be repeated throughout the template.

###### Usage

Simply add the following at the top of the component template:

```
{% load component_tags %}
{% require_script "path/to/script.js" %}
```

#### 2.2. Inject svg

Similar to `includes` but more explicit, and it also supports Alpine bindings, which require the use of `:`.

###### Usage

```html
{% svg "name" x_bind__class="iconClass" %}
```

Will render the svg saved in static files (`svg/name.svg`) like so:

```html
<svg x-bind:class="iconClass" ...>
```

---

### Alert help page

<details>
<summary>Desktop</summary>

[Screencast from 04-17-2025 03:33:42 PM.webm](https://github.com/user-attachments/assets/e7f280b0-6e60-466a-958b-13293a1e47bb)
</details>

<details>
<summary>Mobile</summary>

[Screencast from 04-17-2025 03:36:39 PM.webm](https://github.com/user-attachments/assets/49e84665-943b-4176-b798-50118dabfed0)
</details>

### Updated components page

<details>
<summary>Desktop screenshot</summary>

image

</details>

<details>
<summary>Mobile screenshot</summary>

image

</details>

<details>
<summary>Desktop demo</summary>

[Screencast from 04-21-2025 06:29:23 PM.webm](https://github.com/user-attachments/assets/540083ad-ed82-4ba4-b997-1f0a54e12848)

</details>


cl/custom_filters/templatetags/component_tags.py
```diff
@@ -0,0 +1,46 @@
...
+        nonce = f' nonce="{request.csp_nonce}"'
+
+    script_tags = []
+    for script_path in request._required_scripts:
+        script_url = static(script_path)
+        script_tags.append(
+            f'<script type="text/javascript" src="{script_url}"{nonce}></script>'
+        )
+
+    return mark_safe("\n".join(script_tags))
```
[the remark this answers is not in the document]
elisa-a-v: Each tag is now generated using `format_html` so they are indeed safe. Do we really need to change this?

cl/assets/templates/cotton/code.html
```diff
@@ -0,0 +1,3 @@
+<c-vars class=""></c-vars>
+
+<pre class="w-full max-w-full {{ class }}" {{ attrs }}><div class="w-full px-6 bg-greyscale-50 rounded-[10px] border border-greyscale-300 overflow-auto"><code class="w-full border-none bg-greyscale-50 text-greyscale-800 font-mono">{{ slot }}</code></div></pre>
```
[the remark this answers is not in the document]
elisa-a-v: See [Pass all attributes with {{ attrs }}](https://django-cotton.com/docs/components) in Django Cotton's docs, we're only passing HTML attributes added in the parent on to an element in the component, so this should be safe.

cl/assets/templates/cotton/tabs/index.html
```diff
@@ -0,0 +1,57 @@
+{% load component_tags %}
+{% require_script "js/alpine/components/tabs.js" %}
+{% require_script "js/alpine/components/focus.js" %}
+<c-vars title="" class=""></c-vars>
+
+<div
+  x-data="tabsComponent"
+  class="w-full {{ class }}"
+  {{ attrs }}
```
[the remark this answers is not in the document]
elisa-a-v: Same

cl/assets/templates/cotton/eyebrow.html
```diff
@@ -0,0 +1,2 @@
+<c-vars class=""></c-vars>
+<div {{ attrs }} class="text-sm font-semibold text-primary-600 uppercase {{ class }}">CourtListener{{ slot }}</div>
```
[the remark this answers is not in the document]
elisa-a-v: Same

cl/custom_filters/templatetags/svg_tags.py
```diff
@@ -0,0 +1,57 @@
...
+            key = key.replace("__", ":")  # Convert double underscore to colons
+            key = key.replace("_", "-")  # Convert snake_case to kebab-case
+            svg_content = svg_content.replace("<svg", f'<svg {key}="{value}"')
+
+        return format_html(svg_content.replace("<svg", f'<svg role="img"'))
+
+    except FileNotFoundError:
+        if settings.DEBUG:
+            return f"SVG '{name}' file found but couldn't be opened at '{absolute_path}'"
+        return ""
```
mlissner, line 57: I'd reorganize this to put the try/except together. Something more like:

```suggestion
    try:
        with open(absolute_path, "r", encoding="utf-8") as file:
            svg_content = file.read()
    except FileNotFoundError:
        if settings.DEBUG:
            return f"SVG '{name}' file found but couldn't be opened at '{absolute_path}'"
        return ""
        
    css_class = kwargs.pop("class", None)
    if css_class:
        if 'class="' in svg_content:
            svg_content = svg_content.replace(
                'class="', f'class="{css_class} '
            )
        else:
            svg_content = svg_content.replace(
                "<svg", f'<svg class="{css_class}"'
            )

    for key, value in kwargs.items():
        key = key.replace("__", ":")  # Convert double underscore to colons
        key = key.replace("_", "-")  # Convert snake_case to kebab-case
        svg_content = svg_content.replace("<svg", f'<svg {key}="{value}"')

    return format_html(svg_content.replace("<svg", f'<svg role="img"'))
```

cl/assets/templates/new_base.html
```diff
@@ -98,11 +98,17 @@ <h1>You did not supply the "private" variable to your template.
...
+  <script src="{% static 'js/alpine/alpinejscsp@3.14.8.js' %}" defer nonce="{{ request.csp_nonce }}"></script>
   {% else %}
   <script type="text/javascript"
-        src="{% static "js/jquery/jquery-3.4.1.min.js" %}"></script>
-  <script src="{% static 'js/alpinejscsp@3.14.8.min.js' %}" defer init nonce="{{ request.csp_nonce }}"></script>
+          src="{% static "js/jquery/jquery-3.4.1.min.js" %}"></script>
+  <script src="{% static 'js/alpine/intersect@3.14.8.min.js' %}" defer nonce="{{ request.csp_nonce }}"></script>
+  <script src="{% static 'js/alpine/collapse@3.14.8.min.js' %}" defer nonce="{{ request.csp_nonce }}"></script>
+  <script src="{% static 'js/alpine/focus@3.14.8.min.js' %}" defer nonce="{{ request.csp_nonce }}"></script>
+  <script src="{% static 'js/alpine/alpinejscsp@3.14.8.min.js' %}" defer nonce="{{ request.csp_nonce }}"></script>
```
mlissner, line 111: Did you consider only including the JS files if they are needed? Seems like with your require_scripts trick, some pages might get away without needing *all* of these, and that'd be good.

mlissner (approved):
Looks pretty good to me, but let's get a full review from Eduardo to see what he sees. 

A few quick thoughts/questions...

Tabbed component: 
 - The content within the tab should line up horizontally with the left edge of the first header. I think you can see that in this screenshot, where the bullet lines up almost exactly below the word "District": 

    image

Navigation menu:
 - Looks phenomenal.

Do we have components for the collapsable sections you're using for code on the components page, or documentation for it? I'm not sure I saw it.

cl/simple_pages/templates/v2_help/alert_help.html
```diff
@@ -7,325 +7,288 @@
...
+      {'href': '#creating-search-alerts', 'text': 'Creating Alerts'},
+      {'href': '#enabling-real-time-alerts', 'text': 'Enabling Real Time Alerts'},
+      {'href': '#editing-search-alerts', 'text': 'Editing/Deleting an Alert'},
+      {'href': '#disabling-search-alerts', 'text': 'Disabling an Alert'},
+      {'href': '#courts', 'text': 'Supported Courts'},
+    ]},
+    {'href': '#citation-alerts', 'text': 'Citation Alerts'},
+    {'href': '#coming-soon', 'text': 'Coming Soon'},
+  ]"
+>
```
ERosendo, line 30: I've noticed a rendering issue with this component in Safari. Here's a screenshot: 

image

ERosendo: I also noticed something a little inconsistent with the navigation bar's highlight. The attached GIF shows that while clicking an anchor correctly directs the user to the intended section, the highlight sometimes incorrectly shifts to the next anchor in the bar instead of staying on the one that was clicked.

Screen Recording 2025-04-23 at 5 25 05 PM


I suspect this issue happens when more than one section is visible on the screen at the same time. Scrolling up and down is also affected, but it's not as noticeable


https://github.com/user-attachments/assets/e639721a-dce1-49fb-a740-6a0b74c17966

cl/assets/templates/cotton/eyebrow.html
ERosendo, line 2: While I haven't reviewed all the redesign pages yet, I have a concern about adding a fixed string to this component's content. For example, if I wanted to display 'Big cases bot,' it seems 'courtlistener' would always be added to the slot. Perhaps we should avoid hardcoding this to allow for more flexible content.

cl/simple_pages/templates/v2_help/alert_help.html
elisa-a-v: it's always something with safari 👎🏽 

And yeah the behavior is slightly off precisely when multiple sections are visible at the same time, as the last to enter the viewport will be highlighted.

I haven't been able to come up with a simple enough solution that is worth implementing other than tweaking the margins, which helps a bit but doesn't solve it entirely. This is in a follow-up branch that I haven't pushed yet, but you can swap all `x-intersect.margin.-100px="show"` with `x-intersect.margin.-20%.0.-65%.0="show"` and it looks much better imo.

I think improving this any further would require implementing the intersect plugin from scratch, but I may be missing something. Do you have any ideas?

cl/assets/templates/cotton/eyebrow.html
elisa-a-v: I didn't think too much about it as I think most if not all of the uses will just say "CourtListener", but making this a bit more flexible wouldn't hurt.

What do you say I refactor this so instead of having a hardcoded value with a slot, it'll have a prop `text` with a default value of "CourtListener", so we can override it if necessary, but will show "CourtListener" otherwise which is what we'll want most of the time. Sound good?

cl/assets/templates/new_base.html
elisa-a-v: I didn't give it that much thought but now that you mention it yeah that's a good point and I think it't entirely possible to use the new template tag to include Alpine plugins too, not just our custom components.

I'd have to move the `render_required_scritps` in the base template above the Alpine script as Alpine plugins need to be included before it, but you're right I can just use that to include these too.

cl/assets/templates/cotton/two_column_list.html
```diff
@@ -0,0 +1,11 @@
+{% load static humanize partition_util %}
```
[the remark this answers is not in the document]
elisa-a-v: Oops! No they are not 😬 I'll remove them 🫡

cl/assets/templates/cotton/eyebrow.html
ERosendo: I'd prefer that we always explicitly specify the component's content. I find that more clear and less potentially confusing than having a default value. 

```html
<c-eyebrow>Courtlistener</c-eyebrow>
```

cl/simple_pages/templates/v2_help/index.html
```diff
@@ -10,7 +10,7 @@
 {% block content %}
   <div class="pb-11">
     <div class="flex flex-col gap-4 pr-13">
-      <div class="text-sm font-semibold text-primary-600 uppercase">CourtListener</div>
+      <c-eyebrow></c-eyebrow>
```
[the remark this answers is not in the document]
elisa-a-v: Oh! I see, well I'm not sure I agree, 100% of the instances of the eyebrow in the design say "CourtListener", I think that's the purpose of the component.

I don't think it's a good idea to leave it up to the devs, as they could easily do the same think I initially did which was to add the text all uppercase, instead of `CourtListener`, which will look uppercase anyway due to CSS, but if people copy the text, it will still be "CourtListener". I'm pretty sure we discussed this before, @mlissner , and you suggested adding the component precisely so we wouldn't give devs the chance to do it wrong.

cl/assets/templates/cotton/eyebrow.html
elisa-a-v: Oh! Well I strongly disagree for the reasons explained in [this comment](https://github.com/freelawproject/courtlistener/pull/5400#discussion_r2057037839). For just any other component I'd be totally on board, explicit is better than implicit, but in this case, the purpose of the component is to ensure consistency.

elisa-a-v:
> Do we have components for the collapsable sections you're using for code on the components page, or documentation for it? I'm not sure I saw it.

Oh I forgot to include it, yes we have a component and no, it's not documented in the library. I'll add it!

cl/simple_pages/templates/v2_help/index.html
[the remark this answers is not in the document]
ERosendo: > 100% of the instances of the eyebrow in the design say "CourtListener"

wait, those small red headings are eyebrows, right? Is the Figma design outdated, or am I missing something?


Screenshot 2025-04-23 at 7 50 06 PM

cl/simple_pages/templates/v2_help/alert_help.html
ERosendo: I've been experimenting with the `navigation_menu.html` file and noticed that removing the `-mt-7` class from one of the divs within it seems to improve rendering in Safari.

image

ERosendo: > I think improving this any further would require implementing the intersect plugin from scratch, but I may be missing something. Do you have any ideas?

I don't have any immediate suggestions for the highlight behavior issue. It's possible others are experiencing this too; perhaps there's a bug report in the code repository. 🤔

cl/simple_pages/templates/v2_help/index.html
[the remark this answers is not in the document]
elisa-a-v: 🤦🏽 how did I not see those!! okay I take some of that back.

I'm still inclined to leave the "CourtListener" as default for the reasons described above using a prop, so it'd either be
```html
<c-eyebrow></c-eyebrow>
```
for "CourtListener" or
```html
<c-eyebrow text="Some other text"><c-eyebrow>
```

For anything else.

Now if I'm the only one pushing for the default and we prefer making it explicit every time, which also makes sense to me, we should probably go for a slot as @ERosendo described [here](https://github.com/freelawproject/courtlistener/pull/5400#discussion_r2057032622).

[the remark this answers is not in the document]
mlissner: It should be explicit. If it says CourtListener most of the time, that's probably a fault in the design or the designers not having better ideas. I definitely don't think of it as a default kind of thing. It's really a kind of header, so it shouldn't have any kind of default.

cl/assets/templates/new_base.html
mlissner: Sounds great.

cl/simple_pages/templates/v2_help/index.html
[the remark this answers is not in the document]
elisa-a-v: Okay, explicit it is. I guess I got the wrong idea, my bad!

elisa-a-v:
Thank you both for your feedback! I'll address all of your comments and let you know once it's ready.

ERosendo:
Thanks @elisa-a-v

mlissner:
I leave it to you, @ERosendo!

cl/simple_pages/templates/components.html
```diff
@@ -1,21 +1,276 @@
...
+    <h4 class="mt-3">Demo</h4>
+    <c-eyebrow class="my-4">CourtListener</c-eyebrow>
+
+    <h4 class="mt-3">Props</h4>
+    <p>No props</p>
+    <h4 class="mt-3">Slot</h4>
+    <div class="bg-greyscale-100 rounded-xl p-4 my-3">
+      <div class="flex flex-col md:flex-row border-b last:border-none border-greyscale-200 max-w-full">
+        <span class="p-2.5 max-md:pb-0 md:w-60 md:min-w-60 max-md:font-medium">Default</span>
+        <span class="p-2.5 max-md:max-w-full [overflow-wrap:anywhere]">Text to be appended.</span>
```
ERosendo, line 117: The text is no longer appended within the eyebrow. let's update this description

cl/assets/templates/cotton/navigation_menu.html
```diff
@@ -0,0 +1,155 @@
+{% load svg_tags static component_tags %}
+{% require_script "js/alpine/components/navigation_menu.js" %}
+{% require_script "js/alpine/components/intersect.js" %}
```
ERosendo, line 3: The import path for the `intersect.js` file is incorrect.

```suggestion
{% require_script "js/alpine/composables/intersect.js" %}
```

cl/assets/templates/cotton/tabs/index.html
```diff
@@ -0,0 +1,59 @@
+{% load component_tags %}
+{% require_script "js/alpine/components/tabs.js" %}
+{% require_script "js/alpine/components/focus.js" %}
```
ERosendo, line 3: the path for the `focus.js` file seems to be incorrect as well 😅 

```suggestion
{% require_script "js/alpine/composables/focus.js" %}
```

ERosendo:
Thanks @elisa-a-v. The changes are looking great and the nav component is rendering properly in Safari 🥳. I just have three minor comments for us to address.

> I plan to address the wonky highlight in a later PR where I'll be extending the nav menu component a little bit by adding a sub-component for the sections which will handle highlighting a bit better. I'd rather not add that to this PR as it's already big enough IMO, wouldn't you agree?

Absolutely agree!. Let's create an issue to track that highlight issue. Also, I noticed a few warnings and an Alpine error in the browser console – could we take a quick look at those too?? 

image

elisa-a-v:
@ERosendo thank you so much for your thorough review and for your help tackling those warnings!

Paths are updated, warnings are gone and I also noticed we weren't requiring the `focus` composable in the nav menu so I [added that](https://github.com/freelawproject/courtlistener/pull/5400/commits/444d2b6b6b9d26648b5e709233e3998d027a2513). It was only working because so far all pages with the nav menu also have the `tabs` component which is the one that was importing it. We now don't rely on other components importing it, and make sure the nav menu always has the focus composable available :)

Hopefully this will be the final review, la tercera es la vencida! 😅 thanks again for all your feedback!!

ERosendo (approved):
LGTM :shipit:

PR #5410, opened 2025-04-16
feat(favorites): increase pray-and-pay quota for FLP members

### LINKED ISSUE #5383: Allow FLP members to make a higher number of PACER document requests
Similar to higher quotas for docket alerts, FLP members should also be able to make more requests as part of the pray-and-pay project.

v-anne:
This PR is straightforward. It resolves #5383 by allowing Free Law Project members to make three times the daily requests as part of the pray-and-pay project as regular users, who are currently allotted five requests. Under the quota in this PR, FLP members should now be able to make 15 requests in a 24-hour period.

This form and any other webpage that lists benefits for FLP members will need to be updated after this PR is merged.
https://donate.free.law/forms/membership

### DIFF
--- a/cl/simple_pages/templates/help/prayer_help.html
+++ b/cl/simple_pages/templates/help/prayer_help.html
@@ -102,7 +102,7 @@
   <h3 id="limitations">Limitations</h3>
   <p>Each user can place <span class="bold">{{ daily_quota }}</span> prayers per day. This prevents people from praying for every item on the site.
   </p>
-  <p>Members can place 15 prayers per day.
+  <p>Members can place <span class="bold">{% widthratio daily_quota 1 3 %}</span> prayers per day.
   </p>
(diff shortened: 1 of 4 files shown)


mlissner (commented):
Simple enough! Can we also add a Join FLP button to this section, like we have on the docket alerts page, so people can easily solve their problem if they're reading the docs? E.g.: 

image

cl/simple_pages/templates/help/prayer_help.html
```diff
@@ -102,8 +102,14 @@ <h3 id="creating-prayer">Creating a Prayer</h3>
   <h3 id="limitations">Limitations</h3>
   <p>Each user can place <span class="bold">{{ daily_quota }}</span> prayers per day. This prevents people from praying for every item on the site.
   </p>
+  <p>Members can place 15 prayers per day.
```
ERosendo, line 105: Let's use the `daily_quota` variable instead of hardcoding the number. Additionally, I suggest we apply a consistent highlighting style to the maximum prayer count for members
```suggestion
  <p>Members can place <span class="bold">{% widthratio daily_quota 1 3 %}</span> prayers per day.
```

ERosendo (approved):
Looks great! Just one small suggestion before merging.

ERosendo:
@v-anne I've pushed a commit to fix the failing tests

ERosendo:
> Do you think it's necessary to add a new test that checks to make sure regular users aren't being covered by the FLP members-only quota?

I already addressed this while fixing the failing tests. I adjusted the test_prayer_eligible test to ensure users with memberships receive a higher allowance.

PR #5421, opened 2025-04-17
feat(favorites): Improve prayer availability check for PDF responses

ERosendo:
This PR enhances the `check_prayer_pacer` task to more robustly handle responses when checking prayer availability. Specifically, it now explicitly checks if the response is a PDF document.

Previously, the task might have incorrectly flagged a prayer as unavailable if a PDF was returned directly (without the usual download confirmation page), as this PDF content is not a valid page for parsing. This update ensures that a PDF response is no longer treated as an indication of unavailability.

Here's the [code](https://github.com/freelawproject/juriscraper/blob/2c7a9014a96707623635002461f52fba23c0de22/juriscraper/pacer/download_confirmation_page.py#L49-L53) from Juriscraper that marks pdf responses as invalid: 

```python
        if is_pdf(self.response):
            # Sometimes the PDF document is returned without showing the
            # download confirmation page, not a valid page to parse.
            self.is_valid = False
            return
```

### DIFF
--- a/cl/favorites/tasks.py
+++ b/cl/favorites/tasks.py
@@ -3,6 +3,7 @@ from django.contrib.auth.models import User
 from django.utils.timezone import now
 from juriscraper.lib.exceptions import PacerLoginException
 from juriscraper.pacer import DownloadConfirmationPage
+from juriscraper.pacer.utils import is_pdf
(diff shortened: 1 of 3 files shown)


mlissner (approved):
Nice one. Merging.

PR #5438, opened 2025-04-18
fix(favorites): only sort by date of last_checked in get_top_prayers (not time as well)

### LINKED ISSUE #5436: Pray-and-pay community leaderboard may be incorrectly ranking documents
I think the community leaderboard is incorrectly ranking documents after #5231 and #5367 were merged.

Consider this:

We first sort by whether a document is unavailable, with available documents coming first. Then, we sort by when the documents were last checked, with documents that were last checked less recently coming first as well.

After that, we sort by how many requests a document has. If we're doing that, [Mahdawi v. Trump (2:25-cv-00389)](https://www.courtlistener.com/docket/69888582/mahdawi-v-trump/) should come ahead of the other documents last checked on April 17.

My guess is that this isn't happening because `get_top_prayers` is currently sorting by last_checked values down to the exact minute and second, and that may be confusing because we are only displaying the particular date. Thoughts? If that hypothesis is correct, maybe we don't need to change anything yet, but I wanted to flag.

Also, I am confused about why every document except one was last checked yesterday. Did we wipe the `PrayerAvailability` table of all previous checks before rerunning the script for existing documents?

Image

v-anne:
This PR resolves #5436. While the existing code is not a bug per se, I don't think it is functioning as intended. It seems like it will confuse users to sort by both date and time if only the date is publicly visible to them.

### DIFF
--- a/cl/favorites/utils.py
+++ b/cl/favorites/utils.py
@@ -166,7 +166,7 @@ async def get_top_prayers() -> QuerySet[RECAPDocument]:
                 When(prayeravailability__id__isnull=False, then=Value(True)),
                 default=Value(False),
             ),
-            last_checked=TruncDate("prayeravailability__last_checked"),
+            last_checked=F("prayeravailability__last_checked__date"),
         )
         .order_by(
             "doc_unavailable", "last_checked", "-prayer_count", "-view_count"
--- a/cl/favorites/utils.py
+++ b/cl/favorites/utils.py
@@ -1,6 +1,5 @@
 from dataclasses import dataclass
 from datetime import timedelta
-from typing import Optional
 
 from asgiref.sync import sync_to_async
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


mlissner:
You're on fire!

cl/favorites/utils.py
```diff
@@ -167,7 +166,7 @@ async def get_top_prayers() -> QuerySet[RECAPDocument]:
                 When(prayeravailability__id__isnull=False, then=Value(True)),
                 default=Value(False),
             ),
-            last_checked=F("prayeravailability__last_checked"),
+            last_checked=TruncDate("prayeravailability__last_checked"),
```
ERosendo, line 169: Instead of the `TruncDate` class, we can achieve the same result more directly by accessing the date part of the `prayeravailability__last_checked` field using:

```suggestion
            last_checked=F("prayeravailability__last_checked__date"),
```

Issue #5326, opened 2025-04-01
Make formatting of prayer notification emails identical to docket alerts


v-anne:
I've noticed that the docket alert emails seem to use larger fonts than the prayer notification emails. I know Mike wants to change the templates entirely, but for the time being I think the prayer email template should be adjusted to align with the longstanding alert format.

PR #5440, opened 2025-04-18
fix(favorites): adjusting pray-and-pay email templates to be consistent with docket alerts

v-anne:
Resolving #5326.


ERosendo:
LGTM :shipit:

PR #5441, opened 2025-04-18
fix(donate): Tweak logic to check active status on termination date

### LINKED ISSUE #5425: Membership Renewal: Triggering "Become a member" message one day before expiry.
A user has reported that CL incorrectly prompts them to get a membership to set up more alerts on the very last day their current membership is active.

ERosendo:
This PR fixes the `is_active` check to correctly handle memberships on their termination date.  By comparing only the date portion of the termination date and current time, we ensure that memberships are valid until the end of the day on which they terminate.

Fixes #5425

### DIFF
--- a/cl/donate/models.py
+++ b/cl/donate/models.py
@@ -256,4 +256,4 @@ class NeonMembership(AbstractDateTimeModel):
         if not self.termination_date:
             return True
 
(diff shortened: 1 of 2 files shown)


mlissner (approved):
Nice, thank you.

PR #5454, opened 2025-04-22
feat(opinions): use ordering_key when matching opinions

quevon24:
Use opinions instead of clusters when looking up for citations. Now we use ordering_key value to sort the possible results.


flooie (approved):
Looks good to me - @quevon24

Issue #5458, opened 2025-04-24
Update `get_citations` calls and tests to new `eyecite` version


grossir:
The new eyecite version has added (among many other changes)
- a `clean_steps` argument to `get_citations`
- an internal `Document` object that handles the SpanUpdaters for XML/HTML sources. 

So, we need to 
- update cleanup code (mostly delete it)
- update how we access the `source text` and `clean text` (now, it will be from the Document)
- We can take advantage of this to prevent creating another SpanUpdater in the annotation step, and just access the one in the `Document`. This should speed up things a little

Some of these will require small changes to `eyecite`, that we are just noticing when trying to do the integration

PR #5460, opened 2025-04-24
fix(get_citations): update calls and tests for `eyecite.get_citations`

grossir:
Solves #5458

- Rename `make_markup_kwargs` to `make_get_citations_kwargs`
- Use new `xml` clean step
- Delete `get_and_clean_opinion_text`
- Change `create_cited_html` to use the `document` object instead of attributes previously set by  `get_and_clean_opinion_text`
- Update tests
   - deleted Thompson test because the bug no longer exists
   - updated some annotation tests that have changed


grossir:
Currently this is using an eyecite branch
`docker exec -it cl-django poetry add git+https://github.com/freelawproject/eyecite.git#cl-270-integration-tweaks`

Once the eyecite branch is merged into main and released , we can update to the proper version

flooie:
This looks good to me when you update poetry

flooie:
this is great.

Issue #5385, opened 2025-04-13
Automatically recheck prayer availability after ~90 days


v-anne:
A large fraction of the documents requested by users that are temporarily unavailable for purchase are transcripts. My understanding is that [they are generally under embargo for 90 days](https://pacer.uscourts.gov/help/faqs/are-transcripts-court-proceedings-available-pacer).

I think it would serve users well if we automatically rechecked availability of documents present in the `PrayerAvailability` model after ~90 days. If we fear it will take up too much resources as open requests pile up, we can do this just once for any particular document, as it would clear up the transcripts but avoid repeatedly checking other documents, such as those that are permanently sealed or text-only entries that can't be purchased using RECAP's "Buy on PACER" buttons.

mlissner:
This is a good idea, but kind of a pain right now. We don't have a way of doing something in 90 days unless we create a daemon or cronjob. We could do that, but I'm actually tempted to punt on this until my work in #5238 is complete. The option I'm looking at there actually makes it really easy to make a task that gets executed way down the road. 

If you think this is important, maybe a nightly cronjob is the way to do it, or maybe we can add it to an existing one (like the one that hits the iquery endpoint for lots of items)?

mlissner:
I think this could fit nicely in the pacer_iquery_scraper.py cronjob. Maybe add it there and tweak the file name to be more accurate? Perhaps nightly_pacer_updates.py or something?

PR #5462, opened 2025-04-25
feat(favorites): automating availability checks of RECAPDocuments

v-anne:
Completing #5385:

> A large fraction of the documents requested by users that are temporarily unavailable for purchase are transcripts. My understanding is that [they are generally under embargo for 90 days](https://pacer.uscourts.gov/help/faqs/are-transcripts-court-proceedings-available-pacer).

> I think it would serve users well if we automatically rechecked availability of documents present in the PrayerAvailability model after ~90 days. If we fear it will take up too much resources as open requests pile up, we can do this just once for any particular document, as it would clear up the transcripts but avoid repeatedly checking other documents, such as those that are permanently sealed or text-only entries that can't be purchased using RECAP's "Buy on PACER" buttons.

This PR resolves the issue by adding a function to the scraper to identify PrayerAvailability items that are stale and should be checked again. It does a one-time check at 91 days after a document was filed on PACER, as that is when embargoes for transcripts are commonly lifted (few other documents have an embargo like this). In the future, we can explore an exponential backoff solution, but I don't think it's necessary right now.


mlissner:
Placing in slow queue for review during next sprint. Thank you @v-anne !

cl/recap/management/commands/nightly_pacer_updates.py
```diff
@@ -242,4 +265,30 @@ def handle(self, *args, **options):
...
+
+            rd = (
+                RECAPDocument.objects.select_related("docket_entry__docket")
+                .only(
+                    "pacer_doc_id",
+                    "docket_entry_id",
+                    "docket_entry__docket_id",
+                    "docket_entry__docket__court_id",
+                    "is_sealed",
+                )
```
ERosendo, line 283: The code after this query shows we only need the `pacer_doc_id` and the recap document's id.

cl/recap/management/commands/nightly_pacer_updates.py
```diff
@@ -242,4 +265,30 @@ def handle(self, *args, **options):
...
+                .get(pk=rd_id)
+            )
+
+            pacer_doc_id = rd.pacer_doc_id
+            if pacer_doc_id == "":
+                continue
+
+            else:
+                check_prayer_pacer.delay(rd.pk, None)
```
ERosendo, line 292: This wont work. The current implementation of the `check_prayer_pacer` task needs a user ID. If the document isn't available, it calls the `prayer_unavailable` method to email the user.

cl/favorites/tasks.py
```diff
@@ -31,9 +31,16 @@ def check_prayer_pacer(self, rd_pk: int, user_pk: int):
...
     rd = (
         RECAPDocument.objects.select_related("docket_entry__docket")
-        .defer("plain_text")
+        .only(
+            "pacer_doc_id",
+            "docket_entry_id",
+            "docket_entry__docket_id",
+            "docket_entry__docket__court_id",
+            "is_sealed",
+        )
```
ERosendo, line 40: Is there a reason to only select those fields in the query?

If we end up calling `prayer_unavailable` to send the email, it'll actually cause more database queries down the line. That's because the email needs things like the `docket number`, uses the `best_case_name` function, and pulls other info from the docket entry and recap document models.

cl/recap/management/commands/nightly_pacer_updates.py
ERosendo: Perhaps we can use this PR to make the `user_id` optional and adjust the task's logic at the same time

ERosendo (changes requested):
@v-anne Thanks for this PR. 

I've added inline comments highlighting a few areas we need to address.

cl/recap/management/commands/nightly_pacer_updates.py
mlissner: If it's a system user you need, we sometimes use the RECAP user for stuff like this.

ERosendo: I think the `user_prayer` variable at line 395 will probably be `None`, which will cause a AttributeError exception .

ERosendo: I think that's alright. This approach should work:

```diff

def prayer_unavailable(instance: RECAPDocument, user_pk: int) -> None:
    open_prayers = Prayer.objects.filter(
        recap_document=instance, status=Prayer.WAITING
    ).select_related("user")

    user_prayer = open_prayers.filter(user__pk=user_pk).first()
    
+   if not user_prayer: 
+       return

    email_recipients = [
        {
            "email": user_prayer.user.email,
            "date_created": user_prayer.date_created,
        }
    ]

-  if email_recipients:
   ...
```

cl/favorites/tasks.py
```diff
@@ -29,11 +29,10 @@ def check_prayer_pacer(self, rd_pk: int, user_pk: int):
     :param rd_pk: The primary key of RECAPDocument of interest
     :param user_pk: The primary key of the user who requested the document
     """
-    rd = (
-        RECAPDocument.objects.select_related("docket_entry__docket")
-        .defer("plain_text")
-        .get(pk=rd_pk)
+    rd = RECAPDocument.objects.select_related("docket_entry__docket").get(
+        pk=rd_pk
```
ERosendo, line 33: The initial query didn't include the `plain_text` field. Typically, I only retrieve that field when it's needed.

cl/recap/management/commands/nightly_pacer_updates.py
```diff
@@ -77,6 +80,27 @@ def get_docket_ids_week_ago_no_case_name() -> set[int]:
     )
 
 
+def get_rd_ids_pray_and_pray() -> set[int]:
```
ERosendo, line 83:
```suggestion
def get_rd_ids_pray_and_pay() -> set[int]:
```

cl/recap/management/commands/nightly_pacer_updates.py
```diff
@@ -77,6 +80,27 @@ def get_docket_ids_week_ago_no_case_name() -> set[int]:
...
+    # Filter for documents where:
+    # - the original filing was at least 91 days ago (i.e., embargo likely lifted)
+    # - we have not checked them after the embargo passed
+    queryset = PrayerAvailability.objects.filter(
+        recap_document__docket_entry__date_filed__lte=cutoff_date,
+        last_checked__lt=F("recap_document__docket_entry__date_filed")
+        + timedelta(days=90),
+    )
+
+    return set(queryset.values_list("recap_document_id", flat=True))
```
ERosendo, line 101: How about we update this method to return the `RecapDocument` objects instead of just their IDs?

```python
def get_recap_documents_pray_and_pay() -> QuerySet[RECAPDocument]:
    """
    Retrieve a set of RECAP documents that should be checked for availability
    on PACER as part of the Pray and Pay project
    - This is focused on transcripts that have likely had their embargo
    lifted after the three-month period.
    - In the future, depending on outcomes, we can entertain implementing
    exponential backoff, but I don't think it's necessary right now.

    :return: A set of document matching the criteria.
    """
    cutoff_date = timezone.now() - timedelta(days=91)

    # Filter for documents where:
    # - the original filing was at least 91 days ago (i.e., embargo likely lifted)
    # - we have not checked them after the embargo passed
    return RECAPDocument.objects.filter(
        docket_entry__date_filed__lte=cutoff_date,
        prayeravailability__last_checked__lt=F("docket_entry__date_filed")
        + timedelta(days=90),
    ).only("pacer_doc_id")


```

ERosendo: We also need a test for this method. Here's a suggestion for the current implementation: 

```python
    def test_can_retrieve_old_prayers_to_check(self, mock_cookies):
        """Verifies retrieval of old, unchecked RECAP documents for 'pray and pay'."""
        # Create DocketEntry instances with different filing dates
        ten_days_old_entry = DocketEntryFactory(
            date_filed=now() - timedelta(days=10)
        )
        ninety_one_days_old_entry = DocketEntryFactory(
            date_filed=now() - timedelta(days=91)
        )
        ninety_five_days_old_entry = DocketEntryFactory(
            date_filed=now() - timedelta(days=95)
        )

        # Create RECAPDocument instances associated with the DocketEntry instances
        rd_1 = RECAPDocumentFactory(docket_entry=ten_days_old_entry)
        rd_2 = RECAPDocumentFactory(
            docket_entry=ninety_one_days_old_entry,
        )
        rd_3 = RECAPDocumentFactory(
            docket_entry=ninety_one_days_old_entry,
            document_type=RECAPDocument.ATTACHMENT,
            attachment_number=1,
        )
        rd_4 = RECAPDocumentFactory(
            docket_entry=ninety_five_days_old_entry,
        )
        rd_5 = RECAPDocumentFactory(
            docket_entry=ninety_five_days_old_entry,
            document_type=RECAPDocument.ATTACHMENT,
            attachment_number=1,
        )

        # Simulate PrayerAvailability records with different last_checked dates
        PrayerAvailability.objects.create(
            recap_document=rd_1, last_checked=now() - timedelta(days=3)
        )
        PrayerAvailability.objects.create(
            recap_document=rd_2, last_checked=now() - timedelta(days=2)
        )
        PrayerAvailability.objects.create(
            recap_document=rd_3, last_checked=now() - timedelta(days=1)
        )
        PrayerAvailability.objects.create(
            recap_document=rd_4, last_checked=now() - timedelta(days=6)
        )
        PrayerAvailability.objects.create(
            recap_document=rd_5, last_checked=now() - timedelta(days=2)
        )

        # Retrieve the RECAP documents that should be checked
        recap_ids = get_rd_ids_pray_and_pay()

        self.assertIn(rd_2.id, recap_ids)
        self.assertIn(rd_4.id, recap_ids)
        # Not old enough
        self.assertNotIn(rd_1.id, recap_ids)
        # Not expected: Last Checked on the same day it should have been unsealed
        self.assertNotIn(rd_3.id, recap_ids)
        # Not expected: Checked after it should have been unsealed
        self.assertNotIn(rd_5.id, recap_ids)
```

cl/recap/management/commands/nightly_pacer_updates.py
```diff
@@ -242,4 +265,30 @@ def handle(self, *args, **options):
...
             )
 
+        for i, rd_id in enumerate(get_rd_ids_pray_and_pray()):
+
+            if i % 10 == 0:
+                logger.info(
+                    "Sent %s RECAP Documents to celery for crawling so far.", i
+                )
+
+            rd = (
```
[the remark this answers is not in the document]
ERosendo: @v-anne , I've provided a suggestion in this comment: https://github.com/freelawproject/courtlistener/pull/5462/files#r2079949785 to update the `get_rd_ids_pray_and_pay` function to return recap documents instead of just the IDs. I believe this change should address your concern.

cl/recap/management/commands/nightly_pacer_updates.py
ERosendo: > These are the fields used in `check_prayer_pacer` as well. 

`check_prayer_pacer` is a Celery task, so it runs in a separate context. Instead of passing the entire object, we only pass the record ID. This allows the task to fetch the specific recap document it needs when it's executed by a Celery worker.

ERosendo: > Do you think it is inefficient to directly pass the record into check_prayer_pacer?

Yes! Passing just the ID is definitely the way to go. When we schedule tasks, we store the necessary context in Redis so our Celery workers can grab it later. If we start pushing full records, we'll fill up Redis memory in no time. Plus, think about what happens if we update a record after scheduling the task. The Celery worker would be working with outdated information. Fetching the record within the Celery task ensures we're always using the most current data.

ERosendo:
@mlissner This PR is ready to merge. Just a heads-up that we also need to update the cron job configuration because we're renaming the command file in this PR. I've opened a PR for that update here: https://github.com/freelawproject/kubernetes/pull/26

We can merge this PR once you have time to deploy the cron job update.

mlissner:
Nice. Automerge is set.

v-anne:
Is this working? It's been 3 months since the earliest documents were requested, but their availability hasn't been updated. Some of these documents are transcripts that should be available for purchase now.

https://www.courtlistener.com/docket/69607847/76/national-association-of-diversity-officers-in-higher-education-v-trump/
https://www.courtlistener.com/docket/68524040/58/hesai-technology-co-ltd-v-united-states-department-of-defense/
https://www.courtlistener.com/docket/69503243/142/in-re-paypal-honey-browser-extension-litigation/
https://www.courtlistener.com/docket/69757814/44/khalil-v-joyce/
https://www.courtlistener.com/docket/69754533/23/united-states-institute-of-peace-v-jackson/
https://www.courtlistener.com/docket/69752936/16/rferl-inc-v-lake/
https://www.courtlistener.com/docket/69845539/98/ozturk-v-hyde/

mlissner:
I just checked and the cronjob is indeed running. I guess we'd have to look at it more carefully to know what's up with these.

PR #5464, opened 2025-04-25
feat(citation.tasks): Need to account for no citations found

flooie:
Fix latest changes which didnt account for no citations.  

When no citations were found, we were crashing beacuse there was no document object to use to generate html with citations


quevon24 (approved):
LGTM!

PR #5465, opened 2025-04-25
fix(create_cited_html): case when no resolutions

grossir:
Solves bugs introduced in #5458

Handle case when no resolutions inside of `create_cited_html`

Add citations.tests for
- no citations
- no resolutions
- end to end store_opinion_citations_and_update_parentheticals when no citations will be found


flooie:
Thanks @grossir

PR #5470, opened 2025-04-28
feat(scrapers): Enhance scrape_oral_argument to transcribe audio

ERosendo:
While addressing comments on https://github.com/freelawproject/courtlistener/pull/5447,  I noticed that we're still running a separate `transcribe` command to get audio transcriptions. This adds an extra step when we do our monthly updates to our existing data. So, this pull request changes the `ingest_a_case` method to automatically start the `transcribe_from_open_ai_api` task after an audio file is processed. This makes sure that transcriptions are created as part of the regular process.


mlissner (approved):
Well, this is embarrassingly easy. Merging, thank you.

Issue #1798, opened 2021-10-27
Oral Argument Player Needs Speeds


brianwc:
The embedded audio player on the CL site only allows one to playback at normal speed. Unless I'm listening to music or a movie these days, I virtually always listen and watch at 1.5x to 1.75x speed. When one merely wants to get the information out of a source and is not engaged in aesthetic appreciation of it, this is a really commonplace practice, so the player should support it.

j-f1:
Would it be reasonable to get rid of the audio player all together and just embed an `<audio>` element? HTML5 and mp3 are supported on basically every browser, and recent versions of Chrome, Firefox, and Safari support changing playback rate natively.

mlissner:
Probably not a terrible idea. The other feature we hear requests for is being able to share a link at a particular moment, which our player doesn't support and I assume the built in ones don't either. Is there a page demoing the built in players somewhere?

j-f1:
If you go to the direct URL for an oral argument (like https://storage.courtlistener.com/mp3/2014/04/11/elrico_fowler_v._carlton_joyner_cl.mp3) you should get the built-in browser audio player. In terms of being able to share a link to a particular moment, a short JS snippet could set the [`currentTime` attribute](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/currentTime) to the value of a query parameter or fragment.

flooie:
Is want to add one that allows for transcripts.  I built a small branch a couple months back that let you follow the transcript and jump around based on that.

flooie:
Just my two cents if we’re updating it we should use one that allows for that.

j-f1:
That could be supported by watching the [`timeupdate` event](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/timeupdate_event) (for highlighting in the transcript if desired) and updating `currentTime` (when jumping around based on the transcript).

PR #5480, opened 2025-04-29
feat(oral argument) add in playback speed selection

jbgolds:
Adds in a new bar for user to change playback speed, along with display of the current speed. defaults to 1.0, with options to go from 0.5 to 3x speed.

Original: 
image

After Change:
image

## Note for reviewer:
Changes were successfully tested in Safari with no loss of other features. But in Chrome, even before the change, I was unable to use the seek bar to "fast forward" through the audio track. Was running into issues cloning data from production due to lack of AWS credentials, so messed the script a bit to allow for local storage of audio track; Perhaps the Chrome issue lies somewhere in there.

### DIFF
--- a/cl/audio/templates/oral_argument.html
+++ b/cl/audio/templates/oral_argument.html
@@ -31,15 +31,68 @@
           src="{% static "js/jquery.jplayer.min.js" %}"></script>
   <script type="text/javascript" nonce="{{ request.csp_nonce }}">
     $(document).ready(function () {
-      $("#jquery_jplayer_1").jPlayer({
+      var player = $("#jquery_jplayer_1");
+      var speedBar = $('.jp-speed-bar');
+      var speedBarValue = $('.jp-speed-bar-value');
+
+      // Define min/max speed for the bar
+      var minSpeed = 0.5;
+      var maxSpeed = 3.0;
+
+      player.jPlayer({
         ready: function (event) {
           $(this).jPlayer("setMedia", {
             title: "{{ af|best_case_name }}",
             mp3: "{{ af.local_path_mp3.url }}"
           });
+          // Set initial speed bar position
+          updateSpeedBar(event.jPlayer.options.playbackRate);
         },
-        swfPath: "{% static "swf/Jplayer.swf" %}"
+        swfPath: '{% static "swf/Jplayer.swf" %}',
+        supplied: "mp3",
+        cssSelectorAncestor: "#jp_container_1",
+        useStateClassSkin: true,
+        autoBlur: false,
+        smoothPlayBar: true,
+        keyEnabled: true,
+        remainingDuration: true,
+        toggleDuration: true,
+        playbackRate: 1.0,
+        defaultPlaybackRate: 1.0
       });
+
+      // Function to update speed bar position and display text
+      function updateSpeedBar(currentSpeed) {
+          var percentage = (currentSpeed - minSpeed) / (maxSpeed - minSpeed);
+          // Clamp percentage between 0 and 1
+          percentage = Math.max(0, Math.min(1, percentage));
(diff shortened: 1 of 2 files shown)


cl/audio/templates/oral_argument.html
```diff
@@ -31,15 +31,68 @@
...
         ready: function (event) {
           $(this).jPlayer("setMedia", {
             title: "{{ af|best_case_name }}",
             mp3: "{{ af.local_path_mp3.url }}"
           });
+          // Set initial speed bar position
+          updateSpeedBar(event.jPlayer.options.playbackRate);
         },
-        swfPath: "{% static "swf/Jplayer.swf" %}"
+        swfPath: '{% static "swf/Jplayer.swf" %}',
```
jbgolds, line 51: 🎵 Note: single quote fixes the existing syntax error.

mlissner:
Looks nice and would fix this issue from 2021: https://github.com/freelawproject/courtlistener/issues/1798. 

How'd you figure this out / what was your process?

jbgolds:
Thanks @mlissner, forgot to add in issue link to PR description, added. 

That was my source issue, I had chatted with Jessica from FreeLaw the other day who provided me this link to the volunteer backlog: https://github.com/orgs/freelawproject/projects/31/views/1.

I was looking for some quick small wins that have isolated code changes for quicker reviews. Goal was to get some PRs under my belt for the codebase to gain some level of comfortability and understanding of the review process before getting into larger and more substantial changes.

mlissner:
That's great, thanks for contributing this one. 

I was curious how you figured out the JS and CSS to do. Did you dig around in jplayer's source or something?

---

The next step here to wait for a review. We have a fast queue and a slow queue, depending on the priority of the issue. This one will fall into the slow queue, which means it'll get reviewed as part of our next sprint. I've assigned somebody to do so, and we'll be in touch with that soon.

I did give the code, commits, etc. a glance though and I think you're looking good overall, so if you want to chew on something bigger, I wouldn't wait on this one.

jbgolds:
> I was curious how you figured out the JS and CSS to do. Did you dig around in jplayer's source or something?

Yep, you actually linked it in one of the comments on the issue :)  https://github.com/freelawproject/courtlistener/issues/1798#issuecomment-953341266 

https://jplayer.org/latest/developer-guide/#jPlayer-option-playbackRate

elisa-a-v (changes requested):
@jbgolds thank you very much for the PR! I tested in Chrome, Firefox, and Safari, and it works great! :sparkles: 

I noticed the sliders are already quite compact, so perhaps capping it at 2x or even 1.5x would make it easier to control. What do you think? I’m not sure what typical usage is, do people listen to oral arguments at 3x speed? The size of the sliders also means it's hard to slide back to normal, so we may want to add a "reset to 1x" button. Not sure if this is within the scope for this PR though, @mlissner what do you think? we could open another issue for that.

I think we could merge this once we've addressed the minor comment about the CSS classes. It's a nice feature, thanks again!

After reviewing this I'm a bit concerned about the accessibility of the audio player but I think that's beyond the scope of this PR, so I opened [an issue](https://github.com/freelawproject/courtlistener/issues/5589) to address it.

mlissner:
> perhaps capping it at 2x

Yeah, that makes sense to me. Both the Apple Podcasts app and YouTube only go to 2×.

> Not sure if this is within the scope for this PR though, @mlissner what do you think?

For this one, let's take the improvement we've got and we can think about larger fixes down the road.

jbgolds:
Sorry for the delay here! 

@elisa-a-v thanks for the review, this is ready for re-review again!

I've removed the unused CSS + updated the max playback rate to 2.0 :).

mlissner:
Into the maw!

PR #5481, opened 2025-04-29
Control Redis memory usage by `find_citations` command

grossir:
While trying to analyze the Redis memory spikes when running the `find_citations` command, we noticed that the queue it uses, `batch1`, only contained `find_citations_and_parentheticals_for_opinion_by_pks` tasks, but not any of the children tasks generated by it. The children, mostly `index_related_cites_fields` (there may be some others created by signals), were going to the default queue. 

Since the throttle had a max of 100 tasks, and each parent task `find_citations_and_parentheticals_for_opinion_by_pks` contains 100 opinion ids, this would create at least 10 000 children tasks per full throttle run, which should be the cause of the redis memory spike

The simplest solution to this issue is 
- send the children tasks to the same queue as the parent tasks
- adjust the CeleryThrottle values to the max limit of tasks (parents + childs) we want to be processed at once
- make the queue controlling arguments adjustable, in order to finetune the memory usage 
     - `CeleryThrottle.min_items` (using the default 50 -> 100 max size)
     - `update_documents.chunk_size` (currently 100)

Note that the actual max number of tasks is a function of both queue controlling arguments, not really the `CeleryThrottle.max`

If we just reduced the CeleryThrottle limit to control the parent tasks `find_citations_and_parentheticals_for_opinion_by_pks`, these may get processed quickly and just clog up the default celery queue with their child tasks

We will probably make the `find_citations` command slower, but will have a better idea of it's speed, and we will not crash redis


flooie:
This makes sense and explains why we weren't detecting the run away.

cl/citations/tasks.py
```diff
@@ -125,20 +124,29 @@ def find_citations_and_parentheticals_for_opinion_by_pks(
     )
     for opinion in opinions:
         try:
-            store_opinion_citations_and_update_parentheticals(opinion)
+            # delivery_info does not exist in test environment
```
flooie, line 127: I must be missing something, but why not just pass in the queue from the jump so you dont need to call it here.  and pass it along to `store_opinion_citations_and_update_parentheticals`

cl/citations/tasks.py
```diff
@@ -125,20 +124,29 @@ def find_citations_and_parentheticals_for_opinion_by_pks(
...
+                children_queue,
+            )
         except ResponseNotReady as e:
             # Threading problem in httplib.
             raise self.retry(exc=e, countdown=2)
 
 
 def store_opinion_citations_and_update_parentheticals(
     opinion: Opinion,
+    queue_for_children: str = "celery",
```
flooie, line 142: also is celery the correct default or is batch1

flooie (commented):
I think it looks good but Im curious about two things.  

1. should we just pass in the queue from find_citations and is batch1 the better default value.

cl/citations/tasks.py
grossir: `store_opinion_citations_and_update_parentheticals` is not only called by `find_citations_and_parentheticals_for_opinion_by_pks` from `find_citations`

It is also called in the regular scraping process. I think those tasks should go to the default queue "celery"

cl/citations/tasks.py
flooie: fair enough.

flooie (approved):
Looks good to me - hopefully this is the fix

Issue #583, opened 2016-10-11
Incorrect opinions for some pdf cases


LawGuy2000:
When browsing data I noticed some cases had incorrect opinions. Case in point is this opinion: https://www.courtlistener.com/opinion/871843/an/

If you look at the PDF here: https://www.courtlistener.com/pdf/2011/12/02/an.pdf

You will see the court name and opinion listed on the site is incorrect.

My guess as to what is happening is that text was pulled from the PDF but the only readable text is the notice that it is electronically filed. As a result, the wrong opinion data is used. Instead, the image needs to be ocred.

mlissner:
> You will see the court name and opinion listed on the site is incorrect.

I see that the case name is wrong. That looks like a scraper issue or an issue with the court website. I've corrected that manually, but before I did the case name was simply "An". I don't see that the court name is wrong. Both say Intermediate Court of Appeals of Hawaii.

> My guess as to what is happening is that text was pulled from the PDF but the only readable text is the notice that it is electronically filed. As a result, the wrong opinion data is used. Instead, the image needs to be ocred. 

Yeah, that's exactly right, and super frustrating. What we're currently doing is extracting readable text from every PDF, if it exists. If no readable text exists, then we extract the text from any images. Probably a better way of doing this would be to do OCR on any item that has images.

Thanks for filing this bug. I think we'll prioritize it according to how many examples like this come up. If it's only Hawaii, we might be able to solve it with a specific solution for Hawaii. If it's broader than that, we'll need a more generalized solution. If you find more examples, please feel free to post them.

mlissner:
Just had a user tell us that they missed an important document because it wasn't OCR'ed, and this bug is the root cause. It'd be great to get this fixed and to do the backlog of missed documents. 

The approach to take is probably something like: 

 - Use our current approach to get text in most cases.
 - If there are images in the text, check if their size makes sense to be scans.
 - If so, OCR them.

Something like that. Here's the doc that they had an issue with:

https://www.courtlistener.com/docket/17258559/1/1/united-states-v-lawrence/

mlissner:
Oh, a companion issue to this one is https://github.com/freelawproject/courtlistener/issues/598, which should be easier to fix and probably affects many more documents.

rpdelaney:
[magic-pdf](https://pypi.org/project/magic-pdf/) is heavy-duty compared to most other pdf parsers but it might be worth looking at.

PR #5484, opened 2025-04-29
feat(PACER needs_ocr): update pacer needs OCR regexes

jbgolds:
Fix for #598 

Adds in some additional regexes for checking if there exists any _good_ line. This fixes the few examples provided in the issue:

1. https://www.courtlistener.com/docket/4538240/1/3/bison-resources-corporation-v-antero-resources-corporation/
2. https://ia800304.us.archive.org/4/items/gov.uscourts.cacb.1466705/gov.uscourts.cacb.1466705.1.0.pdf

Which I've included some tests for these above scenarios. 

_Note_: There's further improvements to be made in the case where there's a few good lines, but majority are images, thus even with the update we'd still be skipping OCR. This issue has already been raised here: https://github.com/freelawproject/courtlistener/issues/583, which I'll take a stab at soon. Here's an example where the fix in this PR would **not** help: https://www.courtlistener.com/docket/69323904/30/2/edward-l-spann-iv/.

### DIFF
--- a/cl/lib/recap_utils.py
+++ b/cl/lib/recap_utils.py
@@ -120,19 +120,33 @@ def needs_ocr(content):
         "Appellate",
         "Appeal",
         "Case",
+        "Desc",
+        "Main Document",
         "Page",
+        "Received:",
         "USCA",
     )
     pagination_re = re.compile(r"Page\s+\d+\s+of\s+\d+")
+    doc_filed_re = re.compile(r"Doc\s+\d+\s+Filed")
+    case_num_re = re.compile(r"^\d+:\d+-CV-\d+$")
+
     for line in content.splitlines():
         line = line.strip()
+        if not line:
+            # skip empty lines
+            continue
+
         if line.startswith(bad_starters):
             continue
-        elif pagination_re.search(line):
+        if pagination_re.search(line):
             continue
-        elif line:
-            # We found a line with good content. No OCR needed.
-            return False
+        if doc_filed_re.search(line):
+            continue
+        if case_num_re.match(line):
+            continue
+        
+        # if we get here, we have found a line that is not a "bad" line, meaning we do NOT need OCR.
+        return False
 
     # We arrive here if no line was found containing good content.
     return True
--- a/cl/lib/tests.py
+++ b/cl/lib/tests.py
@@ -65,6 +65,7 @@ from cl.search.factories import (
 )
 from cl.search.models import Court, Docket, Opinion, OpinionCluster
 from cl.tests.cases import SimpleTestCase, TestCase
+from cl.lib.recap_utils import needs_ocr
 
 
 class TestPacerUtils(TestCase):
@@ -1280,11 +1281,6 @@ class TestElasticsearchUtils(SimpleTestCase):
                 "sanitized": 'This "is" unbalanced',
             },
             {
-                "input_str": 'This "is” unbalanced"',
-                "output": True,
-                "sanitized": 'This "is" unbalanced',
-            },
-            {
                 "input_str": 'This "is" unbalanced"""',
                 "output": True,
                 "sanitized": 'This "is" unbalanced""',
@@ -1546,3 +1542,95 @@ class TestLinkifyOrigDocketNumber(SimpleTestCase):
                     expected_output,
                     f"Got incorrect result from clean_parenthetical_text for text: {agency, docket_number}",
                 )
+
+
+class TestRecapUtils(SimpleTestCase):
+    def test_needs_ocr_cacb_example(self):
+        """Test needs_ocr function with multi-line headers from cacb example provided in issue #598
+
+        This text contains headers like 'Case...', 'Doc...Filed...', 
+        'Main Document', and 'Desc'. The function should recognize these
+        as non-content lines and return True (needs OCR).
+        """
+        cacb_text = """
+Case 2:12-bk-17500-TD
+
+Doc 1 Filed 03/01/12 Entered 03/01/12 12:14:26
+Main Document
+Page 1 of 58
+
+Desc
+
+
+Case 2:12-bk-17500-TD
+
+Doc 1 Filed 03/01/12 Entered 03/01/12 12:14:26
+Main Document
+Page 58 of 58
+
+Desc
+
+
+"""
+        self.assertTrue(needs_ocr(cacb_text), msg="cacb example should need OCR")
+
+    def test_needs_ocr_wvnd_example(self):
+        """Test needs_ocr with specific case number format from wvnd example.
+
+        This text contains a case number line ('1:16-CV-107') and a 
+        'Received:' line. The function should recognize these as 
+        non-content lines and return True (needs OCR).
+        """
+        wvnd_text = """
+1:16-CV-107
+
+Received: 06/03/2016
+
+
+"""
+        self.assertTrue(needs_ocr(wvnd_text), msg="wvnd example should need OCR")
+
+    def test_needs_ocr_with_good_content(self):
+        """Test needs_ocr returns False when substantive content is present.
+
+        This text includes standard headers but also lines like 
+        'This is the first line of actual content.', which should cause
+        the function to return False (doesn't need OCR).
+        """
+        good_text = """
+Case 1:23-cv-00123 Document 1 Filed 01/01/2023 Page 1 of 5
+
+This is the first line of actual content.
+Here is another line.
+
+Page 2 of 5
+Some more content here.
+"""
+        self.assertFalse(needs_ocr(good_text), msg="Should not need OCR with good content")
+
+    def test_needs_ocr_only_standard_headers(self):
+        """Test needs_ocr returns True for text with only basic headers/pagination.
+        
+        This tests the original scenario where only 'Case...' lines and
+        'Page X of Y' lines are present. Should return True (needs OCR).
+        """
+        header_text = """
+Case 2:06-cv-00376-SRW Document 1-2 Filed 04/25/2006 Page 1 of 1
+Page 1 of 1
+
+Case 2:06-cv-00376-SRW Document 1-2 Filed 04/25/2006 Page 2 of 2
+Page 2 of 2
+"""
+        self.assertTrue(needs_ocr(header_text), msg="Should need OCR with only headers/pagination")
+
+    def test_needs_ocr_empty_string(self):
+        """Test needs_ocr returns True when the input content is an empty string."""
+        self.assertTrue(needs_ocr(""), msg="Empty content should need OCR")
+
+    def test_needs_ocr_only_whitespace(self):
+        """Test needs_ocr returns True for content containing only whitespace.
+        
+        The function should strip lines, so whitespace-only lines are treated
+        as empty, resulting in True (needs OCR).
+        """
+        self.assertTrue(needs_ocr("  \n\t\n  "), msg="Whitespace-only content should need OCR")


mlissner:
Thank you @jbgolds! This is another one for slow review, but we'll get to it soon. Thank you!

cl/lib/recap_utils.py
```diff
@@ -120,19 +120,33 @@ def needs_ocr(content):
         "Appellate",
         "Appeal",
         "Case",
+        "Desc",
```
albertisfu: One question here.

In the example, `Desc` is not at the beginning of the line, and `bad_starters` are supposed to match only when the line starts with one of the given values.

Screenshot 2025-06-09 at 2 50 13 p m


Also, from your test cases, I noticed that you're treating these strings as multiline. However, when I tested this with a real document, the lines were extracted as a single line. For example:

`Case 2:12-bk-17500-TD   Doc 1 Filed 03/01/12 Entered 03/01/12 12:14:26   Desc`

In this case, the `bad_starter` `Case` alone would be enough to ignore the line.

Have you encountered any real examples where a case like the one in your test case?

```
Case 2:12-bk-17500-TD
Doc 1 Filed 03/01/12 Entered 03/01/12 12:14:26
Main Document
Page 1 of 58
Desc
```

cl/lib/recap_utils.py
```diff
@@ -120,19 +120,33 @@ def needs_ocr(content):
...
+        "Main Document",
         "Page",
+        "Received:",
```
albertisfu: Similarly, `Received:` wouldn’t be matched by the `bad_starters` check since it’s not at the beginning of the line.

Screenshot 2025-06-09 at 2 51 10 p m

I tested this with the real document, and `needs_ocr` returned `False` for this reason.

In this case, I think it’s better to use a regex pattern to detect it, something like:

```
received_re = re.compile(r"Received: ?\d{2}/\d{2}/\d{2}(?:\d{2})?")
if received_re.search(line):
            continue
```

cl/lib/recap_utils.py
```diff
@@ -120,19 +120,33 @@ def needs_ocr(content):
...
         "USCA",
     )
     pagination_re = re.compile(r"Page\s+\d+\s+of\s+\d+")
+    doc_filed_re = re.compile(r"Doc\s+\d+\s+Filed")
+    case_num_re = re.compile(r"^\d+:\d+-CV-\d+$")
```
albertisfu: I think you added this one for the example related to `1:16-CV-107 Received: 06/03/2016`, right?

If that's the case, this type of header would be handled by the "Received" match suggested above.

Also, since this is matched using `case_num_re.match(line)`, it doesn’t match the real example, as the pattern only matches the docket number when it appears alone on a line.

Unless there are examples where matching the docket number is actually helpful (in which case we should also consider appellate and bankruptcy formats), I think we could remove this one.

albertisfu (changes requested):
Thanks, @jbgolds, for working on this.

I’ve just left some questions and suggestions above,  let me know what you think.

mlissner:
@jbgolds just wanted to make sure you saw the review come in on this one.

jacob-coalition:
Will take a look when I get home to my personal laptop mid next week. Again, do not want to block the team so if urgent feel free to make changes to this branch as well.

mlissner:
One more example of this here: https://www.courtlistener.com/docket/69398899/1/ferris-v-amazoncom-services-llc/

Let's make sure it's fixed too, then merge this. If it's possible, we should reprocess all the ones we've missed too. Maybe a query for items where OCR wasn't needed and the text is quite short?

albertisfu:
> Let's make sure it's fixed too, then merge this. If it's possible, we should reprocess all the ones we've missed too. Maybe a query for items where OCR wasn't needed and the text is quite short?

Sure, I gathered some data from the dev DB to make sure the tweaks applied here might be enough to catch the documents that require OCR.

I used this query:

```
SELECT COUNT(*) AS total
FROM "search_recapdocument"
WHERE
    "search_recapdocument"."ocr_status" = 2
    AND "search_recapdocument"."page_count" > 0
    AND "search_recapdocument"."plain_text" IS NOT NULL
    AND COALESCE(LENGTH("search_recapdocument"."plain_text"), 0) <=
        (700 * "search_recapdocument"."page_count");
```

Basically, it checks the documents with status `OCR_UNNECESSARY` and verifies if their `plain_text` is quite small relative to the number of pages the document has.

For this, I took as an example the following header, which I believe is one of the largest ones we can find:

```
Case
 Case9:13-bk-10313-RR
      2:14-cv-01681-DOCDoc
                         Document
                           138 Filed
                                  9 04/10/14
                                     Filed 04/10/14
                                                Entered
                                                     Page
                                                        04/10/14
                                                          1 of 22 15:07:33
                                                                   Page ID #:32
                                                                            Desc
                        Main Document     Page 1 of 22

```

It’s around 678 characters, so I used a threshold of 700 characters per page.

The query basically checks if the `plain_text` has a length less than `700 * page_count`. If it does, it means the document potentially needs OCR.

From this query, I got a count of `885,030` documents that potentially need OCR in the dev DB.

I also took a sample of some of those documents and found a common pattern:

```
Case 19-16776-amc   Doc   Filed 04/01/24 Entered 04/01/24 23:02:17   Desc Main
                          Document      Page 1 of 9
Case 19-16776-amc   Doc   Filed 04/01/24 Entered 04/01/24 23:02:17   Desc Main
                          Document      Page 2 of 9
Case 19-16776-amc   Doc   Filed 04/01/24 Entered 04/01/24 23:02:17   Desc Main
                          Document      Page 3 of 9


                               April 1, 2024
```

```
Case: 25-1158   Document: 00118254067   Page: 1   Date Filed: 02/28/2025   Entry ID: 6703682




      Appeal # 25-1158

      State of New Jersey, et al.
      vs.
      (President) Donald Trump, et al.
      ____________________
      7th EMERGENCY - MOTION to AMEND
      Appellant’s pro se APPEAL BRIEF
      {with specific NEW issues on appeal
      which SUPPORT - Melvin Jones Jr.’s
      pending motion to proceed IFP},,, to
      ADD… new/ additional issues raised
Case: 25-1158   Document: 00118254067   Page: 2   Date Filed: 02/28/2025   Entry ID: 6703682
```

There are many documents where, after removing the common headers, there is still some content. In these cases, the `needs_ocr` method returned False. However, upon examining these documents, it’s clear they still need OCR. For example:

https://www.courtlistener.com/admin/search/recapdocument/415327441/change/
https://www.courtlistener.com/admin/search/recapdocument/516282/change/

Some pages, besides the headers, contain seals or other types of metadata included as text. In other cases, some pages are fully valid text while others in the same document still need OCR.

The new approach I applied to the `needs_ocr` method is as follows:


```
    Case: 08-9007   Document: 00115928542   Page: 1   Date Filed: 07/30/2009   Entry ID: 5364336
    Line 1
    Case: 08-9007   Document: 00115928542   Page: 2   Date Filed: 07/30/2009   Entry ID: 5364336
    Line 1
```

If between two paginator headers there is no valid content (i.e., no other common headers), it means that at least one page needs OCR, so `needs_ocr` returns True.

If there is content between pages but the number of “good lines” is less than `LINE_THRESHOLD_OCR_PER_PAGE` (defaults to 5), the page is considered to potentially need OCR, as the content could belong to a seal or other text metadata.

For instance, see this example:
https://www.courtlistener.com/docket/6382288/126/state-of-texas-v-united-states-of-america/


Paginator headers supported are:

```
Page: x
Page x of y
Pg x of y
```

If no paginator headers are found, the previous approach is applied as a fallback.


### Re-process documents that potentially needs OCR:

For this purpose, I think the best approach is to mark documents that potentially need OCR, following the query described previously, with `OCR_NEEDED`.

To do this, we can run the following script. It will be quite slow.

```
from pathlib import Path

from django.db.models import F, IntegerField, ExpressionWrapper
from django.db.models.functions import Length, Coalesce

from cl.lib.recap_utils import needs_ocr
from cl.search.models import RECAPDocument

potential_rds_ocr = (
    RECAPDocument.objects
    .filter(
        ocr_status=RECAPDocument.OCR_UNNECESSARY,
        page_count__gt=0,
        plain_text__isnull=False,
    )
    .annotate(
        text_len=Coalesce(Length("plain_text"), 0),
        min_text_expected=ExpressionWrapper(
            700 * F('page_count'),
            output_field=IntegerField(),
        ),
    )
    .filter(text_len__lte=F("min_text_expected"))
    .only("id", "plain_text")
)


output_dir = Path(settings.MEDIA_ROOT, "docs_to_review")
output_dir.mkdir(parents=True, exist_ok=True)
output_file = output_dir / "rds_to_review.txt"

with output_file.open("a") as f:
    for i, rd in enumerate(potential_rds_ocr.iterator()):
        if needs_ocr(rd.plain_text):
            rd.ocr_status = RECAPDocument.OCR_NEEDED
            rd.save()
            print(f"Did item {i}: {rd}")
        else:
            f.write(str(rd.pk) + "\n")
            f.flush()
```

Documents that have fewer than `700 * page_count` characters and for which `needs_ocr` returned False will be stored in a .txt file so we can analyze them later and determine if we need to increase the `LINE_THRESHOLD_OCR_PER_PAGE` setting.

Finally, we can run the following command, which I have improved for query efficiency:

`manage.py reprocess_recap_dockets --extract-unextracted-rds --queue celery`

This will schedule the extraction of PDFs that need OCR. We need to consider that this could be a huge number, around 800K items or more.


cl/recap/management/commands/reprocess_recap_dockets.py
```diff
@@ -19,36 +19,30 @@ def extract_unextracted_rds(queue: str) -> None:
...
+            | Q(ocr_status=RECAPDocument.OCR_NEEDED),
+            is_available=True,
+        )
+        .exclude(filepath_local="")
+        .values_list("pk", flat=True)
+        .order_by()
+    )
+    count = rd_needs_extraction.count()
     # The count to send in a single Celery task
     chunk_size = 100
```
mlissner: 100 is a lot to send in a single batch. Maybe 10 would be better?

cl/lib/tests.py
mlissner, line 1: If it's not too late, it'd be nice to include PDF links for each of these so that later developers have easy access to these documents. If that's too difficult, maybe a reference to how you found these would suffice?

mlissner (commented):
Two quick things from me, but the logic of the OCR detection would benefit from a careful eye that I haven't given it. 

This is great work, thanks Alberto!

cl/lib/tests.py
albertisfu: Sure! I've added an example PDF link to each test and also found another variation of the paginator using `PageID: #`, so I added support to detect it.

cl/recap/management/commands/reprocess_recap_dockets.py
albertisfu: Sure. Added the `--chunk-size` parameter with a default value of 10.

cl/lib/recap_utils.py
```diff
@@ -80,6 +80,66 @@ def get_document_filename(
     )
 
 
+PAGINATION_OF_RE = re.compile(r"\bPage\s+\d+\s+of\s+\d+\b", re.I)
+PAGINATION_PG_OF_RE = re.compile(r"\bPg\s+\d+\s+of\s+\d+\b", re.I)
```
ERosendo: Is there a reason to keep these as two separate regexes? Since they’re very similar, we could simplify by combining them into a single pattern, for example:

```python

PAGINATION_RE = re.compile(r"\b(?:Page|Pg)\s+\d+\s+of\s+\d+\b", re.I)
```

cl/lib/recap_utils.py
```diff
@@ -105,29 +165,59 @@ def needs_ocr(content):
...
+    # does not need OCR.
+    lines = (ln.strip() for ln in content.splitlines())
+    in_page = False
+    other_content_count = 0
+    saw_any_page = False
+    for line in lines:
+        if is_page_line(line):
+            if in_page:
+                if other_content_count < settings.LINE_THRESHOLD_OCR_PER_PAGE:
+                    return True
```
ERosendo: I think we can reduce the indentation here by combining the two if statements:

```suggestion
            if in_page and other_content_count < settings.LINE_THRESHOLD_OCR_PER_PAGE:
                return True
```

cl/lib/recap_utils.py
```diff
@@ -105,29 +165,59 @@ def needs_ocr(content):
...
-    # We arrive here if no line was found containing good content.
-    return True
+        # inside a page, count only non-common header lines
+        if not is_doc_common_header(line):
+            other_content_count += 1
+
+    # end of document, close the trailing page
+    if in_page:
+        if other_content_count < settings.LINE_THRESHOLD_OCR_PER_PAGE:
+            return True
```
ERosendo: Same idea here. We can simplify this using a single if statement:

```suggestion
    if in_page and other_content_count < settings.LINE_THRESHOLD_OCR_PER_PAGE:
        return True
```

cl/lib/recap_utils.py
```diff
@@ -105,29 +165,59 @@ def needs_ocr(content):
...
-        "Case",
-        "Page",
-        "USCA",
-    )
-    pagination_re = re.compile(r"Page\s+\d+\s+of\s+\d+")
-    for line in content.splitlines():
-        line = line.strip()
-        if line.startswith(bad_starters):
+    # Minimum number of valid lines between common headers to consider that the page
+    # does not need OCR.
```
ERosendo: This comment looks like a leftover from an earlier commit.

cl/recap/management/commands/reprocess_recap_dockets.py
```diff
@@ -11,44 +11,38 @@
...
+            | Q(ocr_status=RECAPDocument.OCR_NEEDED),
+            is_available=True,
+        )
+        .exclude(filepath_local="")
+        .values_list("pk", flat=True)
+        .order_by()
+    )
+    count = rd_needs_extraction.count()
     # The count to send in a single Celery task
-    chunk_size = 100
```
ERosendo: We should clean up this comment as well.

cl/lib/recap_utils.py
albertisfu: Sure! I've updated the code to use the single `PAGINATION_RE` regex pattern.

cl/lib/recap_utils.py
albertisfu: Sure! I've updated the code with your suggestion, thanks!

cl/lib/recap_utils.py
albertisfu: Sure I moved the comment to the settings where this comment is helpful.

albertisfu:
Thanks for your review @ERosendo I've applied your suggestions.

ERosendo (approved):
LGTM :shipit:

albertisfu (approved):

mlissner:
@albertisfu, the script you've got in [your comment](https://github.com/freelawproject/courtlistener/pull/5484#issuecomment-3177500494) — it'd be easier if you just ran it yourself, so I opened a ticket to grant you and Eduardo access to the cluster: https://github.com/freelawproject/infrastructure/issues/455

Can you make a fresh issue in the infra repo to run that script and assign it to yourself?

branliu0:
Hi, thank you for this -- this is incredibly useful. Just curious though, what is the purpose of the `in_page` variable? It seems to be meant to exclude any content prior to the first page line. But sometimes the first page line could appear at the "bottom" of the first page?

Consider, for instance, this PACER document: [Alkuda v. McDonald Hopkins Co., L.P.A.pdf](https://github.com/user-attachments/files/21958809/Alkuda.v.McDonald.Hopkins.Co.L.P.A.pdf).

When I use the following code to get the text of the first page:

```python
import pymupdf
doc = pymupdf.open(path)
p1 = next(iter(doc))
text = p1.get_text()
```

I get text that looks like this:

```
UNITED STATES DISTRICT COURT
NORTHERN DISTRICT OF OHIO
EASTERN DIVISION
FILED
ocr o 7 2024'
SARAH ELIZABETH ALKUDA
Case No. 1:24-cv-01103
[... removed]
PLAINTIFF'S MEMORANDUM IN
OPPOSITION TO THE DEFENDANT'S
MOTION TO DISMISS
PLAINTIFF'S MEMORANDUM IN OPPOSITION TO THE DEFENDANT'S MOTION
TO DISMISS
I.
INTRODUCTION
This case arises from a breach of fiduciary duty that has resulted in significant financial
and personal harm to the plaintiff. The defendant, McDonald Hopkins Co., L.P .A., through its
involvement in the administration of the trusts, failed to meet the obligations it accepted when
undertaking fiduciary responsibilities on behalf of the trustee. Plaintiff asserts that these
breaches, supported by clear documentation and a sustained timeline of involvement, have
deprived them of critical information and substantial assets. The defendant's motion to dismiss
seeks to obscure these straightforward fiduciary obligations by conflating them with claims of
Case: 1:24-cv-01103-BMB  Doc #: 13  Filed:  10/07/24  1 of 20.  PageID #: 543
```

And `needs_ocr(text)` returns `True` for this first page here even though it should probably be false. I understand that for this document in practice, we'd be calling `needs_ocr` on the text from the entire document, so it'd be fine, but the point is more that I'ive observed the pacer heading appearing *after* the page text, at least when I use `pymupdf` for this document.

I will say, I did not check additional documents, this would only apply for a one-page document, and also I think it's better for `needs_ocr` to err on the side of being overinclusive. So it's not really a big deal, but I'm sharing just in case you find it at all of interest.

branliu0:
I think this is perhaps specific to how `pymupdf` extracts text from these PDFs, and it may not be relevant for courtlistener, but I'm sharing in case it is helpful.

branliu0:
Sorry for debugging out loud here and for all the notifications! Personally I have fixed this by using `get_text(sort=True)` so that the header appears at the top.

PR #5488, opened 2025-04-30
feat(corpus_importer): Enable user retrieval of new/updated records via csv

ERosendo:
This PR implements a helper function that generates and uploads a CSV manifest, enabling users to easily discover and download new or updated records. The manifest provides a mapping between the record ID and its S3 key for download.

Unlike PR #5447, which focused on a manifest for Lambda functions, this version is specifically designed for user consumption. An example of the manifest content is:

```
10000005,ai_case_law_dataset/o_10000005.json
10000032,ai_case_law_dataset/o_10000032.json
```

After downloading the CSV, users can employ scripting tools, such as the following Bash script, to automate the retrieval of records from S3 based on the provided information:

```
awk -F ',' '{print $1, $2}' new_or_updated_4_2025.csv | while read -r record_id s3_key_raw; do
    s3_key=$(echo "$s3_key_raw" | tr -d '\n\r')
    aws s3api get-object --bucket bulk-customer-data --key "$s3_key" "$record_id.json"
done
```

### DIFF
--- a/cl/corpus_importer/management/commands/make_aws_manifest_files.py
+++ b/cl/corpus_importer/management/commands/make_aws_manifest_files.py
@@ -527,6 +527,65 @@ def get_monthly_record_ids_by_type(
             return list(set(record_ids))
 
 
+def upload_list_of_records_for_users(
+    record_type: str, bucket_name: str, record_ids: list[tuple[int]]
+):
+    """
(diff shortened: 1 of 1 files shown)


mlissner (approved):
Works for me!

Issue #5477, opened 2025-04-28
Add oral argument text to the CourtListener website


mlissner:
We spent tens of thousands of dollars extracting text from audio files. We should, ya know, show that text to users? I guess? 

This is one is overdue, but I don't know how to do it. I'll schedule this for a sprint fairly soon, and the first step will be to figure out what options we have for our audio player and for the text itself. The simple thing is to just show the text. The more fun thing is to: 

 - make the text clickable (to skip to the right place in the audio)
 - make the text highlight as you're listening, karaoke style
 - show time stamps
 - etc.

I'm guessing there are some tools for this and we just need to check them out, but I have no idea.

jbgolds:
Looked into this a bit, interested! I dug a bit and it seems that audio metadata that would include the segments is saved here on extraction: https://github.com/freelawproject/courtlistener/blob/84547272320d14eaf1bf5a28e2550d4319293271/cl/audio/tasks.py#L210-L216.

but is not included as part of the API response for the view as it returns just the `Audio` object, without the  `AudioTranscriptionMetadata` object: https://github.com/freelawproject/courtlistener/blob/4df6a7e8be9cde2fb014f9aa7c3c9306bdb671c4/cl/audio/views.py#L15-L54.

Some changes around this will be required for this feature, which is totally ok, but...  I'm not seeing the audio data as part of the monthly bulk data dumps, so curious if a few example data sets can be provided so this can worked on.

mlissner:
Our HTML page doesn't use the API though. I assume we can just include the field in the HTML response from Django, right?

jbgolds:
> Our HTML page doesn't use the API though. I assume we can just include the field in the HTML response from Django, right?

Yup exactly. my bad on the incorrect wording... should have said _the Django Response_. That is the plan! 

Let me know if it's possible to get some example metadata so I can take a deep dive on this one (or any other volunteer without access to the data :-))

mlissner:
@elisa-a-v, can you help provide some sample data here, please?

elisa-a-v:
Sure! Here's a small fixture with the first `AudioTranscriptionMetadata` in the DB, which corresponds to [this audio](https://www.courtlistener.com/api/rest/v4/audio/?id=92711).

[audio_transcription_metadata.json](https://github.com/user-attachments/files/19968677/audio_transcription_metadata.json)

[...]

PR #5512, opened 2025-05-04
5477 Display Audio Transcripts on oral argument page

jbgolds:
#5477 

This change adds in custom JS to support the features requested in the linked issue. I think we can accomplish all the needs without using an existing library as the source data is pretty easy to work with (see issue for example data). Jplayer is also used for our current audio player, which is decently easy to work with as well, so no need IMO to replace or use a custom package. 

NOTE: This does NOT currently support proper diarization display, which the example transcript data in the issue does not include.

If Audio is not available:
Screenshot 2025-05-04 at 12 03 49 PM

Demo (split into two video files as max file size 10mb):
 
Features Demo'd:
1. Display entire transcript with time stamps
2. Highlighting in yellow of current segment of what's playing on `jplayer`, in a "Karaoke" like way 🎤😄 
3. Search box displays # of found segments, highlights the current search selection in green, and other found segments in blue. Note, when there are words in the search box, disables auto scrolling.

https://github.com/user-attachments/assets/8e78d2b8-8dcc-40d7-a27f-d7b1fcb6ef6d

Features demo'd in video 2:
1. Auto smooth scroll to the current transcript segment matching what is currently on the audio player, if not currently in view. 
https://github.com/user-attachments/assets/21c6997b-a40c-4f47-ab4b-2c04667764ba


mlissner:
Wow, this is cool! We saw this come in over the weekend, so although this is slow-tracked generally, you made it into the sprint that started today (called "Eminem"). We'll get you a review soon!

elisa-a-v (commented):
Pretty cool indeed! Thank you @jbgolds for the handy PR and for your patience with the review :pray:

It works nicely; I particularly like the search feature, it's quite helpful. I also appreciate that you added some keyboard navigation which is great for accessibility, thanks for that!

I only noticed one minor quirk. In Firefox and Chrome, the scroll doesn't move until after a line is entirely out of view, so you'll miss a line every few lines unless you manually scroll. This does not happen in Safari, where the scroll is adjusted before the line is hidden.

Here's a demo of what I mean:

[Screencast from 05-28-2025 03:06:19 PM.webm](https://github.com/user-attachments/assets/35c62cb8-69dd-4611-8ff6-318b5ea08007)

Code looks good, I just left one nitpick about a tiny bit of redundancy. I also wonder if maybe we want to log anything when the transcription isn't found, but I'm not sure how useful that would be here.

@mlissner what are your thoughts on the hidden-line behavior and logging? Once we’ve addressed those two and applied the minor code cleanup, I think we'll be ready to merge.

mlissner:
Oh, and logging: I don't know. Can't we just query the DB to see what items are missing? Or did you mean to show something to the user?

jbgolds:
Hey @elisa-a-v thanks for the review! I changed the threshold and margin for when we'd want to auto scroll & added in a feature flag. 

I _believe_ this is working on chrome and Firefox now but I do have some concerns that I was running into locally.
Only in Chrome:
1.  I am unable to seek using the JPlayer duration bar, it is not clickable. 
2. I am unable to seek by clicking on the a row of the transcript lines to fast forward / rewind.

Are you seeing this issue as well?

elisa-a-v:
Nice, thank you @jbgolds for the updates!

I tested successfully in:
- Chrome latest version running in:
	- a linux-based OS (Pop_OS!)
	- Windows
	- Android
	- macOS
- Firefox latest version running in:
	- Pop_OS!
	- Windows
- Safari latest version in macOS

I think it works nicely! The feature flag does its job, and the scroll position is adjusted before the last line is hidden. I can successfully use both the progress bar and the transcript lines to skip to a particular timestamp in all browsers mentioned above.

I actually don't see the issue you're describing, but I did notice a similar issue in Firefox running in macOS: I can't play the audio at all, and I can't use the progress bar to jump to another location, but clicking the transcript lines *does* skip to the correct timestamp (at least as far as I can tell by the time display in the player, which keeps loading forever.) This, however, happens even **without** these changes, so I don't think it should be a blocker for this PR.

Have we had any complaints about that, @mlissner ? If not, this may be just a BrowserStack bug :shrug: but if we have, I can open an issue to address this specifically. What do you think?

mlissner:
No complaints and if it's broken before and after, let's not block this on it?

elisa-a-v (approved):
I say this is ready to be merged then!

mlissner:
OMG, so exciting. We'll need a blog post on this one!

PR #5515, opened 2025-05-05
fix(tests): Avoid enabling django-debug-toolbar during tests

adamchainz:
I spotted this system check warning when running tests:

```
WARNINGS:
?: (debug_toolbar.W001) debug_toolbar.middleware.DebugToolbarMiddleware is missing from MIDDLEWARE.
	HINT: Add debug_toolbar.middleware.DebugToolbarMiddleware to MIDDLEWARE.
```

This comes from django-debug-toolbar, indicating that the app is installed but the middleware isn’t.

This PR refactors the settings files to remove the `debug_toolbar` app during tests, which avoids it triggering the warning. It should also speed up tests a bit since the debug toolbar tracking won’t be installed, although I haven't benchmarked that.

### DIFF
--- a/cl/settings/django.py
+++ b/cl/settings/django.py
@@ -195,8 +195,6 @@ if DEVELOPMENT:
     MIDDLEWARE.append(
         "django_browser_reload.middleware.BrowserReloadMiddleware"
     )
-    if not TESTING:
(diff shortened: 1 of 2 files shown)


albertisfu (commented):
Great, looks good. Merging...

Issue #5493, opened 2025-05-01
Control Redis memory used by the `find_citations` command


grossir:
While running `find_citations` we see Redis memory usage growing rapidly. A first solution attempt was implemented in this [PR](https://github.com/freelawproject/courtlistener/pull/5481). The problem was:
- `find_citations_and_parentheticals_for_opinion_by_pks` was going to the `"batch1"` queue, which was controlled in the command by a `CeleryThrottle`. However, it's children tasks were going to the default queue
- the number of opinions per parent task * the celery throttle limit, were creating 100 tasks per tick in batch1, but 10 000 children tasks `index_related_cites_fields`   in the default queue 
- so, the children tasks were accumulating in the default queue "etl_tasks" increasing the memory usage

The solution
- put the children task into the same queue "batch1" as the parent tasks, so we can throttle them
- make the parameters `DEFAULT_THROTTLE_MIN_ITEMS` and `DEFAULT_OPINIONS_PER_TASK` controllable via user input , so we could tweak them to an optimal rate

--------------

Now, we keep seeing spikes, even if the growth rate is slower, even when a count of the queue elements show sort of stable numbers, going up and down (when the number goes down, there are plateaus in the growth)

![Image](https://github.com/user-attachments/assets/8aabe093-d5e3-486b-8006-1a4d2844a1f9)

```
In [28]: for i in range(100):
    ...:     print(datetime.now(), r.llen(queue_name))
    ...:     time.sleep(60)
    ...: 
2025-05-01 09:51:15.867292 408
2025-05-01 09:52:15.868646 508
2025-05-01 09:53:15.869924 759
2025-05-01 09:54:15.871336 827
2025-05-01 09:55:15.872597 667
2025-05-01 09:56:15.873994 325
2025-05-01 09:57:15.875269 39
2025-05-01 09:58:15.876698 353
2025-05-01 09:59:15.878096 649
2025-05-01 10:00:15.880873 889
2025-05-01 10:01:15.886146 858
2025-05-01 10:02:15.887423 306
2025-05-01 10:03:15.888746 69
2025-05-01 10:04:15.890292 77
2025-05-01 10:05:15.891675 313
2025-05-01 10:06:15.892952 532
2025-05-01 10:07:15.894216 750
2025-05-01 10:08:15.895416 947
2025-05-01 10:09:15.896698 1022
2025-05-01 10:10:15.898039 804
2025-05-01 10:11:15.899332 51
[...]

PR #5527, opened 2025-05-06
fix(find_citations): tweaks to control children tasks

grossir:
Solves #5493

- `find_citations` now has an optional argument `disconnect_elastic_signals` to disconnect ParentheticalGroup signals for ES indexing
- `find_citations_and_parentheticals_for_opinion_by_pks` will no longer restart the whole loop once it fails with an uncontrolled exception
- `store_opinion_citations_and_update_parentheticals` is now a task, so it can be delayed and retried once it hits a deadlock


cl/citations/tasks.py
```diff
@@ -114,42 +119,75 @@ def find_citations_and_parantheticals_for_recap_documents(
...
+                )
+        finally:
+            if disconnect_pg_signals:
+                reconnect_parenthetical_group_signals()
 
 
+@app.task(bind=False, max_retries=5, ignore_result=True)
 def store_opinion_citations_and_update_parentheticals(
-    opinion: Opinion,
+    opinion_or_pk: Opinion | int,
```
mlissner, line 173: Actually, we should really avoid passing opinions for precisely this reason. Let's make a new task that accepts only the ID, and update the code to use it. Then in a few hours, let's remove this task.

cl/citations/parenthetical_utils.py
```diff
@@ -47,3 +48,43 @@ def create_parenthetical_groups(cluster: OpinionCluster) -> None:
         )
         group_to_create.save()
         group_to_create.parentheticals.set(cg.parentheticals)
+
+
+def disconnect_parenthetical_group_signals():
+    """Disconnect ParentheticalGroup ES indexing on save and delete
```
mlissner, line 54: This seems OK, but it should have a loud warning that it disconnects the signals across the entire application or else somebody will think it's safe to use when it's not.

mlissner (changes requested):
A couple thoughts for you, Gianfranco. Looks like you found a good bug with the opinion object getting placed into the celery queue! That'll do it.

grossir:
@mlissner @flooie this is ready for another review; I added tests too. 

A pending task is making a script to re-index all the ParentheticalGroups. 
1. delete all ParentheticalGroupDocuments
2. create new ParentheticalGroupDocuments from existing ParentheticalGroup rows

mlissner:
Thanks Gianfranco. I think it'd be smart to get Alberto's quick review of this to make sure it feels right to him, since I think he wrote most of this code, so let's do that. 

But generally, I think we might as well disable indexing to the parenthetical search until we're ready to work on it.

cl/citations/tasks.py
```diff
@@ -114,28 +119,58 @@ def find_citations_and_parantheticals_for_recap_documents(
...
-    for opinion in opinions:
+    # delivery_info does not exist in test environment
+    children_queue = (self.request.delivery_info or {}).get(
+        "routing_key", settings.CELERY_ETL_TASK_QUEUE
+    )
+
+    for index, opinion in enumerate(opinions):
+        if disconnect_pg_signals:
+            disconnect_parenthetical_group_signals()
+
```
albertisfu, line 144: I think we can avoid connecting and disconnecting the PA signals on every iteration. If we move it outside the loop, we can disconnect and reconnect only once per task? The behavior should remain the same, so any PA save or delete operation executed by the same Celery worker running the task won't trigger PA indexing signals.

cl/citations/parenthetical_utils.py
```diff
@@ -47,3 +52,53 @@ def create_parenthetical_groups(cluster: OpinionCluster) -> None:
...
+    # prevent circular imports between cl.search.signals and cl.citations.tasks
+    from cl.search.signals import pa_signal_processor
+
+    logger.warning(
+        "Disconnecting ParentheticalGroup ES signals. You will need to handle indexing of ParentheticalGroupDocuments yourself"
+    )
+
+    model_name = ParentheticalGroup.__name__.lower()
+    save_uid = pa_signal_processor.save_uid_template.format(model_name)
+    delete_uid = pa_signal_processor.delete_uid_template.format(model_name)
```
albertisfu, line 75: Do you know if any of these field changes can occur during the `find_citations_and_parentheticals_for_opinion_by_pks` process? If so, it might also be necessary to disconnect these signals to avoid triggering them for PA during the process.

Screenshot 2025-05-09 at 11 44 24 a m

albertisfu (commented):
This looks good @grossir !

I just left a couple of comments.

cl/citations/parenthetical_utils.py
grossir: The citation related signals won't happen, since to Citation is created , updated or deleted (reverse.Citation and reverse-delete.Citation)

About OpinionCluster.panel.through; no panels are created

I was expecting m2m.OpinionsCited to trigger, but from the tests I have performed, it doesn't seem to trigger... I will research this further

grossir: I now have a better idea why the `m2m.OpinionsCited` signal is not triggered

From the [docs](https://docs.djangoproject.com/en/5.2/ref/signals/#m2m-changed) for `m2m_changed` and the examples at the end of that same section

>Sent when a [ManyToManyField](https://docs.djangoproject.com/en/5.2/ref/models/fields/#django.db.models.ManyToManyField) is changed on a model instance

`OpinionsCited.objects.filter(...).delete()` will emit a model `post_delete` signal, not a `m2m_changed` field signal. 

For the expected behavior, we would have to do `opinion_instance.opinions_cited.clear()` (would emit a `post_clear` signal, that is not handled in `ESSignalProcessor`) or `opinion_instance.opinions_cited.remove( some_args_to_identify_all_the_opinions )`

So, I think it is better to use `opinion_instance.opinions_cited.clear()` , for it to have the intended behavior? We would have to update the signal processor to also handle `post_clear`, and optionally disconnect that signal receiver when doing bulk work

mlissner:
@grossir, if you've reviewed Alberto's comments, our process is to assign the issue back to him so that it shows up on his to do column. (Sorry if you already know this, but I wanted to make sure we didn't drop this!)

grossir:
Sure @mlissner, I will follow the process.  I was still thinking about this, I tend to edit issues and comments a lot; I find re-reading and editing easier than writing. I usually request for review or assign issues once I am sure I have nothing more to say

cl/citations/parenthetical_utils.py
albertisfu: You're right. The m2m signal is not expected to be triggered in this scenario of deleting cited opinions in bulk. The actual scenarios that can be handled by the m2m signal for parentheticals are those involving `add()` and `remove() `operations, as described in the test `test_keep_in_sync_related_pa_objects_on_m2m_change`.

In this case, I'm noticing that we might not be indexing the parentheticals that are added or updated during this workflow. Since parentheticals are being created using the `bulk_create` method, no signals are triggered. So, a custom task to index them might be required. `Parentheticals` are still in beta, and we need to improve several things like this. I'll open an issue so we don't forget to fix it.

For now, this seems ok, since these signals are not being triggered, there's no need to disconnect them.

albertisfu (approved):
Thanks @grossir this look good now.

Auto-merge enabled...

grossir:
@mlissner the PR can't merge because it has "Requested changes" from your review. Can I dismiss them?

flooie (approved):
Thank you @grossir @albertisfu

mlissner:
Sorry about that. I dismissed my review and it should be good now.

PR #5534, opened 2025-05-08
feat(templates): introduce new footer

### LINKED ISSUE #4974: Do the new footer for new design
Two PRs if it makes sense:
1. For the needed components.
2. For the actual thing

Image

The footer will be visible on every page across CourtListener.

**Requirements:**
- The "Donate Now" button is on the left-hand side of the footer.
- The links in the footer have been reorganized under three headings: Organization, Data, and Legal.
- Some of the names have changed slightly: About -> About Us, Data Services -> Data Consulting
- We've removed "Feeds & Podcasts" from the footer
- We've removed "Donate" from the list of links in the footer (but it will remain as a button in the footer)
- The newsletter subscription is on the right-hand side of the footer.
- Along the bottom of the footer, we provide a link to free.law, and provide icons that link to LinkedIn (https://www.linkedin.com/company/free-law-project/), Bluesky (https://bsky.app/profile/free.law), Github, and Donate.
- The mobile version of the footer displays the same information as the web version, and should be visible at the bottom of every page on CourtListener. (see below)

**MOBILE:**

Image

elisa-a-v:
Closes #4974 

This PR introduces the new footer for the new design. In the issue it was suggested to create two separate issues, one for the components and one for the footer itself, but turns out no additional components were needed, only the footer itself is introduced as a cotton component instead of inline to keep the base template manageable and readable.

Given the way the content overflows and rearranges in different viewport widths, the way the footer looks varies significantly in different screens. While visually this results in multiple layout variations, code-wise only four breakpoints are explicitly defined, but all additional layout variations arise naturally from CSS flexbox behaviors, spacing, and content flow.

The mobile version only looks just like the figma design up to 540px wide. After that, the content rearranges in different ways, and only beyond 1536px it looks exactly like figma again. For the screens in between, the spacing and width of the link columns were tweaked a little bit to make them fit.

I think it turned out a little better than I expected, but I'm open to alternative solutions if these aren't what we want; in that case we could get an intermediate design from the magnet team for screens between 540px and 1536px, or maybe we could just get their feedback on these instead of asking for a whole new design 🤔

Screenshots below display all different ways in which the content is arranged, and a video is included to showcase the fluid rearrangement.

## Screenshots

<details>
<summary>1535px and up</summary>

image

</details>

<details>
<summary>1280px to 1535px</summary>

image

</details>

<details>
<summary>1096px to 1279px</summary>

image

</details>

<details>
<summary>1024px to 1095px</summary>

image

</details>

<details>
<summary>864px to 1023px</summary>

image

</details>

<details>
<summary>768px to 863px</summary>

image

</details>

<details>
<summary>640px to 767px</summary>

image

</details>

<details>
<summary>540px to 639px</summary>

image

</details>

<details>
<summary>Up to 539px</summary>

image

</details>

## Fluid rearrangement recording

[Screencast+from+05-07-2025+08_20_57+PM.webm](https://github.com/user-attachments/assets/1bf04ad7-7a51-471f-bfe1-7b02efaf42df)


mlissner:
I'll leave the full review up to Eduardo, but my design review is that we need a better design for 640-1095px. In this range, things are stacking up weirdly and without a good balance. 

I think it's fine to ship this as is (pending Eduardo's review), but that we should make a plan to do better with this with some help from Magnet.

cl/assets/tailwind/tailwind.config.js
```diff
@@ -22,14 +22,19 @@ module.exports = {
...
         4.5: '18px',
         7.5: '30px',
         13: '52px',
         15: '60px',
+        18: '72px',
         35: '140px',
+        41: '164px',
+        42: '168px',
+        53: '212px',
+        70: '280px',
```
ERosendo, line 37: We should probably use `rem` for these spacing values. I noticed that most Tailwind classes use `rem` instead of `px`. At first, I thought the new spacing might be following a different convention, but it looks like you're using the dafault tailwind spacing scale (one unit is equal to `0.25rem`, which translates to `4px`). So, I think it would be consistent to use rem here as well.

ERosendo (approved):
Thanks @elisa-a-v, the code looks good and the new component is easy to follow. I just have one question about the new spacing values. After that comments is addressed, we can merge the PR.

elisa-a-v:
Yes, the idea was indeed to extend the [tailwind's default spacing scale](https://v3.tailwindcss.com/docs/customizing-spacing#default-spacing-scale), but you're right that it's best we use `rem` instead of `px` not just for consistency but because it's more flexible and allows for better scaling if users rely on accessibility features like zooming or enlarged text, so great catch, thank you!

Values were updated to use `rem` and I left the `px` value as comments for easy reference.

ERosendo (approved):
LGTM :shipit:

PR #5538, opened 2025-05-08
refactor: Convert favorites notes views to sync

adamchainz:
A step towards seeing how hard it would be to convert `cl.favorites` to sync, towards enabling #5537 as a big test optimization.


albertisfu (approved):
Thanks. Looks good, merging!

PR #5546, opened 2025-05-09
chore(ci): Prevent duplicate GitHub Actions runs

adamchainz:
Copy of https://github.com/freelawproject/awscloud/pull/22 for this repository.

Before, tests run twice on each PR:

image

After, just once:

image

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,10 +1,8 @@
 name: "CodeQL"
 
 on:
-  push:
-    branches:
-      - 'main'
   pull_request:
+  push:
     branches:
       - 'main'
   schedule:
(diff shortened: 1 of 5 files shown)


.github/workflows/codeql-analysis.yml
```diff
@@ -1,10 +1,8 @@
 name: "CodeQL"
 
 on:
-  push:
-    branches:
-      - 'main'
   pull_request:
+  push:
     branches:
       - 'main'
```
adamchainz, line 7: Most of these workflows were using:

```yaml
pull_request:
  branches:
    - main
```

Which means "run on PRs targeting `main`", preventing them from running on stacked PRs. That's not ideal, so this PR drops the `branches` filter from the `pull_request` trigger.

albertisfu (approved):
LGTM. Merging.

PR #5561, opened 2025-05-13
feat(recap): Enhances fetch_attachment_page by retrying on ParserError

ERosendo:
While I was debugging a problem reported by a user who was trying to fetch the attachment page for document **436559721** from case [ca3: 25-1555](https://www.courtlistener.com/docket/69849226/atlas-data-privacy-corp-v-we-inform-llc/), I noticed that when `ParserError` exceptions were raised while attempting to retrieve attachment pages, the status of the related fetch queue wasn't being updated. This resulted in the queue appearing to be stuck in status `1`, which indicates it's awaiting processing.  I also noticed these errors often resolve themselves if the task is retried. Therefore, this PR adds the necessary logic to automatically retry the task when these `ParserError` exceptions are encountered. It also includes a mechanism to update the queue's status once the task has reached its maximum allowed number of retries.

Here's the sentry report that helped me notice the issue: 

https://freelawproject.sentry.io/issues/6498118854/?project=5257254


mlissner:
Thanks. Auto-merge set.

PR #5571, opened 2025-05-14
feat(components): introduce data table and copy-to-clipboard components

### LINKED ISSUE #4739: Identify and make components for basic help page

elisa-a-v:
Closes #4739 

This PR introduces:
- two new components: data table and copy-to-clipboard.
- a new Alpine composable to provide JS to copy to clipboard.
- a new template tag filter to escape quotation marks, so we can safely pass text as HTML attributes without breaking syntax.

The `code` component was extended to include the new `copy-to-clipboard` in the top right corner. This has a flag to disable the button if needed.

## Component library screenshots

<details>
<summary>Data table</summary>

image

</details>

<details>
<summary>Copy to clipboard</summary>

image

</details>

<details>
<summary>Updated code block</summary>

image

</details>

### DIFF
--- a/cl/simple_pages/templates/components.html
+++ b/cl/simple_pages/templates/components.html
@@ -12,11 +12,13 @@
         {'href': '#code-block', 'text': 'Code blocks'},
         {'href': '#two-column-list', 'text': 'Two column list'},
         {'href': '#eyebrow', 'text': 'Eyebrow'},
+        {'href': '#copy', 'text': 'Copy to clipboard'},
       ]},
       {'href': '#layout-components', 'text': 'Layout components', 'children': [
         {'href': '#expansion', 'text': 'Expansion panel'},
         {'href': '#navigation-menu', 'text': 'Navigation menu'},
         {'href': '#tabbed-cards', 'text': 'Tabbed cards'},
+        {'href': '#data-table', 'text': 'Data table'},
       ]},
   ]"
 >
@@ -35,12 +37,13 @@
     </ul>
   </section>
 
+  {# CODE BLOCKS #}
   <section class="max-w-full w-full border-t-2 border-greyscale-200" x-intersect="show" id="code-block">
     <h2 class="mt-6 mb-3">Code blocks</h2>
     <p>Note this component uses <code>&lt;pre&gt;</code> tags so all whitespace will be preserved.</p>
 
     <h4 class="mt-3">Demo</h4>
-    <c-code>
+    <c-code disable_copy="true">
 Sample code here.
     </c-code>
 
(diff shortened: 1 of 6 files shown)


cl/simple_pages/templates/components.html
```diff
@@ -271,6 +333,167 @@ <h4 class="mt-3">Slot</h4>
...
+        <span class="p-2.5 max-md:pb-0 md:w-60 md:min-w-60 max-md:font-medium"><code>caption</code></span>
+        <span class="p-2.5 max-md:max-w-full [overflow-wrap:anywhere]">
+          Table caption for accessibility.<br>Not displayed visually, but it's important to provide context for screen readers.
+        </span>
+      </div>
+      <div class="flex flex-col md:flex-row border-b last:border-none border-greyscale-200 max-w-full">
+        <span class="p-2.5 max-md:pb-0 md:w-60 md:min-w-60 max-md:font-medium"><code>safe</code></span>
+        <span class="p-2.5 max-md:max-w-full [overflow-wrap:anywhere]">
+          Marks the contents as safe, so it can render HTML inside the table.
+          <br>Defaults to false.
```
mlissner, line 430: I think a little more explanation/warning might be useful here so that people don't use it incorrectly?

mlissner (approved):
Fun stuff. Looks about right to me, though I'm a bit suspicious of the text filter (probably I'm wrong, but it feels narrow to just fix quotes?). 

I'll let Eduardo do a full review though. 

Thank you!

cl/simple_pages/templates/components.html
elisa-a-v: Yeah actually, that's a really good point!

Would something like this work?

`<strong>Warning:</strong> May allow XSS vulnerabilities if enabled on user-provided HTML, so use only with trusted content.`

ERosendo (approved):
@elisa-a-v thanks for the PR. Your code looks good.

I also have a concern about the `escape_quotations` method. It seems that currently, only quotation marks break the syntax. However, we might encounter other characters in the future that could cause similar issues. With the current approach, we'd probably have to deal with each of those cases separately.

Would it be possible to use something like `json.dumps()` to escape all relevant characters? I believe we can parse the output of `json.dumps()` using `json.parse()` in JavaScript.

elisa-a-v:
Thank you both for your feedback!

Changes:
- We now include an explanatory warning about potential XSS vulnerabilities in the component library entry for the data table's `safe` prop.
- The custom filter was replaced with Django's built-in `escapejs` filter, and the component library entry for the copy to code component was updated accordingly.

To be honest I still can't think of any cases in which other characters could break syntax, because AFAIK everything inside double quotes is considered plain text inside HTML attributes, but I agree it still seems better and cleaner to reduce custom code and instead rely on built-in features whenever possible. Besides, now it _should_ be completely bullet-proof even if some weird unexpected edge-case comes up 👍🏽 

### Updated component library screenshots

<details>
<summary>Data table safe prop</summary>

image

</details>

<details>
<summary>Copy to clipboard text_to_copy prop</summary>

image

</details>

Do you think these changes look good, or do we want to tweak anything else?

mlissner:
Feels about right to me. Eduardo for the final call since I only looked briefly at the code itself.

ERosendo (approved):
Thanks @elisa-a-v. LGTM :shipit:

PR #5574, opened 2025-05-15
Update eyecite reference test

flooie:
PR creates small tweak to test to match updated eyecite reference matching standard. 

A pincite must be followed by an ending punctuation to avoid capturing the start of the next citation.  Previously we did not enforce this with reference citation pincite matches.  

```
        # pin cite must be followed by one of these so it doesn't capture
        # start of next citation
        (?=
            [,.;)\]\\]|  # ending punctuation
            \ ?[(\[]|    # space and start of parens
            $            # end of text
        )
```


grossir (approved):

mlissner:
This is a lovely PR.

PR #5583, opened 2025-05-16
feat(sitemap): Improve performance for large-sitemap index page

ERosendo:
This PR optimizes the generation of large-sitemap index page. It replaces the [`get_latest_lastmod`](https://github.com/django/django/blob/0b2ed4f7c8396c8d9aa8428a40e6b25c31312889/django/contrib/sitemaps/__init__.py#L106-L115) method in the `DocketSitemap` class with a direct database query. This new logic retrieves the `last_modified` timestamp of the most recent record matching the items query, avoiding a full table scan.

This change should significantly reduce the load time of the  large-sitemap index page

PR #5584, opened 2025-05-16
test(pray-and-pay): Fixes flaky test using a mock date

ERosendo:
This PR addresses an issue in the `test_can_retrieve_old_prayers_to_check` test. The test was giving inconsistent results because it depended on the current date. To make sure the test runs the same way every time and gives reliable results, this PR updates the code to use a fixed mock date instead of the real-time date.

This issue is affecting PR #5583 and #5571.


mlissner:
Thanks!

PR #5585, opened 2025-05-16
fix(citation redirector): Use closest citation match

flooie:
If a possible but not exact match is found in the
citation lookup provide that information to the user

Also strip whitespace when searching citations.

### DIFF
--- a/cl/citations/match_citations.py
+++ b/cl/citations/match_citations.py
@@ -82,22 +82,23 @@ def resolve_fullcase_citation(
                     get_clusters_from_citation_str
                 )(volume=volume, reporter=reporter, page=page)
 
-                # exclude self links
-                if getattr(full_citation, "citing_opinion", False):
-                    clusters = [
-                        cluster
-                        for cluster in clusters
-                        if cluster.id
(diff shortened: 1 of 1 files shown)


cl/opinion_page/templates/citation_redirect_info_page.html
```diff
@@ -124,6 +124,27 @@ <h4 class="bottom">{{ name }}</h4>
...
         <div class="col-md-6" id="citation-redirect">
+            {% if closest_match != None %}
+              <div class="alert alert-info">
+                <div class="row">
+                  <div class="col-xs-9">
+                    <p>
+                      <i class="fa fa-info-circle fa-lg"></i>
+                      It looks like you're trying to search for
+                      <a class="alert-link" href="{{ closest_match.absolute_url }}">
+                        {{ closest_match.case_name|safe }}</a>,
```
quevon24, line 135: I don't think we need the `safe` filter on the case name.

```suggestion
                        {{ closest_match.case_name }}</a>,
```

quevon24 (commented):
Everything looks good, just a little comment. There is also a conflict to do the merge.

Issue #3346, opened 2023-11-03
Move docket view count out of docket model


mlissner:
Right now we tally a count to the docket table every time somebody views a docket. I'm sorry to say this is a bad idea. Looking up, locking, and then incrementing a row on a table of this size every time somebody loads the page is difficult. In fact, I've noticed that this query is our worst query according to AWS:

![image](https://github.com/freelawproject/courtlistener/assets/236970/7a1ecbb8-18c7-4a73-968a-b3fcb67a9282)

(You can't really tell, but it's the first one you see there.)

I'm not sure the best fix for this: 
 - We could put this in redis. 
 - We could use our stats app for it. 
 - We could develop something new. 
 - We could use some kind of external API like Plausible or whatever (this seems really slow). 
 - We could use a special stats database for this (Clickhouse?)

We should do something though, I'm afraid....

mlissner:
By the way, the #917 will also fix this problem when we do it (at the cost of losing the accuracy of this stat).

mlissner:
We're using `view_count` to generate sitemaps, so it might not be so easy to remove unless we think of a better thing to use for that. One way to do this better is outlined here: 

https://medium.com/@timanovsky/ultra-fast-asynchronous-counters-in-postgres-44c5477303c3

Basically, rather than incrementing the counter in realtime, you do an insert in some other table. Then, every minute or so a cronjob adds up the rows in the table and increments whichever view_count fields are needed. Interesting approach, and leaves the value in the DB, which lets us continue querying it. Could also be done using redis and a cronjob instead of inserting rows.

When the CDN is in place, I guess we'll need a very fast API for this so that the client can increment the caller.

mlissner:
Note that this also affects the `citation_count` field on `OpinionCluster`s.

mlissner:
- [ ] We need to mark the view_count field as deprecated wherever it is in the DB and make an issue to remove it eventually.

mlissner:
Hm, but we don't need _unique_ counts on each page — or am I lacking imagination of how a HLL could be useful to us?

> Moving the view count to Redis would involve writing additional code to manage the data, specifically for cleaning up Redis and then saving those counts to the database.

Do we need to save these to the DB? What for? 

> We also need to decide if an exact count is important.

It's not, I don't think.

ERosendo:
> we don't need unique counts on each page

You're correct

> or am I lacking imagination of how a HLL could be useful to us?

[...]

PR #5591, opened 2025-05-19
feat(lib): Removes logic to increment view count

ERosendo:
This PR addresses the first round of improvements discussed in issue [3346](https://github.com/freelawproject/courtlistener/issues/3346).

Key changes:

- All usages of the `increment_view_count` helper function and its definition will be removed
- Removes the `suppress_autotime` context manager


mlissner (approved):
Simple enough. Merging.

mlissner:
One note. Visualizations are going to be deprecated in #4724, so we shouldn't think too hard about them as we do this work.

Issue #5580, opened 2025-05-15
Optimize citation finder queries


mlissner:
I don't know how much blood there is to squeeze from this rock, but it's worth checking. These are the top 5 (or four, see below) queries that we've got in the DB:

| Example<br>Query | Calls/s | Rows/s |
|--------|--------|--------|
| `SELECT "search_opinioncluster"."id", "search_citation"."page" AS "cite_page" FROM "search_opinioncluster" INNER JOIN "search_citation" ON ("search_opinioncluster"."id" = "search_citation"."cluster_id") WHERE ("search_citation"."reporter" = 'WL' AND "search_citation"."volume" = 2012)` | 16.23 | 201029 |
| `SELECT set_config('pghistory.context_id', '2f22431c-69ae-44b8-9e77-4a6b354da490', true), set_config('pghistory.context_metadata', '{"user": null, "url": "/opinion/4523334/abdur-rahim-ambrose-v-state-of-mississippi/"}', true); SELECT ("search_opinionscited"."citing_opinion_id") AS "_prefetch_related_val_citing_opinion_id", "search_opinion"."id", "search_opinion"."date_created", "search_opinion"."date_modified", "search_opinion"."cluster_id", "search_opinion"."author_id", "search_opinion"."author_str", "search_opinion"."per_curiam", "search_opinion"."joined_by_str", "search_opinion"."type", "search_opinion"."sha1", "search_opinion"."page_count", "search_opinion"."download_url", "search_opinion"."local_path", "search_opinion"."plain_text", "search_opinion"."html", "search_opinion"."html_lawbox", "search_opinion"."html_columbia", "search_opinion"."html_anon_2020", "search_opinion"."xml_harvard", "search_opinion"."html_with_citations", "search_opinion"."extracted_by_ocr", "search_opinion"."ordering_key" FROM "search_opinion" INNER JOIN "search_opinionscited" ON ("search_opinion"."id" = "search_opinionscited"."cited_opinion_id") WHERE "search_opinionscited"."citing_opinion_id" IN (4300587)` | 13.09 | 169.71 |
[...]

PR #5594, opened 2025-05-20
fix(search.api_views.OpinionViewSet): Prefetch id only for opinions_cited

grossir:
Part of #5580

In a OpinionViewset, use a Prefetch object to get only the id of `opinion.opinions_cited` objects, instead of all opinion fields; since viewset only shows the opinions_cited as a list of hyperlinks 

May help reduce DB `IO:DataFileRead` as seen on RDS Performance Insights

### DIFF
--- a/cl/search/api_views.py
+++ b/cl/search/api_views.py
@@ -252,7 +252,10 @@ class OpinionViewSet(LoggingMixin, viewsets.ModelViewSet):
     ]
     queryset = (
         Opinion.objects.select_related("cluster", "author")
-        .prefetch_related("opinions_cited", "joined_by")
+        .prefetch_related(
+            "joined_by",
+            Prefetch("opinions_cited", queryset=Opinion.objects.only("id")),
+        )
         .order_by("-id")
     )
 


mlissner:
This is cool. It looks good to me. I'll leave it to the case law team to do the review and merge. I think we have tests for how many queries each API endpoint runs, so if that's true and those tests still pass, that's a good sign. 

The thing to watch out for is that if a prefetch is wrong, Django will just go and get the fields it needs, sometimes by making lots of extra queries. To reproduce that in the past I've had to make sure the serialized JSON object has lots of each joined object type (in this case lots of opinions with lots of citations each). 

Anyhow, I'll leave it to y'all. Nice one, Gianfranco.

quevon24 (approved):
Reducing the returned fields to only the necessary fields should be very helpful to reduce the DataFileRead operations because less data needs to be transferred from storage to reduce IO operations and less data being sent between the db and the app. Also the db will use less memory to process and store the result and this should also improve the query execution. 

Great job @grossir

PR #5606, opened 2025-05-22
Add "new" to the RECAP Alert modal

mlissner:
This tweaks the alert modal to say "New" when selecting between new cases or filings:

image

This clarifies that the alert *will* trigger on documents even when you select only on cases. Otherwise, it's a bit confusing. You might think that selecting "notify on cases only" wouldn't find hits in the text of a document, but it will!


albertisfu (approved):
Looks good! Auto merge enabled...

Issue #5609, opened 2025-05-23
`recap_document_into_opinions` does not generate `html_with_citations`


grossir:
`cl.corpus_importer.tasks.recap_document_into_opinions` creates the opinion object but does not use the `store_opinion_citations_and_update_parentheticals` task that creates the `Opinion.html_with_citations`

https://github.com/freelawproject/courtlistener/blob/2d619b665c54840ac2b366a5a32f58569f677f28/cl/corpus_importer/tasks.py#L2966-L2978

This isn't handled on the command calling the task, either (would need a chain). It should be handled by the task above itself, since that task is also called on the `scrape_pacer_free_opinions` command

https://github.com/freelawproject/courtlistener/blob/2d619b665c54840ac2b366a5a32f58569f677f28/cl/corpus_importer/management/commands/recap_into_opinions.py#L84-L92

At the time of writing, 550k opinions with no `html_with_citations` come from this source. See more [here](https://github.com/freelawproject/courtlistener/issues/5015#issuecomment-2905101805)
```sql
 date_part | source |  count  
-----------+--------+---------
      2024 | G      |  510639
      2025 | G      |   40923
```

PR #5610, opened 2025-05-23
fix(recap_document_into_opinions): find_citations for created opinions

grossir:
Solves #5609

Apply the `find_citations_and_parentheticals_for_opinion_by_pks` task after an Opinion has been created. This creates `Opinion.html_with_citations` and all the OpinionsCited, Parenthetical, UnmatchedCitations and related rows. Do not call the citation finding task when called from the bulk command `recap_into_opinions`, to prevent overloading Redis with many small tasks.

### DIFF
--- a/cl/corpus_importer/management/commands/recap_into_opinions.py
+++ b/cl/corpus_importer/management/commands/recap_into_opinions.py
@@ -75,6 +75,12 @@ def import_opinions_from_recap(
             .order_by("id")
         )
 
+        # prevent creating many single-opinion citation finding tasks
+        skip_citation_finding = True
+        logger.warning(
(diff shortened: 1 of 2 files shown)


cl/corpus_importer/management/commands/recap_into_opinions.py
```diff
@@ -75,14 +75,21 @@ def import_opinions_from_recap(
             .order_by("id")
         )
 
+        # prevent creating many single-opinion citation finding tasks
+        skip_citation_finding = True
+        logger.warning(
+            "Opinions created will not have `html_with_citatiosn`. Run `find_citations` over these opinions"
+        )
+
```
flooie, line 83: why not just include it here?

flooie (approved):
Thanks @grossir

PR #5617, opened 2025-05-23
Update alert-api-docs-vlatest.html

s-taube:
Proposed updates for search alerts API.

### DIFF
--- a/cl/api/templates/alert-api-docs-vlatest.html
+++ b/cl/api/templates/alert-api-docs-vlatest.html
@@ -63,7 +63,7 @@
   <p>Search Alerts update you when there is new information in our search engine.</p>
   <p>This system scales to support thousands or even millions of alerts, allowing organizations to stay updated about numerous topics. This is a powerful system when used with <a href="{% url "webhooks_docs" %}">webhooks</a>.
   </p>
-  <p>Search alerts have three required fields:</p>
+  <p>Search alerts have four required fields:</p>
   <ul>
     <li><strong><code>name</code></strong> &mdash; A human-friendly name for the alert.</li>
     <li><strong><code>query</code></strong> &mdash; Search parameters you get from the front end, as a string.
@@ -76,6 +76,7 @@
       <li><code>wly</code> — Weekly</li>
       <li><code>mly</code> — Monthly</li>
     </ul>
+    <li><strong><code>alert_typ</code></strong> &mdash; Whether you want alerts for cases only, or for both cases and filings. For notifications on Cases only: the alert is created with <code>alert_type = SEARCH_TYPES.DOCKETS </code> ("d"). For notifications on both Cases and Filings: the alert is created with the regular <code>alert_type = SEARCH_TYPES.RECAP</code> ("r").</li>
   </ul>
   <p>To learn more about this API, make an HTTP <code>OPTIONS</code> request:</p>
   <pre class="pre-scrollable">curl -X OPTIONS \


mlissner:
I'm not sure if this is quite right, but it makes sense? Let's see what @albertisfu says?

albertisfu:
In this part:

`<p>Search alerts have four required fields:</p>`

I think we can keep three instead of four.

The reason is that this would only apply to RECAP Search Alerts.

For other search types, like Opinions or Oral Arguments, we still get the `alert_type` from the query string, so there’s no need to pass it as a parameter.

Maybe for RECAP we could also make it optional? For example, if users don’t provide an explicit `alert_type=d` and the alert query is for `type=r`, the alert could be created for both filings and cases.

But if the user provides the `alert_type=d` in the API request, then the alert will be saved with the specified type, in this case, `d` (cases only).

I did a test, and this approach would require a tweak in the Alerts serializer. We'd need to prioritize the `alert_type` passed in the request over the one inferred from the query string. Also, some validation may be required to ensure that only RECAP alerts can have an `alert_type` different from the query string type. For example, we want to prevent alerts where the query string says `q=test&type=oa` but the stored alert_type is o (opinions).

Does this sound good to you, @mlissner? If so, I can open an issue to apply these tweaks in the serializer.

Also, if this sounds good, we may update the description of the `alert_type` parameter to say that it’s optional and only required if you explicitly want "Cases Only" alerts for RECAP.

mlissner:
I think it's better to use the current `type` parameter for this purpose, to keep things consistent across the API and search. We *should*, however, document in the API that the way to create a one-per-case alert is to use the `d` type, so people can figure out how to do that.

albertisfu:
> I think it's better to use the current type parameter for this purpose, to keep things consistent across the API and search. We should, however, document in the API that the way to create a one-per-case alert is to use the d type, so people can figure out how to do that.

Do you mean that we should document only that users need to change the type parameter in the query string to `d` in order to set up a case-only alert? And that the `alert_type` parameter in the API request remains optional, being automatically populated from the query string?

Regarding consistency: in the frontend, case-only alerts already work as proposed above. When you save a case-only alert, the query string still uses `type=r`, but the Alert instance is stored with `alert_type=d`. That way, when a user clicks `View more results` for this alert, it behaves exactly like a regular r-type search, rather than being limited to a single nested-child hit as with the `d` type. So this can also be an issue for Alerts saved from the API if we follow that approach.

mlissner:
OK, your explanation about consistency is helpful, thanks, Alberto. So this sounds like the best approach, like you described:

> Maybe for RECAP we could also make it optional? For example, if users don’t provide an explicit alert_type=d and the alert query is for type=r, the alert could be created for both filings and cases.
> 
> But if the user provides the alert_type=d in the API request, then the alert will be saved with the specified type, in this case, d (cases only).

So we wind up with:

| Alert type | `type` GET param | `alert_type` |
|--------|--------|--------|
| Oral arguments | `oa` | null |
| Opinions | `o` | null |
| Cases & Filings | `r` | null |
| Cases only | `r` or `d` | `d` |

Does that look right?

albertisfu:
> So we wind up with:

Almost:

Alert type | type GET param | alert_type
-- | -- | --
Oral arguments | oa | oa
Opinions | o | o
Cases & Filings | r | r
Cases only | r or d | d

The `alert_type` field can't be null in the database. Currently, the form and serializer populate it based on the content of the query string.
For case-only alerts in the frontend, the value provided by the user is prioritized.
In the serializer, we’d need to apply a tweak so that the `alert_type` value for case only alerts provided by the user is stored instead of the one from the query string.
For other search types, users can simply omit this parameter when creating or editing an alert.

albertisfu:
Great! I've created the issue to apply the required tweak in the serializer:

https://github.com/freelawproject/courtlistener/issues/5624

As for the documentation, then we should say something like the `alert_type` parameter is only required when creating a case-only alert, and it should be set to `d`. In all other cases, the `alert_type` is not required and if provided it will be overridden by the `type` value from the query string.

albertisfu:
I've updated the description of `alert_type` based on the agreements above.

Screenshot 2025-06-02 at 9 05 12 a m

The current example is based on a Case Law alert. Do you think we should replace it with a RECAP Case-only search, or add an additional example for RECAP Case-only alerts?

cl/api/templates/alert-api-docs-vlatest.html
```diff
@@ -77,6 +77,10 @@ <h2 id="search">Search Alerts <small>— <nobr><code>{% url "alert-list" version
       <li><code>wly</code> — Weekly</li>
       <li><code>mly</code> — Monthly</li>
     </ul>
+    <li><strong><code>alert_type</code></strong> &mdash; This is an optional field. It is specifically useful for RECAP Search alerts, where you can specify whether you want alerts for cases only or for both cases and filings.
+      For notifications on Cases only, use the <code>d</code> type. For notifications on both cases and filings, use the <code>r</code> type.
+      For other alert types (Case Law or Oral Arguments), this value is not required, and if provided, it will be overridden by the type in the value from the <code>query</code> field.
+    </li>
```
mlissner: How about this:

```suggestion
      For notifications on cases only, use the <code>d</code> type (short for "dockets"). For notifications on both cases and filings, use the <code>r</code> type (meaning all of RECAP).
      For other alert types, this field is ignored.
    </li>
```

albertisfu: Thanks, suggestion applied.

Screenshot 2025-06-02 at 11 21 54 a m

mlissner:
Great. Merging.

PR #5621, opened 2025-05-27
feat(simple_pages): add redesigned version of Terms and Policies page

### LINKED ISSUE #5355: Implement new design for Terms and policies

elisa-a-v:
Closes #5355 

This PR introduces:
- A new Terms History component heavily based off the `includes/terms_history.html` include.
- The redesigned version of the latest Terms and Policies.
- A fix for the notice about archived terms (old templates):
    - v18 had the notice in a different place than the rest, so it was moved for consistency.
    - v19 didn't have the notice so it was added.

## Terms History Component

Unlike the existing include, the new component tweaks the contents depending on whether it's in an archived version or the latest:
- In an archived version, the copy reads "Other versions of these policies..." and includes the link to the latest version.
- In the latest version, the copy reads "Earlier versions of these policies..." and it does not include the link to the latest version.

To identify the version, a flag was added to the old terms view.

Technically we could also use this component in the old templates that currenly have the include, as the code is almost exactly the same, and the tailwind classes that change the link styles aren't present in old templates so it still looks the same. Only difference is it doesn't say "Earlier versions" and it includes a link to the latest version.

For reference this is how it looks:

<details>
<summary>New component in old templates</summary>

Screenshot from 2025-05-26 20-11-54

</details>

Not important at all and might not be worth it, but I thought I'd mention since it bugs me that it says "Earlier versions" even in v2 😅

## Screenshots

<details>
<summary>Desktop</summary>

new terms desktop

</details>

<details>
<summary>Mobile</summary>

new terms mobile

</details>

### DIFF
--- a/cl/simple_pages/templates/v2_terms/latest.html
+++ b/cl/simple_pages/templates/v2_terms/latest.html
@@ -14,7 +14,7 @@
   ]"
 >
   <c-layout-with-navigation.section id="terms">
-    <h1 class="font-sans text-display-xs mb-12">Terms and Policies</h1>
+    <h1 class="max-md:mb-2">Terms and Policies</h1>
     <h2>Terms of Service</h2>
     <p class="lead">By accessing, browsing, or using CourtListener.com, you agree the following terms:</p>
 
--- a/cl/simple_pages/templates/v2_terms/latest.html
+++ b/cl/simple_pages/templates/v2_terms/latest.html
@@ -1,0 +1,149 @@
+{% extends "new_base.html" %}
+
+{% block title %}{{ title }}{% endblock %}
+
+{% block content %}
+<c-layout-with-navigation
+  data-first-active="terms"
+  :nav_items="[
+    {'href': '#terms', 'text': 'Terms of Service'},
+    {'href': '#privacy', 'text': 'Privacy Policy'},
+    {'href': '#removal', 'text': 'Removal Policy'},
+    {'href': '#copyright', 'text': 'Copyright Policy'},
+    {'href': '#versions', 'text': 'Previous Versions'},
+  ]"
+>
+  <c-layout-with-navigation.section id="terms">
+    <h1 class="font-sans text-display-xs mb-12">Terms and Policies</h1>
+    <h2>Terms of Service</h2>
+    <p class="lead">By accessing, browsing, or using CourtListener.com, you agree the following terms:</p>
+
+    <h4>1. We are not your lawyers</h4>
+    <p>CourtListener.com is not intended to be or to provide legal advice. Any information supplied by CourtListener.com or its operators is intended solely as general guidance on the use of the service, and does not constitute professional or legal advice.
+    </p>
+
+    <h4>2. The service may be unreliable and might simply go away</h4>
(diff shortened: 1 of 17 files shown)


mlissner:
> Not important at all and might not be worth it, but I thought I'd mention since it bugs me that it says "Earlier versions" even in v2 😅

Let's not bother. Nobody looks at the old versions anyway, but we keep them *juuuust* in case.

cl/simple_pages/templates/v2_terms/latest.html
```diff
@@ -0,0 +1,149 @@
...
+    {'href': '#terms', 'text': 'Terms of Service'},
+    {'href': '#privacy', 'text': 'Privacy Policy'},
+    {'href': '#removal', 'text': 'Removal Policy'},
+    {'href': '#copyright', 'text': 'Copyright Policy'},
+    {'href': '#versions', 'text': 'Previous Versions'},
+  ]"
+>
+  <c-layout-with-navigation.section id="terms">
+    <h1 class="font-sans text-display-xs mb-12">Terms and Policies</h1>
+    <h2>Terms of Service</h2>
```
ERosendo, line 18: I think we should adjust the style of these two elements. In the new design, the "Terms of Service" heading seems to overpower the primary "Terms and Policies" title due to its size. Additionally, the vertical separation between the two elements is too wide. I looked for a suggestion in Figma, but there isn't a mockup available for this particular page

Screenshot 2025-05-27 at 1 11 16 PM
 
For comparison, here's the current design:

image

In my opinion, the current design offers a better visual hierarchy.

ERosendo (changes requested):
Thanks @elisa-a-v. the new component looks good. Just one minor comment

ERosendo:
@elisa-a-v I just noticed the live version of this page uses a horizontal rule to separate sections, but this element is missing in the PR.  I've included a screenshot for reference: 

Screenshot 2025-05-27 at 1 26 04 PM

cl/simple_pages/templates/v2_terms/latest.html
elisa-a-v: You know what, I completely agree, and I actually changed the heading style in the alerts help page for this same reason, but then didn't keep doing it with the rest of the pages because I've been trying to replicate the figma design as best as I can and it completely slipped my mind, so thank you for pointing this out!

In the alerts help page I used the same styles for the primary heading as the one in the help index:

<details>
<summary>Figma design for alerts help page</summary>

image

</details>

<details>
<summary>Actually implemented alerts help page</summary>

image

</details>

<details>
<summary>Help index</summary>

image

</details>

We should probably use the same style across help pages, and the bigger heading with a different font stands out waaaay better so I'd keep it for all of them. What do you think? Do @mlissner and @s-taube approve?

If so, I can add the minor style fix in all the new help pages in this same PR.

elisa-a-v:
> I just noticed the live version of this page uses a horizontal rule to separate sections, but this element is missing in the PR

It's missing in the PR because it's missing in the new design, but if we want it there we can add it. What do you say, @mlissner ? It would make it easier to tell each section's limits, but I think we could achieve the same purpose adding more spacing between sections and this would look more consistent with the new aesthetic.

PR #5625, opened 2025-05-27
fix(citations.tasks): ingest UnmatchedCitation for opinion that alrea…

### LINKED ISSUE #5619: Allow creation of new `UnmatchedCitations` for an opinion that already has them
The current UnmatchedCitation process does not allow to **add** new UnmatchedCitation to opinions that already have associated UnmatchedCitation rows. 

It assumes that eyecite is static, and once citations are found, no new citations (that may be unmatched) will be found. So, we need to add that logic

https://github.com/freelawproject/courtlistener/blob/4117f4bb23f5b826add2387bfb2663e3b0f97111/cl/citations/tasks.py#L259-L280

grossir:
…dy has them

Fixes #5619

- Move `store_unmatched_citations` and `update_unmatched_citation_status` from `cl.citations.tasks` to `cl.citations.utils`, since they are not tasks

- Add `handle_unmatched_citations` that accounts for opinions already having unmatched_citations, and tries to add only the new ones

- refactored tests to handle updated signatures for the 3 functions mentioned above

- In `store_opinion_citations_and_update_parentheticals` refactor `currently_cited_opinions` query, to prevent an inner join when a covered query is enough
- refactor `opinion_clusters_to_update` to prevent another join when a direct query is enough

### DIFF
--- a/cl/citations/tasks.py
+++ b/cl/citations/tasks.py
@@ -7,7 +7,7 @@ from django.db.models import F
 from django.db.models.query import QuerySet
 from django.db.utils import OperationalError
 from eyecite import get_citations
-from eyecite.models import CitationBase, FullCaseCitation
+from eyecite.models import CitationBase
 from eyecite.tokenizers import HyperscanTokenizer
 
 from cl.celery_init import app
@@ -21,7 +21,6 @@ from cl.citations.match_citations import (
     NO_MATCH_RESOURCE,
     do_resolve_citations,
 )
-from cl.citations.models import UnmatchedCitation
 from cl.citations.parenthetical_utils import (
     create_parenthetical_groups,
     disconnect_parenthetical_group_signals,
@@ -30,7 +29,10 @@ from cl.citations.parenthetical_utils import (
 from cl.citations.recap_citations import store_recap_citations
 from cl.citations.score_parentheticals import parenthetical_score
 from cl.citations.types import MatchedResourceType, SupportedCitationType
-from cl.citations.utils import make_get_citations_kwargs
+from cl.citations.utils import (
+    handle_unmatched_citations,
+    make_get_citations_kwargs,
+)
 from cl.search.models import (
     Opinion,
     OpinionCluster,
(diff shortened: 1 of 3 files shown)


grossir:
About the query optimizations

This 
https://github.com/freelawproject/courtlistener/blob/cb9a5822025bde9a920d77375a0287a0d298bdd9/cl/citations/tasks.py#L222-L224

translates to this
```sql
SELECT "search_opinion"."id" 
FROM "search_opinion" 
INNER JOIN "search_opinionscited" 
ON ("search_opinion"."id" = "search_opinionscited"."cited_opinion_id") 
WHERE "search_opinionscited"."citing_opinion_id" = 11059364 LIMIT 21
```

making a join makes no sense since we only want the "search_opinion.id" which is the same as "search_opinionscited"."cited_opinion_id"

-------

Similarly, this query makes another join, when we actually have the cluster ids available from the opinion objects
https://github.com/freelawproject/courtlistener/blob/cb9a5822025bde9a920d77375a0287a0d298bdd9/cl/citations/tasks.py#L268-L270

flooie (changes requested):
@grossir and I met on this PR.  It's good but could be tightened up.  Mostly we agreed we could simplify the functions by combining the two unmatched lists and update the tests.  This should make it easier to read etc.

cl/citations/tasks.py
```diff
@@ -208,27 +210,10 @@ def store_opinion_citations_and_update_parentheticals(
         opinion.save()
         return
 
-    # Put apart the unmatched citations
+    # Put apart the unmatched citations and ambiguous citations
     unmatched_citations = citation_resolutions.pop(NO_MATCH_RESOURCE, [])
-
-    # Delete citations with multiple matches
     ambiguous_matches = citation_resolutions.pop(MULTIPLE_MATCHES_RESOURCE, [])
```
flooie, line 215: Combining these into a single object here will simplify all the code signatures down the line and should make this more maintainable

flooie (approved):
Changes look good.  Thank you @grossir

PR #5628, opened 2025-05-28
Improve Docker setup

adamchainz:
Copy some improvements from https://github.com/freelawproject/doctor/pull/209, as well as improving the Node.js steps for better caching.


albertisfu (approved):
This look great, thanks.

There is a significant improvement in build time, ~17 seconds compared to the usual ~1.30 minutes.

Screenshot 2025-05-28 at 1 16 44 p m

There were a couple of unrelated failing tests, which I addressed in https://github.com/freelawproject/courtlistener/pull/5629, so this is good to merge.

PR #5655, opened 2025-06-02
fix(api): throttle user

grossir:
User is hitting `/api/rest/v4/opinions/` constantly causing full table counts and high CPU usage, as seen  in RDS Performance Insights


mlissner:
Thanks for this. I reached out to the user to let them know.

grossir:
Count query that was triggered
```
SELECT set_config('pghistory.context_id', '1f827551-d119-4241-9acd-1328f5d7fe18', true), set_config('pghistory.context_metadata', '{"user": ..., "url": "/api/rest/v4/opinions/"}', true); SELECT COUNT(*) AS "__count" FROM "search_opinion"
```

Impacting CPU
image

Issue #5652, opened 2025-06-02
Turn off citation_count temporarily for Opinions


flooie:
Temporarily turn off citation_count while we update html with citations for the corpus.

PR #5656, opened 2025-06-02
fix(find_citations): disable `citation_count` update in DB and ES

grossir:
Solves #5652

- Add an optional argument to `find_citations` "--disable-citation-count-update"
- This will cause OpinionCluster.citation_count to not be updated; and the `index_related_cites_fields` to not be sent in `store_opinion_citations_and_update_parentheticals`
- add a test for the above changes
- also, delete "queue_for_children" behaviour, since it was generating bugs due to ES indexing happening in parallel queues

### DIFF
--- a/cl/citations/management/commands/find_citations.py
+++ b/cl/citations/management/commands/find_citations.py
@@ -7,7 +7,6 @@ from django.core.management import CommandError
 from django.core.management.base import CommandParser
 from localflavor.us.us_states import OBSOLETE_STATES, USPS_CHOICES
 
-from cl.citations.models import UnmatchedCitation
 from cl.citations.tasks import (
     find_citations_and_parentheticals_for_opinion_by_pks,
 )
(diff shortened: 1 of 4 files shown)


grossir:
Note that `index_related_cites_fields` indexes 
- OpinionDocument.citeCount
- OpinionClusterDocument.citeCount
- OpinionDocument.cites

So, disconnecting that task will also leave `OpinionDocument.cites` outdated. 

Splitting  the cites update from the `citeCount`s update is doable; but maybe not worth the effort since we can re-run `index_related_cites_fields` for everything once we are done and bulk update the `OpinionCluster.citation_count` field

cl/citations/tasks.py
```diff
@@ -296,12 +299,13 @@ def store_opinion_citations_and_update_parentheticals(
...
+    if update_citation_count:
+        index_related_cites_fields.delay(
             OpinionsCited.__name__,
             opinion.pk,
-            list(cluster_ids_to_update),
-        ),
-        queue=queue_for_children,
-    )
+            cluster_ids_to_update,
+        )
```
flooie, line 311: is there a reason we would want this to run if the transaction above fails?

grossir: if the transacion fails this won't run, since there is nothing catching errors on the transaction

cl/citations/management/commands/find_citations.py
```diff
@@ -172,15 +180,16 @@ def handle(self, *args: list[str], **options: OptionsType) -> None:
         if options.get("no_html_with_citations"):
             query = query.filter(html_with_citations="")
         if options.get("all"):
-            query = Opinion.objects.all()
```
flooie, line 175: This shouldnt be deleted right?

flooie (approved):
thanks for the update looks good

PR #5663, opened 2025-06-03
Update alert_help.html

s-taube:
I attempted to add the additional updates to the help page. I'm not sure what height/width measurements to include for the screenshot -- I added the dimensions of the original image.


mlissner:
Usually that's the right way to do it: Size the photo correctly for the place where it'll live on the website, and then set those dimensions in your HTML. These are important for slow connections because they let the browser make a space for the image before downloading it (otherwise the page changes shape when the image finishes downloading and you *finally* learn how big it is).

albertisfu (approved):
This looks good to me.

Screenshot 2025-06-05 at 9 07 57 a m


Thanks @s-taube

PR #5666, opened 2025-06-03
Update alert_help.html

s-taube:
Added link to relative dates in v2.


cl/simple_pages/templates/v2_help/alert_help.html
```diff
@@ -184,7 +184,7 @@ <h2>Search Alerts</h2>
   {# Creating Search Alerts #}
   <section x-intersect.margin.-20%.0.-70%.0="show" id="creating-search-alerts">
     <h3>Creating Alerts</h3>
-    <p>To create a Search Alert, begin with a search in CourtListener's Case Law, RECAP or Oral Argument database. In the results page, click the bell icon in the search bar or click the Get Alerts button in the sidebar on the left.
+    <p>To create a Search Alert, begin with a search in CourtListener's Case Law, RECAP or Oral Argument database. You can set static calendar dates, or rolling <a href="{% url "relative_dates_help" %}">relative dates</a> for your alerts. In the results page, click the bell icon in the search bar or click the Get Alerts button in the sidebar on the left.
```
mlissner: This was a good guess for the URL name, but it's actually "relative_dates", as you can see here: 

https://github.com/freelawproject/courtlistener/blob/main/cl/simple_pages/urls.py#L68

So:

```suggestion
    <p>To create a Search Alert, begin with a search in CourtListener's Case Law, RECAP or Oral Argument database. You can set static calendar dates, or rolling <a href="{% url "relative_dates" %}">relative dates</a> for your alerts. In the results page, click the bell icon in the search bar or click the Get Alerts button in the sidebar on the left.
```

mlissner:
Not sure what's up with the test. Perhaps @ERosendo can help if you're happy with the change I suggested.

ERosendo:
@mlissner I've looked into the error and was able to reproduce it on my local env. The error message itself isn't very descriptive, but the core problem is the URL name, which is preventing the alert page from loading. Once that's fixed, the tests should pass without any issues.

mlissner:
Nasty. Thanks for digging in. Stephanie, if you want to commit, wait for tests to pass, and then merge, that'd be great!

Issue #5646, opened 2025-06-02
Pray-and-pay pricing calculation potentially broken


v-anne:
<img width="1157" alt="Image" src="https://github.com/user-attachments/assets/f5843b62-9ae2-42ea-aa3a-81a44ecc50f3" />

The leaderboard says "5,538 prayers granted totaling $5,217.41."

There is something wrong with that price. Either there is a floating point error, or the price of a document is being calculated incorrectly. Is it possible to check whether any documents with granted requests have a price that isn't a multiple of $0.10?

I would not be that surprised if it was just a floating point issue, but I've never seen it before on the page (I recall writing the code in a way to minimize the possibility of that). To be clear, this isn't a major issue by any means, but I would like someone to check just so we can squash any bug that may have arisen.

mlissner:
Well that's sure wonky! @ERosendo, want to give me a couple queries to dig in?

ERosendo:
Hey @mlissner

I noticed that in PR [5286](https://github.com/freelawproject/courtlistener/pull/5286), we updated the code that computes the total cost of prayers to use 91¢ as the default for records where the page count is `null` or `0`.

Before we treat this as a potential floating-point issue, I think we should first confirm there aren't any documents currently using that default in the set of granted prayers.

Here's a short script to query for those records:

```python
from cl.search.models import RECAPDocument
from cl.favorites.models import Prayer
from django.db.models import Q

documents = (
    RECAPDocument.objects.filter(prayers__status=Prayer.GRANTED)
    .filter(Q(page_count__isnull=True) | Q(page_count=0))
    .values_list('id', flat=True)
    .distinct()
)
print(list(documents))

```

If that script returns an empty list, we can then focus on debugging this as a floating-point issue.

mlissner:
Yeah, that's gotta be what it is. Maybe we just round the value on the front end to the nearest dime?

ERosendo:
Sure! I will update the helper method that computes the total.

v-anne:
But what I'm confused about is how documents are being added to RECAP without us knowing how many pages they have. That still seems to be an issue with the calculation.

mlissner:
Oh yeah, that's a good question. :)

ERosendo:
We can still use the script I provided to confirm if such documents exist within the granted prayers set and see what might have happened during their processing.

[...]

PR #5674, opened 2025-06-04
feat(prayers): Rounds prayers total cost to one decimal place

ERosendo:
This PR updates the `compute_prayer_total_cost` to round the aggregated `price__sum` to one decimal place, enhancing readability and consistency in cost presentation.

Implements the suggestion from https://github.com/freelawproject/courtlistener/issues/5646

This tweak will affect the total cost of granted and pending prayers.

### DIFF
--- a/cl/favorites/utils.py
+++ b/cl/favorites/utils.py
@@ -254,7 +254,7 @@ async def compute_prayer_total_cost(queryset: QuerySet[Prayer]) -> float:
         .aaggregate(Sum("price", default=0.0))
     )
 
-    return cost["price__sum"]
+    return round(cost["price__sum"], 1)
 
 
 def send_prayer_emails(instance: RECAPDocument) -> None:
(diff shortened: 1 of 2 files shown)


cl/favorites/utils.py
```diff
@@ -355,8 +355,6 @@ async def get_lifetime_prayer_stats(
     cache_key = f"prayer-stats-{status}"
 
     data = await cache.aget(cache_key)
-    if data is not None:
-        return PrayerStats(**data)
```
mlissner: Did you mean to delete this?

mlissner (changes requested):
One concern, otherwise looks good to me, thanks!

cl/favorites/utils.py
ERosendo: My bad! That was a mistake, thanks for catching it.

mlissner (approved):

Issue #5545, opened 2025-05-09
Detect ongoing duplicate ingestion from scrapers


grossir:
The value of `Opinion.local_path` may give an easy way to detect ongoing duplication due to timestamps in the document that alter the hash everytime it is requested. 

This field is built using the extension, date filed and the case name; and collisions are resolved by adding a sort of counter at the end. So, we could simply send a `logger.error` alert to Sentry when the counter is greater than, say, 5. Sometimes a collision is acceptable:
- when multiple opinion types are filed the same day for the same case (dissent, concurrence, lead, etc)
- for opinion versions (same everything, but different hash due to minor edits, or major edits for official release)

Examples of most recent duplicates, that were found by [chance](https://github.com/freelawproject/juriscraper/issues/1393)
- [local_path example](https://www.courtlistener.com/api/rest/v4/opinions/10861028/):  html/2025/04/25/zelka_h.v.a.c._maintenance_solutions_inc._v._g.m._crisalli__assoc._inc._12.html
- [another local_path](https://www.courtlistener.com/api/rest/v4/opinions/10896955/): html/2025/04/25/zelka_h.v.a.c._maintenance_solutions_inc._v._g.m._crisalli__assoc._inc._60.html

https://github.com/freelawproject/courtlistener/blob/99f99ab0c6e61a51fdd24837aafa520396f52ecd/cl/scrapers/management/commands/cl_scrape_opinions.py#L139-L143

PR #5690, opened 2025-06-06
feat(scrapers.utils):  check_duplicate_ingestion

grossir:
Implements #5545

- Creates `scrapers.utils.check_duplicate_ingestion` that relies on S3 filepath system and `cl.lib.storage.get_name_by_incrementing` to detect suspicious file path repetiton and sent a `logger.error` to Sentry

- integrates with `cl_scrape_opinions` and `cl_scrape_oral_arguments`

- Adds tests


cl/scrapers/utils.py
```diff
@@ -555,3 +556,42 @@ def save_response(site: AbstractSite) -> None:
...
+    date and extension combination
+    'pdf/2025/06/05/state_v._walsh_1.pdf'
+
+    :param local_path_name: filepath of the file in S3
+    :return None
+    """
+    # we may get the same file path for:
+    # - a cluster of opinions, which are actually different documents for the
+    #     same date and case
+    # - a common case name in the same date, for example, "State v. Doe"
```
flooie: can we clean this up.  

Id like to make the first line one line and we should move the comment here in to the docstrings ti think.  Or atleast part of it.

flooie:
@grossir wouldnt it be easier to just 

```
def get_name_by_incrementing(
    instance: Storage,
    name: str,
    max_length: int | None = None,
) -> str:
    """Generate usable file name for storage iterating if needed.

    Returns a filename that is available in the storage mechanism,
    taking the provided filename into account.

    This maintains the old behavior of get_available_name that was available
    prior to Django 1.5.9. This behavior increments the file name by adding _1,
    _2, etc., but was removed because incrementing the file names in this
    manner created a security vector if users were able to upload (many) files.

    We are only able to use it in places where users are not uploading files,
    and we are instead creating them programmatically (for example, via a
    scraper).

    For more detail, see:

    https://docs.djangoproject.com/en/1.8/releases/1.5.9/#file-upload-denial-of-service

    :param instance: The instance of the storage class being used
    :param max_length: The name will not exceed max_length, if provided
    :param name: File name of the object being saved
    :return: The filepath
    """
    dir_name, file_name = os.path.split(name)
    file_root, file_ext = os.path.splitext(file_name)
    count = itertools.count(1)
    while instance.exists(name):
        # file_ext includes the dot.
        name = os.path.join(dir_name, f"{file_root}_{next(count)}{file_ext}")
    return name
```
log the warning here in `get_name_by_incrementing`.

cl/scrapers/utils.py
grossir: Sure; but isn't the current first line of the docstring a single line, or what do you mean?

grossir:
@flooie 
> log the warning here in get_name_by_incrementing

Since we want this warning to be specific to scrapers, I thought it would be better to put it in the scrapers code. `get_name_by_incrementing` is used through `IncrementingAWSMediaStorage` by a lot of other models / objects, too.

Issue #5667, opened 2025-06-03
Oral argument transcript design review fixes


mlissner:
@jbgolds made some great changes to land the transcript feature. It looks great, but, as ever, I have feedback now that I'm seeing it live. 

1. About the search feature, I think I'd remove it for now. I think it's neat, but I think it needs a more careful design review, and I don't think it adds much beyond CTRL+F. Can we remove it now and add it in a subsequent PR?

2. The margins on the page are a bit off. There should be a straight line down from the T of Transcript to the numbers: 

    ![Image](https://github.com/user-attachments/assets/d6080a5c-1eb0-4f9c-a183-a7f0d41315fe)

    This is usually caused by not doing bootstrap divs quite right.

Other than that, I think it's kind of awesome?

@jbgolds, are you game to do make these tweaks?

jbgolds:
yup, will handle it in next day or two 👍 🚀

jbgolds:
yup, will handle it in next day or two 👍 🚀 @mlissner

mlissner:
One more little thing to do here, I think. We got feedback on Bluesky asking if we had audited all 3.2M minutes of the content to make sure it was good, since it was generated by AI. We did not, but we should note somewhere that it's AI generated. Maybe below the transcript, floated right? 

Something like:

> [i] Transcripts generated by AI. Please verify by listening.

I think that should be enough to warn people?

PR #5695, opened 2025-06-07
Jbgolds/5667 oral argument transcript revisions

jbgolds:
https://github.com/freelawproject/courtlistener/issues/5667

image

- Adds in a note saying transcript were generated by AI
- Removed the search feature
- Cleans up the margins (I _think_)


mlissner:
Looks good to me, visually, thank you. I'll let Elisa make the final call. I'm putting this in the fast queue so we can merge during this sprint.

fellowjamess:
How about adding a similar warning box to the one in text, instead of the note?
text

mlissner:
You think the current approach isn't enough @fellowjamess?

fellowjamess:
@mlissner, it may be missed without enough scrolling (of coruse depending on screen size and browser) and the other warning is consistent with other website sections.

mlissner:
I think it's OK as is:

 - The transcript box is relatively short, making it likely to see it; and
 - The risk here is relatively low because all transcripts involve judgement (whether human or computer), so it's normal to listen with your own ears to make sure things are correct.
 - So far the AI errors we've seen have been pretty obvious and of the kind a human might also make (e.g. the "Katz" decision was transcribed as "Cats" decision).

mlissner:
@jbgolds do you mind if Elisa takes this one over the finish line or are you hoping to return to it? We're excited to get this one done!

jacob-coalition:
@mlissner , so sorry for delay! Sorry away from home for a bit and I forgot personal laptop. 

I do not want to slow the team down at all, so of course would not mind! Thanks for keeping me in the loop.

mlissner:
We'll wrap it up. Thanks @jbgolds for getting us so far!

elisa-a-v (approved):
Thanks! I just moved the notice above the box, this is how it looks now:

<details>
<summary>Desktop</summary>

image

</details>

<details>
<summary>Mobile</summary>

image

</details>

Any final notes before we merge this @mlissner ?

mlissner:
Nope. Merging!

PR #5697, opened 2025-06-08
feat(api): Switch to caching raw data over rendered responses

ERosendo:
In #5694, we introduced a mixin class to cache API responses that met certain criteria. Our initial approach was to cache the rendered page for **10 minutes**. However, I overlooked that DRF allows the use of [Query parameter formats](https://www.django-rest-framework.org/api-guide/format-suffixes/#query-parameter-formats) to switch between different available representations. This is problematic because our mixin wasn't designed to change the data's representation when serving requests from the cache once the response was stored.

To fix this, I initially considered tweaking the cache key to include the format, allowing us to store all different representations. However, I realized it's simpler to store the **raw response data** (_queryset results, count url, next page url and previous page url_) in the cache and let DRF's Response class handle the representation. This updated approach also streamlines the mixin class because we no longer need to use the `add_post_render_callback` method to store the response in cache.

This PR updates the mixin class to store the response's data instead of the rendered response. It also includes a small validation to handle existing cached data. I don't believe we'll need to manually tweak the cache, as we've set a short timeout, and this PR adds a check for legacy data.

### DIFF
--- a/cl/api/utils.py
+++ b/cl/api/utils.py
@@ -1,7 +1,6 @@
 import logging
 from collections import OrderedDict, defaultdict
 from datetime import UTC, date, datetime, timedelta
-from functools import partial
 from itertools import batched, chain
 from typing import Any, TypedDict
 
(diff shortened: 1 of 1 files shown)


mlissner (approved):
Actually makes the code simpler too. Thanks.

Issue #5598, opened 2025-05-20
Impossible to delete alert while editing it


mlissner:
Earlier today I got an alert I wasn't interested in anymore: 

![Image](https://github.com/user-attachments/assets/febc338d-5475-4e0c-9050-212e5f6bde1e)

My options in the email are to Disable or Edit it, but what I really want is to Delete it. 

I was hoping that editing it would make it easy to delete, but this is what I see: 

![Image](https://github.com/user-attachments/assets/9a08b3f6-ce53-4a8e-b5ec-05d41b305753)

If I click on save, it lets me delete it: 

![Image](https://github.com/user-attachments/assets/a9316762-1f42-40a1-8b3b-38c4d9e1a4e2)

I mean...OK, I guess. 

----

To make this better: 

1. We can add a delete button to the banner, and maybe convert the Cancel button into an [X]

2. Pressing delete stays on this page. It verifies the action,  does the delete, then shows a toast saying it's complete.

3. Scope creep: Pressing save when editing or creating a new alert should show a toast saying it's complete too, and should have a link to the alerts page or to continue editing (I'm not sure), but saving and going to the Search Alerts page like we do now isn't great.

ERosendo:
> To make this better:
> 
> 1. We can add a delete button to the banner, and maybe convert the Cancel button into an [X]
> 
> 2. Pressing delete stays on this page. It verifies the action, does the delete, then shows a toast saying it's complete.

@mlissner Here's a GIF showing these two tweaks in action

![Image](https://github.com/user-attachments/assets/9c928b8a-fa22-49ff-ae2e-57284292250e)

I'm wondering, since the `delete` button is now in the banner above the search bar when editing an alert, should we consider taking it out of the modal that displays when you click `save`? I've attached a screenshot of the modal for your reference

<img width="604" alt="Image" src="https://github.com/user-attachments/assets/243fe17f-6099-4fb8-9364-403dded5e878" />

PR #5700, opened 2025-06-09
feat(search): Refines UI to edit alerts

ERosendo:
Key changes: 

- **Removes redundant include within the `search.html` template**: `Messages` were appearing twice on search pages because `search.html` extends `base.html`, which already includes these elements.

- **Refines the banner for editing alerts**: This PR adds a delete button to the banner and changes the "Cancel" button to an "X" icon. Additionally, the view handling delete requests now redirects to the search page after an alert record is successfully deleted when the `redirect_to_search_page` input is included in the request 

Here's a GIF showing the new logic in action: 

Image

This PR addresses the first two suggestions from [#5598](https://github.com/freelawproject/courtlistener/issues/5598)


cl/search/templates/search.html
```diff
@@ -174,21 +173,23 @@
...
+                <div class="col-xs-12 flex gap-2">
+                  <div class="flex justify-content-left">
+                    <p class="bottom">You are editing your alert,
+                      "<strong>{{ alert_form.name.value }}</strong>".</p>
+                  </div>
+                  <div class="flex ml-auto gap-1">
+                    <form action="/alert/delete/{{request.GET.edit_alert}}/" method="post" class="float-right">
+                      {% csrf_token %}
+                      <input type="hidden" name="redirect_to_search_page" value="on">
+                      <button type="submit" class="btn btn-danger btn-sm">Delete alert</button>
```
mlissner:
```suggestion
                      <button type="submit" class="btn btn-danger btn-sm">Delete Alert</button>
```

mlissner (dismissed):
This is nice progress, thanks. One typo, I think, and can you make it so that there are three big buttons that do:

    [ Save Alert ] [ Delete Alert ] [x]

So we don't need the little X in the upper right, it can just be a big X button among the others? I *think* this will work OK?

ERosendo:
@mlissner thanks for the feedback. Here's a GIF showing the banner after the refactor: 

Screen Recording 2025-06-09 at 5 32 25 PM

mlissner:
I can't tell from the gif, but is the Delete button slightly taller than the others?

ERosendo:
> is the Delete button slightly taller than the others?

@mlissner I'm glad to know I'm not the only one. I thought I was going crazy! 🫠 This button's height has been bugging me. I've even checked with the dev tools and it confirm the heights are the same, but it still doesn't look right to me. Here's a GIF of what I'm seeing with the dev tools:

Screen Recording 2025-06-09 at 5 48 09 PM

https://github.com/user-attachments/assets/58307063-3c5b-4dbc-836f-c3c7a859a267

mlissner:
It must be an optical illusion caused by the gray shadow vs. the red. 

I'll see if it bugs me once it's live. Let's get this one reviewed. 

Thank you!

cl/search/templates/search.html
```diff
@@ -174,20 +173,26 @@
...
+                    <div>
+                      <a class="btn btn-default btn-sm" data-toggle="modal" data-target="#modal-save-alert">Save Alert</a>
+                    </div>
+                    <form action="/alert/delete/{{request.GET.edit_alert}}/" method="post" class="float-right">
+                      {% csrf_token %}
+                      <input type="hidden" name="redirect_to_search_page" value="on">
+                      <button type="submit" class="btn btn-danger btn-sm">Delete Alert</button>
+                    </form>
+                    <div>
+                      <a class="btn btn-default btn-sm" href="/?{{alert_form.query.value}}">
```
elisa-a-v: I noticed the `href` in the cancel button changed, was this intentional? It was `?{{ get_string_sans_alert }}` before and now it's `/?{{alert_form.query.value}}`. I don't see any practical difference when testing but from what I see, the `get_string_sans_alert` does more than just pull the query value, so I may be missing some testing cases that show the difference?

elisa-a-v (changes requested):
Thanks for the PR! I just left one tiny question about one code change, but it works nicely in all browsers. However, I noticed the issue says:

> Pressing delete stays on this page. **It verifies the action**, does the delete, then shows a toast saying it's complete.

But there's no verification step, which I think is crucial with an action like this that can't be easily undone. Are we sure we're okay without it? Deleting from the modal redirects to the confirmation page instead of deleting straight away:

Screenshot from 2025-06-10 18-43-03

Probably worth to keep it consistent, _especially_ since the new Delete button looks the same as the previous Cancel button, and it's almost in the same place, so heavy users are bound to confuse the two.

> It must be an optical illusion caused by the gray shadow vs. the red

Yep, I think this is exactly it. When you change the border color from light grey to black, they look way more similar to me:

Screenshot from 2025-06-10 18-07-25

And adding a lighter border like `#d1aaa8` to the red button helps:

Screenshot from 2025-06-10 18-16-20

I don't think it matters too much though, what do you think? Wouldn't block merging for this, but I do think we need to address the verification step ☝🏽

mlissner:
I noticed the absence of verification too. I think it might be OK because it's very easy to recreate the alert. You just press the create alert button again on that page. It does seem like an extra step would help though.

cl/search/templates/search.html
ERosendo: Thanks for pointing this out. That wasn't intentional; I must have forgotten to revert the value to its previous state.

ERosendo:
@elisa-a-v I've updated the banner to use the confirmation step. Additionally, I've adjusted the `delete_confirm.html` template to be more flexible. This was necessary because the "go back" button required more elaborate logic to redirect to the previous page. Here's a GIF showing the updated banner:

Screen Recording 2025-06-11 at 5 11 08 PM

Issue #5601, opened 2025-05-21
Optimize `OpinionCluster.citation_count` update queries


grossir:
Also surfaced on RDS Perfomance insights

https://github.com/freelawproject/courtlistener/blob/491664a17ee3db566d2beabcf82d084b88ab91c9/cl/citations/tasks.py#L267-L273

grossir:
Computing the `OpinionCluster.citation_count` is fast enough at around 1 minute. The update will surely take longer...

```sql
SELECT
    cluster_id,
    sum(cited_count) as citation_count
FROM
(
    select 
        cited_opinion_id, 
        count(*) as cited_count
    from search_opinionscited
    group by  cited_opinion_id
) a
INNER JOIN
    (select id, cluster_id from search_opinion) so ON so.id = cited_opinion_id
GROUP BY 
    cluster_id


                                                                                QUERY PLAN                                                                                 
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 HashAggregate  (cost=1354412.02..1356418.58 rows=160525 width=36) (actual time=59422.667..61417.774 rows=3063512 loops=1)
   Group Key: search_opinion.cluster_id
   Batches: 21  Memory Usage: 131129kB  Disk Usage: 93448kB
   ->  Nested Loop  (cost=516037.77..1353609.40 rows=160525 width=12) (actual time=12486.330..57892.482 rows=3126897 loops=1)
         ->  Finalize HashAggregate  (cost=516037.33..517642.58 rows=160525 width=12) (actual time=12486.308..14790.195 rows=3126897 loops=1)
               Group Key: search_opinionscited.cited_opinion_id
               Batches: 5  Memory Usage: 131121kB  Disk Usage: 146392kB
               ->  Gather  (cost=480721.83..514432.08 rows=321050 width=12) (actual time=7940.907..10177.767 rows=6819287 loops=1)
                     Workers Planned: 2
                     Workers Launched: 2
                     ->  Partial HashAggregate  (cost=479721.83..481327.08 rows=160525 width=12) (actual time=7935.191..9373.322 rows=2273096 loops=3)
                           Group Key: search_opinionscited.cited_opinion_id
                           Batches: 5  Memory Usage: 131121kB  Disk Usage: 81096kB
                           Worker 0:  Batches: 5  Memory Usage: 131121kB  Disk Usage: 84744kB
                           Worker 1:  Batches: 5  Memory Usage: 131121kB  Disk Usage: 81136kB
[...]

PR #5704, opened 2025-06-10
feat(citations.models): create model UnmatchedCitationFromRECAP

### LINKED ISSUE #5492: Collect UnmatchedCitations from RECAP
This will need a model change, since we store `UnmatchedCitation.citing_opinion_id`. In this case, the citing document is not an opinion but a `RecapDocument`

grossir:
Solves #5492

- Creates new model UnmatchedCitationFromRECAP, to be used in `recap_citations.store_recap_citations`
- add migrations for that model
- Creates an UnmatchedCitationBase model to re use code
- refactors `unmatched_citations_utils` methods to work with a RecapDocument as a citing object, from working only with Opinion objects
- adds tests
- update signal on Citation creation to update UnmatchedCitationFromRECAP status to FOUND


grossir:
Just realized we don't update `OpinionCluster.citation_count` on the `store_recap_citations` function... should we?
This may be another way in which that field is not accurate
https://github.com/freelawproject/courtlistener/issues/5601

mlissner:
> Just realized we don't update OpinionCluster.citation_count on the store_recap_citations function... should we?

No, at least not for now. I think knowing how many times an opinion is cited by motions and briefs is interesting, and might even be a better signal for whether it's relevant case law, but I think for now we're not going there. :)

cl/citations/models.py
```diff
@@ -1,56 +1,54 @@
...
+from cl.search.models import BaseCitation, Citation, Opinion, RECAPDocument
 
 
-class UnmatchedCitation(BaseCitation):
-    """Keep track of citations that could not be resolved to a cluster on the
-    batch citator run
+class BaseUnmatchedCitation(BaseCitation):
+    """
+    Track citations that couldn't be resolved to a RecapDocument or
+    OpinionCluster
```
quevon24: Opinion instead of OpinionCluster? citing objects point to Opinion or RECAPDocument

cl/citations/models.py
```diff
@@ -124,3 +115,89 @@ def create_from_eyecite(
...
+        :return: an UnmatchedCitation object
+        """
+        unmatched_citation = cls.create_from_eyecite_base(
+            eyecite_citation, has_multiple_matches
+        )
+        unmatched_citation.citing_opinion = citing_opinion
+        return unmatched_citation
+
+
+class UnmatchedCitationFromRECAPDocument(BaseUnmatchedCitation):
```
quevon24, line 161: Register the new model in admin? As is being done with UnmatchedCitation

cl/citations/tests.py
```diff
@@ -629,18 +632,27 @@ def test_unsafe_case_names(self) -> None:
...
+        """
         cls.recap_doc = RECAPDocumentFactory.create(
-            plain_text="In Fisher v. SD Protection Inc., 948 F.3d 593 (2d Cir. 2020), the Second Circuit held that in the context of settlement of FLSA and NYLL cases, which must be approved by the trial court in accordance with Cheeks v. Freeport Pancake House, Inc., 796 F.3d 199 (2d Cir. 2015), the district court abused its discretion in limiting the amount of recoverable fees to a percentage of the recovery by the successful plaintiffs. But also: sdjnfdsjnk. Fisher, 948 F.3d at 597.",
+            plain_text=plain_text,
             ocr_status=RECAPDocument.OCR_UNNECESSARY,
             docket_entry=DocketEntryFactory(),
         )
         # Courts
-        court_ca2 = CourtFactory(id="ca2")
+        court_ca2 = cls.court_ca2 = CourtFactory(id="ca2")
```
quevon24, line 655: why this double assignment?

cl/citations/models.py
```diff
@@ -124,3 +115,89 @@ def create_from_eyecite(
...
+    )
+
+    class Meta:
+        indexes = [
+            models.Index(
+                fields=["volume", "reporter", "page"],
+            )
+        ]
+        unique_together = (
+            ("citing_recapdocument", "volume", "reporter", "page"),
```
quevon24: Why dont use UniqueConstraint instead? this is a new model

cl/citations/unmatched_citations_utils.py
```diff
@@ -49,62 +54,71 @@ def unmatched_citation_is_valid(
...
         else:
             if found.status in [
-                UnmatchedCitation.FAILED,
-                UnmatchedCitation.FAILED_AMBIGUOUS,
+                BaseUnmatchedCitation.FAILED,
+                BaseUnmatchedCitation.FAILED_AMBIGUOUS,
             ]:
                 continue
-            found.status = UnmatchedCitation.FAILED
+            found.status = BaseUnmatchedCitation.FAILED
```
quevon24, line 90: Cant we do this in bulk?

quevon24 (commented):
Just a few comments, nothing blocks the merging of this PR

cl/citations/unmatched_citations_utils.py
grossir: yes, I added this to the new commit

cl/citations/models.py
grossir: added this to the new commit

cl/citations/models.py
grossir: added this to the new commit

cl/citations/models.py
grossir: added this to the new commit

cl/citations/tests.py
grossir: it's just because court_ca2 is used again below in that same func; but we also need to reference it in other funcs. So both names are used

quevon24 (approved):
Thanks for addressing my comments, It is ready.

Issue #5706, opened 2025-06-10
Update `count_citations` command for bulk update


grossir:
We need to update the `OpinionCluster.citation_count` field for all clusters, after we finish running the citation [finder](https://github.com/freelawproject/courtlistener/issues/5015). We turned it off due to performance [issues](https://github.com/freelawproject/courtlistener/issues/5652). 

We already have a [command](https://github.com/freelawproject/courtlistener/blob/main/cl/citations/management/commands/count_citations.py) for updating the `citation_count`, but 
- it takes a list of ids
- it would be to slow

https://github.com/freelawproject/courtlistener/blob/6ec5d1a5d71ebf94ba349bd8066dbe37250674b9/cl/citations/management/commands/count_citations.py#L23-L33

We need to add an option for updating all the counts, and make it more efficient

grossir:
Let's add a way to bulk update the citation count without locking up the whole `OpinionCluster` table during the [update](https://github.com/freelawproject/courtlistener/issues/5015#issuecomment-3128448626)

I am drafting the following
- add `start_cluster_id` and `end_cluster_id` arguments. Use these to generate pairs of STEP_SIZE
- for each pair:
          - compute the counts from OpinionsCited for all clusters between a generated pair

Just do it sequentially. This should avoid locking the whole OpinionCluster table, and still get some bulk speed benefits.
There are 10M opinion clusters, so if STEP_SIZE is 100k, we should have 100 steps
I still need to profile the resulting query to see the speed of the whole process

grossir:
If we use 10_000 rows as step size
-  will need 1000 steps for the 10M opinion clusters
- the update takes 80 seconds
- there is a harcoded 10 seconds sleep between each tick

Total of 90 seconds per step

We can run this during the night to get better stats. Say we start at 10PM server time, and run it for 6 hours, just to be safe:
(6*60*60)/90 = 240

So, the command to run tonight will be

```
nohup ./manage.py count_citations --count-from-opinions-cited --start-cluster-id 1 --end-cluster-id 2400000 --verbosity 3 > /tmp/log_count_citations.txt &
```

If everything remains stable, we should finish the count update in 5 nights, at this rate

-------------

For some reason Grafana was not picking up my pod logs; so I had to use a batch script to get timestamps on the log update, without much granularity. The updates were very fast; so I will just continue running this and see if we can get it done tonight

```
2025-08-15 02:53:04 INFO Finished citation_count update
[...]

PR #5718, opened 2025-06-11
feat(count_citations): add option for bulk update

grossir:
Solves #5706

Add an option `--count-from-opinions-cited` to the command that updates the `citation_count` for all the clusters, in a single query, which should be a lot faster than the existing method

### DIFF
--- a/docker/courtlistener/docker-compose.yml
+++ b/docker/courtlistener/docker-compose.yml
@@ -79,6 +79,8 @@ services:
         - AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
         - AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
         - AWS_SESSION_TOKEN=${AWS_SESSION_TOKEN}
+        - AWS_DEV_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID}
+        - AWS_DEV_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY}
 
   cl-tailwind-reload:
     container_name: cl-tailwind-reload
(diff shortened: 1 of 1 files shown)


grossir:
If you copy the new ORM code into a Django shell and print the query, you get the query below, which are very close to the queries profiled in this [comment](https://github.com/freelawproject/courtlistener/issues/5601#issuecomment-2901938332), so they should be fast enough


```sql
UPDATE "search_opinioncluster" 
SET "citation_count" = COALESCE(
      (
       SELECT COUNT(U0."id") AS "total_count" 
       FROM "search_opinionscited" U0 
       INNER JOIN "search_opinion" U1 
       ON (U0."cited_opinion_id" = U1."id") 
       WHERE U1."cluster_id" = ("search_opinioncluster"."id") 
       GROUP BY U1."cluster_id"
      ), 
0)
```


Sadly, it's a lot slower than the ones on the linked comment, but still will be fast enough

```sql
courtlistener=> explain analyze (
SELECT
COALESCE(
      (
       SELECT COUNT(U0."id") AS "total_count" 
       FROM "search_opinionscited" U0 
       INNER JOIN "search_opinion" U1 
       ON (U0."cited_opinion_id" = U1."id") 
       WHERE U1."cluster_id" = ("search_opinioncluster"."id") 
       GROUP BY U1."cluster_id"
      ), 
0) as new_count, citation_count, id
FROM "search_opinioncluster" 
);
                                                                                      QUERY PLAN                                                                                       
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Seq Scan on search_opinioncluster  (cost=0.00..23752706266.34 rows=9833614 width=16) (actual time=27.561..417298.020 rows=10129447 loops=1)
   SubPlan 1
     ->  GroupAggregate  (cost=0.88..2415.35 rows=2 width=12) (actual time=0.040..0.040 rows=0 loops=10129447)
           Group Key: u1.cluster_id
           ->  Nested Loop  (cost=0.88..2415.26 rows=13 width=8) (actual time=0.022..0.039 rows=7 loops=10129447)
                 ->  Index Scan using search_opinion_a97b1c12 on search_opinion u1  (cost=0.43..12.46 rows=2 width=8) (actual time=0.015..0.016 rows=1 loops=10129447)
                       Index Cond: (cluster_id = search_opinioncluster.id)
                 ->  Index Scan using search_opinionscited_740050e6 on search_opinionscited u0  (cost=0.44..1198.33 rows=307 width=8) (actual time=0.007..0.021 rows=6 loops=10837287)
                       Index Cond: (cited_opinion_id = u1.id)
 Planning Time: 0.295 ms
 Execution Time: 417861.777 ms
(11 rows)

```

flooie:
This looks good to me- we just need to fix the conflict in tests now and we can get this merged today.

PR #5720, opened 2025-06-11
fix(new_header): avoid dropdown flash on load

### LINKED ISSUE #5717: Header drop downs flash on load (new UI)
The new header is landed and looking pretty great, but when you load the page the menus are open at first and then close immediately. This is most noticeable when I hit refresh a bunch of times: 

https://github.com/user-attachments/assets/9e81ccdd-ba8a-4d59-a477-6a9230561a9c

The menus should be closed on page load so there's no flash.

elisa-a-v:
Fixes #5717 

This was actually an easy fix! Alpine provides a directive exactly for this: [x-cloak](https://alpinejs.dev/directives/cloak).

This happens because the dropdowns are hidden by Alpine, so they are shown until Alpine has initialized. In slower connections, the effect can get much worse! Now we hide them by default, and Alpine removes the attributes once it's initialized.

## Demos

To record both demos, I added "Regular 4G" throttle for the effect to be noticeable:

### Before changes

[Screencast from 06-11-2025 02:49:27 PM.webm](https://github.com/user-attachments/assets/de4e53b6-0daf-4d61-b199-c5dac6525971)

### After changes

[Screencast from 06-11-2025 02:42:03 PM.webm](https://github.com/user-attachments/assets/b357ea3b-3ab7-4dea-9ded-1973f63778b8)

### DIFF
--- a/cl/assets/tailwind/input.css
+++ b/cl/assets/tailwind/input.css
@@ -48,6 +48,9 @@
   td {
     @apply p-2;
   }
+  [x-cloak] {
+    display: none !important;
+  }
 }
 
 @tailwind components;
--- a/cl/assets/templates/cotton/header.html
+++ b/cl/assets/templates/cotton/header.html
@@ -38,6 +38,7 @@
           id="header-support-menu"
           x-data="focus"
           x-show="supportMenuExpanded"
+          x-cloak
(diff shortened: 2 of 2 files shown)


mlissner (approved):
Almost too easy, thanks.

Issue #1958, opened 2022-03-10
Ingest opinions from New Hampshire District Court


mlissner:
For some reason, they're all available here:

https://www.nhd.uscourts.gov/sites/default/files/opinions/

I'm putting them all on our disk storage as well and will push them to S3. The bad news is, no metadata that I can see, so actually making use of these is hard, but this seems to have opinions going back to 92, which is more than what PACER has.

mlissner:
6000 items, 3.1GB.

mlissner:
@flooie, this is an interesting one. No idea if we even care, but we've got this data if we do, I think.

grossir:
@mlissner  do you have the S3 address for the files?

mlissner:
Oof, it's been awhile, but they should be in either our storage.cl.com bucket or (more likely) our private bucket. Could you guys take a look?

grossir:
Just found it in `com-courtlistener-private-storage/sources/nhd/`

mlissner:
Interesting, interesting!

grossir:
From exploring both the S3 subfolders and the DB, it seems we can gain at most 5141 opinions between 1992 and 2020; probably less since some of the PDFs will be repeated, and some files are "summaries" of cases published. See table below for more detail

I sampled 5 pdfs per folder and all did not require OCR, which makes it possible to extract with confidence data points to create the proper objects. Sometimes the document date are in the PDF metadata. See this example

![Image](https://github.com/user-attachments/assets/1586e9eb-8cf2-4d77-b751-d51dc05a0316)

- date is on the PDF title
- docket number: Civil No. 13-cv-447-PB
- a neutral citation? : Opinion No. 2014 DNH 175
- case name is extractable by deleting both docket and neutral citation

In other cases the date is on the first line of text

![Image](https://github.com/user-attachments/assets/888e3787-ea4c-4a2e-b14a-94ea2a03ebbe)

The summary table is mixing a SQL query and a summary from a script that counted the PDFs in the bucket (I let Gemini to do the mixing, and it seems OK at a glance)

| year | cluster count in DB | pdf count in s3 bucket | max_date_filed in DB | min_date_filed in DB | sources in DB  |
|-----------|-------|------------------|----------------|----------------|----------|
| 1992      | 50    | 115              | 1992-12-17     | 1992-01-08     | {LU,U}   |
| 1993      | 65    | -                | 1993-12-30     | 1993-01-04     | {L,LU,U} |
[...]

PR #5723, opened 2025-06-11
Add management command to import manually collected court datasets

quevon24:
This new command will allow us to import opinions from manually collected files, originally intended for this issue: https://github.com/freelawproject/courtlistener/issues/1958

With the data already extracted from the NHD PDFs, we can use the following command to load the opinions:

```shell
docker exec -it cl-django python /opt/courtlistener/manage.py cl_import_dataset_opinions --files-dir /opt/courtlistener/cl/assets/media/nhd --court-id nhd
``` 

**JSON Format**

The command expects each JSON file to follow the same format returned by Juriscraper. Example:

```
  {
    "citations": "2018 DNH 136",
    "docket_numbers": "16-cv-33-PB",
    "case_names": "Frank Staples v. NH State Prison Warden, et al.",
    "case_dates": "2018-07-03",
    "download_urls": "https://www.nhd.uscourts.gov/sites/default/files/Opinions/18/18NH136.pdf",
    "precedential_statuses": "Published",
    "blocked_statuses": false,
    "date_filed_is_approximate": false,
    "case_name_shorts": ""
  }

```

**Juriscraper Integration**
If you’ve run a Juriscraper scraper locally and collected both the JSON and opinion files, you can also call the command using the module name:

```shell
docker exec -it cl-django python /opt/courtlistener/manage.py cl_import_dataset_opinions --files-dir /opt/courtlistener/cl/assets/media/nhd --court-id juriscraper.opinions.united_states.federal_appellate.ca1
``` 

**Court id vs. Juriscraper Module arguments**

- Using a court id (e.g., nhd): The command will rely solely on the metadata provided in the JSON files (case name, docket number, date, etc.) and use doctor to extract content from the opinion PDF or other provided file format.
- Using a Juriscraper module: If the module defines custom logic (e.g., extract_from_text method), the command will invoke those methods to extract additional data.

This design allows the command to be reused for other courts or datasets that follow the same format, making it a flexible for manual imports.


flooie:
We used the guts of this - but are shelving this as its own management command, for the import of a number of states, thanks @quevon24

Issue #5724, opened 2025-06-11
Load docket alerts when the user has some and no search alerts


florean:
If my ongoing quest to optimize my use case, I think it makes sense to automatically load the docket alert page for users that have docket alerts but no search alerts.  If the user has no search alerts, they can't create one from the search alert page, so it serves little purpose.  And if they have docket alerts, they immediately click on that link, incurring an additional page load.  I don't know how many users have docket alerts and no search alerts, but I doubt I'm the only one.

PR #5725, opened 2025-06-11
feat(alerts): Load docket alerts when user has some and no search alerts

florean:
This change adds the simple logic of:
If the user has some docket alerts and no search alerts, load the docket alert page.
Otherwise load the search alert page.
To enable that, there is a new view that implements that logic and a new "profile/search-alerts/" path.  I've staged the commits so that all of the logic and required changes are in the first one and subsequent ones just clean up various references that should use the new search alert URL.
Resolves https://github.com/freelawproject/courtlistener/issues/5724

Testing:
I added both the search and docket alert pages to the basic "will it return a 200" test and I manually tested viewing, creating, and deleting all of the possible combinations of alert types.


mlissner:
Looks good at a skim. Scheduled for full review in our next sprint. Thank you!

Issue #5642, opened 2025-05-30
Introduce operators modal in search bar


elisa-a-v:
Should be opened when pressing the brackets button in the search bar.

![Image](https://github.com/user-attachments/assets/d4c83cf0-40d3-43fa-b8f0-8152d7aac78b)

Issue #5640, opened 2025-05-30
Make header search bar functional


elisa-a-v:
First version of the header doesn't include searching capabilities, so this needs to be implemented.

This is only for the search feature. Other related changes will be addressed in their own issue (mobile version of search bar, and operators modal toggled by brackets button in search bar).

PR #5743, opened 2025-06-13
feat(new_header): introduce mobile version of header search bar

### LINKED ISSUE #5641: Introduce mobile version of header search bar
Note: no bell button! (design hasn't been updated yet)

Image

The above modal should appear when clicking the magnifier button in the header, which also needs to be added:

Image

elisa-a-v:
Closes #5641 

This PR introduces a new Corpus Seach component composed of a few sub-components:

| File          | Component call             | Description                                                                                       |
| ------------- | -------------------------- | ------------------------------------------------------------------------------------------------- |
| `button.html` | `<c-corpus-search.button>` | Search button that performs the search.                                                           |
| `index.html`  | `<c-corpus-search>`        | Main component inside which all sub-components should be contained.                               |
| `input.html`  | `<c-corpus-search.input>`  | The input for search terms.                                                                       |
| `mobile.html` | `<c-corpus-search.mobile>` | Wrapper for the contents that displays a button that triggers a dialog with the actual search UI. |
| `scope.html`  | `<c-corpus-search.scope>`  | Selector for the search scope (Case Law, RECAP Archive, Oral Arguments, Judges)                   |

This gives us a lot of flexibility to determine the final contents, style and distribution of the search interface from the parent template, so we now use the same component in both desktop and mobile search bars in the header, despite them looking significantly different.

### Notes

- This PR introduces the first [`Alpine.store`](https://alpinejs.dev/globals/alpine-store) in CourtListener, which provides persistence across components for the search values. This means now both mobile and desktop versions share the same values: changing one will reflect on the other. If we didn't use a store, because they are different instances of the Corpus Search component, they have different `Alpine.data` instances too, so changing the viewport size would change the selected values, which seems confusing imo.
- As was the case in #5643, the header as a whole is still WIP, so some features **are not working yet**:
    - The actual search feature hasn't been implemented so the Search button doesn't do anything yet (https://github.com/freelawproject/courtlistener/issues/5640)
    - The special operators modal hasn't been implemented so the brackets button doesn't do anything yet (https://github.com/freelawproject/courtlistener/issues/5642)
- I couldn't find a design in the figma file for the open dropdown so I used the one in the [Inputs page](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=7710-70513&m=dev).

### Screenshots

<details>
<summary>Dialog closed</summary>

image

</details>

<details>
<summary>Dialog open, dropdown closed</summary>

image

</details>

<details>
<summary>Dialog open, dropdown open</summary>

image

</details>

### Demo

[Screencast from 06-13-2025 03:58:24 PM.webm](https://github.com/user-attachments/assets/9593c726-15fd-409a-a37a-8d937bb6ecc8)


mlissner:
Pretty cool and looks great. I glanced at the code, but I'll let Eduardo do the full look!

cl/assets/templates/cotton/header.html
```diff
@@ -74,92 +74,54 @@
...
+            :request="request"
+            only
+          />
+          <button
+            type="button"
+            class="btn w-full text-primary-600 text-sm font-medium my-2 hover:bg-greyscale-100 flex justify-center"
+          >Advanced Search</button>
+          <c-corpus-search.button class="w-full flex items-center justify-center">Search</c-corpus-search.button>
+        </c-corpus-search.mobile>
+      </c-corpus-search>
```
ERosendo: I've noticed a couple of UI differences between our component and the Figma mockup, specifically concerning the Search button and the spacing around the "Advanced search" link.

Take a look at this side-by-side comparison, with our component on the left:

Screenshot 2025-06-17 at 8 33 36 AM

Upon closer inspection, I found two key issues:

- **Font Size:** The Figma mockup's Search button uses a 16px font size, while our version is currently at 14px.
- **Spacing:** While the "Advanced search" link itself has correct spacing, the gap-2 class applied to the wrapping div is adding an extra separation between the button and the link.

Here's a video further comparing our HTML with the mockup:



https://github.com/user-attachments/assets/c9eacc9e-ee52-4a7b-88f9-6adbe9e5121b

ERosendo (changes requested):
Thanks @elisa-a-v! The new components look good in Chrome, Safari, and Firefox. I've left a minor comment about a small difference between our component and the mockup. Once that's addressed, we can merge this PR.

elisa-a-v:
Oooh good catch, thank you @ERosendo! I've tweaked the spacing to match the design, and thanks to your screenshots I also noticed the search button was slightly bigger in the design, so I fixed that too.

It now looks like this:

image

Issue #5730, opened 2025-06-13
Allow users to click through to PACER for documents that are listed as unavailable on RECAP


v-anne:
Until the pray-and-pay feature drastically increased how many documents were marked as sealed/unavailable for purchase on CourtListener, it was very uncommon that links to PACER were unclickable. But I think users should still be able to access the PACER link on the "This Item is Unavailable" buttons. I think this is necessary because there appear to be some cases where a document is unavailable on RECAP but actually available on PACER. The changes we landed in #5398 attempted to fix that but it wasn't 100% perfect.

<img width="878" alt="Image" src="https://github.com/user-attachments/assets/9d395fa7-9802-4138-956e-bddb2a80a255" />

One potential way to do this is to have a dropdown like "Download PDF", with the option being displayed as "Check PACER" or something like that.

<img width="878" alt="Image" src="https://github.com/user-attachments/assets/2c89bf80-2865-4b86-b398-01b9f3045a37" />

mlissner:
I like it. Maybe the drop down could say "Check Availability"?

v-anne:
This issue was resolved in #5747. The adjacent idea to have the RECAP extension pick up the (un)availability of documents will be addressed in https://github.com/freelawproject/recap/issues/397.

PR #5747, opened 2025-06-15
feat(RECAP): restoring links to PACER for documents listed as unavailable on RECAP

v-anne:
This PR resolves the issue identified in #5730.

Various issues related to the pray-and-pay project are causing a subset of RECAP documents to be incorrectly marked as unavailable on PACER. By allowing users to continue to click through to PACER, they can confirm for themselves whether the document is actually unavailable, without needing to do a convoluted docket search on PACER itself.


mlissner:
Looks pretty OK to me. Some screenshots would be helpful. 

Marking for slow queue unless I'm misunderstanding the importance.

ERosendo:
@mlissner The new dropdown looks great! Here's a quick GIF of it in action:

Screen Recording 2025-06-18 at 11 10 42 AM

I think we should also add the "Check Availability" button to each sealed entry of the docket report. Currently, users can navigate to the document details page from the prayer list. However, from the docket report, this direct access is not possible without manually manipulating the URL, which disrupts the workflow. Adding this button would create a consistent user experience.

Screenshot 2025-06-18 at 11 16 01 AM

Let me know what you think.

mlissner:
I agree. Let's add the dropdown consistently. 

I think it should be the pale blue though, not the dark blue, so it blends in a bit better.

ERosendo:
@v-anne, would you be able to tackle those changes?

v-anne:
@ERosendo, I just added the button to the docket report. I'm having some trouble fixing the bootstrap issues, though. Can you modify the color palette and prevent the dropdown from wrapping to the next line on the docket report? I think it was caused by the "Unavailable on PACER" button currently being the widest button permitted in that spot.

image

ERosendo:
@mlissner this video shows a mockup for the button in the docket report:


https://github.com/user-attachments/assets/c8f37ca8-302e-47ce-a305-9ea7514501a3


I tweaked the button's text so we could keep the dropdown menu button on the same line (As @v-anne pointed out, the previous text made the button too wide for that spot.) Plus, the dropdown button now matches the disabled button's style but remains clickable.

let me know what you think.

mlissner:
We can probably just say "Unavailable" in the button. The abbreviation is pretty weird, I think.

v-anne:
I have to say that I think it's clearer to users that they can click the dropdown when it's a different shade of blue from the rest of the button. But if you disagree, I understand.

mlissner:
I agree it's clearer, but I don't like the way it looks. Tradeoffs....

ERosendo:
@mlissner Here's another video showing the button.


https://github.com/user-attachments/assets/d09a5668-27ce-4775-880c-af25a8826789

mlissner:
Yeah, it looks good to me. I think it's OK that it's a bit hard to find. It's not meant for the casual user, but for one that has been using the site for awhile and explored the corners of it. We can also think more carefully about this during the redesign.

mlissner:
So, Eduardo, if you think this is ready to merge, I'm OK with it aesthetically and am good with a merge too.

ERosendo:
Thanks @mlissner

ERosendo (approved):

PR #5755, opened 2025-06-16
Correctly document how to exclude fields from API responses

mlissner:
Our docs were wrong. I think they were from an earlier time, but just not corrected.


ERosendo (approved):
LGTM :shipit:

Issue #5748, opened 2025-06-16
API Improvement Request: Reduce payload size when listing `docket-entries` (Deadline: Mid July)


speedplane:
The `GET /api/rest/v4/docket-entries/` is capped at `page_size=20`, each entry includes the full OCR text of all attached documents (`plain_text`).

Large dockets (>1000 entries) therefore **require ≈50 sequential calls** to get the docket sheet. This is a problem if you want to pull a docket sheet and analyze it.

## Problem Impact
| Impact           | Consumer side                                 | CourtListener side                            |
| ---------------- | --------------------------------------------- | --------------------------------------------- |
| **Latency & UX** | 30‑60 s waits to list a big docket            | Dozens of redundant TLS handshakes per client |
| **Bandwidth**    | >100× the bytes needed for metadata           | Unnecessary egress costs                      |
| **Server load**  | Clients must use complex retry/parallel logic | High request rate & memory pressure           |

I am not the first to be affected, see #5317 for a related issue.

## Proposed Fix

1. **“Lite” response mode**

   - New query flag `include_plain_text=false` **or** a generic `fields=` selector.
   - Omits `plain_text` (and ideally `recap_documents`) unless explicitly requested.
   - Cuts listing payloads by \~95 %.

2. **Increase **page** limit**

   - Allow up to 100‑250 when `include_plain_text=false`.
   - Keeps request count for a 1000+ dockets small.

3. **Backwards compatibility**

   - Current behaviour (full payload, `page_size=20`) remains default.
   - Lean mode is opt‑in, so existing clients are unaffected.

## Sample request

Request:

```http
GET /api/rest/v4/docket-entries/?docket=123456&include_plain_text=false&page_size=100
```

If full OCR is later required:

```http
GET /api/rest/v4/documents/987654/?include_plain_text=true
```

### Benefits

- **Faster syncs:** 1 000‑entry docket drops from \~45 s ~140 MB → \~4 s ~6 MB.
- **Lower load:** \~80 % fewer requests, order‑of‑magnitude less bandwidth.
- **Simpler integrations:** No need for aggressive concurrency/retry logic.

### Roll‑out suggestions

- Ship as an opt‑in flag; document in API guide & changelog.
- After adoption, consider making lite responses the default when `page_size` > 20.

*Thank you as always for the invaluable public service CourtListener provides! *

mlissner:
[...]

PR #5756, opened 2025-06-17
Feature/5748 docket entries page size

speedplane:
Add page size to docket-entries. Added a test that passes, but have not tested in localhost dev server. See #5748.


mlissner:
Thanks Michael. Let's see if we can do it upstream before doing this approach, if we can.

Issue #5676, opened 2025-06-04
Design: Mobile version of Search Operators Modal


s-taube:
We need a design for the mobile version of the [search operators modal](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=9024-93028&t=LsEywMwnQEHx9kjz-0).

mlissner:
Magnet team, please see https://github.com/freelawproject/courtlistener/pull/5757, where we've implemented this already. If we think that's fine, maybe we don't have to do any more here. if we think it's problematic for some reason, let's chat about what this design might mean. 

Either way, we should probably include this in the figma design, right?

Andrew1451:
We think that looks good! We also think it's good to have in Figma in case you want to reference it in the designs or update the design later. 

We added that in Figma [here](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=10014-62041&t=0wucngxqn5lVwLMH-4).

s-taube:
@Andrew1451 it looks like an older version of the modal was updated. We approved [this one](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=9024-93028&t=LsEywMwnQEHx9kjz-0) and you can find details in [this figma commen](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4?node-id=9024-93028#1140577748)t if helpful.

Irakli12:
@s-taube Updated it to newer version

s-taube:
@Irakli12 it looks like the wild card characters need to be updated.

PR #5757, opened 2025-06-17
feat(new_header): introduce operators modal in search bar

### LINKED ISSUE #5642: Introduce operators modal in search bar

elisa-a-v:
Closes #5642

## Changes

Besides the Operators modal, this PR introduces:
- A Dialog component with its corresponding library entry.
- A Library component to display list sections for props and slots. All previous library entries were refactored to now use this component.
- Minor fixes:
    - In mobile dropdown version of navigation menu, sub-items were only clickable over text. We can now click anywhere on the item row.
    - First item in the Component library is now highlighted on load.

#### Note
I used the design for the confirmation to delete account ([figma](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=6896-42577&m=dev)) for the mobile version of the operators modal, because both have the same layout in their desktop version. @s-taube I noticed #5676, do we need confirmation from the design team?

## Screenshots

<details>
<summary>Desktop Operators modal</summary>

image

</details>

<details>
<summary>Mobile Operators modal</summary>

image

</details>

<details>
<summary>Dialog closed in component library</summary>

image

</details>

<details>
<summary>Dialog open in component library</summary>

image

</details>

## Demos

<details>
<summary>Desktop</summary>

[Screencast from 06-16-2025 11:41:21 PM.webm](https://github.com/user-attachments/assets/790b4db3-602f-46cc-a131-92a393627442)

</details>

<details>
<summary>Mobile</summary>

[Screencast from 06-16-2025 11:43:31 PM.webm](https://github.com/user-attachments/assets/68ea6c43-4e6f-4701-a123-9182d0371b3b)

</details>


mlissner (approved):
Looks OK to me on a skim of the code, and the visuals look great.

elisa-a-v:
Okay first of all, I love that you caught that, because no that was not intended and it's pretty terrible in terms of accessibility, so thank you!

I do believe `x-trap` is the culprit here and I tried to get it to work but to no avail. However, before I could figure out if the issue was on how we were using the directive of if it's even an Alpine bug, I realized the behavior I was trying to achieve is actually **not** WCAG compliant for a menu like the scope selector!

Turns out pressing either <kbd>tab</kbd> or <kbd>shift</kbd>+<kbd>tab</kbd> while a menu item is focused should _not_ move focus between menu items, but rather it should close the menu and move focus **out** of it. `x-trap`  is intended for elements that block interaction within them when open, like [modal dialogs](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-modal), but menus should not prevent interaction outside them.

Given this, we now:
- close the scope menu and focus the input when pressing <kbd>tab</kbd>
- close the scope menu and focus the trigger button when pressing <kbd>shift</kbd>+<kbd>tab</kbd>

When looking at this I realized the accessibility for both the scope menu and the dialog component could be improved too, so a few ARIA attributes were added to achieve this.

Some references for these changes:
- [ARIA Authoring Practices Guide on Menu Buttons](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/)
- [ARIA Authoring Practices Guide on Menus](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/#keyboardinteraction) -> the <kbd>tab</kbd> and <kbd>shift</kbd>+<kbd>tab</kbd> specifications are described here.
- [mdn web docs on ARIA dialog role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/dialog_role)

cl/search/templates/cotton/corpus_search/scope.html
```diff
@@ -26,33 +28,42 @@
...
     x-show="scopeMenuExpanded"
+    tabindex="-1"
     x-cloak
     x-on:click.outside="closeScopeMenu"
-    x-on:keyup.esc="closeProfileMenu"
-    x-on:keyup.right="focusNext"
-    x-on:keyup.left="focusPrevious"
+    x-on:keyup.esc.stop.prevent="closeScopeMenu"
+    x-on:keyup.right.window.prevent="focusNext"
+    x-on:keyup.left.window.prevent="focusPrevious"
```
ERosendo, line 39: After reviewing the links you provided, I think that the `focusNext` and `focusPrevious` actions should be mapped to the up and down arrow keys.

I confirmed this behavior by reviewing the [Editor Menubar Example](https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-editor/) from the APG page. The video below compares its behavior with our current menu bar. For your reference, the bottom-left corner of the video displays the keys I'm pressing.


https://github.com/user-attachments/assets/dbef1032-4aec-4ed6-89b4-a0cc7d1584f5

elisa-a-v: Yeah actually that's a great point, thanks. I wasn't sure when I first read it, but I think you're correct and it does feel natural given the items are stacked vertically.

elisa-a-v:
Navigation using <kbd>up</kbd> and <kbd>down</kbd> arrows is now enabled, not just within the scope menu, but the profile menu too which I also noticed could use a wee bit of polishing, so it now also auto-focuses the first element in the menu when opening it, and is also closed when pressing either <kbd>tab</kbd> or <kbd>shift</kbd>+<kbd>tab</kbd> while a menu item is focused.

Really appreciate the attention to detail ✨ thanks @ERosendo !

PR #5761, opened 2025-06-17
feat(api): Prevents caching responses with dynamic field selection

### LINKED ISSUE #5754: New API cache doesn't factor in the fields parameter
Last week or so, we added a new cache for the root of our bigger models. It is good about not caching when certain fields are applied to it. Alas, we didn't think to include the `fields` parameter, so if you use that on the root of the API, you cache weird stuff! (Don't test this!)

ERosendo:
Introduces logic to explicitly prevent API responses from being cached when the request includes `fields` or `omit` query parameters.

Previously, responses containing dynamic field selection via `fields` or `omit` parameters could be cached. This led to inconsistent data being served from the cache, as different fields or omit configurations would retrieve the same cached response, potentially omitting or including data unexpectedly.

Fixes #5754

### DIFF
--- a/cl/api/tests.py
+++ b/cl/api/tests.py
@@ -3618,3 +3618,36 @@ class CacheListApiResponseTest(TestCase):
 
         # Confirm the cache key still does not exist after the request
         self.assertFalse(self.cache.has_key(fake_cache_key))
+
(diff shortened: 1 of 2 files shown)


mlissner (approved):
Looks great, thanks.

PR #5771, opened 2025-06-20
feat(opinion_page): add a button to access RECAP doc attachments

### LINKED ISSUE #5770: Directly access a document's attachments
Most (all?) RECAP documents will list in their description what attachments they have, but they aren't accessible without going to the docket.  It would be nice to be able to access them directly from the document page.  Likewise, there is no way to directly access the main document from an attachment page without going back to the docket (or manually modifying the URL).

florean:
After thinking for a long while about issue #5770, I think I've come up with a good solution.  This modifies the RECAP document page to display a button on attachments or documents with attachments which links to the next attachment associated with the document or the original document if viewing the last attachment, with a pulldown menu that allows direct access to every attachment.  This enables users to interact with the full docket entry without having to go back to the docket.  And it does it without any additional hits to the database.
image
For the UI, I've made what I think are the best choices, but it is subjective and it's ultimately up to your taste.  Some things I struggled with: is "Attachment/Document #" clearer than using a truncated description?  Should the description length be more or less than 30 characters?  Are carets the best way to highlight the current doc in the dropdown?  Should the current doc be hyperlinked in the pulldown?  I've made my choices (no, no, yes, yes), but I can't say they're right for CourtListener.

Testing:  I've done a lot of manual testing while iterating over different designs, but I have limited and imperfect data.  I have covered documents with 0-3 attachments as well as documents and attachments with and without descriptions.  It'd be nice to see it in a staging environment with the documents that have the most attachments, but I think the design will hold up.

Resolves #5770


mlissner:
Thanks for playing with this. I think you've got an OK direction, but I find it confusing that other documents would be listed as actions for the one I'm currently on. 

What I think might work better could be integrating this into the heading: 

image

For example, what if that has a drop down integrated into it where you could select the other items instead of #2-1? 

For testing, I don't know the item with the most attachments, but it's probably somewhere around 100 or something like that.

florean:
So I can see how it would work in your screenshot if clicking on `Attachment #1` popped up a dropdown to select other attachments.  But how do you get there from the root document?  What does the header look like for a document that has no attachments and a document that has attachments?

mlissner:
Maybe we do a really simple thing and just have a dropdown at the end of that heading. It'd look roughly like this: 

image

In the menu, you could have something like: 

```
Main Document
Att. 1 - Exhibit A
Att. 2 - Civil Cover Sheet
...
```

Notes:
 - I'm not sure if a bullet (`•`), whitespace, a dash or an em-dash is best as the divider in there.
 - I don't think it needs to indicate which you're on. You know that from the header already.
 - I think if we do this, we can make it so that the `#1` in this heading isn't a link anymore, since this is a more useful approach.

What do you think? It's less discoverable, but advanced users will find it.

florean:
Ah, gotcha, I think that clears it up for me.  I'll take a pass at it.

florean:
Okay, it took a lot of tweaking and some custom CSS, but I'm pretty happy with this.
Let me address your notes:
> I'm not sure if a bullet (•), whitespace, a dash or an em-dash is best as the divider in there.

I wasn't clear on what you meant; I think you were referring to a divider between each document's attachment number and its description in the dropdown menu.  I decided to just use the exact same text as the heading.  It is simpler than using a custom formatting and it has a nice continuity when switching between documents.  Of course, the heading can wrap and a menu item can't, but I tested with very long descriptions and it works well enough. 

> I don't think it needs to indicate which you're on. You know that from the header already.

I will respectfully disagree.  Of the three options, leaving the current document in the dropdown list, removing it, or leaving it with a visual indicator highlighting it, I think the latter is better UX.  For the common task of navigating to the next document, the first option requires the user to read each menu item until they find the current one and then select the next item.  The second option requires the user to read each item until they find the one that follows the current document from a menu that changes with each document.  With the last option, the user has a very clear visual indicator where they are in the list that requires no text parsing and can then select one.  It's a tiny thing, but I tried all three and I think it matters.  Plus, it looks much cleaner than the second option.

> I think if we do this, we can make it so that the #1 in this heading isn't a link anymore, since this is a more useful approach.

I think that is orthogonal to this feature, so I left it alone.  This feature helps you navigate documents within a docket entry, but if you want to navigate to other docket entries, the document link is still the best option.  There is always the case name or "Back to Docket" links, but I think it is useful to be able to get directly to the docket entry, especially if it's document `#101` instead of `#1`.  If you do remove it, maybe change "Back to the Docket" to include the docket entry?

Here are some screenshots.  A document with attachments will now look like this:
image
And the dropdown looks like this:
image

I tested it in Safari and Mobile Safari.  I have not looked at it in Chrome.  I can check it out later to make sure it looks okay, but I'm happy to take some else's word for it.
I haven't tested it with a large number of documents.  I don't anticipate any problems, but maybe if I get better at automatically generating some data I'll take a look at it.

florean:
Added a shorter tooltip to help with longer titles.
image

mlissner:
I think this looks great. I have some concerns about how it might perform and about its UX once it goes live, but let's give it a go and see how it works.

I'm putting this on the slow queue, but we have a new sprint starting Monday, so it'll probably get reviewed around then.

Thanks!

florean:
Since I don't think anyone has really reviewed the code yet, do you might if I do a rebase?  Half the commits at this point are merges from `main` and I really don't want to contribute to the 25% of `main` commits that are just merges from `main`.

> I have some concerns about how it might perform and about its UX once it goes live, but let's give it a go and see how it works.

Can you elucidate?  Are you concerned about the menu items?  Would you prefer a `{{ doc.document_number }} – {{ doc.description }}` style?  I kind of liked the consistency with the heading, but that format puts the most important information at the beginning (plus I can move the heading style to the tooltip).  It's really a 50/50 call to me and it's your site.  I'll be a little surprised if you get critical feedback once it goes live, so now's the best time to make any tweaks or changes you want.  At the very least, I can fork this branch and you can compare the two to see what feels best.  (If you don't have a staging server, maybe we could hide both behind Waffle flags and merge them after review?  Not sure how difficult creating new flags is, but it'd be an easy change to `view_recap_document`.)

I'm a little proud of how this turned out, especially being able to do it while increasing performance, and I'd like it to be the best it can be.  I'm new to your design and taste (as evidenced by the initial button) and I'm open to feedback.

Cheers!

P.S.  On, and I'll add a test case for `view_recap_document` before merge.

florean:
BTW, I sent a Contact Us message asking for some API token privileges to help with testing.  Posting here to maybe bump it up the queue.  Thanks!

mlissner:
Yes, a rebase would be welcome, thank you, though we don't sweat the git log too much generally. 

I don't have any concrete concerns, really. The mean thing is whether it might have some issue we hadn't thought of when it meets real data, and often I just notice things when I play with them for the first time. 

If you want to put it behind a waffle, that should be easy (search our HTML files for the word waffle and you'll find examples). If not, that's OK too. I think this is safe and I might just ask for a tweak or two once it's live.

I'll get on the API thing in a sec!

florean:
Alright, I'll address any review feedback and leave this in as the default, but put an alternate behind a Waffle flag.  We'll tweak that once it's live and then integrate the improvements.

Thanks for the API access.  I got some good data from it and it was fortuitous, because I'd just accidentally nuked all of my local data with a Docker build.

elisa-a-v:
Ope, I thought this was ready for review. Are you changing anything else here, @florean ?

florean:
Sorry, @elisa-a-v.  No other changes that affect the code.  I am adding tests for `view_recap_document`.  There is currently only one code path tested, so I'm working to get 100% coverage.  I'll try to get it finished by EOD.

elisa-a-v:
Nice, thank you! I'm marking as draft in the meantime, can you mark as ready for review once all tests are pushed please?

Let me know if there's anything you need feedback on before that anyway :)

florean:
Sorry, took a little longer than expected.

florean:
I'm sure you're busy and it probably doesn't make sense to look at this until you can review the whole thing, but the two things I'm concerned about are the custom CSS and the views code.

Thanks!

cl/search/models.py
```diff
@@ -1546,9 +1546,8 @@ def save(
                         "and we are unable to fix it automatically for "
                         f"rd: {self.pk}"
                     )
-                else:
-                    # Only one duplicate. Attempt auto-resolution.
-                    other = others[0]
+                # Only one duplicate. Attempt auto-resolution.
+                other = others[0]
```
elisa-a-v, line 1550: As per the [commit](https://github.com/freelawproject/courtlistener/pull/5771/commits/d04ea3a8c037677fb0566728e9a99c6108c5a745) message:

> All remaining lines in a block relied on a variable defined in a preceding `else` clause.  Those lines should either be moved into the `else` clause or the clause should be eliminated.  I just removed it.

@mlissner could you double check the logic here? it makes sense to me but if this is wrong we could end up deleting data we would have otherwise kept so I thought it best to make sure this is right.

cl/opinion_page/views.py
```diff
@@ -541,10 +541,10 @@ async def docket_authorities(
...
 
 
-async def make_rd_title(rd: RECAPDocument) -> str:
-    de = await DocketEntry.objects.aget(id=rd.docket_entry_id)
-    d = await Docket.objects.aget(id=de.docket_id)
-    court = await Court.objects.aget(id=d.court_id)
+def make_rd_title(rd: RECAPDocument) -> str:
+    de = rd.docket_entry
+    d = de.docket
+    court = d.court
```
elisa-a-v: I think the removal of `async` is safe here only because we're using `select_related` for those three related objects in both method calls to `make_rd_title`, meaning we don't perform further DB queries here but instead retrieve the Docket Entry, the Docket and the Court, all from memory.

I think using this method on a `RECAPDocument` instance that doesn't have at least one of those would incurr additional DB queries and therefore add I/O delays that we were handling via `async`. Please correct me if I'm wrong, but if this is the case, we may want to document this clearly.

cl/opinion_page/views.py
```diff
@@ -577,7 +577,7 @@ async def make_thumb_if_needed(
             klass=RECAPDocument,
             max_dimension=1068,
         )
-        await rd.arefresh_from_db()
+        await rd.arefresh_from_db(fields=["thumbnail_status", "thumbnail"])
```
elisa-a-v: This makes perfect sense to me but I still wonder, @mlissner is there any reason we may want to refresh the entire `RECAPDocument` instance here?

cl/opinion_page/templates/includes/rd_metadata_headers.html
```diff
@@ -1,12 +1,57 @@
...
+                href="{% if doc != rd %}{{ doc.get_absolute_url }}{% else %}#{% endif %}"
+                style="font-size:21px; font-weight:500; padding:3px 18px;"
+                title="{% if doc.attachment_number %}Attachment #{{ doc.attachment_number }}{% else %}Document #{{ doc.document_number }}{% endif %}"
+              >
+                {% if doc.description %}{{ doc.description }}
+                  <span class="gray">&mdash;</span>
+                {% endif %}
+                {% if doc.attachment_number %}Attachment #{{ doc.attachment_number }}{% else %}Document #{{ doc.document_number }}{% endif %}
+              </a>
+            {% endflag %}
```
elisa-a-v: Unless we address the potential performance issue for docs with many attachments some other way, would it make sense to not just provide an alternative but also hide the feature entirely behind the waffle flag until we've tested to see how it performs in prod?

elisa-a-v (changes requested):
First of all, thank you for the PR and for your patience with the review 🙏🏽 

This is a nice feature that I'm sure will boost UX when navigating through a doc's attachments. I really like what you did with the dropdown design, I fully agree with you that highlighting the current item is better for UX, and CSS looks good to me save for one issue described below.

## Observations from testing

### 1. Docs with many attachments

#### Page overflow

For documents with a common number of attachments the dropdown looks great, but when the attachment list grows, the dropdown grows indefinitely too.

For example I found [this item](https://www.courtlistener.com/docket/17366820/akai-custom-guns-llc-v-kkm-precision-inc/?entry_gte=207#entry-207) with 212 attachments, and it overflows the page:

<details>
<summary>Overflow screenshot</summary>

image

</details>

Adding a `max-height` and `overflow: auto` to the dropdown should probably do the trick.

#### Performance considerations

I noticed the number of DB queries is reduced and docs with fewer attachments are indeed quite quicker to load, however for docs with too many attachments the time it takes the DB to perform those queries is considerably longer than what it was before. Hiding the dropdown for docs with too many attachments seems like the easy way out here, but we may want to wait to see how it performs in prod before doing anything about this. I tested with the dev DB which is not all that powerful so this may not be an issue in prod.

#### Tip on data cloning for local testing

> maybe if I get better at automatically generating some data I'll take a look at it.

I think you should be able to clone that docket with all its entries by running this command:

```
manage.py clone_from_cl --type search.Docket --id 17366820 --add-docket-entries
```

You can read more about it in [the command docstring](https://github.com/freelawproject/courtlistener/blob/main/cl/scrapers/management/commands/clone_from_cl.py), let me know if you need help with this.

### 2. Other RECAPDocument tabs

I only see the dropdown button when in the default PDF/Text pages, but it disappears when in the authorities tab:

<details>
<summary>PDF/Text tabs screenshot</summary>

Image

</details>

<details>
<summary>Authorities tabs screenshot</summary>

Image

</details>

The bug persists when changing tabs after navigating to the authorities tab. I noticed the URL doesn't go back to the main doc page but remains in the `/autorities/` url, I'm guessing these issues are related.

---

Great job overall, thank you for your thoughtful contributions! We'll be ready to merge once we've addressed the observations above and settled on the questions I left inline.

cl/search/models.py
mlissner: Tweak is fine. We check if others exist, and they do, so we check if there are more than 1 other. If so, we abort with a ValidationError. So now we know others is greater than zero (they exist), and not greater than 1 (it didn't crash). That can only mean one thing, that there's only one duplicate and we can attempt auto-resolution.

I can't see how it could be wrong.

cl/opinion_page/views.py
mlissner: FWIW, we're moving away from async in general. There's a good chance we'll be yanking it just about everywhere.

cl/opinion_page/views.py
mlissner: It seems fine, but it tethers the logic of what's updated in the `make_png_thumbnail_for_instance` to the details of the refresh. We only use this function in this one place in the code, but maybe the refresh should move inside the function so they're untethered and it doesn't need to happen here.

cl/opinion_page/views.py
florean: @elisa-a-v, correct, Django will load those and I'm avoiding the three DB call that were there because I know they're already loaded.  But I can add a comment reminding people calling it could result in DB loads.

florean:
### 1. Docs with many attachments
> For example I found [this item](https://www.courtlistener.com/docket/17366820/akai-custom-guns-llc-v-kkm-precision-inc/?entry_gte=207#entry-207) with 212 attachments, and it overflows the page:

Yeah, the UI needs to be able to deal with the worse case, even if it's a non-interactive "...and 150 more attachments" or something.  I'll work on that.

> Hiding the dropdown for docs with too many attachments seems like the easy way out here

Unfortunately, at that point the query is done.  How much of a problem is it?  My guess is that is an extreme example, but I don't know the dataset and I don't know what reasonable targets are.  The questions I can't answer are: what's the median, 95th, and 99th percentile for docket entry documents (technically you'd want to calculate it off of actual page views, but this is sufficient).   Could you run this for me on whatever would have a representative dataset?
```SQL
SELECT
	PERCENTILE_CONT(.5) WITHIN GROUP (ORDER BY entry_count) AS median,
	PERCENTILE_CONT(.95) WITHIN GROUP (ORDER BY entry_count) AS p95,
	PERCENTILE_CONT(.99) WITHIN GROUP (ORDER BY entry_count) AS p99
FROM
	(SELECT COUNT(*) AS entry_count FROM search_recapdocument GROUP BY docket_entry_id) AS _ 
;
```
Because I think you'd take the trade if the 95th percentile page load drops, but the 99th goes up a little.  If this a problem, I think what I would do is:
• only get the pk and attachment number in the first query, which is all that's needed for the first few checks.
• get the full RECAPDocument for the correct one.
• get only what I need for the other attachments.
I don't know, I might try that anyway, I might be able to make things a little faster even in the common case.

> Tip on data cloning for local testing

Thanks, I didn't have API access at the time and hadn't messed with factories, so I was generating everything in the admin.  It was painful.

### 2. Other RECAPDocument tabs
> I only see the dropdown button when in the default PDF/Text pages, but it disappears when in the authorities tab

I didn't test any docs with authorities.  I'll load the one from your screenshot.  There is a completely different view for them, `view_recap_authorities`, so is it a different page load?  I'll take a look.  Glancing at the view, at a minimum, there is a lot of common logic that should be factored out.

florean:
Ah, I don't have authorities.  Is there a process I need to run to generate them?

florean:
Yeah, so the Authorities tab is a page load to `.../authorities`, which is a completely different view and template.  Well, I say completely different, but really they're copies.  It is basically a very roundabout way to avoid making [this query](https://github.com/freelawproject/courtlistener/blob/main/cl/search/models.py#L3499) unless absolutely necessary.  And yeah, that doesn't look like a query you want to be making if you don't have to, but isn't this something we could load dynamically with HTMX?

mlissner:
What we do on the docket page (which has similar design of metadata at the top and tabs a few inches down) is have a common template for the top stuff and then the tabs all just use it. I imagine that'd work well here? I don't think moving to HTMX for that is the best plan just because it's already working and we don't usually use it for the main page contents (we use it more for little things like the attachment page, typically). 

There is a process for generating authorities, but I don't know what it is anymore. It's run automatically when we ingest data, so I suspect it didn't run because adding from admin doesn't trigger that code path. The function will be something like `find_citations...`, if I recall correctly. You could also just fake the data, I imagine.

I'll let Elisa help with the rest. :)

elisa-a-v:
> I didn't have API access at the time and hadn't messed with factories, so I was generating everything in the admin. It was painful.

I can imagine! Let me know if the `clone_from_cl` command works for you.

> I don't have authorities. Is there a process I need to run to generate them?

Yes, you'll need to:
1. Clone the appropriate opinions.
2. Generate the authorities by running `find_citations_and_parantheticals_for_recap_documents` for the `RECAPDocument` instances you need.

To figure out which opinions you need to clone you can go to the authorities tab in a docket and you'll get a list of the opinions cited by the docs in that docket. In the list of docket authorities, the number of filings is a link that searches the entries in the current docket that cite that opinion, so you'll get:
1. The opinion cluster id (which you need to clone)
2. The entries that cite that opinion (which you'll generate the authorities for)

With the huge docket provided above, the docket authorities can be found [here](https://www.courtlistener.com/docket/17366820/authorities/akai-custom-guns-llc-v-kkm-precision-inc/), where you'll see the first item in the list shows the `OpinionCluster` id `1844121`. If you click the "[3 filings](https://www.courtlistener.com/?type=r&q=docket_id%3A17366820%20AND%20cites%3A(1844121))" link, you'll see the entries that cite that opinion in that docket.

You only need to clone a single opinion for the authorities tab to appear so after cloning the docket with its entries, you can simply run

```
manage.py clone_from_cl --type search.OpinionCluster --id 1844121
```

Then in a Django shell:

```
In: from cl.citations.tasks import find_citations_and_parantheticals_for_recap_documents
In: rds = RECAPDocument.objects.values_list('pk', flat=True)
In: find_citations_and_parantheticals_for_recap_documents(rds)
```

Since the docket cloned has loads of entries, you may want to tweak the `RECAPDocument` queryset to run this only for the entries that cite the cloned opinion, but either way it's probably fine.

After doing all this, you should see the authorities tab both in the docket page and in the documents that cite that opinion, for example [this one](https://www.courtlistener.com/docket/17366820/207/180/akai-custom-guns-llc-v-kkm-precision-inc/authorities/) should have the authorities tab visible.

---

As for the attachment number distribution, I'm working on getting you a number but I tried running it over the dev DB and got this output:

image

This makes sense because we're counting the docket entries for each `RECAPDocument`, which is always one since it's a FK from `RECAPDocument` to `DocketEntry`. What we want is the other way around: counting the `RECAPDocument` instances for each `DocketEntry`.

I'm no expert in SQL so please correct me if I'm wrong. I'll try other options and keep you posted, but if you want to suggest other queries you think might work please do share them and I'll give them a try :)

florean:
@mlissner 
> I imagine that'd work well here?

Pretty sure that is what we are doing here.  Just hate to do a full page load when all we need all we need is one API call.  But now at least now the authorities tab is getting the same data as the other ones.

@elisa-a-v 
> I can imagine! Let me know if the clone_from_cl command works for you.

Oh, I [made it](https://github.com/freelawproject/courtlistener/pull/5867) work for me.  :)

Thanks for the authorities process.  That would have taken me quite a while to figure out.


> This makes sense because we're counting the docket entries for each `RECAPDocument`

We shouldn't be.  In `FROM search_recapdocument GROUP BY docket_entry_id` we're generating counts of `RECAPDocument`s per `DocketEntry`.  When I run it in `psql` on my local data, I get:
```
 median | p95 |        p99         
--------+-----+--------------------
      1 |   5 | 15.859999999999971
(1 row)
```
If you run just the subquery, are you seeing counts?  Try it (with an added order_by and limit):
```SQL
SELECT COUNT(*) AS entry_count FROM search_recapdocument GROUP BY docket_entry_id ORDER BY entry_count DESC LIMIT 10;
```
You should see the counts of the largest documents per docket entry.  You can add `docket_entry_id` as a column if you want to be able to verify some of them.  Then that is getting fed into the aggregate functions to generate the distributions.

cl/opinion_page/templates/includes/rd_metadata_headers.html
florean: My latest commits should address the performance issues.

cl/opinion_page/views.py
florean: I hopefully mentioned it in the commit, but I should probably document this: if you refresh the entire document, you destroy any related objects.  And those aren't refreshed using async, so it throws an error in any async function that accesses those related objects when the manager tries to refresh them from the DB.  So this function had some very serious side-effects that weren't documented.

florean: Since it is only called in one place, I could just inline this.

cl/opinion_page/views.py
florean: I've added a comment.

cl/opinion_page/views.py
mlissner: I agree, that sounds best.

cl/opinion_page/urls.py
elisa-a-v, line 1: @ERosendo helped with the review today and pointed out something I had previously missed, and it's that the `document_number` is actually a `CharField`, _and_ some documents could start with a `0`. Wouldn't the `int` type cast remove it?

cl/opinion_page/views.py
```diff
@@ -613,18 +616,70 @@ async def view_recap_document(
...
+    response = await recap_document_context(
+        request,
+        docket_id,
+        doc_num,
+        att_num,
+        slug,
+        is_og_bot,
+        "recap_authorities.html",
+    )
+    if isinstance(response, SimpleTemplateResponse):
```
elisa-a-v, line 628: Looks like we're missing the `SimpleTemplateResponse` import 👀

elisa-a-v (commented):
I haven't performed a thorough review of the latest changes but I wanted to share two small comments before you finish!

elisa-a-v:
Bah, didn't notice you had already finished 🤦🏽 my apologies! I'll provide a full review tomorrow, the comments above are small and unrelated to your latest changes so I think it should be fine.

cl/opinion_page/urls.py
florean: Yes, that's a bug. I was matching the view's type hint, but the model is what matters. I'll remove it.

cl/opinion_page/views.py
florean: Yup. I'll add it.

florean:
> I think what I would do is:
> • only get the pk and attachment number in the first query, which is all that's needed for the first few checks.
> • get the full RECAPDocument for the correct one.
> • get only what I need for the other attachments.

I went ahead and implemented this. I haven't done any serious performance testing, but my intuition is that despite the extra query it will be a bit faster in the common case and much faster in the extreme cases.

I'm not entirely happy with the paging (there isn't any), but it's fine. I arbitrarily limited the number of documents displayed to 20, but that can be easily adjusted. If it doesn't blow out the page, I'd set it to at least the 95th percentile.

I also folded the authorities view in, so we don't have to worry about them getting out of sync again.

florean:
> Bah, didn't notice you had already finished 🤦🏽 my apologies!

They're small changes and I'll push them when I have a sec.

cl/opinion_page/views.py
```diff
@@ -679,8 +742,16 @@ async def view_recap_document(
             if rd.pacer_url and redirect_or_modal:
                 redirect_to_pacer_modal = True
 
-    title = await make_rd_title(rd)
-    rd = await make_thumb_if_needed(request, rd)
+    title = make_rd_title(rd)
+    needs_thumb = rd.thumbnail_status != THUMBNAIL_STATUSES.COMPLETE
+    if all([needs_thumb, rd.has_valid_pdf, is_og_bot(request)]):
```
florean: So we have `is_og_bot()` but we also get passed in an `is_og_bot` bool.  Can I believe the passed in value?  And if not, should it be removed and the call be moved further up?  Right now we use both.

florean: So the `is_og_bot` parameter only comes from `redirect_og_lookup` and I think we can just trust it.  No need to double check, right?

florean:
I decided I should do some actual performance testing and wrote a [little script](https://gist.github.com/florean/a8c296c236b319d950dad3f73d68f83e).  Basically five runs calling `view_recap_document` 1,000 times.  I tested loading the main document against three docket entries that I think represent small, medium, and large:
[Main document only](https://www.courtlistener.com/docket/69777799/3/abrego-garcia-v-noem/)
[Main document + 7 attachments](https://www.courtlistener.com/docket/69777799/1/abrego-garcia-v-noem/)
[Main document + 212 attachments](https://www.courtlistener.com/17366820/207/4/akai-custom-guns-llc-v-kkm-precision-inc/)
And I tested against three checkouts: the current `main` (e53c7bf3); my branch using my old queries, which I'll refer to as "before" (8c99ee20); and my latest (7611d4b), which I'll refer to as "after".

The simple case, which is the best case for my versions because it is basically loading the same data as `main`, was pretty predictable:
```
Main (e53c7bf3):
[11.64735375501914, 14.34264000604162, 17.49547384097241, 21.34751738497289, 26.318782678979915]

Before (8c99ee20):
[5.009087044047192, 4.853335001971573, 4.864667211018968, 4.84330529399449, 4.880911460961215]

After (7611d4b):
[5.687711377046071, 6.120853627973702, 6.949965712032281, 6.725790002965368, 6.701520378002897]
```
My `before` is anywhere from two to five times faster than `main`, with `after` being slightly slower due to the extra query.  I have no idea why there is so much volatility is the `main` results, but it was consistent.  Maybe it has something to do with all the extra async calls, but it's behind a `async_to_sync` and all the tests are being run sequentially, so that doesn't make a lot of sense.  But that definitely is concerning.  That said, the main results were the same for the rest of the tests, so aren't really worth remarking on.

It gets more interesting when we get to the medium case:
```
Main (e53c7bf3):
[11.905972671986092, 14.647029672982171, 17.84963767498266, 21.222732093010563, 24.989681761013344]

Before (8c99ee20):
[11.6302562559722, 11.379111879970878, 11.407496421015821, 11.323862922028638, 11.315761254983954]

After (7611d4b):
[7.1753067940007895, 6.353308295016177, 6.515213627950288, 8.341826420975849, 8.516119462030474]
```
Here, the extra data is obviously slowing `before` down, but it still slightly beats `main` in the worst case and is nearly two times faster in the best case.  Not too bad.  However, `after` has only a slight increase in times and is still handily beating `main` by two to three times.

And large is why knowing your worst case scenario is important:
```
Main (e53c7bf3):
[12.874114172998816, 20.350850759015884, 24.177101802022662, 27.82056984497467, 32.30270176497288]

Before (8c99ee20) [Ran out of memory; ran 100 and multiplied by 10]:
[219.240349259926, 207.74459300970193, 207.25348717998713, 206.7424971802393, 206.64192675962113]

After (7611d4b):
[7.373215545027051, 7.893614544998854, 8.325167628994677, 8.559671211987734, 8.314285753003787]
```
Here, even `main` goes up a little bit, but `before` really struggled: it died due to my Docker memory limits (it hit over 6 GB), so I ran it just 100 times and multiple the results by 10.  Maybe ~220ms for a page isn't horrible, but it definitely isn't great.  On the other hand, `after` barely blinked.  It is still crushing `main` and I'm honestly kind of curious to see at what point it would become slower.

Thanks @elisa-a-v for pointing me to that docket entry and prompting the change, it let to a much better version.

florean:
I just noticed I somehow got some `.github` files in a check-in.  Fixing it.

Edit: yeah, I messed up a merge in 0eb11de; I was trying to merge the upstream main because my local one had some unmerged changes, but I accidentally merged it.  I think the easiest thing will be to revert it and do a force push from there.  Sorry.

elisa-a-v:
@florean are you planning on pushing any more changes? it's perfectly fine if you are, but in that case could you please mark as draft and only mark as ready for review once all changes have been pushed? 🙏🏽

cl/opinion_page/views.py
mlissner: Just want to flag that we had an issue with this template type: https://github.com/freelawproject/courtlistener/pull/5962

florean:
@elisa-a-v No, it's done.  I had to force everything after my failed merge, but they were ones that were already there and I just ran a (correct) merge.  There haven't been any code changes since last night.

Instead of merging `upstream/main` because I have a dirty main in order to test templates, I merged `upstream main`.  🤦‍♂️

florean:
I do have an [open question](https://github.com/freelawproject/courtlistener/pull/5771#discussion_r2224126860), but it's very minor and doesn't affect review.

cl/opinion_page/views.py
mlissner: Yeah, I think you're right. We can just trust what comes in and don't need to run the `is_og_bot` function.

florean:
That question is resolved.

mlissner (approved):
I didn't read in great depth, but this one feels fully cooked. Going to trust Elisa's earlier review and Nathan's efforts and merge. I'll revert if I should have looked more closely.

Issue #5206, opened 2025-03-04
Home Page - Search and Filter Tabs


s-taube:
![Image](https://github.com/user-attachments/assets/29b39853-214c-4876-9373-00e37d288e60)

**Refer to the Figma for the most updated design.**

On the CourtListener home page, users can search and filter for any data type (case law, recap archive, oral arguments, or judges) by selecting the relevant tab.

**Notes on the feature:**

- The Case Law tab is selected by default.
- There should not be a “Financial Disclosures” tab.
- The filters available on each tab vary. 
- The filters should be in the same order as they appear in the Figma.
- The description on each tab (e.g. Case Law “10M+ Opinions”) only appears when the tab is selected.
- The heading above the search changes, depending on which tab is selected. (e.g. for RECAP Archive, the heading is: “Search our database of millions of PACER documents and dockets”)
- “About this data” navigates users to the relevant “About Our Data” / “Coverage” page, depending on which tab is selected.
- Search terms and filters do not transfer to other tabs. So, if you enter a search term in the Case Law search bar, and then click the Judges tab, you **do not** see the search term you entered. If you then click back to the Case Law search, you will also **not see the search term you previously entered**, since you didn’t submit the search.

s-taube:
**Questions on the feature:**
- will descriptions on the tab be manually updated as collections grow? ("10m+ docs" to "11m+ docs")
- on the last bullet point, does the functionality look right? Any changes?

s-taube:
@mlissner Before I add a bunch of issues, I want to see if the organization and level of detail looks about right to you. 

This would be one sub-issue under "home page redesign." Other sub-issues would include the "action" buttons (subscribe, create an alert, find a case, etc.), announcements ("recap suite -- turning pacer around...") and product stats (# API calls made...).

mlissner:
For the homepage, which has so much complexity, yes, this looks perfect to me.

On the last bullet point, yes, I think that's fine. It'll annoy some users, but doing it right is hard and out of scope.

elisa-a-v:
> It'll annoy some users, but doing it right is hard and out of scope.

What would be the right way? With Alpine it's pretty easy to keep the values across tabs, in fact it might even be a little easier than making the form clean up when changing the tab 🤔 which behavior do we actually prefer?

mlissner:
It's hard b/c: 
 - there are different filters on different pages, so you can't just take the filters from one tab and apply them to another
[...]

PR #5782, opened 2025-06-23
feat(new_header): enable search from new header search bar

### LINKED ISSUE #4975: [Parent] Do the new header for the new design
Start with components, then actually do it; probably two PRs.

There are two different versions of the header. The header for the homepage is different from the header on all other pages. The differences are described in the sub-issues.

### LINKED ISSUE #5640: Make header search bar functional

elisa-a-v:
Closes #5640 
Closes #4975 

This PR enables the search feature of the Corpus Search component, and with it, the new header is (almost) complete 🎉 The only thing that is not yet implemented is the Advanced Search in mobile, which will be addressed when adding the homepage search and filters tab in #5206 

---

### Some accessibility notes

As noted by @ERosendo , the new search input (`<input type="search">`) sometimes shows a cross icon to remove the entered value, which is done by the browser:

> The main basic differences come in the way browsers handle them. The first thing to note is that some browsers show a cross icon that can be clicked on to remove the search term instantly if desired, in Chrome this action is also triggered when pressing escape.

Source: [mdn web docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/search#differences_between_search_and_text_types) on the difference between `search` and `text` inputs.

This cross icon is not in the design, and different browsers use different styles (some don't even show it at all, like Firefox). However:
- Using this type of input is better for accessibility because it enhances the semantics and marks the search input as an [important landmark](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Accessibility/WAI-ARIA_basics#signpostslandmarks) in the site, providing more information for screen reader users to find important page elements like the search form.
- As discussed with @ERosendo , the icon is actually pretty useful, so it's probably not worth it to try to remove it.

---

## Demo

[Screencast from 06-23-2025 03:50:10 PM.webm](https://github.com/user-attachments/assets/30e0d415-55d4-473f-945a-127604ac2197)

### DIFF
--- a/cl/assets/static-global/js/alpine/components/corpus_search.js
+++ b/cl/assets/static-global/js/alpine/components/corpus_search.js
@@ -2,7 +2,12 @@ document.addEventListener('alpine:init', () => {
   Alpine.store('corpusSearch', {
     scopeMenuExpanded: false,
     selected: 'Case Law',
-    searchScopes: [{ label: 'Case Law' }, { label: 'RECAP Archive' }, { label: 'Oral Arguments' }, { label: 'Judges' }],
+    searchScopes: [
+      { label: 'Case Law', type: 'o' },
+      { label: 'RECAP Archive', type: 'r' },
+      { label: 'Oral Arguments', type: 'oa' },
+      { label: 'Judges', type: 'p' },
+    ],
   });
   Alpine.data('search', () => ({
     get scopeMenuExpanded() {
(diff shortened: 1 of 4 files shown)


mlissner (approved):
So exciting! LGTM on a skim. Onwards!

ERosendo:
Thanks @elisa-a-v! LGTM :shipit:

PR #5786, opened 2025-06-24
fix(opinions.html): Fix Frontend metadata display for case name full and posture

flooie:
PR makes just a small tweak to the frontend.

Frontend incorrectly displays posture metadata field when nature of suit is not empty.  
Frontend was also showing case name full metadata field when that field is empty which is incorrect.


quevon24 (approved):
LGTM

PR #5788, opened 2025-06-24
feat(search): Refines `cl_compute_token_count` command

ERosendo:
Key changes: 

- The help message for the `--percentage` argument has been refined to clearly state that it expects a float from 0.0 to 100.0 and the new default value is 1.0. 

- Introduces a new required argument, `--type`, which accepts two values `recap` or `caselaw`. Users can now specify which dataset (RECAP documents or Caselaw opinions) they want to compute token counts for, rather than running both by default.

   - The handle method now primarily dispatches to the appropriate helper based on the --type argument.
   - The logic within the `handle` method has been broken down into two dedicated helper methods: `_compute_recap_token_count` and `_compute_opinion_token_count`.

- Adds a new optional flag, `--use-replica`. When this flag is provided, all database queries made by the command will be directed to the replica database connection (if configured) instead of the default primary database.

### DIFF
--- a/cl/recap/tasks.py
+++ b/cl/recap/tasks.py
@@ -1491,6 +1491,13 @@ async def process_recap_acms_docket(pk):
     # Merge parties before adding docket entries, so they can access parties'
     # data when the RECAPDocuments are percolated.
     await sync_to_async(add_parties_and_attorneys)(d, data["parties"])
+
+    # Sort docket entries by their 'document_number'.
+    # We noticed raw ACMS data is not consistently sorted, so we sort by
+    # 'document_number' to match the display order on the docket report.
+    data["docket_entries"] = sorted(
+        data["docket_entries"], key=lambda d: d["document_number"]
+    )
     des_returned, rds_created, content_updated = await add_docket_entries(
         d, data["docket_entries"]
     )
--- a/cl/recap/tests/tests.py
+++ b/cl/recap/tests/tests.py
@@ -1151,9 +1151,9 @@ class RecapUploadsTest(TestCase):
         )
 
         # Confirm Docket entry and RECAPDocument is properly created.
-        self.assertEqual(docket_entries.count(), 2)
+        self.assertEqual(docket_entries.count(), 3)
         recap_documents = RECAPDocument.objects.all().order_by("date_created")
-        self.assertEqual(recap_documents.count(), 2)
+        self.assertEqual(recap_documents.count(), 3)
         self.assertEqual(
             recap_documents[0].pacer_doc_id,
             "46de54cd-3561-ee11-be6e-001dd804e087",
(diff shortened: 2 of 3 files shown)


cl/search/management/commands/cl_compute_token_count.py
```diff
@@ -116,16 +121,82 @@ class Command(VerboseCommand):
     help = "Compute token count for Recap Documents and Caselaw."
 
     def add_arguments(self, parser):
+        parser.add_argument(
+            "--type",
+            type=str,
+            choices=["recap", "caselaw"],
```
mlissner: I think I'd suggest this: 

```suggestion
            choices=["r", "o"],
```

But using the values from the SEARCH_TYPES variable, so it'd be something like:

```suggestion
            choices=[SEARCH_TYPES.RECAP, SEARCH_TYPES.OPINION],
```

cl/search/management/commands/cl_compute_token_count.py
```diff
@@ -116,16 +121,82 @@ class Command(VerboseCommand):
...
+            help=(
+                "Specifies the proportion of the table to be sampled (0.0 to "
+                "100.0 percent). Defaults to 1.0."
+            ),
+        )
+        parser.add_argument(
+            "--use-replica",
+            action="store_true",
+            default=False,
+            help="Use this flag to run the queries in the replica db",
```
mlissner: I'm being picky, but it's so our commands are consistent. I think in our other command we have a flag for `--database` or similar, so that you can put in whatever one you want. Better to be consistent?

cl/search/management/commands/cl_compute_token_count.py
```diff
@@ -170,39 +242,16 @@ def handle(self, *args, **options):
...
-        self.stdout.write(
-            f"The sample represents {sample_size / total_opinions:.3%} of the Caselaw"
-        )
-        self.stdout.write(
-            f"Total number of tokens in caselaw: {intword(total_token_in_caselaw)}"
-        )
+        if dataset_type == "recap":
+            self._compute_recap_token_count(percentage, db_connection)
+        else:
+            self._compute_case_law_token_count(percentage, db_connection)
```
mlissner:
```suggestion
        if dataset_type == SEARCH_TYPES.RECAP:
            self._compute_recap_token_count(percentage, db_connection)
        elif dataset_type == SEARCH_TYPES.OPINION:
            self._compute_case_law_token_count(percentage, db_connection)
        else:
             NotImplementedError etc.
```

mlissner (approved):
A few suggestions. Mostly ergnonomics to make it consistent with our other commands (to the extent that's not a lost war).

Thank you!

ERosendo:
@mlissner Thanks for your feedback! This PR is ready for another review.

mlissner (approved):
Nice. Thank you!

PR #5790, opened 2025-06-25
Improve error handling in find citations

flooie:
Find Citations was silently failing for a handful of reasons.

The retry function was passing kwargs - which appears to need the retry to use position arguments.
This was causing a TypeError and failing retries.  Additionally, this may have been swallowing up 
the error that was causing the retry.  

Now instead, we are logging directly into sentry the specific opinion that failed and why.  This should 
open up our error failing in eyecite.  

Additionally, we are checking for self references, and in opinions where no opinions have been found we 
were causing a bug.  Now we return no match resources if the count remains 0

### DIFF
--- a/cl/recap/tasks.py
+++ b/cl/recap/tasks.py
@@ -1491,6 +1491,13 @@ async def process_recap_acms_docket(pk):
     # Merge parties before adding docket entries, so they can access parties'
     # data when the RECAPDocuments are percolated.
     await sync_to_async(add_parties_and_attorneys)(d, data["parties"])
+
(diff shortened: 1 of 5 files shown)


flooie:
@mlissner could you give this a quick once over?

flooie:
Yes you are right- I've updated the sentry error with your suggestion and tested in in the sentry dev.  Much nicer

albertisfu (approved):
Thanks for the changes @flooie ! This is ready to be merged. Auto-merge enabled.

PR #5808, opened 2025-06-26
feat(cl): Remove Twitter and replace with Bluesky

florean:
I removed references to Twitter and added Bluesky if it wasn't there already.  The biggest change is to the social links in the footer, which I brought over entirely from the new design.  The icons might be a little bigger, but the overall height of the footer is narrower. 
image
The second visual change is to the sharing links.  I'm not totally happy with the spacing from the input box, but I also don't understand what the purpose of the input box is.
image

The other changes are just simple removals of the Twitter links from various e-mail templates that already have Bluesky links.


florean:
@elisa-a-v, when you take a look at this, can you check it in Chrome?  I only used Safari.  Thanks!

cl/assets/templates/includes/social_links.html
```diff
@@ -12,11 +12,11 @@ <h3><span>Share</span></h3>
...
-        <a href="https://x.com/home?status={% get_full_host %}{{ request.path }}"
-           target="_blank"
-           title="Share on X">
-            <i class="fa fa-twitter-square fa-lg gray"></i>
-        </a>&nbsp;
+        <a
+          target="_blank"
+          title="Share on Bluesky"
+          href="https://bsky.app/intent/compose?text={% get_full_host %}{{ request.path }}"
+        >{% svg "logos/bluesky" width="12" height="12" aria_hidden="true" style="vertical-align: -15%;"%}</a>
```
elisa-a-v, line 19: I usually avoid shifting items' positions individually and instead try to use the proper layout, but I see why you landed on this here. I tried tweaking the CSS so this was a row, but it required more changes than I expected, so this is a sensible solution and looks good enough to me, visually.

elisa-a-v (approved):
Thanks for the PR @florean :) I tested in Chrome, Firefox and Safari, and it looks the same in all three.

> I'm not totally happy with the spacing from the input box, but I also don't understand what the purpose of the input box is.

I wouldn't worry about it. I think the input is there for people to copy the content which, granted, is not evident, but it's been like that for over 10 years and we're finally changing the UI, so it won't be there much longer.

Anyway, LGTM. Good riddance x! 🚀

mlissner:
Hm, weirdly these are showing up bright blue for me: 

image

I know we wanted blue sky, but this is a bit literal?

mlissner:
Looks like these all had class="gray" on the anchors before.

PR #5809, opened 2025-06-26
feat(citation.tasks): Speed up retry

flooie:
While I say retry - it really is moving on to the next opinion to process.  And each time we fail we wait 60 seconds.  

Change it to 2 seconds and just move along with our day.  No need to hang for 60 seconds.


mlissner (approved):

flooie:
@mlissner thanks.

PR #5810, opened 2025-06-26
feat(citation.models): Fix create_from_eyecite

flooie:
Create from Eyecite needs to ignore non standard year format.  We dont have 
a way to handle 1995-96 at the moment so it should just ignore it.

I added a test to make sure it works as expected now.  

Also - Updated the error logging to include the error message in sentry.  
The opinion id was not enough to unravel some mysterious bugs.

### DIFF
--- a/cl/citations/models.py
+++ b/cl/citations/models.py
@@ -1,8 +1,10 @@
+import re
+
 from django.db import models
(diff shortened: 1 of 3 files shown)


albertisfu (approved):
Great, this looks good to me! Merging...

PR #5815, opened 2025-06-27
feat(acms): Refines docket entry sorting to handle minute entries

### LINKED ISSUE #5731: Dockets with entries out of order

ERosendo:
This PR updates the logic to sort ACMS docket entries to use '`date_filed`' as the primary key, followed by '`document_number`'. This change specifically addresses cases where '`document_number`' is `None` (minute entries), ensuring these entries are still sorted logically and consistently with the display order on docket reports.

This PR also updates the `MockACMSDocketReport` class to reflect changes from https://github.com/freelawproject/juriscraper/pull/1463

While reviewing the ACMS case that created the error report in https://github.com/freelawproject/courtlistener/pull/5784#issuecomment-3014029943, I noticed that the minute entry was incorrectly pushed to the last row of the table, despite its `date_filed`. The following screenshot shows this issue:

image

This PR resolves this by ensuring minute entries are sorted correctly by their `date_filed`, appearing alongside other entries from the same day and before entries with later filing dates.

Closes #5731


mlissner:
> While reviewing the ACMS case that created the error report in https://github.com/freelawproject/courtlistener/pull/5784#issuecomment-3014029943, I noticed that the minute entry was incorrectly pushed to the last row of the table, despite its date_filed.

Are you saying this is a bug in ACMS itself?

ERosendo:
> Are you saying this is a bug in ACMS itself?

The core issue is in the logic used to sort the entries. They really need to fix how they handle docket entries. JS's loose type comparison probably lets them display the report without completely breaking. However, pushing minute entries to the end of the table is incorrect. In this specific case, there's only one, but if there were more, they'd all end up at the bottom of the table.

mlissner:
What's the docket info for the problematic case? Maybe we can share the bug with them and get some good will.

ERosendo:
oh sorry, I forgot to mention it, you can see the problematic case here: https://ca2-showdoc.azurewebsites.us/23-7577

mlissner:
Nice. I'll reach out.

PR #5824, opened 2025-06-28
Fix RecursionError exception

quevon24:
In the original implementation, when an OperationalError was raised in this part: 

`store_opinion_citations_and_update_parentheticals(opinion)`

we handled it by calling the task recursively using:

`find_citations_and_parentheticals_for_opinion_by_pks.apply_async(...)
`

While this does retry the task, it can lead to a RecursionError if repeated failures occur. This happens because apply_async() spawns a new Celery task but does so within the current task’s call stack, which continues executing. If a task repeatedly schedules itself this way (e.g., due to persistent DB contention or deadlocks), it builds up recursive frames and eventually exceeds Python’s recursion limit, resulting in:

`RecursionError: maximum recursion depth exceeded
`

To prevent this, we now:

- Use self.retry() to requeue only the failed opinion. This method exits the task cleanly and ensures the retry is managed by Celery without adding stack depth.
- Schedule a separate apply_async() task for the remaining opinions, so the batch continues processing from where the failure occurred, avoiding full reprocessing.


cl/citations/tasks.py
```diff
@@ -152,15 +152,33 @@ def find_citations_and_parentheticals_for_opinion_by_pks(
...
+                        args=(
+                            remaining_ids,
+                            disconnect_pg_signals,
+                            disable_citation_count_update,
+                        ),
+                        countdown=5,
+                    )
+
+                # Retry only the failed opinion, use it to avoid RecursionError being raised when retries are exhausted
+                raise self.retry(
```
grossir: this works; the only thing I don't like it's that it will create a new celery task for the `remaining_ids`; which will go into Redis, and the it will be picked up by the workers, etc... When they could just be executed in this task following the current loop

A way of solving this would be keep the loop going and create a new task for the failure but with tweaked [retry](https://docs.celeryq.dev/en/stable/userguide/calling.html#message-sending-retry) parameters

```
find_citations_and_parentheticals_for_opinion_by_pks.apply_async(
                        args=(
                            [failure_id],
                            disconnect_pg_signals,
                            disable_citation_count_update,
                        ),
                        countdown=5,
                        retry=False,
                        max_retries=0,
                    )
```

What do you think? As it is, it currently works too

grossir: So the option above will cause another RecursionError, eventually. I think we should merge your solution as it is

quevon24:
@grossir I've already implemented the changes we discussed earlier and updated the tests, let me know if anything else is needed

cl/citations/tasks.py
```diff
@@ -186,6 +185,20 @@ def find_citations_and_parentheticals_for_opinion_by_pks(
         if disconnect_pg_signals:
             reconnect_parenthetical_group_signals()
 
+    # Schedule a new retry task with the opinions that failed due to OperationalError
+    if failed_ids:
+        logger.warning(
+            "Scheduling retry for %d failed opinions", len(failed_ids)
+        )
+        find_citations_and_parentheticals_for_opinion_by_pks.apply_async(
```
grossir: @quevon24 

shouldn't this be a `self.retry` so it triggers a retry limit? If we have a record that triggers `OperationalError` every time, then this part will cause a MaxRecursionError, right?

quevon24: Yes, you are right, i just updated the code.

PR #5844, opened 2025-06-30
style(tailwind): adjust nested list styling and update input.css

nazarbodak221:
Improved the appearance of nested bullet lists in help pages by setting different bullet styles (`disc`, `circle`, `square`) for each nesting level. Also ensured consistency in spacing and layout across deeply nested content in API documentation.

Refs: https://github.com/freelawproject/courtlistener/issues/5353

### DIFF
--- a/cl/assets/tailwind/input.css
+++ b/cl/assets/tailwind/input.css
@@ -72,7 +72,7 @@
 
    .alert-danger {
     @apply bg-red-100 text-red-700 border-red-500;
-  } 
+  }
 }
 
 @layer components {
--- a/cl/assets/tailwind/input.css
+++ b/cl/assets/tailwind/input.css
@@ -31,8 +31,13 @@
   ul {
     @apply list-disc pl-6;
   }
-  ol {
-    @apply list-decimal pl-6;
+
+  ul ul {
+    list-style-type: circle;
+  }
+
+  ul ul ul {
+    list-style-type: square;
   }
   blockquote {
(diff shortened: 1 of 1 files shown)


mlissner:
Thanks. Can you include some screenshots, please?

nazarbodak221:
> Thanks. Can you include some screenshots, please?

nazarbodak221:
> Thanks. Can you include some screenshots, please?

image

cl/assets/tailwind/input.css
```diff
@@ -31,8 +31,13 @@
...
   }
-  ol {
-    @apply list-decimal pl-6;
+
+  ul ul {
+    list-style-type: circle;
+  }
+
+  ul ul ul {
+    list-style-type: square;
```
elisa-a-v, line 40: Instead of enforcing these in all `<li>` elements, lets use utilities in that particular view only. These tags are used in many other contexts not just text lists, so we don't want to add these styles globally.

cl/assets/tailwind/input.css
```diff
@@ -56,6 +61,20 @@
...
+
+  .alert-warning {
+    @apply alert text-yellow-700 bg-yellow-100 border-yellow-500;
+  }
+
+   .alert-danger {
+    @apply bg-red-100 text-red-700 border-red-500;
+  }
+}
+
```
elisa-a-v, line 77: These were already added in another PR in the proper layer, so we should clean this up.

elisa-a-v (changes requested):
Same as with other API help pages, this one doesn't show even with the flag active. I've also left a couple of comments in the code itself that need to be addressed.

PR #5847, opened 2025-07-01
feat(favorites): Introduce model to track view counts

ERosendo:
This PR introduces a new `GenericCount` model to support generic tracking of simple metrics, such as view counts, across multiple object types. It implements the suggestion from https://github.com/freelawproject/courtlistener/issues/3346#issuecomment-2920002257

Instead of using foreign keys, the model uses a namespaced string key to identify the object and action being tracked. This keeps the schema lightweight and makes it easy to extend view tracking to other object types in the future without schema changes.

This PR only introduces the new model.  The full migration will happen in two stages:

1. Migrate existing view counts from the `docket` and `opinion` tables into this new GenericCount  model.
2. A follow-up PR will update view tracking logic to write to GenericCount.


mlissner:
One last thought from me. Would it be better to have convention be `$object.$id:$action` instead of `$object.$action:$id`. 

Two reasons. First I think this will perform better in a b-tree because it prevents the first branch from being extremely wide. 

Second, I think the semantics are actually better: "Docket 22 was viewed," instead of "Docket viewed; number 22."

ERosendo:
> One last thought from me. Would it be better to have convention be $object.$id:$action instead of $object.$action:$view.

Understood! I'll update the help text attribute

ERosendo:
@mlissner I tried populating the new table using the Django shell, but it kept running into issues and getting terminated. So, I decided to use a direct SQL script to get the data in there instead. Here's the SQL code:

```sql
INSERT INTO favorites_genericcount (label, value)
SELECT
    'd.' || id::TEXT || ':view' AS label,
    view_count AS value            -- Map view_count to value
FROM
    search_docket;
```

_Adding all the rows from the docket table took approximately 20 minutes_

ERosendo:
Thanks, @mlissner. Would you like me to create an issue to track the migration of the view counts from the `docket` and `opinions` tables?

mlissner:
I'll probably just do it, thanks.

PR #5849, opened 2025-07-01
feat(new_homepage): introduce action buttons

### LINKED ISSUE #5360: Home Page - Action Buttons
Image

**Refer to the Figma for the most updated design.**

On the CourtListener home page, the “action” buttons let the users know what types of actions they can take on the site, and redirect them to the relevant pages.

**Requirements:**

- Each action is a clickable button.
- The cursor changes when it hovers over an action, indicating to the user it is clickable.
- For the text “About Free Law Project,” we want the top line to read “About Free” and the bottom line to read “Law Project.”

The user will be redirected to these pages:
- About Free Law Project: https://free.law/about
- Become a Member: https://donate.free.law/forms/membership
- Subscribe to Updates: https://donate.free.law/np/clients/freelawproject/subscribe.jsp
- Create an Alert: https://www.courtlistener.com/help/alerts/
- Find a Case: https://www.courtlistener.com/c/
- Contribute a Document: https://free.law/recap
- Create a Podcast: https://www.courtlistener.com/podcasts/

elisa-a-v:
Closes #5360 

This PR introduces:
- A responsive row of action buttons in the homepage.
- An Alpine component for the homepage that handles the behavior of the arrow buttons that appear in smaller screens.
- A tiny refactor where `bell.svg` is now called `bell_outline.svg`.

#### Note on keyboard navigation
I removed the arrow buttons from the tab order to avoid tabbing through them, which can be confusing and is actually pretty useless (arrows are only focusable when you've already scrolled all the way: right arrow when you get to the end, left arrow when you're at the start.)

Therefore:
- To move scroll through the actions we can simply use the native behavior with the <kbd>left</kbd> and <kbd>right</kbd> arrow keys.
- To move focus we use the native <kbd>tab</kbd> and <kbd>shift</kbd>+<kbd>tab</kbd> navigation, as well as the custom <kbd>home</kbd> and <kbd>end</kbd> keys to move to the start/end.

However, I couldn't get the <kbd>home</kbd> and <kbd>end</kbd> keys to work on Safari, but it's a _really_ minor feature so unless people think it's important I'm not sure it's worth giving it much more time.

## Screenshots

<details>
<summary>Desktop</summary>

image

</details>

<details>
<summary>Mobile</summary>

image

</details>

## Arrow buttons demo

[Screencast from 06-28-2025 12:58:32 AM.webm](https://github.com/user-attachments/assets/429d047b-0599-449b-b455-d5bfc4c72f88)


cl/search/templates/v2_homepage.html
```diff
@@ -13,11 +19,138 @@
...
+        href="https://donate.free.law/np/clients/freelawproject/subscribe.jsp"
+        target="_blank"
+        rel="noopener noreferrer"
+        class="w-28 flex flex-col justify-center items-center gap-4 group scroll-m-10"
+        aria-label="Subscribe to the Free Law Project newsletter (opens in new tab)"
+      >
+        <div aria-hidden="true" class="md:w-15 md:h-15 w-13 h-13 p-3.5 rounded-full bg-greyscale-100 text-greyscale-400 group-hover:bg-greyscale-200">
+          {% svg "bell_plus" class="md:w-8 md:h-8 w-6 h-6" %}
+        </div>
+        <span class="flex items-center justify-center text-center w-28 h-10 text-sm font-medium text-greyscale-900">Subscribe to Updates</span>
```
mlissner: Maybe "Get Our Newsletter"? I forgot what this was, and the bell icon *really* didn't help.

cl/search/templates/v2_homepage.html
```diff
@@ -13,11 +19,138 @@
...
+      <a
+        x-on:focus="scrollIntoView"
+        href="{% url "citation_homepage" %}"
+        class="w-28 flex flex-col justify-center items-center gap-4 group scroll-m-10"
+        aria-label="Use our Citation Lookup Tool"
+      >
+        <div aria-hidden="true" class="md:w-15 md:h-15 w-13 h-13 p-3.5 rounded-full bg-greyscale-100 text-greyscale-400 group-hover:bg-greyscale-200">
+          {% svg "court" class="md:w-8 md:h-8 w-6 h-6" %}
+        </div>
+        <span class="flex items-center justify-center text-center w-28 h-10 text-sm font-medium text-greyscale-900">Find a Case</span>
```
mlissner:
```suggestion
        <span class="flex items-center justify-center text-center w-28 h-10 text-sm font-medium text-greyscale-900">Find a Citation</span>
```

cl/search/templates/v2_homepage.html
```diff
@@ -13,11 +19,138 @@
...
+        x-on:focus="scrollIntoView"
+        href="https://free.law/recap"
+        target="_blank"
+        class="w-28 flex flex-col justify-center items-center gap-4 group scroll-m-10"
+        aria-label="Learn how to contribute documents using RECAP (opens in new tab)"
+      >
+        <div aria-hidden="true" class="md:w-15 md:h-15 w-13 h-13 p-3.5 rounded-full bg-greyscale-100 text-greyscale-400 group-hover:bg-greyscale-200">
+          {% svg "file_text" class="md:w-8 md:h-8 w-6 h-6" %}
+        </div>
+        <span class="flex items-center justify-center text-center w-28 h-10 text-sm font-medium text-greyscale-900">Contribute a Document</span>
```
mlissner: Maybe: "Send us PACER Docs"? That might be more concrete?

cl/search/templates/v2_homepage.html
```diff
@@ -13,11 +19,138 @@
...
+      </a>
+      <a
+        x-on:focus="scrollIntoView"
+        href="{% url "podcasts" %}"
+        class="w-28 flex flex-col justify-center items-center gap-4 group scroll-m-10"
+      >
+        <div aria-hidden="true" class="md:w-15 md:h-15 w-13 h-13 p-3.5 rounded-full bg-greyscale-100 text-greyscale-400 group-hover:bg-greyscale-200">
+          {% svg "microphone" class="md:w-8 md:h-8 w-6 h-6" %}
+        </div>
+        <span class="flex items-center justify-center text-center w-28 h-10 text-sm font-medium text-greyscale-900">Create a Podcast</span>
```
mlissner:
```suggestion
        <span class="flex items-center justify-center text-center w-28 h-10 text-sm font-medium text-greyscale-900">Create Custom Podcast</span>
```

mlissner (dismissed):
Woah. I expected about 3× the JavaScript in this one. I didn't read it, but I'm amazed how short it is. 

Two thoughts from me and I'll let somebody else review properly:

1. The buttons really feel like they're good candidates for components. Lots of repeated code.

2. I tweaked a bunch of the words because I've had some distance from the design and the old words don't make much sense to me anymore.

Thank you!

elisa-a-v:
> I expected about 3× the JavaScript in this one. I didn't read it, but I'm amazed how short it is.

There's probably two factors at play:
1. Alpine is awesome, and a lot of the JS is in the template itself (I think the [last commit](https://github.com/freelawproject/courtlistener/pull/5849/commits/3791e4864dacda93bce335e3deb024ef69467771) illustrates this to an extent)
2. I tried to keep it simple, so the scrolling behavior for the arrow buttons is pretty basic: we just move the scroll position a percentage of the width of the scrolling container. We could do more complex behaviors to make the scrolling position adjust to the items instead, which would probably feel better, but that would indeed require more JS.

IMO this works decently, so if we want to make it better I would suggest we open a separate issue for refining the scrolling behavior of the buttons, as it feels like a good challenge for volunteers wanting to help here. However, I'd be happy to address this now if we feel it's important.

> The buttons really feel like they're good candidates for components. Lots of repeated code.

Yeah I guess you're right, making a component for the action buttons would make it much easier to tweak them in the future, add new ones, etc, so I can do that no problem. Only reason I didn't is because I haven't found any other instances of these buttons anywhere else in the design, but it's probably still worth it, so I will 🫡

cl/search/templates/v2_homepage.html
```diff
@@ -13,11 +19,138 @@
...
+    <div
+      x-ref="container"
+      x-on:scroll="updateButtons"
+      x-on:keydown.home.stop.prevent="focusFirst"
+      x-on:keydown.end.stop.prevent="focusLast"
+      x-on:resize.window="updateButtons"
+      role="toolbar"
+      aria-label="Homepage actions"
+      tabindex="-1"
+      class="relative w-full flex flex-row gap-2 max-w-[1400px] mx-auto overflow-auto justify-between md:p-10 px-4 py-11"
```
ERosendo: Let's update the scrollbar's style. The mockup shows a red scrollbar, but the current PR uses the default style.


image

ERosendo (changes requested):
Thanks for this PR @elisa-a-v. Code looks good. One thing I spotted is that the scrollbar style hasn't been updated. I've added a minor comment on that.

cl/search/templates/v2_homepage.html
elisa-a-v: Oooh you're right! I thought it was a highlight for the buttons 🤦🏽 in fact, we had already [dismissed the red underline](https://github.com/freelawproject/courtlistener/issues/5360#issuecomment-3014667202) 🤪 However, what we dismissed wasn't really the red _scrollbar_. Changing the color is trivial, and it does look pretty nice and cohesive with the rest of the design, so maybe this changes things and we do want the red scrollbar.

The standard `scrollbar-color` CSS property doesn't work in some browsers though, like Safari, so the UI would look slightly different in different browsers, but I'm not sure how relevant that is. What do you think?

elisa-a-v:
Thank you @ERosendo for the review. I've updated the text as requested, made an icon link component, and refactored the homepage to use this new component.

> The buttons really feel like they're good candidates for components. Lots of repeated code.

Yeah I love how much cleaner the code looks now! Good call 👍🏽

cl/search/templates/v2_homepage.html
mlissner: Safari sucks, let's do it for the ones that support it at least. :)

ERosendo (approved):
LGTM :shipit:

PR #5855, opened 2025-07-01
feat(help): apply updated design to Webhooks Start page

nazarbodak221:
Applied the new visual design to the Webhooks Getting Started page (`v2_webhooks-getting-started.html`)

Screenshot of updated design:

image

Refs: https://github.com/freelawproject/courtlistener/issues/5353

### DIFF
--- a/cl/api/templates/v2_webhooks-getting-started.html
+++ b/cl/api/templates/v2_webhooks-getting-started.html
@@ -43,7 +43,7 @@
       <li>Your application should use the <a class="underline" href="{% url 'webhooks_docs' %}#headers">Idempotency-Key included in the event headers</a> to ensure that it only processes the event once.</li>
     </ul>
     <p>Below are examples of endpoints configured in Flask and Django</p>
-    <p><strong>Flask</strong></p>
+    <h3>Flask</h3>
     <c-code>from flask import Flask, request, Response
 app = Flask(__name__)
 
@@ -53,7 +53,7 @@ def respond():
     print(request.json)
     return Response(status=200)
     </c-code>
-    <p><strong>Django</strong></p>
+    <h3>Django</h3>
     <c-code>from django.views.decorators.csrf import csrf_exempt
 from django.http import HttpRequest, HttpResponse
 


cl/api/templates/v2_webhooks-getting-started.html
```diff
@@ -0,0 +1,114 @@
...
+      <li>Your URL must receive POST requests and should not accept any other HTTP verbs.</li>
+      <li>If you are using a web framework like Django or Ruby on Rails, the webhook URL will need to be exempted from the cross site request forgery (CSRF) protection system. If this is not done, most web frameworks will block POST requests from outside domains like ours.</li>
+      <li>To avoid <a class="underline" href="{% url 'webhooks_docs' %}#retries">timeout errors and event retries</a>, your application may need to process events asynchronously.</li>
+      <li>Your application must respond to events with a 2xx HTTP status code or else the event will be retried.</li>
+      <li>Your application should use the <a class="underline" href="{% url 'webhooks_docs' %}#headers">Idempotency-Key included in the event headers</a> to ensure that it only processes the event once.</li>
+    </ul>
+    <p>Below are examples of endpoints configured in Flask and Django</p>
+    <p><strong>Flask</strong></p>
```
elisa-a-v: I think `<h3>` is more semantically correct here than `<p><strong>`:

```suggestion
    <h3>Flask</h3>
```

cl/api/templates/v2_webhooks-getting-started.html
```diff
@@ -0,0 +1,114 @@
...
+    <c-code>from flask import Flask, request, Response
+app = Flask(__name__)
+
+@app.route("/webhooks/a-long-random-url-here/", methods=["POST"])
+def respond():
+    # Do your event processing.
+    print(request.json)
+    return Response(status=200)
+    </c-code>
+    <p><strong>Django</strong></p>
```
elisa-a-v: Same here:

```suggestion
    <h3>Django</h3>
```

elisa-a-v (changes requested):
Looks good overall, thanks! However, as with the [previous review](https://github.com/freelawproject/courtlistener/pull/5825#pullrequestreview-2976172552), this one also doesn't show even when the flag is active. This will probably require the same fix as that one, since both handle template name the same way.

cl/api/templates/v2_webhooks-getting-started.html
nazarbodak221: Yep, that makes sense, I'll change it.

nazarbodak221:
> Looks good overall, thanks! However, as with the [previous review](https://github.com/freelawproject/courtlistener/pull/5825#pullrequestreview-2976172552), this one also doesn't show even when the flag is active. This will probably require the same fix as that one, since both handle template name the same way.

I change middleware to fix it, but to be honest I don't know where to commit, in every branch, or make a PR for this change, because it should be in each PR

elisa-a-v:
> or make a PR for this change

This sounds great! Could you open a PR to update the middleware to support these views please? Then we can merge that first, and the rest should work :)

PR #5860, opened 2025-07-01
feat(cl): add example for creating custom settings file

florean:
There are file patterns for custom Django settings files in the `.gitignore`, but there are no examples or documentation for using them and it is not very discoverable.  I added an example settings file with some documentation and a couple example settings that solve problems I have run into along with the corresponding line in the `.env.dev.example` file.

I also updated the wiki with documentation on using custom settings.  Sorry, I thought it would create a PR or something, I did not know the changes would immediately go live.  If this gets merged, it should be updated again (happy to do it).


florean:
Sorry about the [wiki edit](https://github.com/freelawproject/courtlistener/wiki/Getting-Started-Developing-CourtListener/_compare/63a468387a1813fe6b18d890ffd8a14049f36c04...7a2e395ad6fa3ea254f1907eec83d3e75aad9061).  Do you have a process for that?  I can't believe GitHub still doesn't allow PRs against them.  I can already see that I left off the `Add your problem here...` line.

mlissner:
Hm, thanks for doing this. I thought we could override settings just by tweaking values in the env file. I think you've got a solution that works, but it's sort of layered on top of the simpler solution we'd prefer. Did you try that?

florean:
I haven't dug into the documentation, but my assumption with `django-environ` was that it only overrides settings that have an explicit `env` call.  We could add that to `AWS_S3_CUSTOM_DOMAIN`, but that doesn't help for cases like in my examples, such as modifying the CSP or a setting that takes a function. Basically any settings more complicated than a boolean or string.

I created these because I needed to modify settings (literally the examples I've included), but that is a pain when switching branches often.  So I spent some time figuring out how to use a custom Django settings file with the project's setup and the final step was adding it to the `.gitignore`, which is when I discovered there was already a rule that covered this scenario.  I've got everything I need and maybe nobody else needs custom settings, but I thought I'd contribute what would have saved me the most time in case someone does.  If you have a simpler system that handles my use cases, then I'm happy to add some documentation for it, because I don't know what it is.

florean:
Hey, this isn't a priority, but I do think it is useful for new developers.  At least it was for me and, as mentioned, it enables things that are sometimes required that the simpler system doesn't, like viewing PDFs and being able to use DjTB, which was pretty frustrating.  I think it doesn't even need to be documented; just having well-commented examples will people who are trying to troubleshoot a problem.  The DjTB example in particular would have saved me a bunch of time.

mlissner:
I *think* django-environ is designed to support whatever settings you can dream up. For example, it can do dicts, I believe: 

https://django-environ.readthedocs.io/en/latest/api.html#environ.Env.dict

I appreciate that you figured out how to override settings the nuclear way, but I think the correct way in this project to do this is to put it in the .env file, and if that's not possible (b/c the setting isn't using django-environ yet), then we should do a PR to fix that setting so it can be overridden. 

And if *that's* not enough for some reason *then* we resort to a private settings file like you're doing here. We used to do it more like how you're describing (hence the line in the gitignore), but I don't love it because it makes two places that you can override things. Not great.

florean:
I've read the docs and some of the [source code](https://django-environ.readthedocs.io/en/latest/_modules/environ/environ.html#Env.parse_value).  Yes, we can do the first case.  There is no way to do the second case, which is a [commonly](https://sqlpey.com/python/solved-how-to-fix-django-debug-toolbar-not-showing-up/) [recommended](https://stackoverflow.com/questions/10517765/django-debug-toolbar-not-showing-up) [troubleshooting](https://www.reddit.com/r/django/comments/15k68fa/djangodebugtoolbar_works_in_the_admin_but_not_in/) technique for DjDT.
It is is a method you do support, but it isn't documented.  But it is working for me and maybe someone in the same boat will see this PR when searching for a solution.

PR #5861, opened 2025-07-02
fix(bulk data): apt-get update before attempting apt-get install

branliu0:
Script last succeeded on April 30, 2025.

@mlissner observed this error on `apt install -y sudo`.

```
manual-run-cnk6m Installing utilities                                                                                               
manual-run-cnk6m                                                                                                                    
manual-run-cnk6m WARNING: apt does not have a stable CLI interface. Use with caution in scripts.                                    
manual-run-cnk6m                                                                                                                    
manual-run-cnk6m Reading package lists...                                                                                           
manual-run-cnk6m Building dependency tree...                                                                                        
manual-run-cnk6m Reading state information...                                                                                       
manual-run-cnk6m E: Unable to locate package sudo 
```

1. It's preferred to use `apt-get` in scripts. `apt` is for interactive front-end use.
2. It's standard to use `apt-get update` before installing anything -- inability to locate the package suggests that the package list needs to be updated.

### DIFF
--- a/scripts/make_bulk_data.sh
+++ b/scripts/make_bulk_data.sh
@@ -4,7 +4,7 @@ set -e
 echo "Installing utilities"
 
 # Make sure the sudo package is installed
-apt install -y sudo
+apt-get update && apt-get install -y sudo
 
 # Set up Sentry
 curl -sL https://sentry.io/get-cli/ | bash


branliu0:
Not sure why that test would fail... thoughts?

mlissner (approved):
Looks promising. Let's see if it helps. Thanks.

mlissner:
This did the trick, thank you! 

The only change I had to make was in the cronjob, which has the following command:

```
command:
 - /bin/sh
 - -c
 - "[ $(date -d +1day +%d) -eq 1 ] && /opt/courtlistener/scripts/make_bulk_data.sh"
```

This checks if tomorrow is the first and only runs the cronjob if that's the case, so I had to tweak it to check if tomorrow was the third (it currently is).

The export is running now.

branliu0:
Ah, fantastic! Thank you!

PR #5867, opened 2025-07-03
feat(scrapers): allow cloning of existing dockets

florean:
Currently, `clone_from_cl` crashes if you try to clone a docket that already exists in the database.  So if you want to pull in new docket entries and documents for a docket, you have to manually delete all of the docket's existing entries and documents.  This is the minimal amount of changes required (read: fast and dirty) to allow a docket's entries to be updated.  I've used it a number of times and it's been very useful.

While this works great, I am going to do a larger rewrite of `clone_from_cl` to simplify the logic and make updates much more efficient.  So I have a few questions (and feel free to move this to an Issue or Discussion):

1. Do dockets, docket entries, or RECAPDocuments (and associated metadata) ever change?  My sense is that dockets might be no, docket entries are rarely, and RECAPDocuments are common as PACER documents get added to the system.
1. What other objects should be updated?  Seems like Person is an obvious one.

I think my approach would be to only add new objects and add a `--force` option that would update everything.


mlissner:
This looks fine to me, but @quevon24 is the owner of this bit of code, so I'll leave it to him to merge. 

As for changes, dockets do change, yes. We get them bit by bit some cases, so today's metadata may be less complete than tomorrow's. I think I'd invert your logic and assume that people using this script want the content overwritten. 

Personally, I'd be annoyed at having to look up the `--force` flag every time. Maybe just consider that the default and have a warning:

> Beginning clone process. This will overwrite any local data you may have (use `--no-clobber` to prevent this).
> 
> Press any key to proceed...

This uses the clobbering approach, terminology, and flag from the `mv` command, which might be familiar to folks?

florean:
Sorry, since I don't think this has been reviewed yet and I just got pinged on it, I forgot I was sitting on some other changes I was going to add after this.  But might as well do it all at the same time.

cl/scrapers/management/commands/clone_from_cl.py
```diff
@@ -710,30 +681,25 @@ def clone_recap_documents(
...
     """
     created_tags = []
     for tag_id in tag_ids:
         print(f"Cloning tag id: {tag_id}")
 
-        model = apps.get_model(object_type)
-
         try:
-            tag = model.objects.get(pk=tag_id)
+            tag = Tag.objects.get(pk=tag_id)
```
quevon24, line 695: You forgot to import the Tag model from cl.search.models

quevon24 (changes requested):
Everything looks great, I liked the way you simplified some things, just left a comment on the `clone_tag` function.

The creation of opinions was also refactored and a new function called `clone_opinion` was added. This may need to be updated.

I think that in the documentation at the top of the code it would be good to mention that the docket entries and the recap documents are updated if the argument --add-docket-entries is passed to the command

cl/scrapers/management/commands/clone_from_cl.py
florean: > The creation of opinions was also refactored and a new function called clone_opinion was added.

lol, I looked at the code first and was kicking myself for having missed that.

Good catch on `Tag`.  I thought I would have exercised that when cloning `Docket`s, but I guess none of mine had them.  I also thought `ruff --check` catch it, but I guess not.  I'll start running checking with `flake8`.

Thanks for the review, great suggestions, all implemented.

florean:
> Personally, I'd be annoyed at having to look up the --force flag every time. Maybe just consider that the default and have a warning

Good suggestion and this isn't a priority for me, but I'd like to create an Issue to track it the next time someone touches this code.  What about this: you have to have `--clobber` to automatically overwrite, `--no-clobber` just updates, and if you have neither it displays the warning, tells you about the `clobber` options, and requires you press a key?

Any opinions, @quevon24?

quevon24:
Hmm, I think the option to update them by default is a good idea, and you can only pass a flag if you don't want them to be updated, simply showing that they already exist. Although I don't see any situation where you wouldn't want to update the data, since I believe the idea of this command is to clone courtlistener data in order to replicate an error with a specific cluster, Docker, opinion, etc.

florean:
@quevon24, just so I'm clear, you're saying just a `--no-clobber` and otherwise it just overwrites everything?  I think that's right if a large majority of the usage is clobbering data.  And I hate asking for user input in a command line tool unless necessary.  Final question, does `--no-clobber` still download all the new items?  It must, because otherwise why are you running the tool?
Cool, I'll write up an issue.

mlissner:
Sounds right to me.

quevon24:
Perfect, so this change is ready, when the checks pass I'll merge it

quevon24 (approved):

PR #5882, opened 2025-07-04
feat(search): Adds dismissible recap alerts tooltip

### LINKED ISSUE #5785: Tool Tip for RECAP Search Alerts
We'd like to add a tool tip for RECAP Search Alerts to more directly point users to the feature.

**Requirements:**
- On RECAP Archive search results, a tooltip should point to the alert bell icon.
- The tooltip should appear for all logged in users.
- The tooltip should appear in the RECAP Archive until the user closes it. 
- Once the user closes the tooltip, it should not appear for that user again.

**Content:**
- Heading: RECAP Search Alerts
- Body: Set an alert to get notified when new RECAP data matches your saved search.
- Learn More: links to blog
- x: closes the tooltip permanently

ERosendo:
This pull request implements the recap search tooltip as designed in #5785 . This tooltip is exclusively visible to logged-in users when they interact with the recap search

Once a user closes the tooltip, a new value is stored in local storage to ensure it remains hidden for that user in the future..

Additionally, a new waffle flag, `recap_alert_tooltip_enabled`, has been introduced. This flag allows us to merge the PR now and enable the tooltip independently when ready.

## Screenshots

<details>
<summary>Extra-Large Screens</summary>

image

</details>

<details>
<summary>Large Screens</summary>

image

</details>

<details>
<summary>Medium Screens</summary>

image

</details>

<details>
<summary>Small Screens</summary>

image

</details>

Fixes #5785


mlissner:
Looks good. I'll leave it to Elisa to do the full review.

cl/search/templates/search.html
```diff
@@ -224,6 +229,27 @@
                                    data-target="#modal-save-alert"
                                    title="{% if request.GET.edit_alert %}Edit this alert{% else %}Create new alert{% endif %}"></i>
                                 </span>
+                                {% flag "recap_alert_tooltip_enabled" %}
+                                {% if search_form.type.value == SEARCH_TYPES.RECAP and user.is_authenticated %}
```
mlissner: Do we need to only do it for auth'ed users? Why not do it for anonymous too?

mlissner:
I don't think we have a way of checking for multi-users, but I think the trend is to assume one user per device and I've seen think-pieces talking about how the "Keep me logged in" check box should be increasingly on by default. I feel fine using local storage for this. Worst case, somebody doesn't get our new feature nag.

Eduardo, can you make an issue to clean up this and any other nags/waffles/banners we've used for recap alerts? Or maybe the issue is to convert these nags into Oral Arg nags? Dunno which is better, but it'd be nice to clean as we go.

cl/search/templates/search.html
ERosendo: > Do we need to only do it for auth'ed users? Why not do it for anonymous too?

This was a requirement from issue #5785, but we can easily adjust the template to show the new tooltip to all users. Let me know if you'd like me to implement that.

mlissner: I think that was because I was imagining it set to the user's session, not to their browser. Since it's client-side only, we can do it for everybody unless you can think of a reason that'd be a mistake.

ERosendo: > we can do it for everybody unless you can think of a reason that'd be a mistake.

I believe showing the tooltip to all users is fine. The banner is already visible to everyone, and the tooltip's purpose is to highlight the alerts feature. When anonymous users click the bell icon, they'll be prompted to create an account, which I think is a good flow.

ERosendo:
> can you make an issue to clean up this and any other nags/waffles/banners we've used for recap alerts?

Sure!

ERosendo:
@mlissner I've updated the template to display the tooltip for anonymous users.  Here's the issue to keep an eye on for cleaning up those waffle flags:

https://github.com/freelawproject/courtlistener/issues/5915

elisa-a-v (approved):
LGTM 🫡

PR #5883, opened 2025-07-04
test(tests): serialize ElasticSearch tests

florean:
Add Django's SerializeMixin to ESIndexTestCase to enforce serialization of ElasticSearch tests. A similar technique is used for the Selenium tests.  This helps ensure tests executed in parallel won't clobber each other while using the shared ES server. It has a negligible impact to test speed.
I haven't done a thorough dig through all the tests, but after a once-over it looks like this should cover most if not all of the tests that use ElasticSearch.

### DIFF
--- a/cl/tests/cases.py
+++ b/cl/tests/cases.py
@@ -1,6 +1,5 @@
 import re
 from datetime import datetime
-from pathlib import Path
 from typing import cast
 from urllib.parse import parse_qs, urlparse
 
@@ -124,8 +123,7 @@ class APITestCase(
     ELASTICSEARCH_DISABLED=False,
 )
 class ESIndexTestCase(SerializeMixin, SimpleTestCase):
-    lockfile = __file__ + "-ESIndexTestCase"
-    Path.touch(lockfile, exist_ok=True)
+    lockfile = __file__
 
     @classmethod
     def setUpClass(cls):


mlissner:
Now that I'm looking at this, I remember that we used to have this. I forget why we stopped, but it does seem like a good idea, thank you.

albertisfu (commented):
Thanks, @florean, for looking into this.

However, we don't believe Elasticsearch tests need to be serialized.
The `ESIndexTestCase` already includes a mechanism that creates a separate index for each test class, which helps avoid collisions between different ES test classes running in parallel.

We’re aware that a few ES tests occasionally fail, but we don’t believe these failures are related to test collisions. Those tests likely require a bit more investigation to understand what's going on.

florean:
Ah, yeah, I see that.  Maybe it just needs some fixes, because I get persistent failures that are fixed by rebuilding my `cl-es` container.  So some state is definitely getting saved to ElasticSearch that survives past a single run.  It may be as simple as some missing `super` calls or MRO ordering.  I think I was still learning this part of the code and reached for the simplest solution (which doesn't even solve the problem).  Thanks for the pointer.

What's the best way to ask questions about the code?

mlissner:
> What's the best way to ask questions about the code?

It can be here, but Slack works too, if you want an invite?

florean:
Sure, I think that'd be useful if you don't mind.

PR #5884, opened 2025-07-04
feat(help_pages): change includes to django-cotton components

nazarbodak221:
Convert include files to components

For better code organization, these components were moved to a new `cotton/` directory. This change improves modularity, makes the parent templates cleaner, and simplifies the reuse of UI elements.

Just let me know if you'd like different styles or colors – I'll change them.

Also sidebar to migration guide(includes) I'd like not to write in cotton, but write in migration guide directly. Say if I can do it. 

Affected files:
- `court_endpoint.html`
- `court_id_mappings.html`
- `deprecation_notice.html`
- `docket_endpoint.html`
- `v3_deprecated_warning.html`

Screenshots:
- court_endpoint.html:
image
- court_id_mapping.html:
image
- deprecation_notice.html:
image
- docket_endpoint.html:
image
- v3_deprecated_warnings:
image

### DIFF
--- a/cl/api/templates/cotton/court_endpoint.html
+++ b/cl/api/templates/cotton/court_endpoint.html
@@ -1,0 +1,4 @@
+{# court_endpoint_description.html #}
+<p>This API contains data about the courts we have in our database, and is joined into nearly every other API so that you can know where an event happened, a judge worked, etc.</p>
+<p>To look up field descriptions or options for filtering, ordering, or rendering, complete an HTTP <code>OPTIONS</code> request.</p>
+<p>You can generally cache this API. It does not change often.</p>
--- a/cl/api/templates/cotton/court_endpoint.html
+++ b/cl/api/templates/cotton/court_endpoint.html
@@ -1,4 +1,4 @@
-{# court_endpoint_description.html #}
+{# court_endpoint.html #}
 <p>This API contains data about the courts we have in our database, and is joined into nearly every other API so that you can know where an event happened, a judge worked, etc.</p>
 <p>To look up field descriptions or options for filtering, ordering, or rendering, complete an HTTP <code>OPTIONS</code> request.</p>
 <p>You can generally cache this API. It does not change often.</p>
--- a/cl/api/templates/cotton/deprecation_notice.html
+++ b/cl/api/templates/cotton/deprecation_notice.html
@@ -1,0 +1,18 @@
+{# deprecation_notice.html #}
+<c-callout type="warning" title="API Deprecation Notice">
+  <p>
+    <strong>Heads up!</strong> Versions 1 and 2 of this API are deprecated as of <strong>January 31, 2016</strong>.
+    Please switch to <a href="{% url "rest_docs" %}" class="underline">the latest API version</a> for the best experience.
+  </p>
+
+  <ul class="ml-6 mb-2">
+    <li><strong>February 29:</strong> No new content will be added to these versions.</li>
+    <li><strong>March 31:</strong> These APIs will return empty responses.</li>
+    <li>After that, all related pages will be gradually removed.</li>
+  </ul>
+
+  <p>
+    We know change can be tricky, but we’re here to help! If you have questions or need a hand migrating,
+    <a href="{% url "contact" %}" class="underline">contact us</a> — we’re happy to assist.
+  </p>
+</c-callout>
--- a/cl/assets/tailwind/input.css
+++ b/cl/assets/tailwind/input.css
@@ -57,6 +57,20 @@
 @tailwind components;
 
 @layer components {
+  .alert {
+    @apply px-4 py-3 text-sm rounded-[10px] border;
+  }
+
+  .alert-warning {
+    @apply alert text-yellow-700 bg-yellow-100 border-yellow-500;
+  }
+
+   .alert-danger {
+    @apply alert bg-red-100 text-red-700 border-red-500;
+  }
+}
+
+@layer components {
   .btn {
     @apply py-2.5 px-3.5 rounded-lg w-fit text-sm font-normal flex gap-1 items-center;
   }
--- a/cl/assets/tailwind/input.css
+++ b/cl/assets/tailwind/input.css
@@ -65,12 +65,10 @@
     @apply alert text-yellow-700 bg-yellow-100 border-yellow-500;
   }
 
-   .alert-danger {
+  .alert-danger {
     @apply alert bg-red-100 text-red-700 border-red-500;
   }
-}
 
-@layer components {
   .btn {
     @apply py-2.5 px-3.5 rounded-lg w-fit text-sm font-normal flex gap-1 items-center;
   }
--- a/cl/assets/templates/cotton/callout.html
+++ b/cl/assets/templates/cotton/callout.html
@@ -1,0 +1,10 @@
+{# callout.html #}
+
+<c-vars title="Listen Up!" type="warning" />
+
+<div class="alert alert-{{ type }}">
+    <div>
+        <strong class="font-semibold">{{ title }}</strong>
+        {{ slot }}
+    </div>
+</div>
--- a/cl/assets/templates/cotton/callout.html
+++ b/cl/assets/templates/cotton/callout.html
@@ -3,8 +3,6 @@
 <c-vars title="Listen Up!" type="warning" />
 
 <div class="alert alert-{{ type }}">
-    <div>
         <strong class="font-semibold">{{ title }}</strong>
         {{ slot }}
-    </div>
 </div>
--- a/cl/api/templates/cotton/court_id_mappings.html
+++ b/cl/api/templates/cotton/court_id_mappings.html
@@ -1,0 +1,16 @@
+{# court_id_mappings.html #}
+<c-data-table
+  safe
+  caption="CourtListener court IDs match the subdomains on PACER, except for the following mapping:"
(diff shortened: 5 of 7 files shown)


cl/assets/templates/cotton/callout.html
```diff
@@ -0,0 +1,10 @@
+{# callout.html #}
+
+<c-vars title="Listen Up!" type="warning" />
+
+<div class="alert alert-{{ type }}">
+    <div>
+        <strong class="font-semibold">{{ title }}</strong>
+        {{ slot }}
+    </div>
```
elisa-a-v: Any reason to have that nested `<div>` without any attributes or siblings?

cl/assets/tailwind/input.css
```diff
@@ -56,6 +56,20 @@
...
+
+  .alert-warning {
+    @apply alert text-yellow-700 bg-yellow-100 border-yellow-500;
+  }
+
+   .alert-danger {
+    @apply alert bg-red-100 text-red-700 border-red-500;
+  }
+}
+
```
elisa-a-v: We already have a `components` layer, is there a reason you added a new one? If not, could you move these to the existing one please?

cl/api/templates/cotton/court_endpoint.html
```diff
@@ -0,0 +1,4 @@
+{# court_endpoint_description.html #}
```
elisa-a-v: This is rreally minor but is there a purpose for these comments at the top of the files? If so, this one may need fixing since it says `court_endpoint_description.html` but the file name is `court_endpoint.html`.

cl/api/templates/cotton/deprecation_notice.html
```diff
@@ -0,0 +1,18 @@
...
+    <li><strong>February 29:</strong> No new content will be added to these versions.</li>
+    <li><strong>March 31:</strong> These APIs will return empty responses.</li>
+    <li>After that, all related pages will be gradually removed.</li>
+  </ul>
+
+  <p>
+    We know change can be tricky, but we’re here to help! If you have questions or need a hand migrating,
+    <a href="{% url "contact" %}" class="underline">contact us</a> — we’re happy to assist.
+  </p>
+</c-callout>
```
elisa-a-v: The contents here are not exactly the same as the ones in [the current include](https://github.com/freelawproject/courtlistener/blob/main/cl/api/templates/includes/deprecation_notice.html). Was the new copy taken from anywhere in particular?

@mlissner I guess we only need this notice if we're redesigning the v1 and v2 API help pages, is this something we want? The notice will soon be 10 years old! Maybe [the other warning](https://github.com/freelawproject/courtlistener/pull/5884/files#diff-b3e926b96322104b5562a347addfae6c60906906b46da6decfb923ed7b463dcf) could be enough anyway.

mlissner: v1 and v2 of the API documentation may be removed. ✅ 

I warned nine years ago that that might happen "soon"!

cl/assets/templates/cotton/callout.html
nazarbodak221: Ohh, maybe I wanted to use some classes here, but then delete it, I'll remove this div

cl/assets/tailwind/input.css
nazarbodak221: Sure, I missed that it already exist, I'll move it

cl/api/templates/cotton/court_endpoint.html
nazarbodak221: That may be cpoy-paste error, I'll fix it

cl/api/templates/cotton/deprecation_notice.html
nazarbodak221: So, I should delete this or not?

nazarbodak221:
> Nice, thank you! Most of the code looks great and I really like the callout component you added; I just left a few comments/questions in the code.
> 
> Just one more thing: could you please add entries in the [component library](https://www.courtlistener.com/components/) for the new components? You can use the other entries as a reference.
> 
> We already have a support plea banner that is pure markup, maybe we can rename that last section from "Other components" to "Pure markup" or something similar and include all the API components in there, what do you think?
> 
> I'd place the callout in the Layout components section but feel free to suggest other ideas if you have any :) thanks again!

nazarbodak221:
Also I add callout to component library and rename "Other components" to "Pure markup"

Pls check it and if I need to change something before push it 

Screenshots how it look now :

image

image

elisa-a-v:
> Pls check it and if I need to change something before push it

Looks pretty good, thanks! My only note is the code snippet should correspond with the demos, so your `<c-code>` should include all three versions with the same attributes and content you used. However, the code that's visible in your screenshot looks really nice because it showcases the flexibility of the slot, so maybe use that one for the "warning" type in the demo instead?

It makes sense to me to keep all three types in the demo, it's good to have them visible at a glance!

Oh and if the `title` is optional, please add the `optional` flag (for reference, the props in the Code block component have it)

cl/api/templates/cotton/deprecation_notice.html
elisa-a-v: Yeah let's delete this one. When implementing the redesigned API help pages, we won't do v1 nor v2, so this notice won't be used anywhere. We'll keep v3 with the simpler deprecation warning.

nazarbodak221:
image

If this is what you expected, I'll wait for you to merge this PR and start to refactor other PRs

nazarbodak221:
image

Last check if this is what you expected to see

elisa-a-v:
> Last check if this is what you expected to see

I can't see the updated demo but as long as it uses the same code as the one in the `<c-code>` block you're showing here, then yeah, that looks great, thanks!

nazarbodak221:
@mlissner Since Elisa approved this PR, could you please merge it into main?
I’d like to pull it into other branches to start updating the templates from includes to components.

Otherwise, I’ll need to merge this into each branch separately, which could get messy.

mlissner:
It looks like she requested changes, and it's hers to finalize, so I shouldn't jump the gun, but you should be able to rebase your other branches onto this one pretty easily, I think?

mlissner:
Also, it looks like the lint CI is failing, @nazarbodak221.

nazarbodak221:
> It looks like she requested changes, and it's hers to finalize, so I shouldn't jump the gun, but you should be able to rebase your other branches onto this one pretty easily, I think?

Sure, I can rebase the branches, but just so you know, it'll rewrite history. If you're okay with that, I'll do it
Also I fix bugs and CI is good now

mlissner:
Messing around with history in your branches is fine. I don't see a problem with that, unless I'm misunderstanding?

nazarbodak221:
@elisa-a-v What changes did you request that I haven't done yet? Could you please remind me?

elisa-a-v:
Nothing you haven't done. Latest changes just haven't been reviewed so this hasn't actually been approved yet, but I'll get to it today.

cl/settings/django.py
```diff
@@ -249,3 +250,9 @@
     # Don't warn about HSTS being used
     "security.W004",
 ]
+
+TEMPLATES[0]["OPTIONS"]["loaders"] = [
+    "django_cotton.cotton_loader.Loader",
+    "django.template.loaders.filesystem.Loader",
+    "django.template.loaders.app_directories.Loader",
+]
```
elisa-a-v: I'm really curious about this bit. I read the [commit](https://github.com/freelawproject/courtlistener/pull/5884/commits/d9c9e47c44988f19c0dccd75dfd9cf7027090073) message but it's not clear to me why it's necessary since the system has been recognizing components just fine so far, and everything still works if I remove it.

I think we could get rid of this part, unless I'm missing something.

elisa-a-v (changes requested):
Library entry looks great, thank you very much!

I still have one more comment below, but once we've addressed that (either by removing or documenting why it's needed) I think we will be ready to merge.

cl/settings/django.py
nazarbodak221: This line was added to enable runtime template editing with `django-cotton`, which can be helpful during local development or quick UI iteration. It’s not needed in production — I meant to wrap it in a condition that applies it only in development, but that part slipped through.

nazarbodak221: Thinking about it again, you're right — this probably isn't necessary. I think the cleaner solution is to just remove it.

elisa-a-v:
I think something may have gone wrong with that last force push. I may be missing something but I would have expected just the one new commit removing the code, was there a reason for force pushing?

nazarbodak221:
> I think something may have gone wrong with that last force push. I may be missing something but I would have expected just the one new commit removing the code, was there a reason for force pushing?

This PR includes additional commits because it was based on a newer version of main. This is not an issue, as they already exist on main

mlissner:
Nazar, have you looked at the diff though? Something is quite wrong here. Check out the files tab of this PR, and you'll see it's modifying all kinds of things. Maybe just make a fresh PR with only the needed commits/changes?

nazarbodak221:
> Nazar, have you looked at the diff though? Something is quite wrong here. Check out the files tab of this PR, and you'll see it's modifying all kinds of things. Maybe just make a fresh PR with only the needed commits/changes?

You're right, I messed up here. Thanks for the suggestion — creating a clean PR is a good idea, I'll do it.

PR #5887, opened 2025-07-05
Optimize new homepage actions

### LINKED ISSUE #5886: Optimize homepage action buttons
Now that we've got the actions on the homepage and I can see and think about them, I think they need a few changes:

 - Send us docs --> About RECAP
 - The icon for get our newsletter is inconsistent (it should match the footer)
 - Create custom podcast doesn't say "of what"

I'm going to send a rough PR for this in a sec, but I won't be able to take it over the line, so hopefully @elisa-a-v can finish any missing pieces.

This was a fun exercise. The workflow is smooooooth.

mlissner:
Fixes #5886.

This is a draft PR that does most of the work identified in #5886. I did identify one issue though, which is that I was not able to just use the newspaper logo from the footer because it had specific `fill` values. 

I was able to fix it by swapping those values to use `currentColor` instead, and I added an SVG with that value for the `fill`, but that's not great to have two nearly identical SVGs. 

I think the fix is to use currentColor for the logos in the footer, and to then color them with css, but I didn't have time to figure that out, so I leave that to you, if it makes sense, @elisa-a-v. 

Otherwise, I think this is pretty good and pretty ready to go.

Thank you! 

And a screenshot just for fun:

image


cl/search/templates/v2_homepage.html
```diff
@@ -108,26 +119,16 @@ <h1 class="w-full text-center lg:text-display-sm-cooper text-xl">Search millions
...
-        icon="file_text"
-      >
-        Send&nbsp;us PACER&nbsp;Docs
+        Find a<br>Citation
       </c-icon-link>
 
       <c-icon-link
         x-on:focus="scrollIntoView"
         href="{% url "podcasts" %}"
+        aria-label="Subscribe to our Oral Argument Podcasts, or create your own"
```
mlissner:
```suggestion
        aria-label="Subscribe to our Oral Argument Podcasts or create your own"
```

mlissner:
Thanks Elisa. I made one suggestion because you used to be able to commit your own suggestions (it was a handy tool I lazily used), but the new github review UI doesn't seem to allow that. Anyhow, I did it in my editor, pushed and it's now set for auto-mege. 

Great commit messages by the way. I hadn't noticed before that you're doing such a great job.

PR #5888, opened 2025-07-06
fix(tests): Disable result buffering due to bug

florean:
Using `--buffer` and `--parallel` can cause all test failure log messages, failed test stack traces, and all test results to be erased.  See Django bug [#36491](https://code.djangoproject.com/ticket/36491) for more details.  Until that is fixed and we've upgraded to it, we should use safe defaults.  I'm writing up a Discussion giving an example of how painful this makes test failures and I'll link it when finished.

### DIFF
--- a/cl/tests/runner.py
+++ b/cl/tests/runner.py
@@ -67,7 +67,7 @@ class TestRunner(DiscoverRunner):
         # Default buffering on, to hide output
         # This is disabled due to Django bug #36491.
         # See PR #5888 for more details.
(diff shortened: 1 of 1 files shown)


mlissner:
Didn't see the discussion, but I believe you about this being painful. We'll revert until that bug is fixed, thanks. Too bad.

mlissner (approved):

florean:
Sorry, I had a busy weekend and it just kept getting longer.  It is now up at #5894.

PR #5899, opened 2025-07-07
feat(search): Enhances opinion search to enable semantic search

ERosendo:
Key changes:

- Adds a new `semantic` field to the `SearchForm` class, enabling users to toggle the semantic search feature.

- Adds the `inception-query` entry to the `MICROSERVICE_URLS` dictionary, allowing requests to retrieve embeddings for individual strings from Inception.

- Moves the `clean_up_recap_document_file` helper to `microservice_utils.py` to resolve circular import issues.

- Introduces new settings to fine tune the semantic search behavior:

   - `MAX_EMBEDDING_CHAR_LENGTH`: sets the maximum character length for text to embed. This should align with the corresponding limit in Inception to prevent sending queries that it cannot process.
 
   - `KNN_SEARCH_BOOST`: Adjusts the boost factor applied to KNN-based search scoring, allowing fine-tuning of its influence in hybrid search results.
  
   - `KNN_SEARCH_K`: defines the number of nearest neighbors (`k`) to  retrieve in KNN vector search queries.

- Adds a new exception (`InputTooLongError`) to flag cases where text exceeds the MAX_EMBEDDING_CHAR_LENGTH, allowing for graceful handling and clearer error reporting.

- Introduces `build_semantic_query`, a helper that combines vector-based KNN search with optional exact keyword matching for flexible case law retrieval.

- Excludes embedding fields from opinion queries to reduce payload size. These fields are currently unused in the API response and add unnecessary data transfer overhead

- Semantic Integration in `build_full_join_es_queries`: Refactors the query-building logic to support two semantic search modes when the semantic flag is enabled:

   - Semantic-only Search
  
     Triggered when the user provides a `q` parameter without quoted phrases. The query applies KNN vector search exclusively to the child opinions documents. The structure looks like:

```yaml
query:
  bool:
    should:
      has_child:
        query:
          bool:
            filter:  # post-filtering
            should:
              nested:
                knn:
                  filter:  # pre-filtering
```

   Initially, I tried to preserve the structure we use in regular queries—where we match opinions via a has_child clause and clusters directly—but I realized that matching clusters doesn’t make sense for situational or semantically-driven queries. 

   - Hybrid Search
      Activated when the `q` parameter includes quoted terms, indicating a mix of semantic and exact search intent. The resulting query supports both semantic vector matching and traditional keyword matching on both opinions and clusters:

```yaml
query:
  bool:
    should:
      has_child:
        query:
          bool:
            filter:
            should:
              nested:
                query_string:
                query_string:
                  type: phrase
                knn:
                  filter:
      bool:
        filter:
        should:
          query_string:
          query_string:
            type: phrase
```

### DIFF
--- a/cl/lib/elasticsearch_utils.py
+++ b/cl/lib/elasticsearch_utils.py
@@ -2777,9 +2777,15 @@ def build_full_join_es_queries(
 
     q_should = []
     has_text_query = False
-    is_semantic_query = cd.get("semantic", False) and cd["type"] in [
+    # True if the user explicitly enabled semantic search and the search type
+    # supports it
+    semantic_search_enabled = cd.get("semantic", False) and cd["type"] in [
         SEARCH_TYPES.OPINION
     ]
+
+    # True if semantic search is enabled and the query has content to generate
+    # an embedding
+    has_valid_semantic_query = semantic_search_enabled and cd.get("q", "")
     keyword_text_query = ""
     match cd["type"]:
         case (
@@ -2843,7 +2849,7 @@ def build_full_join_es_queries(
             child_text_query = [mlt_query]
         else:
             string_query = cd.get("q", "")
-            if is_semantic_query and string_query:
+            if has_valid_semantic_query:
                 keyword_text_query, child_text_query = build_semantic_query(
                     string_query,
                     child_fields,
@@ -2919,10 +2925,14 @@ def build_full_join_es_queries(
         # Build the parent filter and text queries.
         string_query = build_fulltext_query(
             parent_query_fields,
-            keyword_text_query if is_semantic_query else cd.get("q", ""),
+            keyword_text_query
+            if has_valid_semantic_query
+            else cd.get("q", ""),
             only_queries=True,
         )
-        has_text_query = True if string_query or is_semantic_query else False
+        has_text_query = (
+            True if string_query or has_valid_semantic_query else False
+        )
 
         # If child filters are set, add a has_child query as a filter to the
         # parent query to exclude results without matching children.
@@ -2967,9 +2977,7 @@ def build_full_join_es_queries(
                     minimum_should_match=1,
                 )
         should_append_parent_query = (
-            parent_query
-            and not mlt_query
-            and not (is_semantic_query and cd.get("q", ""))
+            parent_query and not mlt_query and not has_valid_semantic_query
         ) or keyword_text_query
         if should_append_parent_query:
             q_should.append(parent_query)


cl/lib/elasticsearch_utils.py
```diff
@@ -2679,6 +2685,73 @@ def apply_custom_score_to_main_query(
...
+          and optionally a keyword-based full-text query.
+    :raises InputTooLongError: If the cleaned query string exceeds the maximum allowed length
+        for generating embeddings.
+    """
+    semantic_query: list[Query] = []
+    # Extract quoted phrases from the input string (for exact keyword matching)
+    exact_keywords = re.findall(r'"([^"]*)"', text_query)
+
+    # Join extracted phrases with AND to form a keyword query string
+    keyword_query = " AND ".join([f'"{s}"' for s in exact_keywords])
```
albertisfu: It's not necessary to join the exact terms with `AND`, since the resulting query string uses `AND` as the default operator.


```
{
    "query_string":{
       "query":"\"cases before\"  \"questions which\"",
       "default_operator":"AND",
       "fields":[
          "court",
          "court_id",
          "citation",
          "judge",
          "caseNameFull",
          "status",
          "suitNature",
          "attorney",
          "procedural_history",
          "posture",
          "syllabus",
          "text^1.0",
          "type^1.0",
          "caseName.exact^4.0",
          "docketNumber^2.0"
       ],
       "fuzziness":2,
       "tie_breaker":0.3
    }
 }
```

cl/lib/elasticsearch_utils.py
```diff
@@ -2679,6 +2685,73 @@ def apply_custom_score_to_main_query(
...
+    )
+    embedding_request.raise_for_status()
+    vectors = embedding_request.json()["embedding"]
+
+    # If exact keyword query exists, build and add full-text query to results
+    # This enables hybrid search by combining keyword and semantic results
+    if keyword_query:
+        semantic_query.extend(
+            build_fulltext_query(fields, keyword_query, only_queries=True)
+        )
```
albertisfu, line 2734: I ran some tests regarding the structure of the generated query:

```
query:
  bool:
    should:
      has_child:
        query:
          bool:
            filter:
            should:
              query_string:
              query_string:
                type: phrase
              nested:
                knn:
                  filter:
```

The `minimum_should_match` clause for the should query is set to 1.

This means it can return any document where at least one of the sub-queries matches, either any of the `query_string` queries or the nested vector query.

I'm not sure if that's the intended behavior, but in this case, it can return documents where the nested vector query didn't match at all.

Since the `semantic` parameter is enabled, this feels a bit confusing, as I would expect it to require a match on the semantic query and then refine the results based on exact keywords?

If that's the case, I believe we can fix this by using the following query structure:

```
query:
  bool:
    should:
      has_child:
        query:
          bool:
            filter:
            should:
              query_string:
              query_string:
                type: phrase
             must
              nested:
                knn:
                  filter:
```
In this version, the nested query is wrapped in a `must` clause, ensuring that only results where the vector query matched are returned.

albertisfu (changes requested):
Thanks @ERosendo I just left a couple of suggestions and one extra here:

I have a question regarding the bool cluster query. 
```
query:
  bool:
    should:
      has_child:
          ...
      bool:
        filter:
        should:
          query_string:
          query_string:
            type: phrase
```

In the description you mentioned:
> The resulting query supports both semantic vector matching and traditional keyword matching on both opinions and clusters.

I want to confirm if it's correct to match only clusters in a hybrid search. What I mean is, this would match clusters where the exact terms match any of the `query_string` fields, even if those clusters have nothing to do with the opinion vectors. For instance, if exact keywords match clusters with a high score, they could hide semantic results that actually matched from appearing among the top results.

cl/lib/elasticsearch_utils.py
ERosendo: > It's not necessary to join the exact terms with AND, since the resulting query string uses AND as the default operator.

Got it! Thanks for the reminder. I'll update the code now.

cl/lib/elasticsearch_utils.py
ERosendo: > Since the semantic parameter is enabled, this feels a bit confusing, as I would expect it to require a match on the semantic query and then refine the results based on exact keywords?

That makes sense—I agree it could be bit confusing. I like the updated query you shared. I’ll update the helper function to use a `must` clause instead of `should`.

ERosendo: Thanks for your comment!

While working on this PR, I spent some time digging into how hybrid search works and reviewed several sources. Most define it as combining at least one lexical (keyword-based) query with one semantic query, then merging the results from both.

This [Elastic tutorial](https://www.elastic.co/search-labs/tutorials/search-tutorial/vector-search/hybrid-search) offers a good example using RRF to blend results from the two approaches.

I don’t think the query using must aligns with that definition of hybrid search, since it requires all results to match the semantic query and only boosts those that also include keywords from the query_string. That means purely keyword-based matches would be excluded—so instead of combining multiple sources or criteria, we're effectively filtering by the must clause alone.

I believe this also addresses your question here: https://github.com/freelawproject/courtlistener/pull/5899#pullrequestreview-3040402626, about including results based on keyword matches alone.

Please feel free to correct me if I’m misunderstanding anything!

ERosendo:
@albertisfu Thanks for your review! 

> Another question is whether we'll support highlighting for semantic results, like showing the chunk that matched the vector query.

That's a great question. With our current query structure, we can tweak the nested query and use the `inner_hits` object to display the chunk of text that matched the KNN query.

I tried using the highlight key in the request (similar to how we handle other fields in search request) but I couldn’t get any highlighted fragments to show. I suspect that’s because there’s no exact mapping between the individual words in the chunks and the generated vectors.

Given that, I think the best we can do is rely on `inner_hits` to return the chunk that matched the KNN query.

mlissner:
> Given that, I think the best we can do is rely on inner_hits to return the chunk that matched the KNN query.

I'm just curious if you have an example of what an inner hit looks like?

cl/lib/elasticsearch_utils.py
albertisfu: Thank you for the clarification. I understand better now what hybrid search means: both keyword-based and semantic results should be combined.

Based on that, I have two additional questions:

- One concern with this approach of returning both keyword-based and KNN results using a bool clause is that keyword-based results can obfuscate semantic ones, even if the latter matched.

  This part explains it well.

  https://www.elastic.co/search-labs/blog/hybrid-search-elasticsearch

  > Running a hybrid search query usually boils down to sending a mix of at least one lexical search query and one semantic search query and then merging the results of both. The lexical search results are scored by a similarity algorithm, such as [BM25](https://en.wikipedia.org/wiki/Okapi_BM25) or [TF-IDF](https://en.wikipedia.org/wiki/Tf%E2%80%93idf), whose score scale is usually unbounded as the max score depends on the number and frequency of terms stored in the inverted index. In contrast, semantic search results can be scored within a closed interval, depending on the similarity function that is being used (e.g., [0; 2] for cosine similarity).
  > 
  > In order to merge the lexical and semantic search results of a hybrid search query, both result sets need to be fused in a way that maintains the relative relevance of the retrieved documents, which is a complex problem to solve. Luckily, there are [several existing methods](https://arxiv.org/abs/2210.11934) that can be utilized; two very common ones are Convex Combination (CC) and Reciprocal Rank Fusion (RRF).

  Actually, Rachel mentioned this [here](https://github.com/freelawproject/infrastructure/issues/406#issuecomment-3059528935) 

  For hybrid search, it’s possible to use RRF to properly rank and merge results from keyword-based and KNN queries, but this is an Enterprise feature we don’t currently have access to. So we’re limited to using the standard query tools available.

  Maybe @rachlllg has some thoughts on this, but one simple option could be applying a boost (positive or negative) to either the keyword-based query or the KNN query to adjust their relative relevance. This would work as long as the scores are easily comparable or tunable.

  However, based on the blog post Rachel referenced, merging search results doesn't seem as straightforward as just applying a regular boost. Another option could be crafting our own RRF-like scoring using a painless script via a custom function score. I’m not sure how complex that would be, but it’s something to consider.

- My second question is about the decision to enable hybrid search only when a query contains quoted terms (""). In one of the examples you [shared](https://www.elastic.co/search-labs/tutorials/search-tutorial/vector-search/hybrid-search), it looks like the same search terms are passed to both the keyword-based and semantic queries.

  Was the decision to restrict hybrid search to quoted terms based on observed CL user behavior? If so, I think we should just document that using quotes is the way to trigger hybrid search.

  Another idea I was considering is enabling an explicit hybrid parameter, so the entire query would be passed to both the keyword and KNN queries. However, I’m not sure if that makes sense, especially given the different nature of a regular semantic query.

ERosendo:
> I'm just curious if you have an example of what an inner hit looks like?

@mlissner Here’s an example response from Elasticsearch after tweaking the query to include the chunk that matched the KNN search:

[result_w_inner_hit.json](https://github.com/user-attachments/files/21372366/result_w_inner_hit.json)

The query I used was: `Is a tenant allowed to break a lease unexpectedly?`. 

To find the relevant chunk of text in the response, navigate to:
`inner_hits.embeddings.hits.hits[0].fields.embeddings.chunk`
This field contains the actual fragment from the opinion that semantically matched the query.

The response structure can be a bit hard to follow because Elasticsearch introduces an additional nested level to represent these matching text chunks within the `inner_hits`.

image

cl/lib/elasticsearch_utils.py
ERosendo: > Maybe @rachlllg has some thoughts on this, but one simple option could be applying a boost (positive or negative) to either the keyword-based query or the KNN query to adjust their relative relevance. This would work as long as the scores are easily comparable or tunable.

Yes, this PR introduces the `KNN_SEARCH_BOOST` setting to avoid needing a new PR every time we want to tweak the boost for the KNN component.

> However, based on the blog post Rachel referenced, merging search results doesn't seem as straightforward as just applying a regular boost. 

That sounds right. @rachlllg also mentioned we’ll need to run some experiments to find the most effective boost setting.

> Another option could be crafting our own RRF-like scoring using a painless script via a custom function score. I’m not sure how complex that would be, but it’s something to consider.

That's an interesting idea. I think this is the first time it's come up. I didn’t see it mentioned in any of the related issues. Would your idea involve using the function score to normalize the lexical results so they’re on a similar scale to the KNN scores?

> My second question is about the decision to enable hybrid search only when a query contains quoted terms (""). 

@rachlllg suggested using quotation marks as the trigger for hybrid search, which makes sense given how Elastic describes hybrid search in [this post](https://www.elastic.co/what-is/hybrid-search):

> Typically, hybrid search combines keyword search and [semantic search](https://www.elastic.co/what-is/semantic-search), utilizing advanced [machine learning techniques](https://www.elastic.co/what-is/machine-learning). Semantic search retrieves results based on the meaning of the text, while full-text search focuses on **exact word** matches.

Using quotation marks is a natural way to signal that the user wants exact matches, so it fits the definition well.

> In one of the examples you [shared](https://www.elastic.co/search-labs/tutorials/search-tutorial/vector-search/hybrid-search), it looks like the same search terms are passed to both the keyword-based and semantic queries.

Yes, I noticed that too when reading the article. That said, I’m not sure how accurate the keyword matching would be if we sent full situational queries instead of extracting just the key phrases. Elastic will likely try to match the entire string, which could definitely skew the keyword relevance scores.

> Was the decision to restrict hybrid search to quoted terms based on observed CL user behavior?

I don't think so—it’s more aligned with the definition I mentioned above.

> I think we should just document that using quotes is the way to trigger hybrid search.

Totally agree—that's definitely something we’ll include in the documentation before release.

> Another idea I was considering is enabling an explicit hybrid parameter, so the entire query would be passed to both the keyword and KNN queries. 

Do you mean adding a new field to the `SearchForm` to explicitly toggle hybrid search, instead of relying on quotation marks as the trigger?

albertisfu: > Yes, this PR introduces the KNN_SEARCH_BOOST setting to avoid needing a new PR every time we want to tweak the boost for the KNN component.

Got it! I agree we should start running experiments and try to tune this value to assess the accuracy of the results before considering anything more complex that might not be necessary.

> That's an interesting idea. I think this is the first time it's come up. I didn’t see it mentioned in any of the related issues. Would your idea involve using the function score to normalize the lexical results so they’re on a similar scale to the KNN scores?

Yeah, exactly! Basically, the idea is to introduce a custom score function that reproduces the RFF behavior.

Screenshot 2025-07-22 at 1 01 49 p m

But I think it's better to first assess the accuracy of the results with what we have and only do this if it's necessary.

> Using quotation marks is a natural way to signal that the user wants exact matches, so it fits the definition well.
> Yes, I noticed that too when reading the article. That said, I’m not sure how accurate the keyword matching would be if we sent full situational queries instead of extracting just the key phrases. Elastic will likely try to match the entire string, which could definitely skew the keyword relevance scores.
> Totally agree—that's definitely something we’ll include in the documentation before release.

Got it. Yeah this makes sense. Documentation should be enough to make clear how to trigger hybrid queries.

> Do you mean adding a new field to the SearchForm to explicitly toggle hybrid search, instead of relying on quotation marks as the trigger?

Yeah, but based on your previous explanation, it seems more controllable and reliable to use quotation marks to perform hybrid search.

albertisfu:
Thanks @ERosendo 

- As we talked, we should create an issue to evaluate the use of chunks for highlighting semantic queries, so we can agree on how to display them in both semantic-only and hybrid queries.

- It's a good idea to put the semantic feature behind a waffle flag and also add a setting to disable it by default, since making it work requires setting up a CPU-only Inception instance for embedding queries.

cl/lib/elasticsearch_utils.py
rachlllg: On the RRF calculation, just to clarify, as we can see from the [blog post](https://www.elastic.co/search-labs/blog/hybrid-search-elasticsearch), there are two methods of doing hybrid search, one is using RRF formula, another is using CC formula, each have their own pros and cons, and both are valid methods of performing hybrid search. The natively supported version (by using the query parameter) uses the CC formula. Notably, regardless of which formula we use, we need to tune some hyperparameters (weights for CC, k and ranking quality for RRF) - ie, we need to experiment to find the best value to use. I would not say RRF is a "better" method than CC, they are simply different methods. Since CC is already natively supported via the use of the query parameter, I think we can try it and see how it performs before trying to implement RRF ourselves.

A "side-effect" of hybrid search is (and this is regardless of which formula we use), under the hood, the semantic and the keyword queries are executed in **parallel**, each retrieving their own subset of results. The formulas are used to rerank/rescore the two sets of results and output a final set of results based on the combined score. So the semantic results may not always have the keyword, and the keyword may not always be the one with the highest semantic score. This is the expected behavior. If we want to have the results be semantically similar AND have the keyword, then it would be a "filter", ie, retrieve the highest semantic results, then filter for only the results with the keyword. 

I recommend we use quotation marks to pass the keyword to keyword search and the rest of the query to semantic search. If quotation marks are missing, then we simply perform semantic search. I don't think we should pass the entire phrase to keyword search.

On highlighting, when using hybrid search and the retrieved result has no semantic inner hits, I think the highlight shall be applied to the keyword, otherwise, the highlight would be the entire paragraph returned by the semantic inner hit (~512 tokens).

ERosendo:
@albertisfu thanks for your feedback. this PR is ready for another review.

II've added a waffle flag named `enable_semantic_search` and a setting called `KNN_SEARCH_ENABLED`. The flag allows us to roll out the feature selectively to beta testers and staff, while the setting provides a global toggle to completely disable it. By default, the setting is `False`.

I've also updated the logic to use pre-filtering only for semantic-only queries, as suggested in https://github.com/freelawproject/infrastructure/issues/406#issuecomment-3059528935

cl/lib/elasticsearch_utils.py
```diff
@@ -2831,9 +2918,11 @@ def build_full_join_es_queries(
 
         # Build the parent filter and text queries.
         string_query = build_fulltext_query(
-            parent_query_fields, cd.get("q", ""), only_queries=True
+            parent_query_fields,
+            keyword_text_query if is_semantic_query else cd.get("q", ""),
+            only_queries=True,
         )
-        has_text_query = True if string_query else False
+        has_text_query = True if string_query or is_semantic_query else False
```
albertisfu: We should also update this clause. We need `has_text_query` to be True if `q` has content. We use `has_text_query` to determine whether we need to "replace" or "multiply" the scores from the custom function score. "replace" should be applied when there are only filters (which don't return any score) in the query, and "multiply" when there is a query string or a semantic query, which always returns a score.

I see that if `semantic_query`, `string_query` is empty here, so we might need to check the content of `q` directly.

ERosendo: You're correct. I'll tweak the code

albertisfu (approved):
Thanks @ERosendo this looks great ready to be merged!

Once this is deployed, the next step is to set up the CPU-based Inception instance to embed queries and tweak the required environment variables to start testing it.

PR #5930, opened 2025-07-08
Docket entry link

### LINKED ISSUE #5919: bug: Document page link to docket entry only works for first 165 entries
Expected behavior:
The [document page](https://www.courtlistener.com/docket/59771301/769/in-re-google-rtb-consumer-privacy-litigation/) includes a [link](https://www.courtlistener.com/docket/59771301/in-re-google-rtb-consumer-privacy-litigation/#entry-769) to the docket entry the document is associated with.  When I click on that link, I should be taken to corresponding docket entry on the docket page. (example inline)

Actual behavior:
The link only works if the docket entry is in the first page of paginated results (which in the example is 165).  Amusingly, sorting the docket entries preserves the URL fragment, so if the docket entry is in the last 165, after sorting the page will load at the correct docket entry.

florean:
Adds a new parameter to `view_docket` that allows direct linking to the docket page that contains the given docket entry.  This could be used by individuals, but it is primarily for the document page, which tries to link back to the docket entry in the docket and fails if the entry is not on the first page.  I named it `de`, but if you think it might be used elsewhere, we could call it `entry` or `docket_entry`. 
Updated the document template to use the new parameter.

Testing:
Manual testing so for against a docket with a [lot of entries](https://www.courtlistener.com/docket/59771301/in-re-google-rtb-consumer-privacy-litigation/).  If my approach is acceptable, I'll add some automated tests.

Resolves #5919.

### DIFF
--- a/cl/opinion_page/views.py
+++ b/cl/opinion_page/views.py
@@ -344,18 +344,17 @@ async def view_docket(
     if de_value := request.GET.get("de"):
         # Figure out which page the entry is in
         # This overrides all other parameters
-        entry_number_idx = 0
-        try:
-            entry_number = int(de_value)
-            async for de in de_list:
-                if entry_number == de.entry_number:
-                    break
-                entry_number_idx += 1
-            else:
-                entry_number_idx = 0
-        except ValueError:
-            entry_number_idx = 0
-        page = str(entry_number_idx // entries_per_page + 1)
+        de_entries = [x async for x in de_list.filter(entry_number=de_value)]
+        if de_entries:
+            de_entry = de_entries[0]
+            de_entry_idx = await sync_to_async(
+                de_list.filter(
+                    recap_sequence_number__lt=de_entry.recap_sequence_number
+                ).count
+            )()
+            page = str(de_entry_idx // entries_per_page + 1)
+        else:
+            page = "1"
         form = DocketEntryFilterForm()
         # Remove de param so it won't continue to clobber other ones
         qd = request.GET.copy()
--- a/cl/opinion_page/views.py
+++ b/cl/opinion_page/views.py
@@ -366,13 +366,9 @@ async def view_docket(
 
     @sync_to_async
     def paginate_docket_entries(docket_entries, docket_page):
-        paginator = Paginator(docket_entries, 200, orphans=10)
-        try:
-            return paginator.page(docket_page)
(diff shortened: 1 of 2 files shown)


florean:
The test failure isn't for this PR, but it is mine.  Submitting a fix shortly.

cl/opinion_page/views.py
```diff
@@ -340,39 +337,56 @@ async def view_docket(
...
+        entry_number_idx = 0
+        try:
+            entry_number = int(de_value)
+            async for de in de_list:
+                if entry_number == de.entry_number:
+                    break
+                entry_number_idx += 1
+            else:
+                entry_number_idx = 0
+        except ValueError:
```
mlissner: There can be as many as 20k docket entries in a docket. Pulling them all and iterating won't perform well. Can we do this in the DB instead?

florean: Sure!

florean:
It's up to you.  It's [this](https://www.courtlistener.com/docket/59771301/in-re-google-rtb-consumer-privacy-litigation/?entry_gte=766#entry-766) vs. [this](https://www.courtlistener.com/docket/59771301/in-re-google-rtb-consumer-privacy-litigation/?page=3#entry-766).  One thing I'll note is if you compare the two, you are getting very different results.  But I don't know how often docket entry numbers are out of chronological order.  Is that just erroneous dockets like this one?
I'll hold off on spending time writing tests.

mlissner:
I think it's fine just to do the simple thing for this one. The view_docket function is already pretty beefy.

florean:
This test failure looks like it could be a Redis test collision.  It has nothing to do with the changed code and usually passes.
```
FAIL: test_redis_email_counter (cl.users.tests.CustomBackendEmailTest.test_redis_email_counter)
Test logic to count the number of emails sent by the app
[--snip--]
  File "/opt/courtlistener/cl/users/tests.py", line 2410, in test_redis_email_counter
    self.assertEqual(int(r.get("test-email-counter:temp_counter")), 3)
    ^^^^^^^^^^^^^^^
[--snip--]
AssertionError: 4 != 3
```

mlissner:
> This test failure looks like it could be a Redis test collision.

Yeah it does! I totally agree. I'm going to go ahead and merge.

PR #5932, opened 2025-07-09
refactor(new_templates): remove stale code from new base and help index

### LINKED ISSUE #5405: Get rid of jQuery in new templates
The only thing still requiring jQuery is the anchors:

```html
{% block footer-scripts %}
  {% include "includes/anchors.html" %}
{% endblock %}
```

When we remove the jQuery import, we need to remove this from all templates that extend from `new_base.html`.

If we decide to keep the anchors, we also need to rewrite them using Alpine while keeping the current behavior (anchor uses heading `id` if provided, otherwise generates href with the heading text content), and also extending it by adding the option to disable anchors on a per-heading basis.

elisa-a-v:
Closes #5405 

This PR introduces a few small changes:

- Removed from the new base template:
	- jQuery
	- `base.js` script.
	- FontAwesome stylesheet.
- Removed the anchors include from the new help pages index.
- Removed duplicated copy in new help pages index ("The following help articles are currently available").
- As an accessibility bonus, added a tailwind class to the base new template to prevent smooth scroll behavior when the user prefers reduced motion. This is unrelated to other changes but it's so minor I thought it'd be okay to throw it in here.

Everything should look exactly the same as before except for the new help pages index, which feels a bit cleaner IMO:

<details>
<summary>Before</summary>

screenshot

</details>

<details>
<summary>After</summary>

screenshot

</details>

### DIFF
--- a/cl/assets/templates/new_base.html
+++ b/cl/assets/templates/new_base.html
@@ -56,8 +56,6 @@
   <link rel="manifest" href="{% static "json/manifest.json" %}">
   {% endblock %}
 
-  <link href="{% static "css/font-awesome.css" %}" rel="stylesheet">
-
   {% block privacy %}
     {% if private %}
       <meta name="robots" content="noindex, noodp, noarchive, noimageindex" />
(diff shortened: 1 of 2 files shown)


mlissner (approved):

mlissner:
Nice, thanks!

PR #5940, opened 2025-07-10
fix(audio): Use select_related to join on the related Docket to preve…

### LINKED ISSUE #2752: Oral argument sitemap times out
This is kind of surprising since it's the smaller of our sitemaps, but who knows. 

We should fix it somehow. Maybe we pre-generate our sitemaps using cursors? Maybe we make it have fewer items so it generates more quickly? Right now it's 50k items, which probably takes a while to generate as XML.

amattalols:
…nt N+1 query. Fixes #2752

Resources for QuerySet optimization:

* https://kenyattaanthony.medium.com/mastering-django-query-optimization-with-select-related-prefetch-related-and-fixing-n-1-5814ac31d169
* https://medium.com/codex/n-1-queries-problem-a-django-example-fd57e3a1761e
* https://docs.djangoproject.com/en/5.2/ref/models/querysets/#select-related

### DIFF
--- a/cl/audio/sitemap.py
+++ b/cl/audio/sitemap.py
@@ -12,7 +12,13 @@ class AudioSitemap(sitemaps.Sitemap):
     limit = 50_000
 
     def items(self) -> QuerySet:
-        return Audio.objects.filter(blocked=False).order_by("pk")
+        q = (
+            Audio.objects.filter(blocked=False)
+            .select_related("docket")
+            .only("date_modified", "pk", "docket__slug")
+            .order_by("pk")
+        )
+        return q
 
     def lastmod(self, obj: Audio) -> datetime:
         return obj.date_modified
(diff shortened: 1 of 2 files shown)


mlissner:
Thank you, this is great. Jeez, it's frustrating to learn that the thing messing these up for so long as an N+1 query...

amattalols:
The debug_sql query logs were very helpful in pointing it out, at least for Oral Arguments. That model doesn't have its own slug, so the URL location is based on the docket slug. I suspect latency/timeout issues with the other sitemaps will have different causes, but I wanted to keep this fix small.

The test failure (`FAIL: test_positions_connected_to_person (cl.people_db.tests.TestPersonWithChildrenFactory.test_positions_connected_to_person)`) appears to be a flake, but I'm not able to get failures locally.

ERosendo (approved):
Thanks @amattalols. LGTM :shipit:

PR #5956, opened 2025-07-13
test(favorites): Adds patch to mock the `check_prayer_pacer` task

ERosendo:
This PR adds a mock for the `check_prayer_pacer` task in the `RECAPPrayAndPay` and `PrayerAPITests` test classes. The mock prevents the test suite from attempting to check the availability of documents linked to newly created prayers during test execution.

The task is mocked using the `@patch` decorator with the `new` argument, which avoids the need to add the mock as an extra argument to each test method in the decorated classes.

This issue with the prayer tests was originally reported by @florean  here: https://github.com/freelawproject/courtlistener/discussions/5898#discussioncomment-13740151

### DIFF
--- a/cl/favorites/tests.py
+++ b/cl/favorites/tests.py
@@ -1,7 +1,7 @@
 import time
 from datetime import date, datetime, timedelta
 from http import HTTPStatus
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import time_machine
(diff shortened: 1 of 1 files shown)


florean:
I'd love to see this merged.  It adds about 90 seconds to every test run I make and while I'm working on tests, it really adds up.

florean:
Thank you!

PR #5938, opened 2025-07-10
feat(pacer_session): Tweaks SAML auth cookie settings for proxy compatibility

ERosendo:
This PR overrides the `_get_saml_auth_request_parameters` method to ensure that cookies obtained during the SAML authentication workflow are compatible with proxy connections that do not use HTTPS. Specifically, it disables the `secure` flag on cookies so they are not stripped when routed through our proxy.

Fixes #5921 

----

### Additional notes

Once this PR is merged, you can verify that ACMS authentication works as expected by running the following script. It should execute without raising any exceptions:

```python
from django.conf import settings
from cl.lib.pacer_session import ProxyPacerSession

session = ProxyPacerSession(
    username=settings.PACER_USERNAME,
    password=settings.PACER_PASSWORD,
    get_acms_tokens=True,
)
session.login()
```

PR #5960, opened 2025-07-13
feat(recap): Enhances appellate docket purchase to support ACMS cases

### LINKED ISSUE #5154: Review: Implementing logic to purchase ACMS dockets

ERosendo:
Key changes: 

- Introduces a new helper function that centralizes logic to determine whether a given `court_id` requires ACMS-specific handling

- Updates `purchase_appellate_docket_by_docket_number` to detect ACMS cases and switch to `ACMSDocketReport` when appropriate. Falls back to `AppellateDocketReport` for standard PACER courts.

- Tweaks `merge_pacer_docket_into_cl_docket` to save raw ACMS data as `.json` instead of `.html` (**ACMS** does not return a traditional **HTML** response)

- Adds `FakeAcmsCaseSearch`, `FakeAcmsDocketReport`, and related fakes for isolated ACMS test behavior.

Depends on #5938 and https://github.com/freelawproject/juriscraper/pull/1494

Fixes #5154

### DIFF
--- a/cl/recap/tasks.py
+++ b/cl/recap/tasks.py
@@ -2425,7 +2425,7 @@ def purchase_appellate_docket_by_docket_number(
     acms_case_id = None
 
     if should_check_acms_court(court_id):
-        acms_search = AcmsCaseSearch(court_id="ca9", pacer_session=session)
+        acms_search = AcmsCaseSearch(court_id=court_id, pacer_session=session)
         acms_search.query(docket_number)
         acms_case_id = (
             acms_search.data["pcx_caseid"] if acms_search.data else None
--- a/cl/recap/tasks.py
+++ b/cl/recap/tasks.py
@@ -2445,6 +2445,16 @@ def purchase_appellate_docket_by_docket_number(
     docket_data = report.data
     if not docket_data:
         raise ParsingException("No data found in docket report.")
+
+    if acms_case_id:
+        docket_data["docket_entries"] = sorted(
+            docket_data["docket_entries"],
+            key=lambda d: (
+                d["date_filed"],
+                d["document_number"] is None,
+                d["document_number"],
+            ),
+        )
     return create_or_update_docket_data_from_fetch(
         fq, court_id, None, report, docket_data
     )
--- a/cl/recap/tasks.py
+++ b/cl/recap/tasks.py
@@ -2350,7 +2350,7 @@ def create_or_update_docket_data_from_fetch(
     fq: PacerFetchQueue,
     court_id: str,
     pacer_case_id: str | None,
-    report: DocketReport | AppellateDocketReport,
+    report: DocketReport | AppellateDocketReport | ACMSDocketReport,
     docket_data: dict[str, Any],
 ) -> dict[str, str | bool]:
     """Creates or updates docket data in the database from fetched data.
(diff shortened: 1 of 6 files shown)


mlissner:
Not much code to add something so big. Nice to see, thanks.

cl/recap/tasks.py
```diff
@@ -2420,12 +2422,39 @@ def purchase_appellate_docket_by_docket_number(
...
+
+    if acms_case_id:
+        docket_data["docket_entries"] = sorted(
+            docket_data["docket_entries"],
+            key=lambda d: (
+                d["date_filed"],
+                d["document_number"] is None,
+                d["document_number"],
+            ),
+        )
```
albertisfu: I remember we have this sorting logic in a different method. Can we create a method with this logic that can be reused?

cl/recap/tasks.py
```diff
@@ -2420,12 +2422,39 @@ def purchase_appellate_docket_by_docket_number(
     :param fq: The PacerFetchQueue object
     :return: a dict with information about the docket and the new data
     """
-    report = AppellateDocketReport(map_cl_to_pacer_id(court_id), session)
-    report.query(docket_number, **kwargs)
+    acms_case_id = None
+
+    if should_check_acms_court(court_id):
+        acms_search = AcmsCaseSearch(court_id="ca9", pacer_session=session)
```
albertisfu: Should this be?

```suggestion
        acms_search = AcmsCaseSearch(court_id=court_id, pacer_session=session)
```

cl/recap/tasks.py
ERosendo: You're correct. I'll add the new helper method and refactor this code

cl/recap/tasks.py
ERosendo: you're right! thanks for catching this.

cl/recap/tasks.py
albertisfu: btw when testing the fetch attachment page PR I found this docket:
https://www.courtlistener.com/docket/68490335/united-states-of-america-v-raji/

Which I retrieved using the fetch API in my local env.

As you can see the entry 17 is in the wrong place due to date Field is being prioritized over the document number.

Screenshot 2025-07-16 at 2 45 32 p m

I checked in PACER and the date filed is correct.

Would it be possible to prioritize the document number, and use the date filed if it's None?

ERosendo: Since we just introduced a method to sort ACM entries, I think we could check whether all docket entries in the input have a document_number. If they do, we can sort purely by that number. Otherwise, we fall back to the existing sort function. What do you think?

albertisfu: Yes, I think this makes sense. We can leverage the fact that all entries are available when fetching the docket.

ERosendo: cool! i'll refine the new helper method

ERosendo:
@albertisfu Thanks for the review. I’ve addressed all your comments and also added a final refinement to the logic for sorting ACMS entries.

> However, I’m curious where get_acms_tokens is being set to True to trigger ACMS authentication. I couldn’t find it in the code.

Currently, the token is generated only when a user attempts to fetch an ACMS document. I’m planning to open a follow-up PR to store the ACMS credentials in cache so we can reuse them and avoid recomputing the token..

albertisfu (approved):
Thanks @ERosendo this looks ready to be merged!

PR #5961, opened 2025-07-14
Refactor TestCases to mixins

florean:
There are a lot of `TestCase` that basically just setup some useful Factory objects and don't really need to be TestCases.  This can cause problems when you want to use a TransactionTestCase and one of these TestCases and makes the class hierarchy more complicated than it needs to be.  Converting them into mixin classes simplifies things quite a bit.

### DIFF
--- a/cl/tests/mixins.py
+++ b/cl/tests/mixins.py
@@ -38,6 +38,7 @@ from cl.people_db.models import Race
 from cl.search.factories import (
     CitationWithParentsFactory,
     CourtFactory,
+    DocketEntryFactory,
     DocketFactory,
     OpinionClusterFactory,
     OpinionFactory,


florean:
@mlissner, as I mentioned in #5954, I'd like to convert some TestCases into mixins.  Does this approach look good to you?  I branched from my existing test PR, so check out commits [574d24c](https://github.com/freelawproject/courtlistener/pull/5961/commits/574d24c2f2a6d510a787c6ff36715951b181f1aa) and [c22da50](https://github.com/freelawproject/courtlistener/pull/5961/commits/c22da50c9d55805b5e1d7a33b58638171609b737).

(sorry about the force pushes, my alternate test patch fix kept getting in there)

mlissner:
I think the conversion to mixins feels about right. Two thoughts I had:

1. It seems like a lot of these mixins relate directly to models. In that case, it feels like these mixins should be in the app, rather than centralized in a huge file. Maybe we should have something like `cl.audio.test_utils.AudioDataMixin` instead of `cl.test.mixins.AudioMixin`. 

    I'm not sure that's right, but something along those lines feels right to me. Maybe they live in `cl.audio.test_mixins` instead, but that feels overly specific. Or perhaps there's `cl.audio.tests`, inside which has tests.py, mixins.py, and (in some cases) the various other test files.

1. I'm no git pro, but ideally we'll be able to merge this separately from your PR fixing SimpleTestCase and whatnot. That'd be a lot easier to review, rather than putting it all in one PR (as this appears to be).

florean:
> 1. It seems like a lot of these mixins relate directly to models.

Yeah, now that I've got most things converted to mixins, I'm going to start pulling out ones that are more model specific.  If they're only used by one app's tests, I'm just moving them into the tests file like [this](https://github.com/freelawproject/courtlistener/pull/5961/commits/a4aa2f272e985ecc61c933079e7f50bb843dce0c#diff-29d1090279225b6de52404f4142581dad9e4d90ace688e7c9341490f2c377b90L1465).  Although I should move that to the top of the file.  The `PrayAndPayMixin` is next and there's a few others.  I don't think it's worth creating a bunch of single class mixin files.  `RECAPSearchMixin` might be an exception, though, because it is used by multiple files and `search` already has a `test` directory.

 >2. I'm no git pro, but ideally we'll be able to merge this separately from your PR fixing SimpleTestCase and whatnot.

I forked this from #5954 for two reasons: I didn't want to duplicate work from it and this isn't worth looking at until that is merged, anyway: that one actually fixes broken tests, this one is just a refactor to simplify some things.  I just put it up as a draft to be able to point to some specific commits to make sure this was going in the right direction.  I don't want to ambush you with a massive refactor that you don't want.  And I don't expect this to be the final change.  If you decide it's better to spread them across apps, we can do that.  But this is a large step on the path to enabling that.
Honestly, most of the commits are a mess, so I'll probably squash it before review anyway (my personal philosophy is any commit on `main` should be deployable).

mlissner:
This all sounds good. Hold off on going too deep here, I think, until we've caught up with you. That point we'll have a better sense of where things stand? I'm +1 on mixins, but I think the dev team wants a chance to catch up before having opinions.

florean:
I think it is mostly done and I've moved on to some other [things](https://github.com/freelawproject/courtlistener/issues/5969).  I'll probably squash this down to a single commit and force push it, just to make it easier to read while #5954 is still hanging out.

PR #5962, opened 2025-07-14
The alerts profile page doesn't have user context

tiedyedguy:
User Alerts was using SimpleTemplateResponse which did not pass over request context and such the user context was not passed.

Since it was not passed the "Profile" link in the upper right was showing "Sign in / Register" and that bugged me.

This removes the usage of SimpleTemplateResponse from profile alerts view and the reference to SimpleTemplateReponse because that was the only usage of it.

### DIFF
--- a/cl/citations/group_parentheticals.py
+++ b/cl/citations/group_parentheticals.py
@@ -110,6 +110,8 @@ def compute_parenthetical_groups(
     )
 
     parenthetical_groups: list[ComputedParentheticalGroup] = []
+    # start the visited_nodes set here to prevent recomputing components, since
+    # the same component will be found when starting from any of it's nodes
     visited_nodes: set[str] = set()
     for node, neighbors in similarity_graph.items():
         if component := get_graph_component(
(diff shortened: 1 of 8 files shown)


mlissner:
I see what you mean: 

image

tiedyedguy:
> I see what you mean:
> image

Right? That bugged me and I had to figure out why. Took a while to figure out because I don't know django :).

mlissner:
I just kicked off your tests again. Looks like docker hub might have briefly failed? We'll see in a sec. 

Thanks for tracking this down!

tiedyedguy:
I realized a mistake. I didn't update it to a AuthenticatedHttpRequest type of request. I will fix later and resubmit.

tiedyedguy:
I don't think this mattered, but it was using HttpRequest type instead of AuthenticatedHttpRequest type to make the request variable in the alerts section. 

I do not think it really matter because line 130 showed `docket_alerts = request.user.docket_alerts.filter` which was working so the request variable had the user information in it already.

So this just make it look more consistent.

mlissner:
Elisa, do you mind taking a look at this one? Might be a welcome break from your usual? :)

mlissner:
Thanks all. Merging!

PR #5985, opened 2025-07-16
feat(cl): make --keepdb mean something

### LINKED ISSUE #5969: Using --keepdb causes test failures
Using the `--keepdb` options for testing, which saves about a minute of applying migrations, causes around a hundred test failures.  It appears that some of the data created by migrations used by tests, namely users and courts, is getting deleted from the test database.  If you also use `--parallel`, Django runs all tests against clones of the test database (`test_courtlistener_1`, etc.).  In that case, the test database isn't changed and deleting those clones fixes the problem, but it isn't automatic, so you have to manually do that between test runs.  So some test is deleting that data outside of a transaction.

One solution might be to convert tests to only use data they explicitly create, rather than relying on implicitly created objects from migrations, but it would probably be worth tracking down and eliminating the root cause.

Example runs using the command `docker exec -e SELENIUM_DEBUG=1 -e SELENIUM_TIMEOUT=30 cl-django ./manage.py test cl --verbosity=3 --exclude-tag selenium --parallel=1 --timing --durations 10 --keepdb`

initial run:
```
Ran 1510 tests in 261.842s

OK (skipped=53)
Preserving test database for alias 'default' ('test_courtlistener_1')...
Preserving test database for alias 'default' ('test_courtlistener_2')...
Preserving test database for alias 'default' ('test_courtlistener')...
Total database setup took 62.094s
  Creating 'default' took 61.754s
  Cloning 'default' took 0.181s
  Cloning 'default' took 0.159s
Total database teardown took 0.000s
Total run took 327.935s
```

second run:
```
Ran 1421 tests in 142.359s

FAILED (failures=5, errors=186, skipped=53)
Preserving test database for alias 'default' ('test_courtlistener_1')...
Preserving test database for alias 'default' ('test_courtlistener_2')...
Preserving test database for alias 'default' ('test_courtlistener')...
Total database setup took 0.693s
  Creating 'default' took 0.672s
  Cloning 'default' took 0.010s
  Cloning 'default' took 0.011s
Total database teardown took 0.000s
Total run took 145.636s
```

Some snipped example errors look like:
```
ERROR: test_make_a_docket_alert (cl.alerts.tests.tests.DocketAlertAPITests.test_make_a_docket_alert)
Can we make a docket alert?
  File "/opt/venv/lib/python3.13/site-packages/django/test/testcases.py", line 1400, in setUpClass
    cls.setUpTestData()
    ^^^^^^^^^^^
  File "/opt/courtlistener/cl/alerts/tests/tests.py", line 2600, in setUpTestData
    cls.court = Court.objects.get(id="scotus")
    ^^^^^^^^^^^^^^^
cl.search.models.Court.DoesNotExist: Court matching query does not exist.
```

```
ERROR: test_nda_get_document_number_fallback (cl.recap.tests.test_recap_email.GetDocumentNumberForAppellateDocuments.test_nda_get_document_number_fallback)
This test verifies if we can get the PACER document number for
  File "/usr/local/lib/python3.13/unittest/case.py", line 603, in _callSetUp
    self.setUp()
    ~~~~~~~~~~^^
  File "/opt/courtlistener/cl/recap/tests/test_recap_email.py", line 2906, in setUp
    self.user = User.objects.get(username="recap-email")
    ^^^^^^^^^^^^^^^
django.contrib.auth.models.User.DoesNotExist: User matching query does not exist.
```

florean:
This modifies our custom test runner so that tests will never permanently modify the test database.  This means --keepdb will work even if we use a TransactionTestCase.

To accomplish this, we continue to let the test runner setup the test database, running migrations and whatever else it might decide to do in the future.  But then we create the clones ourselves and erase any trace of the original test database.

This will always create an extra cone when --parallel is greater than one, but trying to fix that behavior is messy and I'm fine with an extra 200ms in that case.  

Resolves #5969.

### DIFF
--- a/cl/tests/runner.py
+++ b/cl/tests/runner.py
@@ -3,6 +3,7 @@ import sys
 import warnings
 from unittest import TestLoader
 
+from django.conf import settings
 from django.test.runner import DiscoverRunner
 from override_storage import override_storage
 
@@ -69,14 +70,59 @@ class TestRunner(DiscoverRunner):
         # See PR #5888 for more details.
         # parser.set_defaults(buffer=True)
 
-    def setup_databases(self, **kwargs):
+    def setup_databases(self, *args, **kwargs):
         # Force to always delete the database if it exists
         interactive = self.interactive
         self.interactive = False
+
         try:
-            return super().setup_databases(**kwargs)
+            if self.keepdb:
+                # --keepdb doesn't really work. See Django bug #25251:
+                # https://code.djangoproject.com/ticket/25251. In addition to
+                # TransactionTestCases, it appears TestCases can also delete
+                # migration data. To resolve that, this modifies the test
+                # runner to never run against the actual test database if we're
+                # trying to keep it. Instead, always run against a clone that
+                # gets destroyed at the end of testing.
+
+                # Because we don't want to run through migrations twice,
+                # we use the normal process to create the actual test database, then
+                # manually clone it and the parallel clones.
+                parallel = self.parallel
+                self.parallel = 1
+                old_config = super().setup_databases(*args, **kwargs)
+                self.parallel = parallel
+                # Create the clones ourselves
+                for c in (x for x, y, z in old_config if z):
+                    # Create test_database_clone and replace the real test database.
+                    db_clone_name = c.settings_dict["NAME"] + "_clone"
+                    with self.time_keeper.timed(f"  Cloning '{c.alias}'"):
+                        c.creation.clone_test_db(
+                            "clone", verbosity=self.verbosity
+                        )
+                    settings.DATABASES[c.alias]["NAME"] = db_clone_name
+                    c.settings_dict["NAME"] = db_clone_name
+                    # Create parallel clones.
+                    if parallel > 1:
+                        for index in range(1, self.parallel + 1):
+                            with self.time_keeper.timed(
+                                f"  Cloning '{c.alias}'"
+                            ):
+                                c.creation.clone_test_db(
+                                    str(index), verbosity=self.verbosity
+                                )
+                return old_config
+            else:
+                return super().setup_databases(*args, **kwargs)
         finally:
             self.interactive = interactive
+
+    def teardown_databases(self, *args, **kwargs):
+        keepdb = self.keepdb
+        # Always delete the cloned DBs.
+        self.keepdb = False
+        super().teardown_databases(*args, **kwargs)
+        self.keebdb = keepdb
 
     @override_storage()
     def run_tests(self, *args, **kwargs):
--- a/cl/tests/runner.py
+++ b/cl/tests/runner.py
@@ -45,6 +45,8 @@ class TestRunner(DiscoverRunner):
     def __init__(self, *args, enable_logging, **kwargs):
         super().__init__(*args, **kwargs)
         self.enable_logging = enable_logging
+        self.use_clones = kwargs["use_clones"]
+        self.keepdb = self.use_clones or self.keepdb
 
     @classmethod
     def add_arguments(cls, parser):
@@ -55,6 +57,14 @@ class TestRunner(DiscoverRunner):
             default=False,
             help="Display all log lines",
         )
+
+        parser.add_argument(
+            "--use-clones",
+            action="store_true",
+            default=False,
+            help="Runs all tests against fresh clones of the test DB.",
+        )
+
         super().add_arguments(parser)
 
         # Modify parallel option to default to number of CPU cores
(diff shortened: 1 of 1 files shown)


mlissner:
This feels a bit janky and like a pretty big workaround vs. making our tests use factories instead of the migrated fixtures. 

I feel like the real solution is to go fix that (like it seemed like you were working on?), rather than have a pretty hefty workaround we are stuck with for years. How bad is fixing the tests to use factories instead?

cl/tests/runner.py
```diff
@@ -69,15 +70,60 @@ def add_arguments(cls, parser):
...
+                        for index in range(1, self.parallel + 1):
+                            with self.time_keeper.timed(
+                                f"  Cloning '{c.alias}'"
+                            ):
+                                c.creation.clone_test_db(
+                                    str(index), verbosity=self.verbosity
+                                )
+                return old_config
+            else:
+                return super().setup_databases(*args, **kwargs)
```
florean: Here's a more minimal version of this fix, but it runs migrations twice.  
```suggestion
            if self.keepdb:
                parallel = self.parallel
                self.parallel = 1
                old_config = super().setup_databases(*args, **kwargs)
                self.parallel = parallel
                for c in (x for x, y, z in old_config if z):
                    c.creation.clone_test_db("clone", verbosity=self.verbosity)
                    settings.DATABASES[c.alias]["TEST"]["NAME"] = c.settings_dict["NAME"] + "_clone"
                    c.settings_dict["TEST"]["NAME"] = c.settings_dict["NAME"] + "_clone"
            return super().setup_databases(*args, **kwargs)
```
If Django were to implement something like this, the worst case is you are generating an unnecessary clone until you remove the code.

florean:
>This feels a bit janky and like a pretty big workaround vs. making our tests use factories instead of the migrated fixtures.

Janky seems a bit harsh.  It is very minimal and it only uses APIs and conventions that have been stable for over a decade.  It doesn't do anything hacky or use internal APIs.  I would even argue that it is an elegant solution.  But we can agree to disagree.  I made a comment with an even more minimal version .  I would just ask that you try it out the next time you're having to do multiple test runs:  `git cherry-pick a9527bd0`

This is orthogonal to moving tests to factories.  It does alleviate a symptom by ensuring any clobbering won't impact subsequent runs, but it won't fix any test failures or fragility.


>I feel like the real solution is to go fix that (like it seemed like you were working on?), rather than have a pretty hefty workaround we are stuck with for years. How bad is fixing the tests to use factories instead?

I think getting everything off the default users is somewhat painful.  I did some work on that and moving off the default courts, but I'm either basing it off my existing branches or I'm doing a bunch of duplicate work and it's getting a bit confusing.  So I'm holding off on new fixes while I wait for my open ones to get resolved.  No point adding to the backlog.

mlissner:
I don't mean to impugn your code. It's clever. I think the problem is that it changes people's understanding of the way the system works. It's no longer that you have parallel DB's running tests, now it's parallel DBs spawning more DB's that run tests and get torn down. Feels like a big layer of abstraction on top of the existing system that would eventually bite us somehow. 

The current situation with tests isn't great (as you continue demonstrating!), but I think this layer feels like a last resort kind of thing, not something we'd want to reach for at this time. I think.

florean:
> It's no longer that you have parallel DB's running tests, now it's parallel DBs spawning more DB's that run tests and get torn down.

Sorry, just have to make a correction here.  It isn't parallel DBs spawning more DBs.  In the non-parallel case, it is one copy, the minimum possible.  In the parallel case, it is n + 1, because there is no clean way to destroy clones without also destroying the database (scratch that, I just thought of one).  My biggest problem with it is that it breaks the semantics of `--keepdb`, because while it does keep the test db, it deletes the databases that were tested against, which complicates forensic analysis of test failures.

Let me explain a bit about why I was motivated to do this, because I haven't really done that.  The `--keepdb` option saves me a minute of time per test run, which is pretty substantial when I am trying to iterate on tests.  However, there are a large number of tests that break the test DB and those are some of the ones I'm working on.  I need to use both single and parallel options and `--keepdb` has a different pathology for each path.

So here is my workflow: I make a copy of the database called `test_courtlistener_main`.  Whenever a solo run finishes, I switch to a PSQL console, drop `test_courtlistener`, and copy over the backup.  If I'm running parallel, the original database is fine, I just need to drop all the clones so they get recreated.  So I copy and paste my command to drop all clone databases over to PSQL and then I'm ready to run again.  It isn't a long process, but when tests run in less than a second, it's still a large multiple of the actual task and it's a friction point for me.  In addition, it is cumbersome and I'm often unsure of whether I copied/dropped the database or whether I did the correct one.  And there is no way to tell just from just looking at the results.

I should probably just wrap all of that into a script, but I like to solve root causes, especially if I think other people will find it useful.  For my [Django PR](https://github.com/django/django/pull/19648), I created a new flag, as there is no way to accomplish the correct behavior using the existing flags without breaking semantics or functionality.  What if I just moved this under the same flag, `--use-clones`?  Then it has no impact on any existing workflows and it can be ignored, but it's there is someone needs it.

If you really don't think it will be useful to anyone else working on this repo, then go ahead and close the PR and I'll figure out how to get Git to ignore it locally.

florean:
Okay, that's much cleaner and basically replicates my Django version.  See what 24 hours of learning can do?  :)
It no longer removes DB clones after testing, just automatically replaces them the next run.  And it no longer creates an extra copy for parallel runs. It uses the exact same names as `--parallel`. Plus it eliminates the `teardown_databases` override.

Maybe one of the confusing points is the shorthand [old_config loop](https://github.com/freelawproject/courtlistener/pull/5985/files#diff-774b9f9ac112c5b194c44da7ec94de0f40929bd48980a53c410d12f9be2718d1R105)?  I can give the variables longer names, but it's not super important.  It's tuples of a database connection (`x`), the original database name (`y`), and whether it is an actual connection or a mirror (`z`, False if mirror).

I think that is the best version of this.

mlissner:
Definitely an interesting solution, but the thing that irks me about it (and that I suspect the Django devs might not like either) is that it is mostly a workaround instead of a fix. The correct thing for Django to do is fix that bug, not create clones to minimize its impact, right? We might be able to convince them that after ten years that bug isn't going anywhere, so we should just do a fix instead, but that'll take some consensus building, I fear.

A different question, I think, is what we should do on our side, given that the bug exists and is unlikely to be fixed anytime soon. Some options include:

 - Stop doing data migrations before tests
 - Fix our tests that wipe out data
 - Patch our test runner to work around it (this PR)j

I think my preference is to stop doing data migrations rather than patching our test runner, but maybe it's helpful to patch the runner temporarily if it helps get rid of the data migrations? I just don't love the idea of having an unusual test runner long term, though it is mitigated by putting it behind a flag that we can document in our dev guide as a special testing tool. Hm.

florean:
> The correct thing for Django to do is fix that bug

My PR does fix that bug.  The problem is that it isn't really a bug, it's that there are two valid and useful interpretations of "keep" and Django only implements one (unfortunately, the less useful one).  The definition `keepdb` is using is "to stay or continue in".  It *keeps* the database ***where*** it is.  But there's also "preserve or maintain".  My option *keeps* the database ***what*** it is.  If you have large test databases (more common with something like GeoDjango) and a very stable test suite, then you're willing to trade some reliability for speed.  In almost every other situation, you should prioritize consistency and reliability over speed (and again, we're talking a couple hundred milliseconds for Postgres and a couple for SQLite).

> Not create clones to minimize its impact

I'm guessing you mean additional clones, because `--parallel` is already creating clones and not touching the original test database.  My only change is to simply not reuse them.  It's the exact same code path, I literally just flip a flag in one place and not the other (in Django; in the test runner it's a copy of the code).  I am creating a novel clone when running sequentially, but you literally have to create clones if you care about ***what*** the database is rather than ***where*** it is while allowing writes.  There's just no way around it.

> I think my preference is to stop doing data migrations rather than patching our test runner

Although it does ease the immediate pain, this will still be useful to me even after the tests are fixed.  It's about having a consistent and reliable test environment.  Building from scratch takes way too long and `--keepdb` has bit me too many times over the now decades to trust it.  I just didn't realize how painless it would be to fix it.  I will gladly pay a few hundred milliseconds to know that my test databases are in a known-good state.

It's disappointing, but this isn't code that really impacts CourtListener.  Your CI/CD system will always build from scratch.  It clutters up your custom test runner and it adds complexity that you have to maintain.  I'll just close this; it's a distraction and I have blocking PRs I'd rather have reviewed addressing your other bullet points.

mlissner:
Thanks for the explanation and understanding. I'll flag this for others on our team to see where they land, and I'm open to reconsidering it down the road too, if you keep liking it and we keep needing it. I'm curious to see where your Django PR goes too.

PR #5989, opened 2025-07-16
fix(citations): Update match citations

flooie:
Fixes bug allows self citations for single matches

### DIFF
--- a/cl/citations/match_citations.py
+++ b/cl/citations/match_citations.py
@@ -113,9 +113,12 @@ def resolve_fullcase_citation(
 
         # If there is one search result, try to return it
         if len(db_search_results) == 1:
-            result_id = db_search_results[0]["id"]
+            match = db_search_results[0]
+            if match["cluster_id"] == full_citation.citing_opinion.cluster.pk:
+                # Self Citation
+                return NO_MATCH_RESOURCE
             try:
-                return Opinion.objects.get(pk=result_id)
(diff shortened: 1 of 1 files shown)


cl/citations/match_citations.py
```diff
@@ -113,9 +113,12 @@ def resolve_fullcase_citation(
 
         # If there is one search result, try to return it
         if len(db_search_results) == 1:
-            result_id = db_search_results[0]["id"]
+            match = db_search_results[0]
+            if match["cluster_id"] == full_citation.citing_opinion.cluster.pk:
```
grossir, line 117: Tests are failing on [this](https://github.com/freelawproject/courtlistener/actions/runs/16327754414/job/46125036925?pr=5989) `full_citation.citing_opinion.cluster` 

```
  File "/opt/courtlistener/cl/citations/match_citations.py", line 117, in resolve_fullcase_citation
    if match["cluster_id"] == full_citation.citing_opinion.cluster.pk:
    ^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'cluster'
```

I think this happens because the functions is called in the recap_citations workflow too, so it may not have the "citing_opinion" attribute; instead being called "citing_document"

grossir, line 117: Shouldn't the check be on the match having the same opinion id? I thinks opinions in the same cluster can cite each other? And it would be correct unless they cite themselves?

PR #5996, opened 2025-07-17
feat(search.models): add ClusterRedirection model

### LINKED ISSUE #5983: Redirect dead opinion links to proper pages
A Courlistener opinion link is defined by the cluster id. Example:

https://www.courtlistener.com/opinion/10633255/kenneth-wright-v-state-of-florida/?type=o&type=o&q=&order_by=score+desc

If we delete the cluster `10633255`, the link will be dead, leading to a 404 page ([example](https://www.courtlistener.com/opinion/9999999999/ss/)). This may result in poor user experience, if the user copied the link and saved it somewhere (which is the expected behavior).

A solution is to create a redirection table, from a deleted cluster to an active cluster. Deletions may happen due to:

| Deletion reason  | Expected behavior |
| Delete clusters from same hash duplicates  | Redirect to the proper cluster object |
| Delete clusters due to opinion [versioning](https://github.com/freelawproject/courtlistener/issues/3803) | Redirect to the main version cluster |
| Delete clusters for consolidating spread opinion types spread through many clusters (happens in scrapers)  | Redirect to the proper cluster object |
| Manual sealing of opinions | Redirect to a "this record is sealed" page |

For the first 3, we may use a [permanent redirect status code](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/308)

--------------------

A draft model:

```python
class ClusterRedirection(models.Model):
    VERSIONING = 1
    DUPLICATE = 2
    CONSOLIDATION = 3
    SEALED = 4
    REDIRECTION_REASON = (
        (VERSIONING, "Opinion versions clusters were consolidated in a single one"),
        (DUPLICATE, "Clusters were duplicated"),
        (CONSOLIDATION, "Opinion types clusters were consolidated in a single one"),
        (SEALED, "Cluster was deleted due to sealing"),

    )    
    deleted_cluster_id = models.IntegerField()
    cluster = models.ForeignKey(
        OpinionCluster,
        help_text="The existing cluster the deleted clusters now point to",
        related_name="merged_clusters",
        on_delete=models.DO_NOTHING,
    )
    reason = models.SmallIntegerField(
        help_text="The reason why the old cluster was deleted", choices=REDIRECTION_REASON
    )        
```

grossir:
Fixes #5983

- Creates a model to store deleted cluster ids and point them to existing cluster ids; useful to handle versioning, consolidation and deletion of duplicates without affecting users that may have saved links
- Adds a decorator to handle opinion_page views that look for an OpinionCluster
- Overrides OpinionClusterViewSet.retrieve to handle API redirection
- Adds tests for both cases


cl/search/models.py
```diff
@@ -3786,3 +3786,41 @@ class Meta:
...
+    REDIRECTION_REASON = (
+        (
+            VERSIONING,
+            "Opinion versions clusters were consolidated in a single one",
+        ),
+        (DUPLICATE, "Clusters were duplicated"),
+        (
+            CONSOLIDATION,
+            "Opinion types clusters were consolidated in a single one",
+        ),
```
flooie: does this mean merged.  and how is it different then versioning?

grossir: these 2 clusters should be a single cluster. one of the clusters will be deleted due to "consolidation", and should redirect to the one we keep. It's not versioning, since they are actually different opinions
- https://www.courtlistener.com/opinion/10618524/ken-paxton-in-his-official-capacity-as-attorney-general-of-texas-and-greg/?q=court_id%3Atex&type=o&order_by=dateFiled+desc&stat_Published=on&stat_Unpublished=on&stat_Unknown=on
- https://www.courtlistener.com/opinion/10618523/ken-paxton-in-his-official-capacity-as-attorney-general-of-texas-and-greg/?q=court_id%3Atex&type=o&order_by=dateFiled+desc&stat_Published=on&stat_Unpublished=on&stat_Unknown=on

this happens in scrapers, mostly

flooie: gotcha.  thanks.

flooie:
I haven’t done a full review yet, but I do have some thoughts on the language around the new table.

My suggestion would be to use these four reasons and explanations:

VERSION   = 1
DUPLICATE = 2
MERGED    = 3
SEALED    = 4

REDIRECTION_REASON = (
    (VERSION,   "Cluster replaced by a newer version of the same opinion"),
    (DUPLICATE, "Cluster removed as a duplicate"),
    (MERGED,    "Cluster merged into another cluster (e.g., a dissent added to a cluster containing the majority opinion)"),
    (SEALED,    "Cluster removed by court order"),
)

Nothing major just makes it a bit more descriptive to me.  I was lost earlier about reasons.  


I’m a bit torn on the table name. `ClusterRedirection` feels slightly off, since in some cases, it’s not really a redirection (ie sealed). That said, I also see the logic behind the name.

I asked ChatGPT for some alternatives, and it suggested:
	•	ClusterRedirect
	•	OpinionClusterArchive
	•	RetiredCluster
	•	ClusterForwarding

Not sure any are clearly better, but I think I lean toward OpinionClusterArchive cc @mlissner

I think we need to add a constraint to prevent multiple redirections from pointing to and from the same places.  I was able to create duplicates, and it throws errors.

Also should we include a date_created

cl/opinion_page/decorators.py
```diff
@@ -0,0 +1,41 @@
...
+                redirection = await ClusterRedirection.objects.aget(
+                    deleted_cluster_id=kwargs["pk"]
+                )
+                cluster_id = redirection.cluster_id
+
+                if not cluster_id:
+                    # TODO: should redirect to sealed page
+                    raise exc
+            except ClusterRedirection.DoesNotExist:
+                raise exc
```
flooie: sorry, here 

                if redirection.reason == ClusterRedirection.SEALED:
                    html = await sync_to_async(render_to_string)(
                        "410.html", request=request
                    )
                    return HttpResponse(html, status=410)

I think something like this could handle the sealed example

cl/search/api_views.py
```diff
@@ -239,6 +243,36 @@ class OpinionClusterViewSet(
...
+            return Response(serializer.data)
+        except Http404 as exc:
+            try:
+                redirection = ClusterRedirection.objects.get(
+                    deleted_cluster_id=pk
+                )
+                cluster_id = redirection.cluster_id
+
+                if not cluster_id:
+                    # TODO return sealed page
```
flooie: I think I would use 410 Gone.  for sealed opinions

flooie (changes requested):
I think it's close. But I think we need to improve some language around the new table and we should implement the 410 for sealed records instead of leaving it for another PR.  I think something could be added to the templates quickly enough. Otherwise it works great.  

Aside from changing the language a bit, do you think we should include date_created and a unique constraint on the table.

flooie:
What do you think 
Screenshot 2025-07-23 at 12 50 41 PM

mlissner:
That's fun, Bill, but I think the `410 U.S.` joke might not work since people don't know the 410 status code like they do the 404 status code — alas! Also, it's a more serious topic, so probably best not to make jokes. The rest looks great to me though. I assume you'll return a 410 HTTP code with it?

flooie:
@mlissner I would return a 410 yes.

flooie:
@mlissner do you have thoughts on the table name or fields?

grossir:
@flooie

Just waiting for final input on the model name, to avoid doing too many migrations variations

About the `REDIRECTION_REASON` values
I still think `MERGED = 3` is too generic, and that it should be `CONSOLIDATED = 3`

Also, can you send the exact texts for the `"410template.html"`, or even paste the HTML template you want to use. I wanted to leave it for future work because I expect the design and copy decisions to take more time and didn't want to delay this PR

flooie:
I think the html page should just be `410.html` similar to other files.  I think.  

Screenshot 2025-07-23 at 4 02 27 PM

cl/api/tests.py
```diff
@@ -22,6 +22,7 @@
 from django.test.utils import CaptureQueriesContext
 from django.urls import reverse
 from django.utils.timezone import now
+from rest_framework import status
```
mlissner: This used to be needed back in the day, but Python packaged these into the http lib (I think), so it's probably better to use that since it's more official. At some point I think I swapped all of our existing imports to use it, IIRC.

cl/search/models.py
```diff
@@ -3786,3 +3786,41 @@ class Meta:
...
+    SEALED = 4
+    REDIRECTION_REASON = (
+        (
+            VERSIONING,
+            "Opinion versions clusters were consolidated in a single one",
+        ),
+        (DUPLICATE, "Clusters were duplicated"),
+        (
+            CONSOLIDATION,
+            "Opinion types clusters were consolidated in a single one",
```
mlissner: Possibly?

```suggestion
            "Opinion clusters were consolidated in a single one",
```

cl/search/models.py
```diff
@@ -3786,3 +3786,41 @@ class Meta:
...
+        ),
+        (SEALED, "Cluster was deleted due to sealing"),
+    )
+
+    deleted_cluster_id = models.IntegerField()
```
mlissner: This needs an index, right?

mlissner (approved):
Model looks OK to me, but I only really looked at the field and table names. I think it should be OK though, if you guys have thought through the various model field options.

cl/search/models.py
grossir: I think setting the `unique` flag will create a unique index for the field
```python
deleted_cluster_id = models.IntegerField(unique=True)
```

From [Postgres](https://www.postgresql.org/docs/current/indexes-unique.html) docs
> PostgreSQL automatically creates a unique index when a unique constraint or primary key is defined for a table. The index covers the columns that make up the primary key or unique constraint (a multicolumn index, if appropriate), and is the mechanism that enforces the constraint.
> There's no need to manually create indexes on unique columns; doing so would just duplicate the automatically-created index.


And our current DB itself, for the `Court` model, without a `Meta.indexes` attribute
```python
    position = models.FloatField(
        help_text=(
            "A dewey-decimal-style numeral indicating a hierarchical "
            "ordering of jurisdictions"
        ),
        db_index=True,
        unique=True,
    )
```

On the DB shows as
```psql
\d search_court
Indexes:
    "search_court_position_key" UNIQUE CONSTRAINT, btree ("position")
```

mlissner: Huh! That's news to me. Cool, thanks.

cl/search/api_views.py
flooie: is this todo complete?  cant we return a sealed message

grossir: sorry forgot this one

flooie: great. thanks

flooie:
I added two labels.  as this was created after the new template.  @grossir did I do it right?   @chaco-fl

flooie (approved):
This looks good to me.

PR #5997, opened 2025-07-17
Fix slugify reporter variations

quevon24:
Instead of solving this in get_canonicals_from_reporter(), I fixed it directly in slugify_reporter().

Why? Because slugify_reporter() is the first and central place where reporter input is normalized. By adding a final fallback check for known variations (using `VARIATIONS_ONLY`), we can cleanly disambiguate most cases before they hit more complex logic later.

**What works now**
Here are cases that now redirect correctly from the variation to the intended slug:

Input Variant | Redirected Slug
-- | --
Vr. | vroom
V.R. | vt
Black Rep. | black
Black. Rep. | blackf
Cal. App. 2d Supp | cal-app-2d
Cal. App. 2d Supp. | cal-app-supp-2d
CLR | conn-l-rptr
Cl.R. | cl-ch
Dec. Commr. Pat. | dec-com-pat
Dec. Comm'r Pat. | dec-commr-pat
Hayw. & H. | hayw-hdc
Hayw.& H. | hay-haz
Johns.(N.Y.) | johns-ch
Johns.N.Y. | johns
Mt. | mont
mt | mt
Pa.C. | pa-commw
Pac. | p
Sc. | scam

**What still can't be disambiguated**
Some abbreviations are too ambiguous to resolve safely. For example:

"B.R." is used both as a standalone reporter for Bankruptcy Reporter and as a variation for Baltimore City Reports ("Balt. C. Rep."). When we receive "B.R.", it exists in REPORTERS, so the function assumes it's the Bankruptcy Reporter and returns its slug "br". But this could be incorrect, the user might've meant "Balt. C. Rep." (slug "balt-c-rep"), which also uses "B.R." as a variation.

Because we can't safely guess the intent, we continue suggesting an alternative reporter:

image

Or continue showing a page with HTTP 300 like this if disambiguation is not possible:

image

Here's the list of cases that cannot be automatically disambiguated:

Input Variant | Canonical(s) | Status | Why?
-- | -- | -- | --
B.R. | br, balt-c-rep | 300 | "B.R." is both a reporter and a variation of another
BR | br, balt-c-rep | 300 | Same as above, unpunctuated version
Wash. | wash, wash-terr | 300 | Could refer to Washington Reports or Washington Territory Reports
WASH | wash, wash-terr | 300 | Uppercase version, same ambiguity
HOW | how, howard | 300 | Could be Howard's Supreme Court Reports Howard's Reports or Mississippi Reports, Howard.
How. | how, howard | 300 | Same as above
OKla. | okla, okla-crim | 300 | Could be Oklahoma Criminal Reports or Oklahoma Reports
Okla. | okla, okla-crim | 300 | Same as above
S.C. | s-ct, scam | 300 | Could be West's Supreme Court Reporter or South Carolina Reports

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,0 +1,62 @@
+name: Bug Report
+description: File a bug report
+body:
+  - type: markdown
+    attributes:
+      value: |
+        <p align="center"><em>Thank you for filing a bug report!</em></p>
+
+        > [!IMPORTANT]
+        > - Please provide descriptive titles. Good titles help us address your report faster.
+        > - Please search existing issues to avoid creating duplicates.
+        > - For enhancement requests, please use GitHub Discussions.
+        > - Please fill out the template below to the best of your ability.
+        > - Please describe the issue as much as possible and provide steps to reproduce it screenshots if possible.
+  - type: textarea
+    id: what-happened
+    attributes:
+      label: What happened?
+      description: Describe the bug clearly and concisely.
+      placeholder: Tell us what you see!
+    validations:
+      required: true
+  - type: textarea
+    id: expected-behavior
+    attributes:
+      label: Expected behavior
+      description: Please write a description of what was supposed to happen.
+    validations:
+      required: false
+  - type: textarea
+    id: actual-behavior
+    attributes:
+      label: Actual behavior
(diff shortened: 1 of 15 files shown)


flooie (approved):
Makes sense to do It here. 
but I wonder if we need to do some cleanup for some of those courts-db slugs

PR #6005, opened 2025-07-18
fix(cl): fixed unrandom Court names

florean:
> I can't figure out why this test would fail [like](https://github.com/freelawproject/courtlistener/actions/runs/16156761254/job/45600782427?pr=5930#step:17:3758) [this](https://github.com/freelawproject/courtlistener/actions/runs/16156761254/job/45600782427?pr=5930#step:17:4620).

Good news, everybody!  Thanks to that [PR](https://github.com/freelawproject/courtlistener/pull/5933) that added more detailed debugging information leading to this very readable [failure](https://github.com/freelawproject/courtlistener/actions/runs/16360675735/job/46227845669#step:17:4630) on one of my other [PRs](https://github.com/freelawproject/courtlistener/pull/5867), I know exactly why this test is failing: 6.2% of the time, roughly 1 in 16, at least two of the three courts are getting the exact same short_name (and name), which this test sorts on.  That means it is solely dependent on how Postgres wants to order equivalent items; let's call that fifty-fifty.

The names were cute, but not worth the test failures.  I did salvage the Eruptanyom name, since it seemed to have some sentimental value.

Sorry, this gets the fire emoji.  It's literally a one minute review and it has a meaningful impact on tests.  In fact, 93.8% of the review is just reading this PR description.

### DIFF
--- a/cl/tests/providers.py
+++ b/cl/tests/providers.py
@@ -50,7 +50,7 @@ class LegalProvider(BaseProvider):
                 "Eruptanyom",  # Kelvin's pretend world
             ]
         )
-        last_word = f"{last_word}-{str(uuid4)[:8]}"
+        last_word = f"{last_word}-{str(uuid4())[:8]}"
 
         return " ".join([first_word, mid_word, last_word])
 
--- a/cl/tests/providers.py
+++ b/cl/tests/providers.py
@@ -1,4 +1,5 @@
 import random
+from uuid import uuid4
 
 from faker import Faker
 from faker.providers import BaseProvider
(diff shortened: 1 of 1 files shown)


mlissner:
Nice find. I'm sad to see somewhat human-ish court names removed from the factory though, since that data is used in the front end by developers. It feels sort of useful to me to have those values seem at least somewhat realistic so that there's a certain UX experience the devs have. 

Any chance you looked into Faker or FactoryBoy to see if they can enforce uniqueness in a provider like this one? Seems doable if we add another 10-20 first and last words?

florean:
> Any chance you looked into Faker or FactoryBoy to see if they can enforce uniqueness in a provider like this one? Seems doable if we add another 10-20 first and last words?

I did not.  Even if we take the high end of your numbers, it's still a .25% chance of collisions.  That's too high.  And the less frequent they are, the harder they are to track down.  Random test failures suck up developer time. 

From a triage point of view, I'd like to see this merged to stop the test fails and then we can figure out a better solution.  I'm fine truncating the UUIDs to a friendlier length for now if that helps.

But maybe automated tests and UX need different factories?  Those are serving very different purposes.

florean:
I cleaned it up a little and added some comments.  It should be good, any reason not to merge? I think this should be considered a testing hot fix given its size and that the bug is causing real PR checks to fail.

Thanks!

Issue #6009, opened 2025-07-18
Update CourtListener deploy script to use many celery deployments


mlissner:
In freelawproject/infrastructure#410, I'm splitting Celery across many deployments. We need to update our deploy script accordingly or else who knows what will happen.

PR #6010, opened 2025-07-18
Update deployment to do new celery deployments

mlissner:
## Fixes

Fixes: #6009 

Context: freelawproject/infrastructure#410

## Summary

This does three things in two commits. In the first commit:

1. It refactors our deployment script so that it's more modular, has better parallelism, and uses loops rather than repeated code.

2. It adds another deployment that can be skipped with the `skip-daemon-deploy` label, which I've created.

In the second commit: 

3. It updates the deployment script to deploy our new celery deployments created in https://github.com/freelawproject/kubernetes/pull/34. This is in the second commit.

## Deployment

This PR should not skip any steps so that we know it works properly.

It should not be deployed until freelawproject/infrastructure#410 is deployed.


chaco-fl (approved):
LGTM

albertisfu (approved):
Great. Looks good to me!

PR #6012, opened 2025-07-18
refactor(pacer): rewrite do_pacer_fetch using match-case syntax

### LINKED ISSUE #6008: Migrate fetch requests off main celery queue
We had an issue yesterday that our PACER fetches are all on the default celery queue, which is clogged up. This is bad. 

In freelawproject/infrastructure#410 we're creating a bunch of new deployments and queues. I am going to create a deployment specifically for pacer fetches and give it the queue name of `recap-fetch`

This issue is to update the `do_pacer_fetch` function to use the new queue.

ERosendo:
## Summary

This PR updates the `do_pacer_fetch` function to route PACER fetch tasks to a dedicated Celery queue, with the queue name now configurable via the new `CELERY_PACER_FETCH_QUEUE` setting. Additionally, it refactors the helper function to use Python’s `match-case` statement for clearer branching logic. 

## Fixes

Fixes #6008 

## Deployment

<details>
<summary>Instructions</summary>

---

The following labels control the deployment of this PR if they’re applied. Please choose which should be applied, then apply them to this PR:

| Label                 | Description                             | Use case                                                                                                                         |
| --------------------- | --------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- |
| `skip-deploy`         | The entire deployment can be skipped.   | This might be the case for a small fix, a tweak to documentation or something like that.                                         |
| `skip-web-deploy`     | The web tier can be skipped.            | This is the case if you're working on code that doesn't affect the front end, like management commands, tasks, or documentation. |
| `skip-celery-deploy`  | Deployment to celery can be skipped.    | This is the case if you make no changes to tasks.py or the code that tasks rely on.                                              |
| `skip-cronjob-deploy` | Deployment to cron jobs can be skipped. | This is the case if no changes are made that affect cronjobs.                                                                    |

**If deployment is required:**

- What extra steps are needed to deploy this beyond the standard deploy?
- Do scripts need to be run or things like that?
- If this is more than a quick thing, a new issue should be created in our infra repo: https://github.com/freelawproject/infrastructure/issues/new (if you don’t have access to it, just put the steps here)

---

</details>

**This PR should:**

- [x] <code>skip-celery-deploy</code>
- [x] <code>skip-cronjob-deploy</code>

Extra steps to deploy this PR:
1. The new setting defaults to the existing `celery` queue to allow merging this change without disruption. After merging, environment configurations should be updated to set the value of `CELERY_PACER_FETCH_QUEUE` to the new queue name  for fetch requests.

### DIFF
--- a/cl/recap/tasks.py
+++ b/cl/recap/tasks.py
@@ -175,29 +175,33 @@ def do_pacer_fetch(fq: PacerFetchQueue):
     :return: None
     """
     result = None
-    if fq.request_type == REQUEST_TYPE.DOCKET:
-        # Request by docket_id
-        court_id = get_court_id_from_fetch_queue(fq)
-        c = (
-            chain(fetch_appellate_docket.si(fq.pk))
-            if is_appellate_court(court_id)
-            else chain(fetch_docket.si(fq.pk))
-        )
-        c = c | mark_fq_successful.si(fq.pk)
-        result = c.apply_async()
-    elif fq.request_type == REQUEST_TYPE.PDF:
-        # Request by recap_document_id
-        rd_pk = fq.recap_document_id
-        result = chain(
-            fetch_pacer_doc_by_rd.si(rd_pk, fq.pk),
-            extract_recap_pdf.si(rd_pk),
-            mark_fq_successful.si(fq.pk),
-        ).apply_async()
-    elif fq.request_type == REQUEST_TYPE.ATTACHMENT_PAGE:
-        result = chain(
-            fetch_attachment_page.si(fq.pk),
-            replicate_fq_att_page_to_subdocket_rds.s(),
-        ).apply_async()
+    match fq.request_type:
+        case REQUEST_TYPE.DOCKET:
+            court_id = get_court_id_from_fetch_queue(fq)
+            c = (
+                chain(fetch_appellate_docket.si(fq.pk))
+                if is_appellate_court(court_id)
+                else chain(fetch_docket.si(fq.pk))
+            )
+            c = c | mark_fq_successful.si(fq.pk)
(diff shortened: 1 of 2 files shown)


cl/recap/tasks.py
```diff
@@ -175,29 +175,33 @@ def do_pacer_fetch(fq: PacerFetchQueue):
...
+                fetch_pacer_doc_by_rd.si(rd_pk, fq.pk),
+                extract_recap_pdf.si(rd_pk),
+                mark_fq_successful.si(fq.pk),
+            )
+        case _:
+            raise NotImplementedError(
+                f"Unsupported request_type: {fq.request_type}"
+            )
+
+    result = c.apply_async(queue=settings.CELERY_PACER_FETCH_QUEUE)
```
mlissner, line 204: I'm not sure this works (though it should). Looking at other chains in the codebase, we usually do something like:

```
chain(
  task1.s(xyz).set(queue=q), 
  task2.s(xyz).set(queue=q),
).apply_async()
```

I hope yours works too, but I'm not sure?

ERosendo: 🤔 I was a bit hesitant too, but I found this test in the Celery repository—it's related to groups though:

https://github.com/celery/celery/blob/bf1c98baba7431ee6a60d5972c5759699473acad/t/smoke/tests/quorum_queues/test_quorum_queues.py#L29-L36

mlissner: Works for me, thanks.

mlissner (approved):

mlissner:
Deployment complete. Application code updating now to get new secret.

ERosendo:
Thanks! I’ll keep an eye on it using Loki.

PR #6022, opened 2025-07-21
feat(back_scrape_citations): Add citation extraction from EFT

flooie:
Use extract from text to update citations from new versions.

## Fixes

Adds support for citation matching from inside documents.

## Summary

Checks if extract from text is implemented before using the more cumbersome process.

## Deployment

No deployment needed.

</details>

**This PR should:**

- [ ] <code>skip-celery-deploy</code>
- [ ] <code>skip-cronjob-deploy</code>
- [ ] <code>skip-daemon-deploy</code>


flooie:
This PR should allow us to update citations from older opinions in georgia and any other court that updates citations into the final opinion document 

cc @mlissner 

This issue was raised recently from a user that wanted GA citations.

PR #6027, opened 2025-07-21
Ensure orchestration runs

mlissner:
I noticed in a few of our runs that if we skip earlier steps, the `needs` parameter of the orchestration step makes it get skipped too. That's wrong. 

The idea of the orchestration step is that until it finishes, this action cannot be started a second time, so if it is skipped while a deployment is still happening, that's bad.

You can see an example of this here: 

https://github.com/freelawproject/courtlistener/actions/runs/16422096726

In that run, the orchestration step should not have been skipped as soon as an earlier step was skipped. The orchestration step should have run at the end, as designed.

The fix here is to make sure that it runs even if earlier deployments are skipped. Adding `if: always()` should accomplish this.

## Deployment

I'd like to do a minimal test to see if this works. That means skipping most things, but doing at least one of them. So:

- [x] <code>skip-web-deploy</code>
- [x] <code>skip-celery-deploy</code>
- [x] <code>skip-cronjob-deploy</code>
- [ ] <code>skip-daemon-deploy</code>


mlissner:
Perfect: 
image

Issue #6043, opened 2025-07-24
Implement new design for mobile scope selector


elisa-a-v:
The current implementation of the mobile scope selector in the Corpus Search component is out of date.

We need to update the design to match the following but without the question mark icon:

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/fb3e2388-88da-4c86-b52b-850f80dd7f85" />

Note the design introduces the total number of elements in each search scope, which is not included in the current implementation. For reference, some stats were added to the new homepage in #6173. And FYI, the numbers in Figma are orders of magnitude wrong (for example we don't have ~140k opinions but >10M).

Issue #6002, opened 2025-07-17
Implement date selectors


elisa-a-v:
First version of the advanced filters uses bare text inputs for date fields, but we need to implement the widget to select dates using a calendar, with a toggle between calendar and relative dates.

[Link to Figma (Desktop)](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=10106-67577&m=dev)

[Link to Figma (Mobile)](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=10130-80678&m=dev)

<details>
<summary>Screenshots</summary>

<img width="205" height="312" alt="Image" src="https://github.com/user-attachments/assets/4fd98a3a-9ec6-4b0a-9cba-394a8da7ab5b" />

<img width="205" height="231" alt="Image" src="https://github.com/user-attachments/assets/97843e91-4e2c-4bbf-aa81-09dd6e26da20" />

<img width="229" height="499" alt="Image" src="https://github.com/user-attachments/assets/cf095407-5beb-475c-b506-9a44c9d563e4" />

</details>

##### Question

From #5599 I understand the relative dates are available for all date fields, but the design only uses it for some of them. Is this what we want? It would make sense to me to allow users to filter all date fields by relative dates especially since this is possible via API, and I guess it might make it ever-so-slightly simpler to implement.

mlissner:
> From https://github.com/freelawproject/courtlistener/pull/5599 I understand the relative dates are available for all date fields, but the design only uses it for some of them. Is this what we want?

Yes. Relative dates are only useful for some fields, but it's a nice trick for oddball developers with weird ideas to make them work everywhere, even though the front end doesn't expose that.

Issue #6000, opened 2025-07-17
Implement Jurisdictions selector


elisa-a-v:
First version of the search tabs in the homepage uses a bare text input for the court field, but we need to implement the modal to select courts to filter results by.

Advanced filters should use this new selector everywhere (header search bar and homepage for now, search results once that's implemented).

[Link to Figma](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=6789-9664&m=dev)

<img width="3024" height="1964" alt="Image" src="https://github.com/user-attachments/assets/463389a8-89a0-4a46-b9b8-9dafddfcacbe" />

mlissner:
I think we can remove the check all buttons and search input in this iteration. The checkboxes in the column can suffice. It may bug power users, but I think it's OK.

flooie:
@mlissner have we settled on this layout?

mlissner:
See here for latest discussion: https://github.com/freelawproject/courtlistener/issues/6039

PR #6044, opened 2025-07-24
feat(new_homepage): introduce corpus search UI with advanced filters

elisa-a-v:
## Fixes

Relates to #5206 but doesn't fully resolve it. Pending tasks described below.

## Summary

This PR introduces the search UI in the homepage using the same components from the header search bar.

To achieve this, this PR also:
- Introduces a new `CorpusSearchForm` class and adds it to all new templates.
- Refactors and extends the Corpus Search component.
- Extends the Dialog component by introducing a sub-component for a trigger button.

**Minor improvements:**
- Added custom style for checkboxes which is automatically added to any input with `type="checkbox"`.
- Introduced a new utility `focus-container` that provides a focus style to an element when a child input is focused.

**What's still pending:**
- #6000
- #6002
- #6043

#### New `CorpusSearchForm` class

The new form class is heavily based on the existing `SearchForm`, but only includes the fields needed so far, and changes styles to use tailwind classes and match new design.

We use the `IncrementalNewTemplateMiddleware` to introduce the `search_form` with an instance of the new `CorpusSearchForm` to the context in all new templates. This is required because we now support searching from every single page, so we need the form everywhere.

Before adding the new `CorpusSearchForm` instance, we check the existence of the new template, and only modify the context when it does exist, otherwise we return the response as-is. This is important to avoid breaking some old views like the Case Law advanced search.

#### Corpus Search component updates

1. **Refactor mobile version.**
   Mobile dialog is more easily reused with a customizable trigger button, so we use the same component in the header (with the magnifier icon trigger button), and in the homepage (with the "Filters" trigger button)

2. **Introduce fully responsive advanced filters with scoped fieldsets.**
   Fieldsets are disabled unless the scope is active. This prevents entered values in one scope from also being sent when submitting the form in a different scope.

3. **Introduce new scope selector variant.**
   By default the scope selector remains a simple dropdown (soon to me enhanced with new design), and we now also support a tabbed variant which is the one used in the desktop version of the new homepage.

## Deployment

**This PR should:**
- [ ] `skip-deploy`
- [ ] `skip-web-deploy`
- [x] `skip-celery-deploy`
- [x] `skip-cronjob-deploy`
- [x] `skip-daemon-deploy`

## Screenshots

<details>
<summary>Mobile homepage</summary>

Screenshot from 2025-07-24 14-30-48

</details>

#### Case Law

<details>
<summary>Mobile</summary>

Screenshot from 2025-07-24 13-50-10

</details>

<details>
<summary>Desktop</summary>

image

</details>

#### RECAP Archive

<details>
<summary>Mobile</summary>

Screenshot from 2025-07-24 13-50-18

</details>

<details>
<summary>Desktop</summary>

image

</details>

#### Oral Arguments

<details>
<summary>Mobile</summary>

image

</details>
<details>
<summary>Desktop</summary>

image

</details>

#### Judges

<details>
<summary>Mobile</summary>

Screenshot from 2025-07-24 13-59-01

</details>

<details>
<summary>Desktop</summary>

image

</details>

## Demos

<details>
<summary>Mobile homepage</summary>

[Screencast from 07-24-2025 03:16:56 PM.webm](https://github.com/user-attachments/assets/fbca01f1-6576-4839-8d67-0122749f5081)

</details>

<details>
<summary>Accessing advanced search from header</summary>

[Screencast from 07-24-2025 04:09:44 PM.webm](https://github.com/user-attachments/assets/8bf85f09-d1a4-48f3-bb87-6ed21ab0d720)

</details>

<details>
<summary>Responsive layout</summary>

[Screencast from 07-24-2025 04:21:18 PM.webm](https://github.com/user-attachments/assets/e7c4018a-b152-4709-a265-16064c5041f3)

</details>


mlissner:
I want to start giving @albertisfu some of these front end PRs so that he can start getting more up to speed on the new architecture, but I think this is a particularly big one, so it's better if Eduardo goes for it. But I'm going to keep an eye out for one for you, Alberto. :)

cl/lib/middleware.py
```diff
@@ -92,7 +96,15 @@ def process_template_response(self, request, response):
         ):
             old_template = response.template_name
             if isinstance(old_template, str):
-                new_template = f"v2_{old_template}"
-                response.template_name = [new_template, old_template]
+                new_template_name = f"v2_{old_template}"
+
+                try:
+                    # verify the new template actually exists
+                    select_template([new_template_name])
```
ERosendo: Given that we only need to check a single template name, we can use [get_template](https://docs.djangoproject.com/en/5.2/topics/templates/#django.template.loader.get_template) instead of `select_template`

cl/assets/templates/cotton/dialog/index.html
ERosendo, line 1: The console is showing a few warnings related to this component. Here's a screenshot for reference:

image

cl/search/templates/cotton/corpus_search/scope.html
```diff
@@ -3,9 +3,54 @@
...
+            class=""
+          ></span>
+          <span
+            x-show="isActiveScope"
+            x-bind:data-scope="scope.label"
+            x-text="scope.shortDescription"
+            class="text-xs text-greyscale-500"
+          ></span>
+        </span>
+      </button>
```
ERosendo: I noticed that our tabs currently use the same color for both the **selected** and **unselected** states. At first, I thought it might just be my impression, but I double-checked the Figma mockup and confirmed it that the inactive tab should use a slightly warmer tone. Here's a video for reference:



https://github.com/user-attachments/assets/a0f612c0-323e-4cba-a6a9-8591e2ae9ec6

cl/search/templates/cotton/corpus_search/filters/judges.html
```diff
@@ -0,0 +1,76 @@
...
+        class="text-ellipsis overflow-hidden text-nowrap"
+        title="{{ search_form.born_before.label }}"
+      >
+        {{ search_form.born_before.label_tag }}
+      </div>
+      <div class="focus-container input-text flex justify-stretch items-stretch gap-2">
+        {% svg "calendar" class="w-5 h-5" %}
+        {{ search_form.born_before }}
+      </div>
+    </div>
```
ERosendo: Is there a specific reason we’re wrapping these date fields in a `<div>`? On extra-large screens, I noticed this causes a slight increase in the spacing between the search bar and the first row of filters. After inspecting the HTML, I saw that the `labels` inside the `<div>` are slightly taller than regular labels.

Here's a video showing the issue:

https://github.com/user-attachments/assets/06db1f15-8d8e-43d0-ab14-fe16d7e490ad


As shown in the video, a standard label has a height of 20px, while the label wrapped in a `<div>` ends up with 24px. If the intent behind the `<div>` is just to apply a class to the label, I believe we can simplify it like this:

```html
      <label for="{{ search_form.born_after.id_for_label }}" class="text-ellipsis overflow-hidden text-nowrap">
        {{ search_form.born_after.label }}
      </label>
      <div class="focus-container input-text flex justify-stretch items-stretch gap-2">
        {% svg "calendar" class="w-5 h-5" %}
        {{ search_form.born_after }}
      </div>
```

cl/assets/templates/cotton/header.html
```diff
@@ -100,24 +101,13 @@
...
-            Search
-          </c-corpus-search.button>
-        </c-corpus-search.mobile>
+        <c-corpus-search.mobile.dialog :request="request" :search_form="search_form" only>
+          <c-slot name="trigger_button">
+            <c-dialog.trigger-button>
+              {% svg "magnifier" class="text-greyscale-400 w-5 h-5" %}
+            </c-dialog.trigger-button>
+          </c-slot>
+        </c-corpus-search.mobile.dialog>
```
ERosendo, line 110: I noticed that using the Advanced Search from the header triggers an uncaught `TypeError` in the console. Here's a video that shows the error:


https://github.com/user-attachments/assets/ce0a5767-44ab-43e5-8035-95f7650fdce7

ERosendo (changes requested):
@elisa-a-v thanks for this PR! The new filters component looks very close to the mockups, and tab navigation works as expected. In addition to my previous comments, I have two more notes:
 
----

1. In the Figma mockups, there's a component next to the scope selectors that appears to be an anchor link. I'm attaching a screenshot highlighting the element I'm referring to.

    This PR doesn’t introduce that highlighted element. I checked the document provided in issue #5206 and didn’t see any mention of it being removed. It’s also not listed among the pending tasks, so I’m not sure if it’s expected to be added in a follow-up PR or if it was intentionally dropped from the design, though it’s still visible in the mockups.

Screenshot 2025-07-26 at 10 34 30 PM

   


----

2. I noticed that the "Filters" trigger button is always present in the search bar within the header. On mobile, it appears below the keywords input. However, for larger breakpoints, the trigger should be integrated inside the keywords input itself.

   This PR is introducing the trigger button for the mobile version , but the trigger is missing on larger screen sizes. This task isn’t listed as pending, and there’s no clear indication that it was intentionally left out. I'm attaching a screenshot highlighting the missing element:

Screenshot 2025-07-26 at 10 52 05 PM

 Just wanted to flag this in case the desktop version is still planned or was deferred.

cl/assets/templates/cotton/dialog/index.html
elisa-a-v: Yeah I wouldn't worry about those to be honest. They were there before and we don't have any issues with the dialogs, it's just the CSP build being noisy.

Some built-in features aren't supported so some warnings here and there will be inevitable with our setup.

cl/assets/templates/cotton/header.html
elisa-a-v: This happened because the desktop header search bar doesn't have advanced filters yet, so the function that updates the fieldsets threw the error when trying to update a non-existent element. Now we simply return when no element is found.

cl/search/templates/cotton/corpus_search/filters/judges.html
elisa-a-v: Nah those are probably just a leftover from various layout experiments, no need for them so I removed them 🫡 good catch

elisa-a-v:
Thanks @ERosendo for your feedback! 🙏🏽 I've addressed all comments.

> This PR doesn’t introduce that highlighted element.

Yeah forgot to add it, it's in now! Despite [not being in the figma mockup yet](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4?node-id=10130-73692&m=dev#1352595515), I've also included it below the "Search" button in mobile, both in the homepage and the header, as discussed in meetings with the design team. This is replacing the question mark icon in the new design of the scope selector that will be implemented in #6043 

<details>
<summary>Mobile</summary>

image

image


</details>

I've also hidden the "About this data" in a small range of widths in which the text didn't fit. When no text is visible the icon looks a bit bigger.

<details>
<summary>Comparison</summary>

image

vs

image

</details>

> I noticed that the "Filters" trigger button is always present in the search bar within the header.

That's right, we're adding that when we introduce the advanced filters in the desktop header. I missed that one when listing the pending stuff so thanks for pointing it out, the new issue can be found here: #6072

cl/search/templates/cotton/corpus_search/mobile/dialog.html
```diff
@@ -49,11 +49,18 @@
...
+        <div class="p-4 flex flex-col gap-4">
           <c-corpus-search.button class="w-full max-md:btn-xl flex items-center justify-center">
             {% svg "magnifier" class="text-white w-5 h-5 mr-1.5" aria_hidden="true" %}
             Search
           </c-corpus-search.button>
+          <a
+            href="{% url "coverage" %}"
+            class="flex items-center justify-center text-sm text-greyscale-600 gap-1 hover:text-greyscale-500"
+          >
+            {% svg "circle_question_mark" class="text-greyscale-400 hover:text-greyscale-300 h-4 w-4" %}
```
ERosendo, line 61: @elisa-a-v It seems the SVG file is missing.

elisa-a-v: omg you're right! 🤦🏽 my bad, it's pushed now 🫡

elisa-a-v:
Thanks again Eduardo for all your feedback!

Latest changes:
- Renamed the components library template to begin with `v2_`, following our standard and thus having the new search form injected in its context. This fixed an issue in which the advanced filters were not working in that page.
- We now close all dialogs when the window is resized past the mobile breakpoint to avoid weird UX (dialogs appearing out of nowhere just by resizing the window can be confusing), and to fix an accessibility issue in which the desktop form remained hidden to AT if the dialog was not closed when the window was resized.
- Refactored the `IncrementalNewTemplateMiddleware` to use early return and reduce indentation.

I've also opened the following issues for more pending work:
- #6085
- #6086

ERosendo (approved):
@elisa-a-v Thanks for the latest tweaks! I confirmed that the advanced filters are working on the components page, and the updated dialog behavior when resizing the window looks great. I also appreciate the refactor to use early returns, it's much cleaner.

:shipit:

Issue #914, opened 2018-12-06
Count user queries and then prompt big users for contributions


mlissner:
When people search the RECAP Archive and don't have the extension installed, we bug them to install it with the idea being that if they're querying the DB, they should contribute to it. 

A suggestion today was that we should also prompt people for contributions if they do more than a certain number of queries. It's about awareness that FLP is a non-profit and about getting people to think about whether they can/should contribute. 

I'm tempted to build this out as a search history feature. For some time, I've wanted to keep track of people's queries so that they can find their old work again (and so, perhaps we can pop it into the query box for folks). For now, that's complicated and has privacy issues, so I think we can just keep track of their query *count*, and then prompt using that.

I'd say the best way to do this is to start collecting the per-user counts first, then when we know that, we can think about when to put the prompt. 

It'd also be nice to do this for unauthenticated users, though I don't know if that's easy or worth it.

mlissner:
This was a good idea and we have search history now. Putting this on the backlog.

tclancy:
Would this just be a matter of grouping the `SearchQuery` by user and making it available for display? Would you want to filter by date range at all? Feels like this should be stuck in the cache and invalidated/ incremented when a new `SearchQuery` is made by the user.

mlissner:
Yeah, the search query table is very large, so trying to count the user's number of requests won't perform, but it's not crazy to put it in redis. I think if we did that, we'd want: 

 - A django command or script to backfill redis based on the current searchquery table. 
 - A bit of code to increment the counts as needed (excluding API queries). This might be able to follow the pattern that's used by the API logger, which keeps this for API queries (maybe it'd even integrate to prevent code duplication — that might be a bad idea, but it's worth checking?) 
 - A little alert thing in the search results that pops up according to the milestones we have in the stats module starting at maybe 50 queries? The messages we send could be different based on the number of queries they've made, but with a light touch? And the message shouldn't show up for donors or members.

Then we just check the count when the user does a query, increment it, and if it hits a milestone, we show the appropriate thing to please donate or become a member.

All in all, pretty simple, I think?

PR #6046, opened 2025-07-24
Add RECAP query counter to track user searches

killerdevildog:
- Add recap_query_count field to UserProfile model
- Increment counter when users perform RECAP searches via website or API
- Track searches only for authenticated users and successful queries
- Add proper error handling to prevent breaking search functionality

Addresses issue #914: Count user queries for future contribution prompting

Fixes

Fixes: #914
Summary

This PR implements the first phase of tracking user queries to eventually prompt heavy users for contributions.
Approach:

Added a recap_query_count field to the UserProfile model to store the count
Implemented increment logic in the search views/API endpoints when authenticated users perform successful RECAP searches
Added proper error handling to ensure counting failures don't break search functionality

Key decisions:

Only tracking authenticated users (as suggested in the issue)
Only counting successful queries to avoid inflating counts from errors/retries
Using a simple integer counter field for now, which can be extended later with more detailed tracking if needed
Not implementing the prompting UI yet - just the counting mechanism as suggested in the issue

Deployment
This PR should:

 skip-deploy

 skip-web-deploy

 skip-celery-deploy

 skip-cronjob-deploy

 skip-daemon-deploy

Run database migration to add the new recap_query_count field to UserProfile model
No data migration needed - all users will start with a count of 0


mlissner:
@killerdevildog, thanks for this contribution, but did you see that we're logging user queries now? This comment mentions it: 

https://github.com/freelawproject/courtlistener/issues/914#issuecomment-2480279126

So I don't think we need this as a denormalized field in the user profile and probably we just need to figure out the right way to ask for donations.

Sorry! I think we should probably close this one?

Issue #6048, opened 2025-07-24
Adjust INCEPTION_HOST variable to handle CPU and GPU


mlissner:
In https://github.com/freelawproject/kubernetes/pull/35 I created a new deployment for the inception CPU instance, and I renamed the deployment for the GPU instance. 

In our code, we have a setting for INCEPTION_HOST, which I changed to point to the new CPU instance, but we need to tweak our code to use CPU when appropriate and GPU when appropriate. 

So. I've already created INCEPTION_CPU_HOST and INCEPTION_GPU_HOST secrets in AWS. They're ready to go. Now we just need to:

1. Split the current variable into INCEPTION_CPU_HOST and INCEPTION_GPU_HOST depending on which is correct.
2. Deploy this code.
3. Someday: Delete INCEPTION_HOST (not that it *really* matters, it's just being tidy).

PR #6050, opened 2025-07-25
ci(settings): split INCEPTION_HOST to CPU vs GPU

rachlllg:
## Fixes
Close: https://github.com/freelawproject/courtlistener/issues/6048

## Summary

This PR splits the INCEPTION_HOST setting to INCEPTION_CPU_HOST and INCEPTION_GPU_HOST. It also assigns the appropriate host for query vs batch requests, where query uses the CPU host while batch uses the GPU host.

## Deployment

**This PR should:**

- [ ] `skip-deploy`

- [ ] `skip-web-deploy`

- [ ] `skip-celery-deploy`

- [X] `skip-cronjob-deploy`

- [X] `skip-daemon-deploy`


rachlllg:
@ERosendo  helped me better understand the settings and the difference between local and production env and the different deployments we have, thank you!

mlissner (approved):
Nice. Good to see this was straightforward. Almost like we planned on it!

Issue #5980, opened 2025-07-15
Running tests in reverse causes 70+ failures


florean:
If tests are run in reverse order using the `--reverse` flag, there are a ton of failures.  This is related to #5969, but could be resolved independently of it.  Because `--reverse` preserves [test grouping](https://docs.djangoproject.com/en/5.2/topics/testing/overview/#order-of-tests), the root problem here is TestCases that delete objects from the database, whereas #5969 has the additional complication of TransactionTestCases truncating the entire database.  And while I think it would be perfectly reasonable to just write off `--keepdb` as not working, this definitely needs to be fixed because any new test could run into this depending on where it is in the default order.

Example results for `./manage.py test cl --verbosity=0 --exclude-tag selenium --parallel=1 --timing`:
```
----------------------------------------------------------------------
Ran 1510 tests in 316.320s

OK (skipped=53)
Total database setup took 57.088s
  Creating 'default' took 57.088s
Total database teardown took 1.154s
Total run took 377.016s
```

With `--reverse`:
```
----------------------------------------------------------------------
Ran 1510 tests in 305.167s

FAILED (failures=70, errors=2, skipped=53)
Total database setup took 60.243s
  Creating 'default' took 60.243s
Total database teardown took 1.130s
Total run took 371.754s
```

mlissner:
Nice one. Simple way to catch this stuff. Thanks for your focus on this.

PR #6052, opened 2025-07-25
Fixes for tests that were clobbering migration data

florean:
Fixed some tests were deleting the `Court` objects that are installed by a migration and upon which other tests rely.

## Fixes

This makes progress on #5980, but doesn't resolve it.  But it does knock it down to 69 errors!  Nice!
```
Ran 1530 tests in 316.306s

FAILED (failures=69, errors=2, skipped=53)
```

## Summary

The `DRFCourtApiFilterTests` used factories to create Court objects, but it manually set the IDs to ones used by the migration fixture, so it started by deleting all Court objects.  This does not get rolled back by TestCase, so subsequent tests that rely on them will fail (I need to figure out if that is a Django bug and file it; I don't believe I came across an open one, but it seems like a big problem).  This seems to normally run later in the run, which is why it results in more failures when running in reverse.
It also includes some miscellaneous fixes: an import that was getting clobbered and thus not being tested, a test that unnecessarily used `TransactionTestCase`, and an unused variable.

## Deployment

**This PR should:**

- [X] `skip-deploy`

- [ ] `skip-web-deploy`

- [ ] `skip-celery-deploy`

- [ ] `skip-cronjob-deploy`

- [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/api/tests.py
+++ b/cl/api/tests.py
@@ -871,8 +871,7 @@ class FilteringCountTestMixin:
 class DRFCourtApiFilterTests(TestCase, FilteringCountTestMixin):
     @classmethod
     def setUpTestData(cls):
-        Court.objects.all().delete()
-
+        super().setUpTestData()
         cls.parent_court = CourtFactory(
             id="parent1",
             full_name="Parent Court",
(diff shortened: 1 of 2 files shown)


albertisfu (approved):
Thanks @florean this looks good and ready to be merged.

> This does not get rolled back by TestCase, so subsequent tests that rely on them will fail (I need to figure out if that is a Django bug and file it; I don't believe I came across an open one, but it seems like a big problem)

Yeah this is indeed surprising.

PR #6053, opened 2025-07-25
Implement opinion versions in frontend

### LINKED ISSUE #5451: Show opinion versions in opinion page
From Mike's comments in the parent issue 

> We need a UI that shows previous versions. I imagine this could be a page like /opinion/23232423/{slug}/versions/, but if it's really simple, it might be able to fit into a drop down like on Github comments.

> Either way, ideally it would show the dates of earlier versions and have an interface to do diffs. There is a for-profit co that does PDF diffs in browser extremely well. I think they'll give us free usage of their tool.

> We need to hide duplicates from search except, perhaps, to have a button next to some results that says "Versions" or something like that.

> We need the field to make sense in the API, and we need to document it accordingly.

> Discovery: How do users know that there are multiple versions of a decision, and how do they find them? I imagine this would go on the opinion page itself somewhere, perhaps as a pretty small button/icon/thing, but then people will want to be able to read such decisions, so we'll need a way to do that. Ideas welcome here. This implies that the secondary opinions will need their own URLs.

### LINKED ISSUE #6800: Make more than 1 sub opinion PDF reachable on cluster view
### Describe the issue.

Only the first PDF is considered, so users can't reach the other PDFs in the cluster

See an example https://www.courtlistener.com/opinion/10775389/shamrock-enterprises-llc-dba-frsteam-gulfcoastla-v-top-notch-movers/pdf/

https://github.com/freelawproject/courtlistener/blob/e1ee62ca13adfb0d7ea2c61dddd48d0cd9922b2b/cl/opinion_page/views.py#L866-L887

### Related Issue

_No response_

### Screenshots

_No response_

Research

quevon24:
## Fixes

None

## Summary

This PR introduces a new feature that allows users to view different versions of opinions. 

- Users can now access and view historical versions of opinions. 
- A cached_property named opinions_versions and opinion_main_version has been implemented in the model. This efficiently fetches and caches sub-opinion versions, ordered by creation date, to optimize performance. Also fetches main version of a cluster.
- Up to a maximum of 10 versions are displayed.

## Deployment

**This PR should:**

- [ ] `skip-deploy`

- [ ] `skip-web-deploy`

- [x] `skip-celery-deploy`

- [x] `skip-cronjob-deploy`

- [x] `skip-daemon-deploy`

<details closed>
<details open>
image
In this screenshot you see the same dates because i cloned the data and the date_created value is almost the same for each version

</details>
<details open>

</details>


mlissner:
This is looking pretty good. One thing I don't see, that I think is important, is some explanation of why we have so many versions. In an earlier version of this there was a page that listed the older versions. Maybe something like that would have space to explain the situation? Another option could be a modal that the button opens which shows the versions, but that might not scale well if there's more than one or two (the current design has this problem too).

flooie:
@quevon24 I havent touched this as its still listed as a draft but - did you make any changes to the download pdf dropdown
Screenshot 2025-07-30 at 1 52 36 PM

I imagine that this is actually where I would want to identify versions.  ... like download ... pdf from x date or something

mlissner:
I like the direction, but I wouldn't click "Download PDF" if I was trying to get to an older version, and I wouldn't expect a download button to take me to an HTML page for an earlier version (I'd expect a PDF). Hm.

grossir:
@quevon24 

Please check my latest comment on the related issue https://github.com/freelawproject/courtlistener/issues/5451#issuecomment-3946106171

quevon24:
I implemented a collapsible accordion in the PDF tab when a cluster has multiple independent opinions with PDFs (e.g. lead, dissent, concurrence). Clusters with a single PDF or where all opinions are versions of the same document continue to display the single PDF view as before. Also updated the version labels in the Download PDF dropdown to show the opinion type and date.

Dropdown:
image

Collapsible:
image

cl/opinion_page/templates/includes/add_download_button.html
```diff
@@ -41,5 +41,14 @@
                 </li>
             {% endif %}
         {% endfor %}
+    {% if cluster.opinions_versions %}
+        <li role="separator" class="divider"></li>
+        <li><a href="#"><b>Versions</b></a></li>
```
grossir: This could be a `  <li class="dropdown-header">Versions</li>`  already packaged on bootstrap

cl/opinion_page/views.py
```diff
@@ -898,29 +898,52 @@ def get_attachment_values(
...
-                break
-            elif sub_opinion.download_url:
-                has_downloads = True
-                pdf_path = None
-
-    return {"has_downloads": has_downloads, "pdf_path": pdf_path}
+        download_file_path = cluster.filepath_pdf_harvard.url
+        embeddable_pdfs.append(
+            {
+                "label": "Harvard PDF",
```
grossir: I think the label should be something like "Case Law Access Project scan" or something, to be consistent with other usages

Image

cl/search/models.py
```diff
@@ -2813,6 +2813,18 @@ def ordered_opinions(self):
         # If there's only one or no sub-opinions, return the main opinion
         return sub_opinions
 
+    @cached_property
+    def opinion_main_version(self):
+        # Fetch opinion main version of the cluster
+        return self.sub_opinions.filter(main_version__isnull=True)
```
grossir, line 2827: We should type annotate the return or change the name or add a better docstring, since the name makes it look like it will return a singular opinion, when a cluster with opinion types may return many.

grossir (commented):
Some small changes, it's looking good

grossir (approved):
LGTM

PR #6073, opened 2025-07-29
fix(merge_opinion_versions): limit versioning to SOURCES.COURT_WEBSITE clusters

### LINKED ISSUE #6045: Stop merging version clusters that have Harvard data
From the tests on `ohio` opinion [versions](https://github.com/freelawproject/courtlistener/issues/5768#issuecomment-3114233531) we got 2 clusters with different Harvard data merged, losing some Harvard data. 

This was downstream of a bad merger; where the case "Cite as [In re Resignation of Roser](https://storage.courtlistener.com/harvard_pdf/2842992.pdf), 144 Ohio St.3d 1213, 2015-Ohio-3273" was assigned the same "download_url" as the case "Cite as [In re Mason](https://storage.courtlistener.com/harvard_pdf/2841898.pdf), 143 Ohio St.3d 1295, 2015-Ohio-3579."

The shared URL actually belongs to "[In Re Mason](https://storage.courtlistener.com/pdf/2015/09/03/in_re_mason.pdf)"

I expect there to be more bad Harvard - Scraper sources mergers, and we don't want to complicate more the issue by merging them as versions. So, if both main and version clusters have different harvard data, the merge should be stopped and a warning sent to Sentry. 

Also, we should regenerate the deleted cluster with harvard data. Luckily, it was a single case for `ohio`

```
WARNING Unexpected difference in filepath_json_harvard: 'law.free.cap.ohio-st-3d.144/1213.4043319.json' 'law.free.cap.ohio-st-3d.143/1295.4353884.json'. opinioncluster: 2842992, 2841898
WARNING Unexpected difference in filepath_pdf_harvard: 'harvard_pdf/2842992.pdf' 'harvard_pdf/2841898.pdf'. opinioncluster: 2842992, 2841898
```

grossir:
## Fixes
Fixes #6045

Since, for now, we use `Opinion.download_url` for identifying possible versions, and this is only populated for scraped opinions, we are limiting possible versions to scraped opinions.

This solves the problem in the referenced issue, where we found bad Harvard-Scraper merges, which would lead to data loss when versioning

## Summary
Fixed by adding filter clauses to the queryset that finds candidate versions by URL

## Deployment

**This PR should:**

- [ ] `skip-deploy`

- [x] `skip-web-deploy`

- [ ] `skip-celery-deploy`

- [x] `skip-cronjob-deploy`

- [x] `skip-daemon-deploy`

1. To deploy...


flooie (commented):
Can we add a test where it wouldnt merge because something is not solely sourced from a website?  Otherwise the code looks good to me

grossir:
@flooie  just added the test with a candidate version with `source=SOURCES.COURT_M_HARVARD`

flooie (approved):
Looks good.

PR #6074, opened 2025-07-29
feat(merge_opinion_versions): ingest versioned clusters into ClusterRedirection table

### LINKED ISSUE #6063: Save versioned clusters to `ClusterRedirection` table
### What happened?

Once the `ClusterRedirection` table is [merged](https://github.com/freelawproject/courtlistener/pull/5996), we should make deleted version clusters redirect to the main version cluster

We should also pick up from the history tables the already deleted cluster ids, and include them in the `ClusterRedirection` table. Creating this as a sub issue

### Expected behavior

Version clusters are deleted, and their ids saved in the `ClusterRedirection` table

### Actual behavior

Version clusters are deleted, their ids lost

### Steps to reproduce

_No response_

### Screenshots

_No response_

Unknown

grossir:
## Fixes

Fixes #6063

When a cluster is deleted through versioning, a `ClusterRedirection` row with the deleted cluster id and the remaining main cluster is created

## Summary

In the transaction block that deletes the versioned cluster, create a `ClusterRedirection` entry

## Deployment

**This PR should:**

- [ ] `skip-deploy`

- [X] `skip-web-deploy`

- [ ] `skip-celery-deploy`

- [X] `skip-cronjob-deploy`

- [X] `skip-daemon-deploy`

### DIFF
--- a/cl/scrapers/management/commands/merge_opinion_versions.py
+++ b/cl/scrapers/management/commands/merge_opinion_versions.py
@@ -23,6 +23,7 @@ from cl.search.models import (
     SOURCES,
     BankruptcyInformation,
     Claim,
+    ClusterRedirection,
     Docket,
     DocketEntry,
     DocketPanel,
@@ -484,6 +485,15 @@ def merge_opinion_versions(
             ES_CHILD_ID(version_opinion.id).OPINION,
             version_cluster.id,
         )
+
+        if not ClusterRedirection.objects.filter(
+            deleted_cluster_id=version_cluster.id
+        ).exists():
+            ClusterRedirection.objects.create(
+                deleted_cluster_id=version_cluster.id,
+                cluster=main_opinion.cluster,
+                reason=ClusterRedirection.VERSION,
+            )
 
         # Both Opinion and Citation have a ForeignKey to OpinionCluster, with
         # on_delete=models.CASCADE. Also, there are signals (see
--- a/cl/scrapers/tests.py
+++ b/cl/scrapers/tests.py
@@ -69,6 +69,7 @@ from cl.search.models import (
     SEARCH_TYPES,
     SOURCES,
     Citation,
+    ClusterRedirection,
     Court,
     Docket,
     Opinion,
@@ -1110,6 +1111,7 @@ class OpinionVersionTest(ESIndexTestCase, TransactionTestCase):
             other_dates=other_dates,
             summary="",
         )
+        cluster2_id = cluster2.id
         cluster3 = OpinionClusterFactory.create(
             docket=version_docket, other_dates="", summary=summary
         )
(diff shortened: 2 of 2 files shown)


cl/scrapers/management/commands/merge_opinion_versions.py
```diff
@@ -485,6 +486,15 @@ def merge_opinion_versions(
...
 
+        if not ClusterRedirection.objects.filter(
+            deleted_cluster_id=version_cluster.id
+        ).exists():
+            ClusterRedirection.objects.create(
+                deleted_cluster_id=version_cluster.id,
+                cluster=main_opinion.cluster,
+                reason=ClusterRedirection.VERSION,
+            )
+
```
flooie: Would it be easier to just do this?  
```
ClusterRedirection.objects.get_or_create(
    deleted_cluster_id=version_cluster.id,
    cluster= main_opinion.cluster,
    reason= ClusterRedirection.VERSION,
)
```

grossir: `deleted_cluster_id` has a unique constraint, so when using `get_or_create` you risk trying to create another entry with the same `deleted_cluster_id`, since it's also trying to filter by the other 2 fields; and that would raise an error

Now, that should never happen because we are actually deleting the cluster... so it should work

Still, I have to add some more logic here, for deleting a cluster that already has `ClusterRedirection` entries pointing to it

cl/scrapers/tests.py
```diff
@@ -1472,3 +1486,28 @@ def test_string_merging(self):
...
+        )
+
+        self.assertTrue(
+            ClusterRedirection.objects.filter(
+                deleted_cluster_id=existing_redirection_id,
+                cluster=to_keep,
+                reason=ClusterRedirection.DUPLICATE,
+            ).exists(),
+            "Existing re-direction was not re-assigned to new cluster",
+        )
```
quevon24, line 1564: I think you could also test something like this to verify that not only the ClusterRedirection object was created in L1496, but also to test that `create_from_clusters` method created the ClusterRedirection object with `VERSION` reason, not only updates object with `DUPLICATE` reason:

```
to_delete_id=888888
self.assertTrue(
            ClusterRedirection.objects.filter(
                deleted_cluster_id=to_delete_id,
                cluster=to_keep,
                reason=ClusterRedirection.VERSION,
            ).exists(),
        )
```

quevon24 (commented):
I think it looks good, I just have a suggestion for a test, I just need feedback to know if what I suggest is valid.

cl/scrapers/tests.py
grossir: It's a good question @quevon24 ; I had thought about it but it's better to write it

If cluster1 is a duplicate of cluster2, and then cluster2 is marked as a version of cluster3
- cluster1.sub_opinions will not exist; because we deleted them as exact duplicates
- cluster2.sub_opiniosn will still exist, because we keep them as different versions

In this case, I think we should keep the "DUPLICATE" reason for cluster1 redirection to cluster3. If we replaced it with "VERSION", in the future we may wonder why did the version was deleted together with it's cluster

In general, I think we should keep the original deletion reason; instead of adopting the transitive redirection reason

quevon24: Hmm, but then is it correct that there are two objects of type ClusterRedirection, one with VERSION and the other with DUPLICATE in the test? I suppose that doesn't matter since either of those two will ultimately redirect to the existing cluster. The difference is that depending on which deleted cluster id you use, you'll be able to determine the reason for the redirect.

grossir: Yes, we should keep all the deleted_cluster_ids; but know the reason why they were deleted; I think that shouldn't change once the redirection is re - assigned to another cluster

quevon24: Perfect, then everything is correct. I'll re-request a review myself so I can approve it.

PR #6080, opened 2025-07-29
fix(api_views): fix overriden OpinionClusterViewset.retrieve

grossir:
## Fixes

Fixes a bug introduced while implementing #5983 in #5996

Sentry Issue: [COURTLISTENER-AB9](https://freelawproject.sentry.io/issues/6777665971/?referrer=github_integration)

```
TypeError: OpinionClusterViewSet.retrieve() got an unexpected keyword argument 'format'
(3 additional frame(s) were not displayed)
...
  File "rest_framework/viewsets.py", line 125, in view
    return self.dispatch(request, *args, **kwargs)
  File "rest_framework/views.py", line 509, in dispatch
    response = self.handle_exception(exc)
  File "rest_framework/views.py", line 469, in handle_exception
    self.raise_uncaught_exception(exc)
  File "rest_framework/views.py", line 480, in raise_uncaught_exception
    raise exc
  File "rest_framework/views.py", line 506, in dispatch
    response = handler(request, *args, **kwargs)
```

## Summary

Overriden `OpinionClusterViewset.retrieve` was not accounting for extra kwargs, which was causing an error when passing extra DRF args such as "format". Now using standard (*args, **kwargs) and calling super().retrieve

## Deployment

**This PR should:**

- [ ] `skip-deploy`

- [ ] `skip-web-deploy`

- [X] `skip-celery-deploy`

- [x] `skip-cronjob-deploy`

- [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/api/tests.py
+++ b/cl/api/tests.py
@@ -3804,13 +3804,28 @@ class ClusterRedirectionTest(TestCase):
         """Test that a deleted cluster redirects to an existing cluster"""
         url = reverse(
             "opinioncluster-detail",
-            kwargs={"version": "v4", "pk": self.deleted_cluster_id},
+            kwargs={
+                "version": "v4",
+                "pk": self.deleted_cluster_id,
+                "format": "json",
+            },
         )
         api_client = await sync_to_async(make_client)(self.user.user.pk)
         response = await api_client.get(url)
         self.assertEqual(response.status_code, HTTPStatus.MOVED_PERMANENTLY)
+        redirect_url = response.headers["Location"]
 
-        redirect_response = await api_client.get(response.headers["Location"])
+        # Check that extra kwargs are passed
+        self.assertTrue(
+            "v4" in redirect_url,
+            "'version' extra kwarg not passed in redirection",
+        )
+        self.assertTrue(
+            ".json" in redirect_url,
+            "'format' extra kwarg not passed in redirection",
+        )
+
+        redirect_response = await api_client.get(redirect_url)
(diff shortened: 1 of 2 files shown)


mlissner:
Looks like a good candidate for a couple tests, no?

grossir:
It is, I did put some tests in place in the original PR; but I assumed "version" and "pk" where the only possible kwargs; but they are not. I will expand the tests

grossir:
@mlissner I think this is ready; I just added some tests to check the propagation of the "format" kwarg  on the redirection, which also implies that the kwarg can be handled.

The kwargs passed to the ModelViewSet.retrieve are defined by the [DefaultRouter](https://www.django-rest-framework.org/api-guide/routers/#defaultrouter), and have this shape `{prefix}/{lookup}/[.format]` for this `{basename}-detail` endpoint. `{lookup}` will just be the `pk`  kwarg.  Was missing the `.format` kwarg, but I will just leave it with (*args, **kwargs) in case something is added in the future 

Relevant code is in the `api/urls` file

https://github.com/freelawproject/courtlistener/blob/1116b6e2e813fa357fb028089eb101f45eab030e/cl/api/urls.py#L201-L202

quevon24 (approved):
Everything looks good, this should fix the unexpected keyword argument error.

mlissner (approved):

mlissner:
Thanks for jumping in, Kevin. I added `skip-daemon-deploy` too. I think we should be good. Merging and thank you both!

PR #6088, opened 2025-07-30
fix(merge_opinion_versions): delete objects related to citation annotation

### LINKED ISSUE #5982: Handle `html_with_citations` and related objects for opinion versions
- delete `html_with_citations` for opinions marked as versions
- make the `find_citations` command not execute for versions
- when an opinion is versioned, clean up it's OpinionsCited entries, and other related objects (is there anything to related to ES here?

grossir:
Fixes #5982

- Makes `find_citations` exclude version opinions

- Adds a `delete_version_related_objects` method to deleted Parenthetical, OpinionsCited and UnmatchedCitations objects related to the versioned opinion. It also decreases the OpinionCluster.citation_count of cited clusters, and takes care to recompute ParentheticalGroups where needed

- Updates test to test behaviour of the above function.


quevon24 (approved):
I think everything looks good, versions are correctly excluded in the `find_citations` command, and the new `delete_version_related_objects` method correctly removes objects related to the opinion version (Parenthetical, OpinionsCited, UnmatchedCitations) and correctly updates the citation count using an F expression to increment the count so we don't have a possible race condition.

PR #6100, opened 2025-07-31
feat(search): Adds support for unlimited search alerts 

### LINKED ISSUE #3470: Allow unlimited realtime search alerts via user profile model
We've got a checkbox in the UserProfile model to allow unlimited docket alerts. Let's add a similar one for unlimited real time search alerts. 

Once we add the checkbox, we'll have to add its logic to the front end where people edit alerts, so that it actually takes effect, but I think that code is nicely centralized.

ERosendo:
## Fixes

This PR fixes #3470 

## Summary

- Updates validations in the `CreateAlertForm` class to support users with unlimited search alerts.

- Adds a new hasUnlimitedAlerts flag to the `alerts_context` object.

- Integrates the `hasUnlimitedAlerts` flag into the existing `search-alerts.js` logic 

## Deployment

**This PR should:**

- [ ] `skip-deploy`

- [ ] `skip-web-deploy`

- [ ] `skip-celery-deploy`

- [ ] `skip-cronjob-deploy`

- [x] `skip-daemon-deploy`

### DIFF
--- a/cl/alerts/forms.py
+++ b/cl/alerts/forms.py
@@ -64,22 +64,27 @@ class CreateAlertForm(ModelForm):
         )
         quotas = RECAP_ALERT_QUOTAS[quotas_key]
         is_member = self.user.profile.is_member
+        has_unlimited_alerts = self.user.profile.unlimited_docket_alerts
         if is_member:
             level_key = self.user.membership.level
         else:
             level_key = "free"
 
         flp_membership = "https://donate.free.law/forms/membership"
-        # Only members can create RT alerts
-        if rate == Alert.REAL_TIME and not is_member:
+        # Only members or user with unlimited alerts can create RT alerts
+        if rate == Alert.REAL_TIME and not (is_member or has_unlimited_alerts):
             msg = format_html(
                 "You must be a <a href='{}' target='_blank'>member</a> to create Real Time alerts.",
                 flp_membership,
             )
             raise ValidationError(msg)
 
-        # Only check quotas for RECAP or DOCKETS alerts
-        if self.current_type in {SEARCH_TYPES.RECAP, SEARCH_TYPES.DOCKETS}:
+        # If the user doesn't have unlimited alerts, only check quotas for
+        # RECAP or DOCKETS alert types
+        if not has_unlimited_alerts and self.current_type in {
+            SEARCH_TYPES.RECAP,
+            SEARCH_TYPES.DOCKETS,
+        }:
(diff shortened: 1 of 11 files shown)


albertisfu (changes requested):
Thanks @ERosendo, this is on the right track!

Just a couple of points:

- In addition to allowing the creation of unlimited alerts in the frontend, we also need to allow users who have unlimited alerts set but don't have a membership to actually receive these alerts. For that, we'll need to tweak the following conditions:

https://github.com/freelawproject/courtlistener/blob/6bd89769348d406c3d0644ffb4dde67dd2419d2e/cl/alerts/management/commands/cl_send_alerts.py#L286

https://github.com/freelawproject/courtlistener/blob/6bd89769348d406c3d0644ffb4dde67dd2419d2e/cl/alerts/management/commands/cl_send_recap_alerts.py#L646

https://github.com/freelawproject/courtlistener/blob/6bd89769348d406c3d0644ffb4dde67dd2419d2e/cl/alerts/tasks.py#L671

- And here, to allow users with unlimited alerts to edit them.
https://github.com/freelawproject/courtlistener/blob/6bd89769348d406c3d0644ffb4dde67dd2419d2e/cl/alerts/templates/includes/search_alerts/table_row.html#L44

- Should we add a banner similar to the one we have for unlimited docket alerts (🎉 Your account has unlimited docket alerts) in https://github.com/freelawproject/courtlistener/blob/main/cl/users/templates/profile/alerts.html ?


Finally, I noticed that the original issue #3470 says: 
> Let's add a similar one for unlimited real time search alerts.

Is that still true?

If so, with the logic in this PR, we're allowing users with unlimited alerts set to not only have unlimited RT RECAP Search Alerts but also unlimited daily, weekly, and monthly RECAP Search Alerts. Is that expected?

For Opinions Search Alerts and OA Search Alerts, this is not an issue since we don't check limits for daily, weekly, or monthly alerts for non-members.

mlissner:
> If so, with the logic in this PR, we're allowing users with unlimited alerts set to not only have unlimited RT RECAP Search Alerts but also unlimited daily, weekly, and monthly RECAP Search Alerts. Is that expected?

Yep. This is the freebie to give to very special people (like staff that need it for testing!)

ERosendo:
@albertisfu Thanks for your review. 

This PR is ready for another pass. I've added a new property to the profile model to centralize the eligibility check for users to access RT alerts, and I've also updated all the if statements you mentioned.

PR #6116, opened 2025-07-31
fix(api_serializer): Strip chars from op api

### LINKED ISSUE #6114: UnserializableContentError: Control characters are not supported in XML 1.0
Sentry Issue: [COURTLISTENER-77X](https://freelawproject.sentry.io/issues/5328551394/?referrer=github_integration)

```
UnserializableContentError: Control characters are not supported in XML 1.0
(3 additional frame(s) were not displayed)
...
  File "rest_framework/response.py", line 70, in rendered_content
    ret = renderer.render(self.data, accepted_media_type, context)
  File "rest_framework_xml/renderers.py", line 35, in render
    self._to_xml(xml, data)
  File "rest_framework_xml/renderers.py", line 51, in _to_xml
    self._to_xml(xml, value)
  File "rest_framework_xml/renderers.py", line 59, in _to_xml
    xml.characters(force_str(data))
  File "django/utils/xmlutils.py", line 27, in characters
    raise UnserializableContentError(
```

flooie:
DISALLOWED_XML_CHARS = re.compile(r"[\x00-\x08\x0B\x0C\x0E-\x1F]") Remove these characters from the api
they are invisible and not allowed in
xml

## Fixes
#6114 6114

## Summary

Cleans Opinion API of disallowed invisible characters in xml.

## Deployment

**This PR should:**

- [ ] `skip-deploy`

- [ ] `skip-web-deploy`

- [x] `skip-celery-deploy`

- [x] `skip-cronjob-deploy`

- [x] `skip-daemon-deploy`


grossir (commented):
Looks good; can you add a minimal test, that has the different types? And see that this doesn't break something

- string with the bad character
- ints
- dates
- nested types (lists?)

grossir:
@mlissner I think this is ready for review. Can you review it, should I request a review from someone else on the Case Law team?

I put the cleanup on a custom `Renderer` because the error is triggered in the "rendering" step, not in the "serialization"


Current changes:

fix(OpinionViewset): add custom renderer to handle invalid characters for XML

- Add a class SafeXMLRenderer that inherits from the regular XMLRenderer. It will clean all the text fields on a `UnserializableContentError`
- Save opinion ids that trigger the XML error on a redis set
- Assign the new renderer instead of the regular XMLRenderer on the OpinionViewset
- Add tests

cl/search/api_renderers.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def render(self, data, accepted_media_type=None, renderer_context=None):
+        if data is None:
+            return b""
+
+        try:
+            return super().render(data, accepted_media_type, renderer_context)
+        except UnserializableContentError:
+            # Save problematic IDs to Redis for future correction
+            r = get_redis_interface("STATS")
```
mlissner: Is this the right one to use? I would have guessed CACHE, but I haven't looked at this in a little bit.

quevon24: I think it's fine in "STAT", since what is stored in "CACHE" has the expiration time set

quevon24 (approved):
Everything looks good. I liked that the `clean_xml_data` method handles the possible cases in which it might receive data. I also liked the idea of storing the ids in Redis, since there could still be some characters not allowed by XML and this could help us to identify them.

cl/search/api_renderers.py
grossir: Currently we define these options; I thought STATS was the less wrong option; but perhaps we should name another DB?

https://github.com/freelawproject/courtlistener/blob/eb34ed09b24c8fbddf3f60c1768d70abd6d332a9/cl/settings/third_party/redis.py#L16-L23

mlissner: I think CACHE is where we usually put stuff like this, with STATS reserved for counters of some kind. CACHE doesn't have to have a time limit, does it?

grossir: After looking for any explicit expiration; I found that Redis keys expiration is explicitly set using `r.expire(key, duration)`; so, if we don't do it, the set with the opinion ids won't expire. I have changed the DB to CACHE

PR #6117, opened 2025-07-31
fix(oa_transcripts): prevent MultipleObjectsReturned error when fetching metadata

### LINKED ISSUE #6113: AudioTranscriptionMetadata.MultipleObjectsReturned: get() returned more than one AudioTranscriptionMetadata -- it returned 2!
This looks like a data issue that we're uncovering b/c people are looking at our data more. Can we take a quick look and see what it'll take to fix it? 

----

Sentry Issue: [COURTLISTENER-ABC](https://freelawproject.sentry.io/issues/6782409530/?referrer=github_integration)

```
AudioTranscriptionMetadata.MultipleObjectsReturned: get() returned more than one AudioTranscriptionMetadata -- it returned 2!
(1 additional frame(s) were not displayed)
...
  File "cl/audio/views.py", line 38, in view_audio_file
    metadata_obj = await AudioTranscriptionMetadata.objects.aget(
```

Filed by @mlissner

elisa-a-v:
## Fixes

Fixes #6113

## Summary

This assumes the last `AudioTranscriptionMetadata` is the best one.

I checked both objects related to the audio that triggered the error in the dev DB, and the first one (id 1638) looked like this:

Image

Whereas the last one (id 88337) looks like this:

Image

So I guess last is best?

This fix prevents the error and displays the page properly, however **I could not get that audio to play at all** using the dev DB in my local machine, which I can do for other audios no problem. It seems [the file does not exist](https://dev-com-courtlistener-storage.s3.amazonaws.com/mp3/2024/03/07/vfs_leasing_co._v._markel_insurance_company_cl.mp3).

## Deployment

**This PR should:**

- [ ] `skip-deploy`

- [ ] `skip-web-deploy`

- [x] `skip-celery-deploy`

- [x] `skip-cronjob-deploy`

- [x] `skip-daemon-deploy`

### DIFF
--- a/cl/audio/views.py
+++ b/cl/audio/views.py
@@ -35,9 +35,10 @@ async def view_audio_file(
     )(request, "transcript_feature")
     if transcript_active:
         try:
-            metadata_obj = await AudioTranscriptionMetadata.objects.aget(
+            metadata_qs = AudioTranscriptionMetadata.objects.filter(
                 audio=af
-            )
+            ).order_by("-id")
+            metadata_obj = await metadata_qs.afirst()
             # Extract the 'segments' list instead of 'words'
             segments_list = metadata_obj.metadata.get("segments", [])
             # Validate if segments_list is actually a list
--- a/cl/audio/views.py
+++ b/cl/audio/views.py
@@ -34,19 +34,19 @@ async def view_audio_file(
         waffle.flag_is_active, thread_sensitive=True
     )(request, "transcript_feature")
     if transcript_active:
-        try:
-            metadata_qs = AudioTranscriptionMetadata.objects.filter(
-                audio=af
-            ).order_by("-id")
-            metadata_obj = await metadata_qs.afirst()
+        # Get the latest metadata. There can be many metadata rows for a single
+        # audio, if the transcription failed as hallucinated and was retried
+        metadata_qs = AudioTranscriptionMetadata.objects.filter(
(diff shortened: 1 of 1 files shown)


mlissner:
Hm, is it better to clean up the bad items, if we can figure out how to find them?

elisa-a-v:
Well yeah, that sounds like the actual solution to the problem. I just thought what we needed right now was a fix to lets users see the page properly and read the transcript without seeing an error:

image

Would it make sense to get the case law team to look at the data and figure this one out?

mlissner:
Yeah, that's a good idea. They're very good at the data problems. I think we can wait for that since this isn't happening *too* much. If it works for you, let's close this one and I can ask for their help on the original issue?

elisa-a-v:
Sure, sounds great!

mlissner:
Per the analysis in https://github.com/freelawproject/courtlistener/issues/6123#issuecomment-3145326189, multiple transcripts are expected when there are problems, so grabbing the first one was a good solution. Re-opening, and I'll assign to Gianfranco, in case he has any last minute thoughts.

mlissner:
@grossir, @elisa-a-v is out today. Do you think you could tweak this one to get it landed, if you still have time today?

grossir:
sure @mlissner , I can

grossir (approved):

Issue #5889, opened 2025-07-07
Improve Citation Model: String Volume Support and Table Citation Flag


flooie:
@mlissner 

We should enhance the BaseCitation model to handle real-world citation edge cases more accurately by:
	
1.	Allowing non-numeric values in the volume field (e.g., 1915C, 12-A)
	
2.	Adding a boolean flag to indicate whether a citation is a table-only reference (no full opinion text)

----
1. Volume Field (small int)

We are missing lots of non number volume citations; volume is a SmallIntegerField, which breaks for valid citations like: `Am.Ann.Cas. 1915C, 483`

Eyecite parses this as:

```
FullCaseCitation(
    'Am.Ann.Cas. 1915C, 483',
    groups={'reporter': 'Am.Ann.Cas.', 'volume': '1915C', 'page': '483'},
    ...
)
```

However, our model cannot store this citation as-is due to the volume field constraint. This isn’t just limited to this format — other jurisdictions also include hyphens or alphanumeric values in volumes (e.g., 12-A, Vol. IV).

To address this, I propose changing the volume field to a CharField:

This balances flexibility and performance, while keeping compatibility with short, indexable strings. Sorting and filtering by volume may require casting in certain contexts.

----
2. Flag for Table Citations

Some citations are labeled as “(Table)” and do not reference a full opinion. For example:
`20 N.Y.3d 930, 981 N.E.2d 287, 957 N.Y.S.2d 690 (Table), 2012 WL 6213299`

Being able to identify if a citation is from a `table` indicates it will have duplicate overlapping citations and could help us improve merging and adding citations. 

I propose something like this 

```
is_table_citation = models.BooleanField(
    default=False,
    help_text="True if the citation is from a table-only opinion (no full text)."
)
```

This would also help us find opinions later that should be merged.  if a citation should be unique ie is_table_citaiton=False - we should be able to hone in on duplicates more easily.  

I assume @mlissner this is less controversial than 1. 

I imagine there could be other improvements made to citation, but these are the two main ones I have in my head this morning.  @grossir @quevon24 - any others you can think of?

mlissner:
These don't seem crazy to me, though we'll want to scope out the impact of the citation change and notify our downstream consumers about it. 

For other changes, I usually tag them with the `database` tag. Want to look through it and see what others might apply? https://github.com/freelawproject/courtlistener/issues?q=state%3Aopen%20label%3A%22database%22

quevon24:
[...]

Issue #4844, opened 2024-12-19
Add word count to models


mlissner:
We've occasionally thought it would be useful to have word count for every opinion. A user just asked me for it too. We should do this, with an index, and probably make it searchable too.

I'm on the fence about whether we should add this to the RECAP Documents. Maybe. I could see it being useful when searching documents, but we already have `page_count` in RECAP, so that's probably fine.

We can do this later, next time we do major schema updates.

Issue #4843, opened 2024-12-19
Update Citation Model


flooie:
@mlissner 

I propose we update the citation model to include an optional `page_end` field. This change would provide several benefits:
	1.	Filtering by Substantive Opinions: Users could filter out table citations or focus on opinions based on their length, enhancing usability.
	2.	Data Analysis: It could help identify gaps in the data. While this wouldn’t work in all cases, it could highlight discrepancies in certain situations.

Additionally, I believe we already have much of this data in the Harvard JSON, which we chose not to capture initially because the data was embedded in the page numbers. In hindsight, that decision seems like a missed opportunity.

mlissner:
Seems fine to me, but let's prioritize carefully. When we do this, we should look for other issues with the `database` label and see which we want to include at the same time: 

https://github.com/freelawproject/courtlistener/labels/database

I don't know what you mean by "Table citations" though? Example?

flooie:
I think these are sometimes referred to as table citations.  Citations that appear by reference only and without the actual opinion.  

![Image](https://github.com/user-attachments/assets/3bc10ce0-672f-4671-9f51-4ec339c573e9)

mlissner:
Got it, thanks for the photo! 

So you said that an advantage of having the page length of each opinion is that we'd be able to identify table citations. I think adding the word count would accomplish this as well (see #4844). Similar idea. Worth doing both.

flooie:
could even add a is_valid flag - because we now see revoked citations - or removed citations.

mlissner:
What do you mean?

flooie:
I came across a revoked citation.

mlissner:
What does that mean?

flooie:
This amazing gem - 

Apparently they withdraw citations from the bound volumes which I verified is actually correct.   I assume its in subsequent publications its removed.  I wonder if the advance sheets include removed citations?  

![Image](https://github.com/user-attachments/assets/8cd0bb40-f0ce-40a2-a7fd-df20d9c886f8)

mlissner:
I assume advanced sheets contain the most, and subsequent volumes might remove things here and there. 

In any case, we can add this to our citator instead of the citation model.

PR #6119, opened 2025-08-01
5144 valueerror field volume expected a number but got 1975 2 in find citations and parentheticals for opinion by pks

quevon24:
## Fixes

The impossibility of saving volume numbers with characters. e.g. `71A A.F.T.R.2d (RIA) 3011`

## Summary

- This PR updates the volume field type in BaseCitation model from SmallIntegerField to TextField.
- Tests are updated to use volume string instead of volume integers.
- New tests were added to validate the change.
- resolve_fullcase_citation function updated
- unmatched_citation_is_valid function updated
- get_prev_next_volumes function updated
- API serializer Updated
- API views updated
- Urls updated
- New migrations were generated

## Deployment

**This PR should:**

- [ ] `skip-deploy`

- [ ] `skip-web-deploy`

- [ ] `skip-celery-deploy`

- [x] `skip-cronjob-deploy`

- [x] `skip-daemon-deploy`


cl/alerts/tests/tests.py
```diff
@@ -1728,21 +1728,21 @@ def setUpTestData(cls):
...
             cls.dly_opinion.cluster.panel.add(cls.person_1)
             cls.lexis_citation = CitationWithParentsFactory.create(
-                volume=10,
+                volume="10",
                 reporter="Yeates",
                 page="4",
                 type=Citation.LEXIS,
                 cluster=cls.dly_opinion.cluster,
             )
             cls.neutra_citation = CitationWithParentsFactory.create(
```
grossir, line 1748: not an error from this PR, but it should be `cls.neutral_citation`

cl/citations/unmatched_citations_utils.py
```diff
@@ -36,17 +37,12 @@ def unmatched_citation_is_valid(
...
-        return False
-
     # This would raise a DataError, we have seen cases from bad OCR or
     # citation lookalikes. See #5191
-    if int(groups["volume"]) >= 32_767:
-        return False
+    if match := re.match(r"^(\d+)", groups["volume"]):
+        vol_num = int(match.group(1))
+        if vol_num >= 32_767:
+            return False
```
grossir: I am not sure if this validation is still valid. I  think this was more to prevent uncontrolled exceptions from bad volume values; but those values won't raise errors now that they are text fields.

There is a deeper problem, from the [issue](https://github.com/freelawproject/courtlistener/issues/5191#issue-2892573525); we are finding some very bad volume values, and checking for smallint max value (32_767) won't cover all cases. Ideally we would have a valid range of volume values for each reporter.

So, should we keep this validation that may give a false sense of security, or try to research the problem and solve it in another way? If it's the second option, it is better to leave it as an issue for future development, not for this PR

cl/opinion_page/tests.py
```diff
@@ -834,6 +834,58 @@ async def test_volume_pagination(self) -> None:
         self.assertEqual(volume_previous, None)
         self.assertEqual(volume_next, None)
 
+        # Create new test data
+        await sync_to_async(CitationWithParentsFactory.create)(
+            volume="71",
+            reporter="A.F.T.R.2d (RIA)",
```
grossir, line 840: I have seen a lot of tests with this reporter; but maybe we can also use some of the other non integer volume reporters that follow another format (if they exist, I am not sure), not `"\d+A?"`, to introduce some variety, maybe catch unexpected cases?

cl/citations/unmatched_citations_utils.py
grossir: I just took a dive into this; and I think this guard should be deleted since its a bad solution to a bigger problem

https://github.com/freelawproject/courtlistener/issues/6192

grossir (commented):
I think this PR is mostly ready. 

My main question is: will the new sorting logic work for all the new volume formats we will ingest? I see that most of the examples are from the same reporter, and I wonder if  there are no other volume formats where that may break down

grossir:
Related, what happened with adding `date_created = models.DateTimeField(auto_now=True)` and `date_modified`? I think those fields would be very useful at least for the `search.Citation` model; and doing it now that we will apply a migration will save work to the person that deploys model changes

mlissner:
It's worth looking at the various items that have the `database` tag. Probably you don't want to think about them in this PR, but I don't think we can migrate this field without providing notice to our clients first because it may impact their systems the way it's impacting ours. If that's the case, then it makes sense to batch any other anticipated changes into this PR too. 

And that means that once we've got this PR approved, we will need emails to our customers to explain the change(s) and we'll need to set a schedule for making this change. 

The good news about replication is that it helps a lot of organizations, but the bad news is it makes this kind of thing hard. Sorry!

quevon24:
@mlissner I have already reviewed all the issues with the label database and we would need to discuss whether to include these changes:

- word_count field https://github.com/freelawproject/courtlistener/issues/4844
- page_end field https://github.com/freelawproject/courtlistener/issues/4843
- is_table_citation field https://github.com/freelawproject/courtlistener/issues/5889

Regarding citations with nominative reporters, there are ways to implement it to save those citations without having to make changes to the model fields (modify the model save method to add a 1 as volume, adjust __str__ method, create a list with the known nominative reporters to avoid having to make changes to reporters-db, etc.)

@flooie @grossir What do you think?

mlissner:
I'm not hearing any push for any of those three changes, so I think we should only do them if they're important to the case law team.

grossir (approved):
Looks good! However, it seems this will need to be split in 2 PRs for client deployment (check the email @quevon24 , but let's wait for confirmation anyway)
- 1 PR with volume changes
- 1 PR with date changes

cl/opinion_page/urls.py
```diff
@@ -133,12 +133,12 @@
         name="citation_homepage",
     ),
     re_path(
-        r"^c/(?:(?P<reporter>.*)/(?P<volume>\d{1,10})/(?P<page>.*)/)?$",
+        r"^c/(?:(?P<reporter>.*)/(?P<volume>.*)/(?P<page>.*)/)?$",
```
flooie, line 136: is this fine?  do we need to have something stricter?

cl/search/migrations/0044_alter_citation_volume_alter_citationevent_volume.sql
```diff
@@ -0,0 +1,10 @@
+BEGIN;
+--
+-- Alter field volume on citation
+--
+ALTER TABLE "search_citation" ALTER COLUMN "volume" TYPE text USING "volume"::text;
```
grossir: It seems this statement will force an index rebuild; which may cause server problems?

Perhaps a better way to do it will be 
- `DROP INDEX` for indexes affected by the data type change
- execute data type change `ALTER TABLE`
- `CREATE INDEX CONCURRENTLY` 
- `ANALYZE` the affected tables
-------


From the Postgres [docs](https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-DESC-SET-DATA-TYPE)

> When this form is used, the column's statistics are removed, so running [ANALYZE](https://www.postgresql.org/docs/current/sql-analyze.html) on the table afterwards is recommended. For a virtual generated column, ANALYZE is not necessary because such columns never have statistics.

From the Notes Section in the same docs

> Changing the type of an existing column will normally cause the entire table and its **indexes to be rewritten**. As an exception, when changing the type of an existing column, if the USING clause does not change the column contents and the old type is either binary coercible to the new type or an unconstrained domain over the new type, a table rewrite is not needed. However, indexes will still be rebuilt unless the system can verify that the new index would be logically equivalent to the existing one. For example, if the collation for a column has been changed, an index rebuild is required because the new sort order might be different. However, in the absence of a collation change, a column can be changed from text to varchar (or vice versa) without rebuilding the indexes because these data types sort identically.

cl/search/models.py
```diff
@@ -3000,7 +3000,7 @@ class BaseCitation(models.Model):
             "72 Soc.Sec.Rep.Serv. 318)",
         ),
     )
-    volume = models.SmallIntegerField(help_text="The volume of the reporter")
+    volume = models.TextField(help_text="The volume of the reporter")
```
grossir, line 3003: now that this will be a text field; it should have `null = False, default = ''` ? 

`null = False` to prevent double "empty" values: `null` and `''`

https://docs.djangoproject.com/en/5.2/ref/models/fields/#null

quevon24: Good point, but in this case, null=False is already the default for TextField, so it’s redundant to include it explicitly.
I also didn’t add a default='' because this field should always have a valid value (for citations with nominative reporters we can set it to 1, this will also help to display them in the citation lookup page, when selecting a nominative reporter because it uses volume numbers for the links), an empty string wouldn’t be meaningful for volume.

cl/search/migrations/0044_alter_citation_volume_alter_citationevent_volume.sql
quevon24: Good point, since both indexes and the unique constraint include volume, the ALTER TABLE will trigger a full table and index rewrite, which could block access given the table’s size.

Django relies on Postgres to rebuild those automatically, but that happens synchronously. To avoid long locks, we can do it manually updating the sql migration like this:

```
-- 
-- Drop indexes and unique constraint
--
DROP INDEX CONCURRENTLY IF EXISTS search_citation_volume_ae340b5b02e8912_idx;
DROP INDEX CONCURRENTLY IF EXISTS search_citation_volume_251bc1d270a8abee_idx;
ALTER TABLE search_citation DROP CONSTRAINT IF EXISTS search_citation_cluster_id_7a668830aad411f5_uniq;
--
-- Alter field volume on citation
--
ALTER TABLE "search_citation" ALTER COLUMN "volume" TYPE text USING "volume"::text;
--
-- Recreate constraint and indexes concurrently
--
ALTER TABLE search_citation ADD CONSTRAINT search_citation_cluster_id_7a668830aad411f5_uniq
    UNIQUE (cluster_id, volume, reporter, page);
CREATE INDEX CONCURRENTLY search_citation_volume_ae340b5b02e8912_idx
    ON search_citation (volume, reporter, page);
CREATE INDEX CONCURRENTLY search_citation_volume_251bc1d270a8abee_idx
    ON search_citation (volume, reporter);
--
-- Refresh planner statistics
--
ANALYZE search_citation;
```

And something similar for unmatchedcitation migration:

```
-- 
-- Drop indexes and unique constraint
--
DROP INDEX CONCURRENTLY IF EXISTS citations_u_volume_da4d25_idx;
ALTER TABLE citations_unmatchedcitation DROP CONSTRAINT IF EXISTS citations_unmatchedcitat_citing_opinion_id_volume_ca9f46d3_uniq;
--
-- Alter field volume on unmatchedcitation
--
ALTER TABLE citations_unmatchedcitation ALTER COLUMN volume TYPE text USING volume::text;
--
-- Recreate constraint and indexes concurrently
--
ALTER TABLE citations_unmatchedcitation
    ADD CONSTRAINT citations_unmatchedcitat_citing_opinion_id_volume_ca9f46d3_uniq
    UNIQUE (citing_opinion_id, volume, reporter, page);
CREATE INDEX CONCURRENTLY citations_u_volume_da4d25_idx
    ON citations_unmatchedcitation (volume, reporter, page);
--
-- Refresh planner statistics
--
ANALYZE citations_unmatchedcitation;
```

Or simply run it in parts to avoid updating the sql migration. What do you think?

grossir: > to avoid updating the sql migration

I think all that we want to happen should be in the SQL; perhaps what we can leave out is the ANALYZE?


Also, the .py migrations should be modified too, since it will be run when using `migrate`?

cl/citations/migrations/0002_alter_unmatchedcitation_volume.py
```diff
@@ -0,0 +1,55 @@
...
+class Migration(migrations.Migration):
+    atomic = False
+
+    dependencies = [
+        ('citations', '0001_add_unmatched_citation'),
+    ]
+
+    operations = [
+        # Drop indexes and unique constraint
+        migrations.RunSQL(
```
grossir: Any reason not to use the special operators. I think those come with their reverse actions already implemented

`from django.contrib.postgres.operations import AddIndexConcurrently, RemoveIndexConcurrently`

grossir (commented):
SQL migrations look OK. They match what is generated by `sqlmigrate`

For Python migration files I have an observation below


```
docker exec -it cl-django python /opt/courtlistener/manage.py sqlmigrate citations 0002

docker exec -it cl-django python /opt/courtlistener/manage.py sqlmigrate search 0044
```

quevon24:
We refactored the volume field migration into three separate migrations to safely handle triggers and concurrent indexes. First, we remove the old indexes and constraints non-atomically, then we rename and alter the volume column inside an atomic transaction, and finally we recreate the indexes non-atomically. This approach avoids issues with triggers depending on the column and ensures that data is preserved during the migration. Given the table sizes, the batch updates and concurrent index operations are safe and won't cause downtime.

For the smaller unmatchedcitation table, the migration is simpler and in a separate app. Since it has fewer rows and no triggers, we can safely update the volume field type and handle indexes and constraints in a single migration without splitting into multiple steps.

cl/search/migrations/0044_remove_citation_indexes_customers.sql
ERosendo, line 1: I noticed this migration file has the same content as `0044_remove_citation_indexes.sql`. In this case, we should include only one .sql file since they’re identical.

cl/citations/migrations/0002_alter_unmatchedcitation_volume_customers.sql
ERosendo, line 1: The comment at https://github.com/freelawproject/courtlistener/pull/6119/files#r2452162974 applies here as well. The `.sql` migration files look identical (correct me if I’m wrong), and in those cases we usually just include one of them.

ERosendo (changes requested):
@quevon24 Thanks for refactoring the migrations. I've added two small comments on the `.sql` migration scripts. Once those comments are addressed, I think this is ready to merge.

cl/search/migrations/0044_remove_citation_indexes_customers.sql
quevon24: Thanks, that's exactly what I was wondering a few days ago, but I couldn't find the answer in the wiki for this case.

quevon24:
Thanks @ERosendo for all the guidance in this process. The changes are now ready. I also saw that sql from 0046 migration for replica and customers are the same.

ERosendo (approved):
Thanks @quevon24 LGTM

PR #6125, opened 2025-08-01
Remove style attrs and tweak words

mlissner:
## Summary

#5771 added a cool new dropdown to the attachment page, but:
 - It uses style attrs, which we usually avoid (though they don't matter *so* much anymore).
 - The size of the text was really big
 - The words could use a bit of cleanup to align with norms

This PR fixes these three little things by using a CSS class we already had, removing the style tags, and tweaking the language a bit.

## Deployment

- [x] `skip-celery-deploy`

- [x] `skip-cronjob-deploy`

- [x] `skip-daemon-deploy`

<details closed>
<details open>
<summary><h4>Before</h4></summary>
image

</details>
<details open>
<summary><h4>After</h4></summary>
image
</details>

### DIFF
--- a/cl/opinion_page/templates/includes/rd_metadata_headers.html
+++ b/cl/opinion_page/templates/includes/rd_metadata_headers.html
@@ -1,47 +1,47 @@
 {% load tz %}{% load waffle_tags %}
-<div class="btn-group" style="display:block;">
-  <h3 style="margin-top: 10px;">
-    {% if rd.description %}{{ rd.description }}
-      <span class="gray">&mdash;</span>
-    {% endif %}
-    Document
-    <a href="{{ rd.docket_entry.docket.get_absolute_url }}?entry_gte={{ rd.document_number }}#entry-{{ rd.document_number }}">
-      #{{ rd.document_number }}</a>{% if rd.document_type == rd.ATTACHMENT %},
-      Attachment #{{ rd.attachment_number }}{% endif %}
+<h3 style="margin-top: 10px; margin-left:0px;">
+  {% if rd.description %}{{ rd.description }}
+    <span class="gray">&mdash;</span>
+  {% endif %}
+  Document
+  <a href="{{ rd.docket_entry.docket.get_absolute_url }}?entry_gte={{ rd.document_number }}#entry-{{ rd.document_number }}">
+    #{{ rd.document_number }}</a>{% if rd.document_type == rd.ATTACHMENT %},
+    Attachment #{{ rd.attachment_number }}{% endif %}
     {% if attachments %}
-      <button
-        class="btn btn-default btn-lg dropdown-toggle"
-        type="button"
-        data-toggle="dropdown"
-        aria-expanded="false"
-        style="vertical-align:bottom; padding:1px 10px; border-radius:3px;"
-      >
-        <span class="caret"></span>
-      </button>
+      <div class="btn-group">
+        <button
+          class="btn btn-default btn-lg dropdown-toggle"
+          type="button"
+          data-toggle="dropdown"
+          aria-expanded="false"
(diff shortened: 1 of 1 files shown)


cl/opinion_page/templates/includes/rd_metadata_headers.html
```diff
@@ -1,48 +1,47 @@
...
-      <button
-        class="btn btn-default btn-lg dropdown-toggle"
-        type="button"
-        data-toggle="dropdown"
-        aria-expanded="false"
-        style="vertical-align:bottom; padding:1px 10px; border-radius:3px;"
-      >
-        <span class="caret"></span>
-      </button>
+      <div class="btn-group">
```
mlissner, line 11: This diff looks big, but it's mostly indentation caused by moving the div on line 2 to line 11.

albertisfu (approved):
Changes look good!

I found another issue which I know is unrelated to the fixes this PR applies, but I wanted to flag it.

On mobile, sometimes the dropdown aligned to the left falls outside the viewport:

Image

Even on desktop, when the short description is too long, the dropdown can extend beyond the main layout:

Image


Perhaps we can limit the short description’s length?

I think we can merge these changes if you agree, and tackle the other issues in a separate issue.

florean:
> On mobile, sometimes the dropdown aligned to the left falls outside the viewport:

Yeah, that's why I left aligned it with the description and rest of the headers.  It would still blow out the viewport, but to the right so you just lost some of the description.  Although, I thought I had limited the width to the viewport max.  🤔 Maybe I didn't commit it?  I'll take a look.

mlissner:
Yeah, let's merge. There's a CSS rule for doing ellipsis automatically when you run out of space. Maybe it's also possible to wrap the text?

cl/opinion_page/templates/includes/rd_metadata_headers.html
```diff
@@ -1,48 +1,47 @@
...
+          {% for doc in attachments %}
+            <li {% if doc.attachment_number == rd.attachment_number or not doc.url %}class="disabled"{% endif %}>
+              {% if not doc.url %}
+                {{ doc.description }}
+              {% else %}
+                <a
+                  href="{% if doc.attachment_number != rd.attachment_number %}{{ doc.url }}{% else %}#{% endif %}"
+                  title="{% if doc.description %}{{ doc.description }}{% elif doc.attachment_number%}Attachment #{{ doc.attachment_number }}{% else %}Document #{{ rd.document_number }}{% endif %}"
+                >
+                  {% if doc.attachment_number %}Att. {{ doc.attachment_number }}{% else %}Entry {{ rd.document_number }}{% endif %}
```
florean, line 32: Is there a reason to use "Entry" rather than "Doc."?  I think of a docket entry as encompassing the main document and all attachments, so it seems odd.  If I didn't know, would it be unreasonable to think that might take me back to the docket entry?  And while "Attachment" gets abbreviated to "Att.", "Document" becomes "Entry", which is not what I would expect.

florean:
Also wanted to mention my best test documents:
[Long descriptions](https://www.courtlistener.com/docket/17403874/18/5/stadtmauer-v-nordlicht/)
[213 documents](https://www.courtlistener.com/docket/17366820/207/2/akai-custom-guns-llc-v-kkm-precision-inc/)

cl/opinion_page/templates/includes/rd_metadata_headers.html
mlissner: I've found that Entry is usually the word folks use to refer to these. They say "Entry 22" or "Entry 22-2" etc. Saying "Document 22" is something we've done for a while, but I think it's not what folks use in practice.

PR #6136, opened 2025-08-06
feat(scrapers): create OriginatingCourtInformation objects from scrapers

### LINKED ISSUE #5453: Create OriginatingCourtInformation objects from scraped data
See parent issue for description

grossir:
Fixes #5453

Use juriscraper values to populate OriginatingCourtInformation objects
- "lower_court_id" goes to Docket.appeal_from_id
- "lower_court_numbers" goes to OriginatingCourtInformation.docket_number
- "lower_court_judges" goes to OriginatingCourtInformation.assigned_to_str
- Update tests to check creation of OriginatingCourtInformation and confirm updates if they exist


cl/corpus_importer/management/commands/import_tn.py
```diff
@@ -120,7 +120,7 @@ def import_tn_corpus(
                 op.cluster.case_name,
             )
 
-        docket, opinion, cluster, citations = make_objects(
+        docket, opinion, cluster, citations, _ = make_objects(
```
Luis-manzur: Recommendation:

Add a name to the value just for context `originating_court_info` or add a comment if feeling is to long, so others can understand the value without going into the func if needed in the future.

cl/scrapers/utils.py
```diff
@@ -597,3 +615,35 @@ def check_duplicate_ingestion(local_path_name: str) -> None:
...
+def update_or_create_originating_court_information(
+    docket: Docket, lower_court_number: str, lower_court_judge: str
+) -> OriginatingCourtInformation | None:
+    """
+
+    :param docket:
+    :param lower_court_number:
+    :param lower_court_judge:
+    """
+    if not (lower_court_judge or lower_court_number):
```
Luis-manzur, line 629: this may be confusing because arguments are not optional, so leave a comment saying that you're validating "" empty strings or change the definition to optional.

Luis-manzur: At first glance it seems like you are doing unnecessary validation because you are validating if there are some arguments that are required in the function.

grossir: Even if the args are required for this func, they come from a process that may not find them, that's why we do the check

the way is done, it is standard validation to check if there is any non falsey value ("" or None)

If both of them are, then we don't have to do anything else and should return early

I think a comment is too verbose for something standard / obvious

grossir: I did miss the docstring
```

    :param lower_court_number:
    :param lower_court_judge:
```

Luis-manzur (approved):
done

PR #6143, opened 2025-08-07
fix(scrapers.utils): fix `appeal_from_id` logger.error for empty strings

grossir:
Solves a bug introduced in #6136

The `logger.error` is triggered if the `Court.id` does not exist; but it's not checking if the "appeal_from_id" value is a non empty string; so, it is triggering for empty strings creating many false positives


grossir:
Sentry Issue: [COURTLISTENER-AGQ](https://freelawproject.sentry.io/issues/6796989833/?referrer=github_integration)

```
Docket.appeal_from_id has non existing Court.id '' as value
```

PR #6157, opened 2025-08-09
feat: New bulk data pipeline see discussions/6131

devlux76:
This pull request introduces a new, modular Bash-based bulk data export pipeline for database tables as discussed in discussions/6131. The pipeline is split into distinct phases, each managed by its own script, and orchestrated by a master script. Major improvements include robust argument parsing, environment validation, parallelized data export, and clear separation of schema and data handling. The design emphasizes reliability, maintainability, and performance, with strong logging and error handling throughout.

**Pipeline orchestration and environment setup:**

* Added the master orchestration script `bulk-data-pipeline.sh` to coordinate all phases, including argument parsing, environment setup, preflight checks, metadata generation, and execution of each pipeline phase.
* Implemented comprehensive preflight checks for environment variables, required binaries, database connectivity, and output location, ensuring reliable execution.

**Schema export and import script generation:**

* Added `001-export-schema.sh` to export the database schema in split files (`pre-data` and `post-data`), with support for S3 uploads and version compatibility checks between client and server. This will ensure that imports no longer fail on foreign key constraints.
* 
* Added `002-generate-import-script.sh` to generate a bulk import script using a template, with robust variable substitution, table definition generation, and output validation.

**Parallelized bulk data export:**

* Added `003-bulk-data-export.sh` to export table data in parallel, sorting tables by size for optimal resource usage, and compressing outputs with `lbzip2`. Includes S3 upload support (Note: I am unable to test S3) and detailed logging.
* The export phase calculates conservative parallel worker counts based on available CPU cores, and includes safety checks to prevent writing to system directories.

**Metadata and configuration management:**

* Metadata about tables (names, filenames, sizes) is generated and stored for use in subsequent phases, ensuring consistent and accurate processing across the pipeline. Tables to export and the file names they are exported to are all contained in a .csv so tables may be added or removed without impacting the code.

These changes together implement a maintainable bulk data export pipeline, with modular scripts for each phase and strong operational safeguards.

### DIFF
--- a/scripts/003-bulk-data-export.sh
+++ b/scripts/003-bulk-data-export.sh
@@ -156,7 +156,7 @@ export_worker() {
     local filename=$(echo "$line" | cut -d'|' -f3)
 
     # Debug output
-    echo "DEBUG: Processing table: $table_name, filename: $filename" >&2
+    log "DEBUG" "Processing table: $table_name, filename: $filename"
 
     export_table "$table_name" "$filename" "$EXPORT_DIR" "$LBZIP2_THREADS"
 }


devlux76:
Just a note, even though this focuses on parallelization, you can limit it to a single thread of execution using --workers=1  conversely you can use all available cores (defaults to 80%) for work by specifying --workers=-1 and you can set any number that you wish, although my testing shows 80% or -1 is the sweet spot here.

The import script will run a fixed number of jobs specified by --workers as well. But it uses lbzip2 for decompression of the .bz2 files and will use 80% of total cores to decompress each file much faster. Everything else about workers remains the same as described above.

The net result is an export on my 2018 Macbook Pro (32 GB of RAM and 12 cores) takes around 3 hours while an import is around 4 hours.

I have not been able to rigorously test this because the total amount of files (bz2, csv, etc) exceeds my available disk space. However I have been able to manually clean up files mid flight and get it to work.  It is doubtful anyone other than a tester/debugger will need to have all stages on disk at the same time. The total compressed size is identical to the existing path (around 80GB). While the total uncompressed size is around 400GB, plus importing it into postgres needs an additional 400GB to work with although the final postgres datastore comes in at 284GB due to the way postgres stores data. 

This is not new though, it's the exact same figures with the existing make_bulk_data path.

So in a nutshell, this change speeds up the bulk data process across the board. It decreases actual real world time by about a factor of 10 on both the export and import while making the tables to files mapping easier to maintain. Additionally it allows for easier debugging if something does go wrong.

mlissner:
Thanks for pulling this together! @flooie, I think Kevin has been the main person working on the export script. Do you think he might have some time to evaluate this? We'll want to think through things like if it suits our needs, and if we think we can maintain it — things like that, and then do the standard PR review, if we like it in general. It's a big change and a lot of bash, but seems like it's probably helpful?

PR #6170, opened 2025-08-13
feat(profile): Add profile page with user's CL ID

### LINKED ISSUE #5924: Add profile page with user's CL ID
This is a fun and easy one. We just need a page at `/profile/id/` that looks like a plain-jane profile page and contains:

 - The user's ID in big text
 - A copy text button to the right of it

Some explanation that says, "Your CourtListener user ID is..." or something like that that makes sense.

We'll ask users to copy this value into Neon to make the mapping between Neon and CL stronger.

ERosendo:
## Fixes

This PR fixes #5924 

## Summary

This PR introduces a new input field to the form in the profile page that displays the CourtListener ID as a read-only text box. A copy button is included next to the input, allowing users to easily copy the ID to their clipboard. The button visually indicates success by switching icons and text upon copying.

Here's a GIF showing the new field and button in action: 

Screen Recording 2025-08-13 at 1 01 57 AM

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


mlissner:
Sorry to be a bit picky about this, but I don't think we want to have it on all the profile pages, since it will just confuse people. Can we do it like the API key page: 

image

In other words, just a page that says, "Your CourtListener ID" and then provides it in big font (no clipboard or anything). It doesn't need to have the navbar from the other profile pages or anything like that. Just a really plain, boring, simple page with a simple URL? 

It'll be linked from Neon, but nowhere else, so it just needs to say the number and that's it. Too simple?

ERosendo:
@mlissner I’ve updated this PR. It now adds a new template and view to display the user ID. The new URL is `/profile/id/`, and it looks like this:

image

mlissner (approved):
Yeah, that's the stuff, Eduardo. Thanks!

PR #6164, opened 2025-08-12
feat(delete_duplicates): new command

grossir:
Fixes #4376

- Add a new command to delete same hash duplicates
- refactor `merge_metadata` to introduce a strict mode that raises an error if there is an unexpected difference
- add a new exception `MergingError`
- refactor `Opinion.joined_by` merging; to prevent false positive "unexpected difference" warnings and properly merge objects
- add tests for `Opinion.joined_by` merging
- add tests for new command

PR #6171, opened 2025-08-13
fix(delete_duplicates): handle versioned cluster

grossir:
Fixes bug not accounted for in #6164

Takes into account the case when duplicated same hash opinions are already grouped in a single cluster due to versioning. In that case, we can't delete the cluster, since doing that would delete all the opinions already grouped together, even then ones we want to keep

Labelling this as `skip-*` because 
- I will run it in my maintenace pod which I update manually
- doesn't affect any code used by any of the automated processes


quevon24 (approved):
The changes look good.

PR #6173, opened 2025-08-13
feat(new_homepage): introduce product stats

### LINKED ISSUE #5362: Home Page - Product Stats
Image

**Refer to the Figma for the most updated design.**

On the CourtListener home page, we will share product statistics.

The general idea is:

- Seven statistics
- Advertise RECAP, Case law, oral args and system usage
- The large spot on the left is for Case Law (spot 1)
- Next three for PACER, PACER, and Oral args (spots 2 - 4)
- Bottom row for system usage (spots 5 - 7)

Image

**REQUIREMENTS**

The boxes should use the following language:

1. 10.1M Number of precedential decisions in our system
2. 508M Number of federal cases in the RECAP Archive
3. 60M Number of federal filings in the RECAP Archive
4. 2,139 Days of oral argument recordings in our system
5. 182,357 Number of searches conducted in the last ten days
6. 55,974 Number of alert emails sent over the last ten days
7. 2.6M Number of API calls made in the last ten days

How to access the data:

1. Number of precedential opinions comes from the opinions index in Elastic, with an approximate count.
2. Number of federal cases comes from Elastic with an approximate query on the docket index.
3. Same, but the documents index.
4. Oral arg days we've already got on the CourtListener homepage.
5. Searches come from the SearchQuery table.
6. Alert emails we've got on the CourtListener homepage.
7. API calls we've got on the CourtListener homepage.

**Designs needed:**

- [ ] Mobile

### LINKED ISSUE #5857: Minor fixes in new header
- The gap between the donate button and its dropdown menu in the homepage header is too big, it should be 8 px.

Image

- The hover style in the scope button in desktop clashes with the input color. We should either remove it or change it to something else.

Image

- The focus style in the search input should be the same shade of red as the search button (read more about it [here](https://github.com/freelawproject/courtlistener/issues/5795#issuecomment-3005741094))

elisa-a-v:
## Fixes

Closes #5362 
Fixes #5857 

## Summary

This PR:
- Adds the product stats in the new homepage.
- Introduces a new router view for the homepage.
- Introduces a new `utils.py` file in the `search` app with the methods required to get stats for both homepage versions.
- Tweaks some header styles (See [commit](https://github.com/freelawproject/courtlistener/commit/ccc05c257dd853f2ffcb39f5497747174a05521b) and #5857)

### About the new router view

- The context required in the new homepage differs from the one required in the live homepage.
- The view that handles the live homepage (`show_results`) is also responsible for other pages, like the search results, as well as alert editing.

Instead of tweaking `show_results` to handle even more stuff, we introduce a new view just for the new homepage, and a new router view that determines which view to use for a given request. This way `show_results` is only in charge of old templates, and new templates get their own view to get the exact context they need, and nothing more.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

<details closed>
<details open>
<summary><h4>Desktop large</h4></summary>

image

</details>

<details open>
<summary><h4>Desktop small</h4></summary>

image

</details>

<details open>
<summary><h4>Mobile large</h4></summary>

image

</details>

<details open>
<summary><h4>Mobile small</h4></summary>

image

</details>

<details open>
<summary><h4>Profile button hover before / after</h4></summary>

image

image

</details>

<details open>
<summary><h4>Scope selector hover and dropdown before / after</h4></summary>

image

image

</details>
<details open>
<summary><h4>Donate button dropdown before / after</h4></summary>

image

image

</details>
</details>

### DIFF
--- a/cl/citations/management/commands/count_citations.py
+++ b/cl/citations/management/commands/count_citations.py
@@ -1,8 +1,10 @@
+import time
+
 from django.db import transaction
 from django.db.models import Count, IntegerField, OuterRef, Subquery, Value
 from django.db.models.functions import Coalesce
 
-from cl.lib.command_utils import VerboseCommand
+from cl.lib.command_utils import VerboseCommand, logger
 from cl.search.models import OpinionCluster, OpinionsCited
 
 
@@ -20,7 +22,28 @@ class Command(VerboseCommand):
             "--count-from-opinions-cited",
             action="store_true",
             default=False,
-            help="Flag to compute the citation_count for all clusters from the OpinionsCited table",
+            help="Flag to compute the citation_count from the OpinionsCited table",
+        )
+        parser.add_argument(
+            "--start-cluster-id",
+            type=int,
+            nargs="*",
+            default=None,
+            help="An OpinionCluster.id to start the batch updates from",
+        )
+        parser.add_argument(
+            "--end-cluster-id",
+            type=int,
+            default=None,
+            nargs="*",
+            help="An OpinionCluster.id where the batch updates end",
+        )
+        parser.add_argument(
+            "--step-size",
+            type=int,
+            default=10_000,
+            nargs="*",
+            help="Number of ids to consider in a batch update",
         )
 
     def handle(self, *args, **options):
@@ -46,6 +69,38 @@ class Command(VerboseCommand):
         if not options.get("count_from_opinions_cited"):
             return
 
+        if not (
+            options.get("start_cluster_id") and options.get("end_cluster_id")
+        ):
+            raise ValueError(
+                "Must pass values for start-cluster-id and end-cluster-id"
+            )
+
+        # In 2025, we have a little more than 10.1M clusters
+        step_size = options["step_size"]
+        for start_id in range(
+            options.get("start_cluster_id"),
+            options.get("end_cluster_id"),
+            step_size,
+        ):
+            end_id = start_id + step_size
+
+            logger.info(
+                "Updating citation_count for clusters with ids between %s and %s",
+                start_id,
+                end_id,
+            )
+            self.update_cluster_citation_count_from_opinions_cited(
+                options["start_cluster_id"], options["end_cluster_id"]
+            )
+            logger.info("Finished citation_count update")
+
+            # Give some time for anything waiting for the clusters' locks
+            time.sleep(10)
+
+    def update_cluster_citation_count_from_opinions_cited(
+        self, start_cluster_id: int, end_cluster_id: int
+    ) -> None:
         # Group by OpinionCluster.id and count all the OpinionsCited rows for
         # all of its subopinions
         count_by_cluster_subquery = Subquery(
@@ -53,7 +108,9 @@ class Command(VerboseCommand):
                 # 'pk' refers to the OpinionCluster.id of the row being updated
                 # OuterRef joins to a key from the query enveloping the Subquery,
                 # in this case, the update statement below
-                cited_opinion__cluster_id=OuterRef("pk")
+                cited_opinion__cluster_id=OuterRef("pk"),
+                cited_opinion__cluster_id__gte=start_cluster_id,
+                cited_opinion__cluster_id__lt=end_cluster_id,
             )
             # Group by the cluster ID
             .values("cited_opinion__cluster_id")
@@ -65,7 +122,9 @@ class Command(VerboseCommand):
         )
 
         with transaction.atomic():
-            OpinionCluster.objects.update(
+            OpinionCluster.objects.filter(
+                id__gte=start_cluster_id, id__lt=end_cluster_id
+            ).update(
                 citation_count=Coalesce(
                     count_by_cluster_subquery,
                     Value(0),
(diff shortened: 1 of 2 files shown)


elisa-a-v:
This is what it looks in an iPhone mini now. I think it works, what do you think?

image

[Screencast from 08-13-2025 06:01:48 PM.webm](https://github.com/user-attachments/assets/2dc494b5-4bfc-49a3-a4aa-a212c7e9b04a)

mlissner:
Looks great! Approach looks good too. Onward to Eduardo for review. Thank you!

mlissner (approved):

cl/search/templates/v2_homepage.html
```diff
@@ -205,5 +205,67 @@ <h1 class="w-full text-center lg:text-display-sm-cooper text-xl" x-text="tagline
...
+  >
+    <h2 id="stats-title" class="sr-only">CourtListener Statistics</h2>
+    <dl class="flex flex-col md:gap-5 gap-2.5 md:flex-row">
+      <div class="rounded-[20px] border border-greyscale-200 py-7 md:px-8 px-6 md:gap-6 gap-3 flex flex-col md:max-w-96">
+        {% svg "page_search" aria_hidden="true" class="text-greyscale-500" %}
+        <dt class="order-2 text-md text-greyscale-600">Number of precedential decisions in our system</dt>
+        <dd class="order-1">
+          <data class="md:text-display-md text-display-sm font-cooper font-semibold" value="{{ opinion_count }}">{{ opinion_count|intword }}</data>
+        </dd>
+      </div>
```
ERosendo, line 222: This element looks different in the Figma mockup. In the mockup, the number is placed **after** the description. Here’s a screenshot for reference:

image


In this PR, however, the number is displayed **before** the description. Here’s a screenshot of the section as rendered in Chrome:

image

I’m not sure if we intentionally changed the layout for consistency, but this change isn’t mentioned in the issue or the PR, so I’m flagging it for visibility.

elisa-a-v: Yes this was intentional, it was actually changed in the very last commit, but I forgot to mention in the comment 😅 We talked about this with @mlissner and decided to change it for consistency and because it looked slightly off in smaller screens.

Oh and something else I failed to mention, we also decided it was fine to keep the full word "million" instead of the bare "M". It was trivial code-wise to just use the full word, and it's more explicit and therefore clearer for users.

ERosendo: Thanks @elisa-a-v!

cl/search/utils.py
```diff
@@ -0,0 +1,114 @@
...
+
+    homepage_stats = {
+        "alerts_in_last_ten": alerts_in_last_ten,
+        "queries_in_last_ten": queries_in_last_ten,
+        "api_in_last_ten": api_in_last_ten,
+        "minutes_of_oa": minutes_of_oa,
+        "opinion_count": get_total_estimate_count("search_opinion"),
+        "docket_count": get_total_estimate_count("search_docket"),
+        "recap_doc_count": get_total_estimate_count("search_recapdocument"),
+        "private": False,  # VERY IMPORTANT!
```
ERosendo: This isn’t exactly an issue, but I feel this key (`private`) isn’t really part of the stats. it’s part of the template’s context. It also seems to suggest that the stats are private, and it doesn’t quite align with the rest of the dictionary, which mostly contains integers representing counts.

elisa-a-v: > this key isn’t really part of the stats. it’s more part of the template’s context.

> it doesn’t quite align with the rest of the dictionary, which mostly contains integers representing counts.

Yeah you're right. I wondered about it for a bit but didn't give it much thought and just left it there for consistency with the previous method, but I can move it to the view itself instead, which I agree makes more sense.

Bottomline, that key is not a stat, so it shouldn't be in the stats dict!

Do you think it would be worth it to also refactor the legacy view? Or would that be out of scope and maybe not worth it?

ERosendo: I don’t think it would be worth refactoring if we’re planning to remove it in the future. it’d be better to focus our effort on the parts we intend to keep.

elisa-a-v: Nice, that was my intuition too, thank you. In that case this is ready once again :)

elisa-a-v:
Thank you for your feedback! The comment has been addressed 🫡

elisa-a-v:
I'm not sure how that ApiError could be related to this PR, but I see Sentry auto-assigned the issue to @albertisfu and flagged [this commit](https://github.com/freelawproject/courtlistener/commit/e62a60059915924f3aedd3019dfc5747f763d547) as suspect, so I'm tagging in case this context is indeed relevant.

ERosendo:
@elisa-a-v I don’t think that error is directly caused by the changes in this PR. 🤔

albertisfu:
Yeah, that error is not related to the changes in this PR. I think it was linked because the PR referenced the `show_results` method, but that’s an old issue https://github.com/freelawproject/courtlistener/issues/3660  related to users using invalid search syntax.

elisa-a-v:
Oh okay, yeah I figured it was unrelated to this, but I thought it was best to make sure anyway. That all makes sense, thank you both! :)

PR #6174, opened 2025-08-13
fix(count_citations): citation_count update is now done in batches

grossir:
Solves #5706

OpinionCluster.citation_count can now be updated in batches of consecutive ids. This should prevent locking the whole table during the update

Add arguments to control behavior of batches
- start-cluster-id
- end-cluster-id
- step-size; default 10_000

Update tests

-------

Labelling this as `skip-*` because

- we intend to run it on the maintenace pods which we update manually
- doesn't affect any code used by the automated processes

--------------

The query now looks like below, with `start = 0` and `end = 10_000`

```sql
UPDATE "search_opinioncluster" 
SET "citation_count" = COALESCE(
    (SELECT COUNT(U0."id") AS "total_count" 
    FROM "search_opinionscited" U0 
    INNER JOIN "search_opinion" U1 ON (U0."cited_opinion_id" = U1."id")
    WHERE (U1."cluster_id" = ("search_opinioncluster"."id") AND U1."cluster_id" >= 0 AND U1."cluster_id" < 10000) 
    GROUP BY U1."cluster_id"), 0) new_count
FROM
    search_opinioncluster
WHERE ("search_opinioncluster"."id" >= 0 AND "search_opinioncluster"."id" < 10000)
```

I actually ran this from the PR code, in a django shell, to get proper stats, and got 80 seconds duration for the above update

and 2 seconds duration for an update with step size 1000

### DIFF
--- a/cl/citations/management/commands/count_citations.py
+++ b/cl/citations/management/commands/count_citations.py
@@ -1,8 +1,10 @@
+import time
+
 from django.db import transaction
 from django.db.models import Count, IntegerField, OuterRef, Subquery, Value
 from django.db.models.functions import Coalesce
 
-from cl.lib.command_utils import VerboseCommand
+from cl.lib.command_utils import VerboseCommand, logger
 from cl.search.models import OpinionCluster, OpinionsCited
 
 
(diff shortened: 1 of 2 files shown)


quevon24 (approved):
Everything looks good.

I liked the idea of having to specify the start and end id, in the worst case that something fails you just need to run it again specifying the range.

And it will definitely help to not lock the entire cluster table.

Issue #6176, opened 2025-08-13
Command to re index citation related fields in Elastic


grossir:
### Describe the issue.

Since we turned off  ES `OpinionDocument.citeCount` and `OpinionClusterDocument.citeCount` and `OpinionDocument.cites` indexing, we need to index these updated fields again. 

This was done through `cl.search.tasks.index_related_cites_fields`. 

It should hit
- 7608285 / 10884615 opinions
- 3271789 /  10173920 clusters

This is still blocked until we [update](https://github.com/freelawproject/courtlistener/pull/6174) the `OpinionCluster.citation_count` 

Also, we should have some strategy to perform this update without impacting other Elastic indexing tasks; or the general performance of Elastic

```sql
courtlistener=> select count(distinct(citing_opinion_id)) from search_opinionscited;
  count  
---------
 7608285
(1 row)

courtlistener=> select count(*) from search_opinioncluster where citation_count  <> 0;
  count  
---------
 3271789
(1 row)

courtlistener=> select count(*) from search_opinion;
  count   
----------
 10884615
(1 row)

courtlistener=> select count(*) from search_opinioncluster;
  count   
----------
 10173920
(1 row)

```

grossir:
Now we have to run this, probably nightly as to not impact ES servers. Will run some tests and update this issue

mlissner:
I'm curious how hard this was. Can you share a few details?

grossir:
I didn't run the commands; I intended to leave this to be auto cleaned up by the regular reindex process. 

I can't find the comment where we talked about this happening; it was probably during a meeting. 

I haven't check if the regular reindex process has actually caught up even with the older opinions / clusters affected by the citation count turn-off, described by the query below.

```sql
select 
distinct(so.id) 
from search_opinion so 
inner join search_opinionscited on citing_opinion_id = so.id 
where 
-- PR turning off citation count updates was merged on 2025-06-03
date_modified::date > '2025-06-02'::date 
-- we finished running the bulk updates in 2025-07-28
and date_modified::date < '2025-07-29'::date;
```

mlissner:
Gotcha. We're re-indexing everything again, too, for #4381. Thanks.

grossir:
Sorry for the confusion should have written about it while closing the issue

PR #6177, opened 2025-08-13
feat(citations): command to reindex_es_cite_fields

grossir:
Solves #6176

- Add a new command `reindex_es_cite_fields`. This command reindexes
OpinionClusterDocument.citeCount and OpinionDocument.cites; which
may be out of sync due to find_citations with a --disable-citation-count
flag

- processing will happen in batches, sequentially, to avoid overloading ES

- Refactors cl.search.tasks to reuse logic building the update documents

- Adds tests


cl/citations/tests.py
```diff
@@ -3406,3 +3411,64 @@ def test_count_citations_from_opinions_cited(self):
...
+    def test_cites_reindex(self):
+        """Can we update the OpinionDocument.cites field?"""
+        cluster = OpinionClusterWithChildrenAndParentsFactory.create()
+        opinion = cluster.sub_opinions.first()
+        index_parent_and_child_docs([cluster.id], SEARCH_TYPES.OPINION)
+
+        another_cluster = OpinionClusterWithChildrenAndParentsFactory.create()
+        another_opinion = another_cluster.sub_opinions.first()
+
+        # prevent triggering the signal using a straightforward .create
```
quevon24: If i remember correctly, create() fire signals but not update()

quevon24 (approved):
Everything looks good, just a detail in a code comment

cl/citations/tests.py
grossir: yes, thanks ; the comment should be for this line and for `.update`
```python
        OpinionsCited.objects.filter(id=created.id).update(
            citing_opinion=opinion
        )

```

PR #6187, opened 2025-08-15
Prevent double click when submit document in court upload portal

quevon24:
## Fixes

This fixes the creation of duplicates in the courts upload portal (me, mo, miss, etc)

## Summary

This PR prevents double clicking when submitting the form. It disables the submit button and replace the text with: "Processing" with a spinner. 

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

1. To deploy...

### DIFF
--- a/cl/opinion_page/templates/publish.html
+++ b/cl/opinion_page/templates/publish.html
@@ -30,6 +30,13 @@
 {% block footer-scripts %}
 	{% include "includes/date_picker.html" %}
 	<script src="{% static "js/jquery.bootstrap-growl.js" %}"></script>
+    <script type="text/javascript" nonce="{{ request.csp_nonce }}">
+        document.getElementById('court-upload-form').addEventListener('submit', function(e) {
+          const submitButton = this.querySelector('button[type="submit"]');
+          submitButton.disabled = true;
+          submitButton.innerHTML = '<i class="fa fa-spinner fa-spin"></i>&nbsp;Processing...';
+        });
+    </script>
 {% endblock %}
 
 {% block content %}
@@ -71,7 +78,7 @@
 				<br>
 				<form method="post" action=""
 					  enctype="multipart/form-data"
-					  class="form-horizontal">
+					  class="form-horizontal" id="court-upload-form">
 					{% csrf_token %}
 					{% for field in form.hidden_fields %}
                         {{ field }}


mlissner:
Just chiming in to say that this is usually called "debouncing". We have an example of it in input.html, which is part of our new UI. I haven't done enough research to know if using that code is best or if your solution is just simple enough to run with, but I thought I'd highlight it for you in case you want to chat with Elisa and there's an opportunity to be more consistent in how we debounce.

quevon24:
> Just chiming in to say that this is usually called "debouncing". We have an example of it in input.html, which is part of our new UI. I haven't done enough research to know if using that code is best or if your solution is just simple enough to run with, but I thought I'd highlight it for you in case you want to chat with Elisa and there's an opportunity to be more consistent in how we debounce.

Thanks @mlissner, I proposed this solution since it seemed to me the simplest to solve the problem, a more complex alternative could be to generate a one-time token for each form submission or maybe rate limiting the form submit.
I'll ask Elisa what she thinks.

mlissner:
I'd definitely reach for JS to solve this problem. I just thought we might have some sort of debounce library/code we're already using.

elisa-a-v:
Thanks for looping me in, happy to contribute! I don’t think debouncing really fits here. It's great for repeated events like typing or scrolling, but a form submission should happen immediately and only once, so we shouldn't wait for the user to stop clicking/pressing <kbd>enter</kbd> to send the form.

I think Kevin's solution is reasonable for this use case and providing user feedback is **great**, but indeed only disabling the button would probably not be enough because users might still be able to resend the form by:
1. Pressing <kbd>enter</kbd>.
2. Navigating back/forward in the browser.
3. Reloading the confirmation page.

To address # 1, we could add a flag to mark as submitting:

```js
    form.addEventListener('submit', function (e) {
      if (form.dataset.submitting === 'true') {
        e.preventDefault();
        return;
      }
      form.dataset.submitting = 'true';

      // disable button and provide user feedback
    });
```

This would prevent any repeat submissions, no matter the source of the event.

To address # 2 and # 3, there's a design pattern called POST/Redirect/GET (PRG) which might be worth considering. It doesn't seem hard to implement: Normally when submitting a form, the browser sends a POST request to the server, and the server responds with a new rendered template. Instead, with PRG the server responds first with a 303 status code and then redirects to a proper GET URL. Then users can reload or navigate however they want and the form will not be implicitly resubmitted by the browser.

Bottomline my recommendation would be:
- Keep the form-level submit listener.
- Keep disabling the button and showing feedback to the user.
- Add a submitting flag to prevent further submissions with <kbd>enter</kbd>.
- Consider PRG server-side.

Some resources about PRG in case you want to learn more about it: [TheServerSide](https://www.theserverside.com/news/1365146/Redirect-After-Post), [Wikipedia](https://en.wikipedia.org/wiki/Post/Redirect/Get), [Seobility](https://www.seobility.net/en/wiki/post-redirect-get).

quevon24:
Thanks for this information @elisa-a-v, especially for the PRG references. I'm not entirely sure that PRG helps prevent double submissions in this specific case (click the submit button quickly several times), but it can be helpful for users so they don't see the traditional alert when they reload a page that has posted a message and the browser wants to resubmit the data with the alert (If the first submission was successful, even if you try to resubmit the form when reloading the page, a validation error will appear due to the file's SHA.).

I think the PRG part can be implemented without problems in the view in this part:

```
if await sync_to_async(form.is_valid)():
    cluster = await sync_to_async(form.save)()
    goto = reverse("view_case", args=[cluster.pk, cluster.slug])
    messages.info(
        request, "Document uploaded successfully.", extra_tags=goto
    )
    return HttpResponseRedirect(
        reverse("court_publish_page", kwargs={"pk": pk})
    )
```

Instead of redirecting back to the form upon success, you could redirect to the court's homepage and meet PRG requirements with something like this:

```
if await sync_to_async(form.is_valid)():
    cluster = await sync_to_async(form.save)()
    goto = reverse("view_case", args=[cluster.pk, cluster.slug])
            messages.info(
                request, "Document uploaded successfully.", extra_tags=goto
            )
    # Redirect to court homepage to complete PRG pattern
    return HttpResponseRedirect(
        reverse("court_homepage", kwargs={"pk": pk})
    )
```

This along with the additional change you suggested to avoid the post when pressing enter.

Just one question, @mlissner . Doesn't the view redirect back to the form, since courts can upload multiple documents in one day? Is this done to avoid additional clicks to return to the form from the upload portal main page?

The solution I propose above in the view involves an additional click to return to the form after a successful upload.

elisa-a-v:
> I'm not entirely sure that PRG helps prevent double submissions in this specific case (click the submit button quickly several times)

Yeah I think you're right, if that's the only possible cause, then PRG is no use here.

> If the first submission was successful, even if you try to resubmit the form when reloading the page, a validation error will appear due to the file's SHA.

I guess I just assumed if it was possible to bypass that validation by clicking twice, it could theoretically be possible to do it by navigating back and forth before the submission is fully processed. Now I realize even if that is the case, PRG wouldn't help because PRG would only kick off _after_ the submission has been processed 🤔 am I getting this right?

In light of all this I'm not sure I see the point in using PRG here anymore, might complicate things more than help. Maybe the safeguard for the submission with <kbd>enter</kbd> is enough for this particular issue.

mlissner:
> In light of all this I'm not sure I see the point in using PRG here anymore, might complicate things more than help. Maybe the safeguard for the submission with enter is enough for this particular issue.

I think I agree!

grossir (approved):
Tested it on the dev environment, and I think it works as expected

Example URL http://0.0.0.0:8000/court/tennworkcompcl/new-opinion/

image

PR #6195, opened 2025-08-19
feat(admin): optimize OpinionAdmin

elisa-a-v:
## Summary

I tried using the Opinion admin panel and it took forever to load, so this PR optimizes the Opinion admin panel by joining the clusters with `select_related`.

When testing with the dev DB, this reduced the number of DB queries from 103 to 3, and the time went from 20s to only 2. Still long, but a 10-fold improvement is not bad 🤷🏽 

### Before

Screenshot from 2025-08-19 14-19-30

### After

image

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/search/admin.py
+++ b/cl/search/admin.py
@@ -49,6 +49,9 @@ class OpinionAdmin(CursorPaginatorAdmin):
         "date_modified",
     )
 
+    def get_queryset(self, request):
+        return super().get_queryset(request).select_related("cluster")
+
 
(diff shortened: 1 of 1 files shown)


mlissner (approved):
LGTM, thanks

Issue #6217, opened 2025-08-25
Selecting a date from either the next or the previous month is highlighted with wrong color


elisa-a-v:
### What happened?

New date selectors with date pickers were recently added to the corpus search component. For the date pickers, we used Flatpickr, and we customized the styles to match our Figma design by overriding classes in our Tailwind input file.

### Expected behavior

When a date is selected in the date picker, it gets highlighted with our primary red circle.

### Actual behavior

When selecting a date from either the previous or the next month in the date picker, the selected date looks light blue instead of our primary red until clicking somewhere else.

### Steps to reproduce

1. Open one of the date pickers.
2. Click a date from either the previous or the next month.
3. The highlight will look light blue.
4. Clicking anywhere turns it our primary red.

#### Design / Current

<img width="350" alt="Image" src="https://github.com/user-attachments/assets/15cb75dd-a7fe-4958-8b95-0c6a658d74cb" />
<img width="350" alt="Image" src="https://github.com/user-attachments/assets/fe4c238f-428b-49b4-aacd-0f3572cc365e" />

https://github.com/user-attachments/assets/2df2477e-ae76-46d2-b852-85eb3c9145ed

### What is the severity of the bug?

Trivial

Issue #6215, opened 2025-08-25
Resizing window in new homepage breaks advanced filters in mobile


elisa-a-v:
### What happened?

When introducing the date selectors to the corpus search component, the advanced filters overflow to the right in mobile after resizing from desktop to mobile.

### Expected behavior

When resizing the window:
- Layout should be preserved.
- The site should remain completely usable without needing to reload the page.

### Actual behavior

Resizing the window breaks layout, rendering the advanced filters very hard to use in mobile after this happens unless reloading the page.

### Steps to reproduce

1. Go to the new homepage.
2. In desktop, click the RECAP Archive tab.
3. Click any of the calendar date inputs and open the picker popup.
4. Resize the window to mobile width.
5. On the RECAP Archive filters, the layout breaks.

### Expected / Actual

<img width="350" alt="Image" src="https://github.com/user-attachments/assets/cea18bc9-50a5-4289-95d3-52bc92674040" />

<img width="350" alt="Image" src="https://github.com/user-attachments/assets/7ac9a740-0e4c-44e0-947b-e6b6d0f9e82a" />

### What is the severity of the bug?

Low

mlissner:
This is a bit of a corner case and users can work around it fairly easily. Let's push or hope a volunteer picks it up.

Issue #6214, opened 2025-08-25
Introduce mobile version of date selector component


elisa-a-v:
### Describe the issue.

Current version of date selector works exactly the same in mobile and desktop: clicking/focusing the date inputs opens the calendar popup, and the user can browse the calendar and select a date, which is only entered in the input once the user clicks "Apply", otherwise the selected date gets cleared when the picker is closed.

However, the mobile version is different in the design: instead of a popup, a full dialog appears, with the same design as the operators modal but with an inline calendar instead.

We could introduce the same `<c-dialog>` component in the date selector component, and include two more hidden inputs in it so we can attach the flatpickr calendars to them.

#### Requirements
- Tapping the calendar date inputs should open the dialog.
- On open, preload the calendar from the live input value (if any).
- The user can select any date in the calendar without changing the value in the input. Only after tapping "Apply" the value gets saved.
- When closing the dialog, focus should return to the triggering input.
- Multiple instances should be supported, so no hardcoded global `id`, always use `x-id` to ensure unique IDs.

These requirements are already working in desktop, but we need to make sure they work correctly in both versions.

### Related Issue

#6002 

## Current / Design

<img width="350" alt="Image" src="https://github.com/user-attachments/assets/81632be0-af50-4d7b-9617-74ebe04f38d3" />

<img width="350" alt="Image" src="https://github.com/user-attachments/assets/049e8d48-36f3-41b2-867c-3a9713b28819" />

### What type of issue is this?

To Do

mlissner:
We pushed this to our "Someday" pile because it'll work OK using the regular desktop approach and we want to get the new UI out the door.

PR #6220, opened 2025-08-25
feat(components): introduce date selector sub-component for the Corpus Search component

### LINKED ISSUE #6002: Implement date selectors

elisa-a-v:
## Fixes

Closes #6002 

## Summary

This PR introduces:
- A range date selector component as a sub-component of Corpus Search.
- New date selectors in the advanced filters for the corpus search component:
    - **Case Law**: calendar dates for filing dates.
    - **RECAP Archive**: relative dates for Docket Date and Entry Date.
    - **Oral Arguments**: calendar dates for Argument date.
    - **Judges**: calendar dates for date of birth.
- The Flatpickr library for the calendar date pickers.
- Three new READMEs:
    - Third party CSS: all upstream CSS required for third party libraries **for new templates**.
    - Third party JS that is **not** related to Alpine: all upstream code **for new templates**.
    - Alpine: some upstream code (core + plugins) and our code (components + composables).

### About the changes

- For the date picker calendars, the Flatpickr library was used due to its lack of other dependencies and "easy" [integration with Alpine](https://alpinejs.dev/component/flatpickr). It kinda does the job but it's more brittle than expected (tweaking one thing breaks another easily) and I haven't been able to use keyboard navigation within it, but keyboard users can still type the date so this doesn't feel critical.
- The Mobile dialog is not introduced with this PR: it currently works the same as the desktop version with the calendar popup, but it should open a dialog similar to the Operators Modal.
    - Will be addressed here: https://github.com/freelawproject/courtlistener/issues/6214
- https://github.com/freelawproject/courtlistener/issues/6215.
    - I don't think this should block this PR since it's an edge case and reloading fixes it again. Not ideal UX but good enough?
- https://github.com/freelawproject/courtlistener/issues/6217
    - I don't think this one should block this PR either since I already tried to fix it and got nowhere, so given how minor it is, I'd defer for later.
- I found a very minor issue that I don't think warrants any action cause it's barely noticeable, but just want to mention anyway just in case: Firefox doesn't center the styled radio buttons to the pixel, for some reason both the inner white circle and the outer focus ring are ever-so-slightly off, and not in a consistent way (some are off to the right, some to the left, some kind of diagnoally). Safari looks better for once! 😅 

Firefox:

image

Safari:

image

## Question

The date inputs in the Oral Arguments are the only ones that span two cells (one per input) instead of one cell for both. I know we want this to avoid having a single cell in the last row, but how critical is it to have this now? tbh this completely slipped my mind so they are currently sharing the same cell, just like the rest. The solution that comes to mind isn't hard but it isn't trivial either and I think could be addressed in a followup PR.

I think this would involve a refactor, since the date selector introduced in this PR is actually for date ranges, not for single dates. I don't think it would be hard to extract the logic for the calendar date inputs into its own standalone component, and use it in our date range selector, but I'd rather not have that block this PR so we get things moving. What do you think, @mlissner ? If this is critical we can include it in this PR, otherwise I can open a new issue for it, your call.

<details>
<summary>This is what I mean</summary>

#### Design
Screenshot from 2025-08-25 17-21-19

#### This PR
Screenshot from 2025-08-25 17-21-45

</details>

## Note to reviewer

An issue that I ran into while developing was some elements getting blocked by closed flatpickr calendars because they sit on top of the content even when hidden. `pointer-events-none` was added so they now shouldn't block the content below, but flatpickr performs live CSS changes so it's something to keep in mind.

This means when reviewing this PR we need to make sure all other features are still usable:
- Search tabs in homepage should be clickable and switch tabs when clicked.
- All inputs should be focusable by clicking on them.
- Header buttons are clickable both in desktop and mobile.

## Deployment

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

---

<details closed>
<details open>

Screenshot from 2025-08-25 10-01-38

Screenshot from 2025-08-25 10-01-19

Screenshot from 2025-08-25 10-01-00

image

</details>

<details open>

Screenshot from 2025-08-25 10-40-33

Screenshot from 2025-08-25 10-59-34

</details>
</details>

## Demo

Demo doesn't show it but the relative dates input is a normal text input that is only prefilled by the selection, it can be edited freely.

[Screencast from 08-25-2025 10:53:48 AM.webm](https://github.com/user-attachments/assets/5b842137-630f-4a76-9aa1-a9194cb32492)


cl/assets/static-global/css/third_party/README.md
```diff
@@ -0,0 +1,7 @@
+# Third-Party CSS
+
+This directory contains stylesheets for third-party libraries used in new templates.
+
+- All files are **upstream code** and must **not be edited directly**.
+- If the stylesheet is widely used and needs to be available globally, import it in Tailwind's input file. This ensures the code is minified.
```
mlissner, line 6: Is there a minified version of the flatpicker.css file that we should be using?

mlissner (commented):
I just noticed one little thing, but this looks cool otherwise. 

Should this be added to the components library (I think so)?

mlissner:
And thanks for all the detail in the description. I don't have much to add: These sound like reasonable compromises to get things done, and it's annoying that Firefox is letting us down. Lame!

cl/assets/static-global/css/third_party/README.md
elisa-a-v: Actually yes, there is a minified version, but we import `flatpickr.css` in Tailwind's input file, not using any `<link>` tags, so in production it will end up minified in the same bundle anyway.

I think for CSS that should be available in _every_ page this is best, and only in cases where the third party CSS won't be used in every page I would add both minified and non-minified, and handle them manually via conditional `<link>` tags in the `head` block of the specific templates that require them. How does that sound?

mlissner: Sounds great. Want to memorialize that somewhere that makes sense?

elisa-a-v:
> Should this be added to the components library (I think so)?

Yes, I hadn't thought about it because I thought it wouldn't be used anywhere else, but even if that's the case, I agree it makes sense to document it, so I'll add the entry to the library.

mlissner:
I'm a little torn on it. Maybe you're right and we only put re-usable ones in the library.

cl/assets/static-global/css/third_party/README.md
elisa-a-v: Sure, this new README seems like the right place, and maybe the Wiki could have something about this in the Tailwind section too 🤔

elisa-a-v:
> Maybe you're right and we only put re-usable ones in the library.

I think bigger stuff like the whole corpus search component doesn't need to be in there, but there's a case for adding these sub components for maintainability. I think there's a couple of things that aren't so obvious about how to use the component that might be worth documenting for future reference, unless the likelihood of future changes to the filters is near zero.

mlissner:
It's *pretty* unlikely, and maybe the usage could go in a comment? I really don't have a strong opinion and trust your judgement on it. My only concern is proper documentation for it, I think.

elisa-a-v:
> It's pretty unlikely, and maybe the usage could go in a comment?

Yeah that sounds way better given how unlikely it is.

We need to strike a balance between thorough documentation, and keeping it relevant to devs who will use it, so it makes sense to keep the library for re-usable ones only, and use docstrings for the rest.

So no new library entry then, but I'll add a docstring with the non-obvious stuff.

cl/search/templates/cotton/corpus_search/filters/opinions.html
```diff
@@ -33,22 +33,16 @@
...
-        {{ search_form.filed_before }}
-      </div>
-    </div>
-  </div>
+  <c-corpus-search.date-selector
+    label="Filed After"
+    second_label="Filed Before"
+    :request="request"
+    date_after_field_name="filed_after"
+    date_before_field_name="filed_after"
```
ERosendo: Some of the date input fields in the search interface have incorrect names.

```suggestion
    date_before_field_name="filed_before"
```

cl/search/templates/cotton/corpus_search/filters/oral_arguments.html
```diff
@@ -26,18 +26,15 @@
...
-    <div class="focus-container input-text">
-      {{ search_form.argued_before }}
-    </div>
-  </div>
+  <c-corpus-search.date-selector
+    label="Argued After"
+    second_label="Argued Before"
+    :request="request"
+    date_after_field_name="argued_after"
+    date_before_field_name="argued_after"
```
ERosendo:
```suggestion
    date_before_field_name="argued_before"
```

cl/search/templates/cotton/corpus_search/filters/judges.html
```diff
@@ -19,22 +19,16 @@
...
-        {{ search_form.born_before }}
-      </div>
-    </div>
-  </div>
+  <c-corpus-search.date-selector
+    label="Born After"
+    second_label="Born Before"
+    :request="request"
+    date_after_field_name="born_after"
+    date_before_field_name="born_after"
```
ERosendo:
```suggestion
    date_before_field_name="born_before"
```

cl/search/templates/cotton/corpus_search/filters/recap.html
```diff
@@ -33,43 +33,23 @@
...
-        </div>
-      </div>
-    </div>
-  </div>
+  <c-corpus-search.date-selector
+    label="Docket Date"
+    relative
+    :request="request"
+    date_after_field_name="filed_after"
+    date_before_field_name="filed_after"
```
ERosendo:
```suggestion
    date_before_field_name="filed_before"
```

elisa-a-v:
> I noticed a small issue with the date field names that could affect the search function.

Sweet of you to say it like that 😂 that was no minor oversight, thanks for noticing!

I've addressed all comments, and added a [new section in the wiki](https://github.com/freelawproject/courtlistener/wiki/New-Frontend-Architecture#third-party-css) about third party CSS.

I also noticed a bug where the closed calendars were sitting below the footer and would overflow the page, so it was possible to scroll past it:

image

They now have `display: none` when they are not open, so it looks right again:

image

ERosendo (approved):
LGTM :shipit:

Issue #6179, opened 2025-08-14
`recap_into_opinions` sometimes creates duplicate  opinions


grossir:
Seems to be happening from June 2024, and ongoing (up to May 2025). The query below returns 898 rows for 344 unique hashes; so 554 of them should be deleted

```sql
courtlistener=> select a.date_created, a.id, source,  sha1, download_url from search_opinion a inner join search_opinioncluster b ON a.cluster_id = b.id where sha1 in  (select sha1 from search_opinion where sha1 <> '' group by sha1 having count(*) > 1) and source = 'G';
         date_created          |    id    | source |                   sha1                   | download_url 
-------------------------------+----------+--------+------------------------------------------+--------------
 2024-06-25 16:41:16.143557+00 | 10292293 | G      | ebaf77392410a0090fccbcde67442630bc7c2bc4 | 
 2024-06-25 16:41:16.279814+00 | 10292294 | G      | ebaf77392410a0090fccbcde67442630bc7c2bc4 | 
...
 2025-05-06 08:48:05.558558+00 | 10995106 | G      | 995afdaee7855321b840b5f6d274921b75e26989 | 
 2025-05-06 08:48:11.140761+00 | 10995109 | G      | 995afdaee7855321b840b5f6d274921b75e26989 | 
```

Most duplicates are time clustered, so, it seems to me that the `Opinion.sha1` existence check should be redone closer to the creation of the objects themselves
https://github.com/freelawproject/courtlistener/blob/e51752a5a4f8b45ba21a831838eb72a0a2db489c/cl/corpus_importer/tasks.py#L2973

grossir:
The majority of the duplicates seem to be happening due to the same opinion existing one time for each of the dockets in a consolidated case. 

This may be a data modelling difference between the PACER and the Case Law parts of the DB. I see that `search_recapdocument` has no sha1 unique constraint; and it uses the hash to identify a document, not to check for duplication

https://github.com/freelawproject/courtlistener/blob/b9a20e63e1a0f15db14ab6e3cfdfeea1fe9c4a5d/cl/recap/tasks.py#L428-L442

See an example [opinion](https://storage.courtlistener.com/recap/gov.uscourts.nynd.117996/gov.uscourts.nynd.117996.50.0.pdf)

The clusters below have the duplicated (same-hash) opinion linked above

[Reece](https://www.courtlistener.com/opinion/9845764/x/), No. 1:19-CV-219;
[...]

PR #6221, opened 2025-08-25
fix(recap_into_opinions): improve duplicated hash checks

grossir:
Helps solve #6179

- introduce a second hash check in `recap_document_into_opinions`, just before object creation

- add a explicit hash check in the `recap_into_opinions` command, to prevent enqueuing the same hash multiple times in a same call

- introduce a random countdown to the chain in `scrape_pacer_free_opinions` (the ongoing process that calls `recap_document_into_opinions` ). This should help the hash checks work. A explicit hash deduplication (as for the `recap_into_opinions` command) cannot be done since the getting the pdf and the hash is part of chain, and no sha1 is available at the start of it

### DIFF
--- a/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
+++ b/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
@@ -356,7 +356,7 @@ def get_pdfs(
         # we accept same hash RecapDocuments; but we don't want duplicates in
         # Opinions. Introduce some time variability between consecutive tasks
         # to help the hash checks in `recap_document_into_opinions` work
-        c.apply_async(countdown=random.uniform(5, 30))
+        c.apply_async(countdown=random.uniform(3, 10))
         completed += 1
 
         if completed % 1000 == 0:
--- a/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
+++ b/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
@@ -2,6 +2,7 @@ import argparse
 import datetime
 import inspect
 import math
+import random
 import time
 from collections.abc import Callable
 from typing import cast
@@ -335,6 +336,7 @@ def get_pdfs(
                 f"before starting the next cycle."
             )
             time.sleep(1)
+
         logger.info(f"Processing row id: {row.id} from {row.court_id}")
         c = chain(
             process_free_opinion_result.si(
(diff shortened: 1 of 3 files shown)


cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
```diff
@@ -351,8 +353,12 @@ def get_pdfs(
             delete_pacer_row.s(row.pk).set(queue=q),
         )
 
-        c.apply_async()
+        # we accept same hash RecapDocuments; but we don't want duplicates in
+        # Opinions. Introduce some time variability between consecutive tasks
+        # to help the hash checks in `recap_document_into_opinions` work
+        c.apply_async(countdown=random.uniform(5, 30))
```
quevon24: 30 isn't a lot? How could the highest value be estimated?

grossir: the highest countdown value would be 30 seconds. It's a uniform distribution, so it has equal probability for anything between [5,30], with the average being 17.5 seconds. Note that this is not a global throttle, but a per-worker "wait"; so the "wait" impact will be less than first apparent and shouldn't delay the process by a lot

We ingest at most 500 recap opinions a day, split among 2 or 3 consecutive hours, so this should amount to an average of `( 500 * 17.5 ) / 60 / 60 / 12 = 0.20 hours`  where 12 is (the number of celery pods * worker concurrency) for the `celery-default` deployment. Still, this may be a lot of wait time... I will reduce it to [3, 10]

```sql
courtlistener=> select date_created::date, date_part('hour', date_created), count(*) from search_opinioncluster where source  = 'G' group by date_created::date, date_part('hour', date_created) order by 1 desc limit 100;
 date_created | date_part | count 
--------------+-----------+-------
 2025-08-27   |         8 |   314
 2025-08-27   |         9 |    75
 2025-08-26   |         8 |   297
 2025-08-26   |         9 |   130
 2025-08-25   |         8 |   232
 2025-08-25   |         9 |    96
 2025-08-24   |         8 |   107
 2025-08-24   |         9 |   131
 2025-08-23   |         8 |    88
 2025-08-23   |         9 |   100
 2025-08-22   |         9 |   113
 2025-08-22   |        10 |    54
 2025-08-22   |        11 |    63
 2025-08-21   |         8 |   173
 2025-08-21   |         9 |    96
 2025-08-21   |        10 |    68
 2025-08-21   |        11 |    28
 2025-08-20   |         8 |   154
 2025-08-20   |         9 |   117
 2025-08-20   |        10 |     7
 2025-08-19   |         8 |   251
 2025-08-19   |         9 |   147

```

grossir:
Setting this to skip deployment since we are going to deploy to every deployment on https://github.com/freelawproject/courtlistener/pull/6231

PR #6237, opened 2025-08-28
feat(admin): seal opinion clusters action

quevon24:
## Summary

This PR implements the action to seal an opinion cluster, which implies:

- Deleting the Opinion Cluster if possible. Upon deletion, a corresponding `ClusterRedirection` is created to maintain reference integrity.
- Deleting the associated Docket only if it is also eligible for deletion, but only after the cluster has been deleted.

Eligibility for deletion depends on the related objects:

- **Cluster deletion blockers:**  
  The cluster cannot be deleted if related objects exist in these models. Instead, the user is shown a detailed confirmation page listing these blocking objects:  
  - `favorites.DocketTag`  
  - `favorites.Note`  
  - `alerts.DocketAlert`  
  - `visualizations.SCOTUSMap`

- **Docket deletion blockers:**  
  The docket cannot be deleted if related objects exist in:  
  - `audio.Audio`  
  - `people_db.AttorneyOrganizationAssociation`  
  - `people_db.PartyType`  
  - `people_db.Role`  
  - `search.BankruptcyInformation`  
  - `search.Claim`  
  - `search.DocketEntry`

- **Non-blocking related objects:**  
  The following related objects can be deleted alongside the cluster or docket without blocking deletion:  
  - `scrapers.PACERMobilePageData`  
  - `search.DocketPanel`  
  - `search.DocketTags`  
  - `search.Citation`  
  - `search.Opinion`  
  - `search.OpinionClusterNonParticipatingJudges`  
  - `search.OpinionClusterPanel`

This approach ensures that sealing cleanly removes the cluster (and creating the redirection) and its docket if possible, while clearly indicating to users any blocking dependencies that prevent deletion.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

1. To deploy...

<details closed>
<details open>

image

image

image

image

</details>

### DIFF
--- a/cl/assets/templates/admin/seal_cluster_blocking_confirmation.html
+++ b/cl/assets/templates/admin/seal_cluster_blocking_confirmation.html
@@ -1,0 +1,37 @@
+{% extends "admin/base_site.html" %}
+
+{% block content %}
+
+    {% if has_blocking %}
+
+        <p>The cluster <strong>{{ cluster }}</strong> contains the following blocking related items that prevent sealing:
+        </p>
+
+        <ul>
+            {% for label, items in blocking_relations.items %}
+                {% if items.exists %}
+                    <li><strong>{{ label }} ({{ items.count }})</strong>
+                        <ul>
+                            {% for item in items %}
+                                <li>
+                                    {% if item.admin_url %}
+                                        <a href="{{ item.admin_url }}">{{ item }}</a>
+                                    {% else %}
+                                        {{ item }}
+                                    {% endif %}
+                                </li>
+                            {% endfor %}
+                        </ul>
+                    </li>
+                {% endif %}
+            {% endfor %}
+        </ul>
+
+        <p>Please remove or reassign these objects before attempting to seal the cluster.</p>
+
+    {% else %}
+        <p>The cluster <strong>{{ cluster }}</strong> doesn't contains blocking related items that would prevent sealing.
+        </p>
+    {% endif %}
+
+{% endblock %}
--- a/cl/search/admin.py
+++ b/cl/search/admin.py
@@ -1,7 +1,13 @@
+from typing import Any
+
 from admin_cursor_paginator import CursorPaginatorAdmin
-from django.contrib import admin
-from django.db.models import QuerySet
-from django.http import HttpRequest
+from django.contrib import admin, messages
+from django.db import transaction
+from django.db.models import Q, QuerySet
+from django.http import HttpRequest, HttpResponse
+from django.shortcuts import get_object_or_404, render
+from django.urls import path, reverse
+from django.utils.html import format_html
 
 from cl.alerts.models import DocketAlert
 from cl.lib.admin import build_admin_url
@@ -28,6 +34,7 @@ from cl.search.models import (
     RECAPDocument,
     SearchQuery,
 )
+from cl.visualizations.models import SCOTUSMap
 
 
 @admin.register(Opinion)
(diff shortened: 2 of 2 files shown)


cl/assets/templates/admin/seal_cluster_blocking_confirmation.html
```diff
@@ -0,0 +1,37 @@
...
+
+        <ul>
+            {% for label, items in blocking_relations.items %}
+                {% if items.exists %}
+                    <li><strong>{{ label }} ({{ items.count }})</strong>
+                        <ul>
+                            {% for item in items %}
+                                <li>
+                                    {% if item.admin_url %}
+                                        <a href="{{ item.admin_url }}">{{ item }}</a>
```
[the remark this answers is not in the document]
grossir: Checking this semgrep comment: it seems to be asking you to use 
```
<a href="{% url "name_of_url" %}">
```

There is a name "opinioncluster_blocking_confirmation" defined in `get_urls()` of the Opinion admin, but will it work for this use case? Or should it be in a `urls.py` file?

cl/search/admin.py
```diff
@@ -91,6 +98,201 @@ class OpinionClusterAdmin(CursorPaginatorAdmin):
...
+        "audio.Audio": lambda cluster: cluster.docket.audio_files,
+        "people_db.AttorneyOrganizationAssociation": lambda cluster: cluster.docket.attorneyorganizationassociation_set,
+        "people_db.PartyType": lambda cluster: cluster.docket.party_types,
+        "people_db.Role": lambda cluster: cluster.docket.role_set,
+        "search.BankruptcyInformation": lambda cluster: getattr(
+            cluster.docket, "bankruptcyinformation", None
+        ),
+        "search.Claim": lambda cluster: cluster.docket.claims,
+        "search.DocketEntry": lambda cluster: cluster.docket.docket_entries,
+    }
```
grossir, line 130: missing a key for "search.OpinionCluster"

cl/search/admin.py
```diff
@@ -91,6 +98,201 @@ class OpinionClusterAdmin(CursorPaginatorAdmin):
...
+                    "Problem sealing cluster id: {}{}",
+                    cluster.pk,
+                    format_html(
+                        ' - <a href="{}" target="_blank">View Dependencies</a>',
+                        url,
+                    )
+                    if url
+                    else "",
+                )
+                self.message_user(request, message, messages.WARNING)
```
grossir, line 300: I think it would be better to have the `messages.ERROR` level? I see in the screenshot this looks green / success colored, and it shouldn't

grossir (commented):
Looks very good; just a few comments below

cl/search/admin.py
quevon24: I thought the same thing, but in the Django configuration these styles were set for the messages:

```python
MESSAGE_TAGS = {
    message_constants.DEBUG: "alert-warning",
    message_constants.INFO: "alert-info",
    message_constants.SUCCESS: "alert-success",
    message_constants.WARNING: "alert-warning",
    message_constants.ERROR: "alert-danger",
}
```

And that works for the frontend because it uses Bootstrap, but the admin doesn't have Bootstrap, so regardless of the level, it always displays as success. This could be solved by adding an additional CSS with those classes and having it only load in the admin. What do you think, @mlissner ?

mlissner: I think this should be as simple as possible and doesn't need to look good. So if it just says "ERROR" and is green or whatever, that's fine. It's just for us. Low standards.

cl/assets/templates/admin/seal_cluster_blocking_confirmation.html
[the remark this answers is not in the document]
quevon24: In this case admin_url is generated for each object in the blocking objects list, so you can go directly to admin page of the object and decide what to do (edit, delete, etc).

Also there shouldn't be any issues with XSS attacks because the url is being generated using django's `reverse()` in `get_blocking_relations` method. This should be safe.

The url pattern was added in `get_urls()` to make sure the view is protected by admin middleware.

Issue #6204, opened 2025-08-21
Handle unique constraints when merging clusters' related objects


grossir:
When merging 2 clusters during deletion or versioning, some related tables have unique constraints that are not handled by the merging code, which causes unhandled errors and merging failure

For example 

```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "search_opinioncluster_panel_opinioncluster_id_judge_id_key"
DETAIL:  Key (opinioncluster_id, person_id)=(10587390, 7668) already exists.
```

Couldn't find the Sentry issue, but here is a sample traceback
INFO Processing group {'sha1': '5b91540ac5d02b9da9957158504c9eb4d5a8df04', 'number_of_rows': 2}

```
  File "/opt/venv/lib/python3.13/site-packages/django/db/backends/utils.py", line 105, in _execute
    return self.cursor.execute(sql, params)
           ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
  File "/opt/venv/lib/python3.13/site-packages/psycopg/cursor.py", line 97, in execute
    raise ex.with_traceback(None)
psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "search_opinioncluster_panel_opinioncluster_id_judge_id_key"
DETAIL:  Key (opinioncluster_id, person_id)=(10587390, 7668) already exists.

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/opt/courtlistener/./manage.py", line 15, in <module>
    main()
    ~~~~^^
  File "/opt/courtlistener/./manage.py", line 11, in main
    execute_from_command_line(sys.argv)
    ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^
  File "/opt/venv/lib/python3.13/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line
    utility.execute()
    ~~~~~~~~~~~~~~~^^
  File "/opt/venv/lib/python3.13/site-packages/django/core/management/__init__.py", line 436, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^
  File "/opt/venv/lib/python3.13/site-packages/django/core/management/base.py", line 413, in run_from_argv
    self.execute(*args, **cmd_options)
[...]

PR #6247, opened 2025-09-02
fix(merge_opinion_versions): handle related objects with unique constraints

grossir:
Solves #6204

This PR abstracts the handling of unique constraints for related objects when merging main and version opinions, clusters and dockets.

Before the 2 special cases handled were
- Opinion.joined_by, which has a unique constraint over (person_id, opinion_id)
- Citation, which has a unique constraint over (cluster_id, reporter,page,volume)

Both had tests; which now become the tests of the improved `update_referencing_objects` function that handles the constraints while merging `models_that_reference_docket`, `models_that_reference_cluster`, `models_that_reference_opinion`

### DIFF
--- a/cl/api/templates/rest-change-log.html
+++ b/cl/api/templates/rest-change-log.html
@@ -31,6 +31,62 @@
   <h1 id="about">REST API Change Log</h1>
   <ul>
     <li>
+      <p><strong>v4.3</strong> — Enhances security by enforcing authentication in V4 API endpoints. Anonymous requests
+        now receive a <code>401 Unauthorized</code> response. This change affects the following endpoints:
+      <ul>
+        <!-- Audio -->
+        <li><strong>Audio:</strong> <a href="{% url "oral_argument_api_help" %}#audio-endpoint">audio</a></p>
+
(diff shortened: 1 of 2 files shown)


cl/scrapers/management/commands/merge_opinion_versions.py
```diff
@@ -335,9 +346,7 @@ def update_referencing_objects(
 ):
```
quevon24: This change is not here but Opinion typing is missing from update_referencing_objects params

cl/scrapers/management/commands/merge_opinion_versions.py
```diff
@@ -346,22 +355,69 @@ def update_referencing_objects(
...
+            model._meta.model_name,
+            main_object._meta.model_name,
+            main_object.id,
+        )
+        if not unique_together:
+            related_to_version_qs.update(**update_query)
+            continue
+
+        if isinstance(unique_together, str):
+            exclude_existing_for_main = (
```
quevon24: I think it would be better named something like `existing_for_main` since you haven't excluded it here yet.

PR #6261, opened 2025-09-03
fix(coverage_utils): handle court with no courthouse

grossir:
Solves https://github.com/freelawproject/courtlistener/issues/6239

Previously, when a court had data or descendants, and had no courthouse,
it would cause an error and the coverage page wouldn't load

with this PR:
   - log a logger.error that will go to Sentry for someone to manually review it
   - skip the court, so it doesn't break the page load
   - add a signal to the Court model, that only works on prod and warns about a
	new court needing a new courthouse


quevon24 (approved):
everything looks good

PR #6263, opened 2025-09-04
Add webhook link to top of alert docs

mlissner:
## Summary

Our webhooks are really cool and they're mentioned in the documentation here and there, but Eri pointed out that it'd be better if it were more prominent, so this accomplishes that.

## Deployment

We can skip deployment of this and let is land in another PR later. Seems OK to me! 

## Screenshots

I should do some, but I don't have CL up and running and I just made these through the Github website. Hopefully you'll forgive me!

### DIFF
--- a/cl/simple_pages/templates/help/alert_help.html
+++ b/cl/simple_pages/templates/help/alert_help.html
@@ -45,7 +45,6 @@
     <h3>Table of Contents</h3>
     <ul>
       <li><a href="#about">Overview</a></li>
-      <li><a href="#receiving-alerts">Receiving Alerts</a></li>
       <li>
         <a href="#recap-alerts">Docket Alerts</a>
         <ul>
@@ -69,6 +68,7 @@
         </ul>
       </li>
       <li><a href="#citation-alerts">Citation Alerts</a></li>
+      <li><a href="#receiving-alerts">Receiving Alerts</a></li>
     </ul>
   </div>
 </div>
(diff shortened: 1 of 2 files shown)


mlissner:
I think I'd shy away from saying too much about webhooks outside their own documentation because they're such a technical feature. I think we want people to know they exist, and for folks that know what they are, they can click through and learn more, but I don't think we want to confused the normies.

elisa-a-v:
What part do you think could confuse the normies? 😅 

Maybe this works better?

image

mlissner:
I think I'd take out the "but this requires..." part, but sure, the rest seems good to me. 👍🏻

elisa-a-v:
I imagine what you're trying to do is prevent users from getting prematurely intimidated and decide alerts aren't for them? Am I getting that right?

Cause I think not giving any warning that webhooks are highly technical might end up confusing/frustrating them even more if they get excited by a shiny feature ("oooh live updates in my apps? I want that!") but then realize it's all technical jargon. If they're non-technical, we want to keep them in the alerts page and keep reading, and not get distracted by something that will be out of reach for them.

How about something like this?

> For more advanced uses, webhook events can be sent to your applications or services. This option is mainly for developers. Check out the webhook documentation to learn more.

mlissner:
Sure!

elisa-a-v:
Okay, pushed those changes in both templates, and added an underline to the link you added in the new template:

image

image

mlissner:
I think it might be better to add this new section to the end of the page. It breaks the flow of the intro to put it here because the intro says, "There are three kinds, you can get them in different ways." 

From there it should list the three kinds (docket alerts, search alerts, and citation alerts), and then list the ways of getting those things. For almost all readers, adding this section will only slow them down and complicate things, instead of allowing them to race ahead to the information they need about using CL itself. The only people that benefit from this section are techies, and then (we) don't get to put stuff at the top, I don't think.

Can we try that last tweak?

eriodiah:
@elisa-a-v , @mlissner I think if anyone reading it, is non-technical, they would forward it to their technical lead, or maybe even attempt to figure it out. If they can't figure it out on their own, they would likely then reach out to us. This may potentially give us an opportunity for sell consulting services where we might help them implement the webhooks. But not sure if that's a service we currently have capacity to offer.

mlissner:
Right. This is the help page for alerts, not the API docs, so we don't want to make it technical beyond just a small gesture.

eriodiah:
Ahhhh...ok, gotcha!

elisa-a-v:
> Can we try that last tweak?

Sure! Is this better?

image
image

mlissner:
Looks good to me, but I can't approve my own PR. Want to approve and merge it?

PR #6264, opened 2025-09-04
Feat: Corpus mobile scope design change

bnjdev:
Closes [#6043](https://github.com/freelawproject/courtlistener/issues/6043).

## Summary

This PR addresses [#6043](https://github.com/freelawproject/courtlistener/issues/6043).
The new design follows the one shared in the issue discussion and in the Figma design.

## Changes

1. scope.html
2. corpus_search.js 

## Screenshot

Screenshot 2025-09-04 at 5 42 18 PM

### DIFF
--- a/cl/search/templates/cotton/corpus_search/scope.html
+++ b/cl/search/templates/cotton/corpus_search/scope.html
@@ -64,7 +64,7 @@
     x-bind:aria-controls="menuId"
     aria-haspopup="menu"
     x-bind:aria-expanded="scopeMenuExpanded"
-    class="md:h-[--corpus-search-height] btn-outline border-0 md:border hover:text-greyscale-900 hover:border-greyscale-300 hover:bg-greyscale-50 justify-between md:rounded-xl md:border-greyscale-200 px-0 pb-2.5 md:p-2 {{ class }}"
+    class="md:h-[--corpus-search-height] btn-outline border-0 md:border hover:text-greyscale-900 hover:border-greyscale-300 !bg-transparent justify-between md:rounded-xl md:border-greyscale-200 px-0 pt-0 pb-[8px] md:pb-2.5 md:p-2 {{ class }}"
     aria-label="Open menu to select the scope of your search"
   >
     <span
@@ -106,7 +106,7 @@
     x-transition:leave="transition ease-in duration-100"
     x-transition:leave-start="opacity-100"
     x-transition:leave-end="opacity-0"
-    class="z-10 absolute mt-2 border border-greyscale-200 shadow-lg rounded-lg bg-white py-2 {{ menu_class }}"
+    class="z-10 absolute mt-0 sm:mt-2 border border-greyscale-200 shadow-lg rounded-lg bg-white py-2 {{ menu_class }}"
     role="menu"
     aria-label="Select the scope of your search"
   >
--- a/cl/search/templates/v2_homepage.html
+++ b/cl/search/templates/v2_homepage.html
@@ -39,7 +39,7 @@
         <c-corpus-search class="px-4 py-2.5 w-full !flex-col" only>
           <c-corpus-search.mobile.top-controls :request="request" only />
 
-          <c-corpus-search.mobile.dialog class="w-full mt-2" :request="request" :search_form="search_form" only>
+          <c-corpus-search.mobile.dialog class="w-full mt-2.5" :request="request" :search_form="search_form" only>
             <c-slot name="trigger_button">
               <button
                 x-on:click="open"
--- a/cl/search/templates/cotton/corpus_search/input.html
+++ b/cl/search/templates/cotton/corpus_search/input.html
@@ -3,7 +3,7 @@
 <c-vars rounded class=""></c-vars>
 
 <div
-  class="focus-container -outline-offset-2 bg-greyscale-50 flex flex-row gap-2 items-center pl-4 pr-2 flex-grow h-[--corpus-search-height] border-y border-greyscale-200{% if rounded %} rounded-lg border-x{% endif %} {{ class }}"
+  class="focus-container -outline-offset-2 h-[52px] bg-greyscale-50 flex flex-row gap-2 items-center pl-4 pr-2 flex-grow md:h-[--corpus-search-height] border-y border-greyscale-200{% if rounded %} rounded-lg border-x{% endif %} {{ class }}"
 >
   {% svg "magnifier" class="text-greyscale-400 flex-grow-0 h-5 w-5" aria_hidden="true" %}
   <div class="bg-greyscale-100 flex-grow" x-id="corpusInputIdGroup">


mlissner:
Two little things I see, and I'm being nitpicky because it's better to be that way now so that future PRs are smooth:

1. Your git commits should follow our best practice that's in our developer guide. See number 3, here: https://github.com/freelawproject/courtlistener/wiki/Getting-Started-Developing-CourtListener#guidelines-for-contributions

    Don't bother cleaning them up in this PR, we're not strict about it, but for future ones it'd be great to follow the convention.

2. Still need a signature on the CLA, please: 

https://github.com/freelawproject/courtlistener/pull/6264#issuecomment-3255914727

I think that's it for me, at least for now. Looks like this will be a good one for ironing out the process and little stuff.

Thanks!

cl/search/templates/cotton/corpus_search/scope.html
```diff
@@ -56,24 +56,37 @@
...
+      type="button"
+      x-bind:id="triggerButtonId"
+      x-on:click="openScopeMenu"
+      x-bind:aria-controls="menuId"
+      aria-haspopup="menu"
+      x-bind:aria-expanded="scopeMenuExpanded"
+      class="btn-outline border-0"
+      aria-label="Open menu to select the scope of your search"
+    >
+      <span class="text-sm" x-text="selectedScope"></span>
```
elisa-a-v: This is a duplicate of the one above and is not in the design.

cl/search/templates/cotton/corpus_search/scope.html
```diff
@@ -56,24 +56,37 @@
...
-      class="transition-transform duration-200 flex justify-center items-center w-4 h-4"
-      aria-hidden="true"
-    >{% svg 'chevron' class="w-full h-full text-greyscale-400" aria_hidden="true" %}
-    </span>
-  </button>
+  <div class="flex justify-between items-center mb-2">
+    <div class="flex flex-col items-start">
+       <span
+          x-text="selectedScope"
+          class=""
```
elisa-a-v: Let's not add empty attributes to keep the noise to a minimum please

elisa-a-v (changes requested):
Thanks for the cleanup. Sadly, this doesn't match the design yet, and also breaks the desktop header version.

### In the homepage:
- Chevron icon should be bigger.
- The entire selector should be clickable, not just the right half.
- There should be a single label for the selected scope, not two.
- Font should be bolder in the scope name label.
- The top padding in the card should be `10px` instead of `16px`.
- Instead of `margin-bottom: 8px`, this should have `padding-bottom: 10px`:

Image

<details>
<summary>Screenshots</summary>

#### Homepage version: design / this PR

Image

Image

</details>

### In the header:
- All the issues in the homepage are also present here.
- Spacing should be bigger between logo and scope selector.

<details>
<summary>Screenshots</summary>

#### Header version: design / this PR

Image

Image

</details>

### Dropdown menu

This one is on me cause I initially thought we didn’t have a design, but Figma does include one:

Image

## Other notes

- Since you’re also working from Figma, I’d expect any mismatches to be noticed by you too. If something isn’t clear, or what you see in Figma differs from what either I said or what is stated in an issue, please raise it so we can confirm together.
- Please keep markup semantic and meaningful. Divs and spans don’t convey information, and important elements like the selected scope need proper semantic tags and ARIA roles/attributes.
- This PR also breaks the desktop search bar:

Image

Please review the design carefully and address these points before we can move forward with this PR.

elisa-a-v:
I should note, the dropdown design is actually the same as the one that is currently used in the desktop version, so it's probably just a matter of cleaning up a few tailwind classes.

elisa-a-v:
> Looks like this will be a good one for ironing out the process and little stuff.

Totally! On that note, we use the assignee field quite a bit for PR reviews with the following simple workflow:
- Reviewers are assigned when the PR is ready.
- If the reviewers request changes, the assignees are cleared and the author is assigned.
- Rinse and repeat.

cl/search/templates/cotton/corpus_search/scope.html
bnjdev: Do you mind explaining what you mean by duplicate in the code?

elisa-a-v: I mean we have two spans showing the same `selectedScope`. The one above is the correct one (lines 61-64), but this one is not in the design.

cl/search/templates/cotton/corpus_search/scope.html
```diff
@@ -56,20 +56,33 @@
...
+
   <button
     type="button"
     x-bind:id="triggerButtonId"
     x-on:click="openScopeMenu"
     x-bind:aria-controls="menuId"
     aria-haspopup="menu"
     x-bind:aria-expanded="scopeMenuExpanded"
-    class="h-[--corpus-search-height] btn-outline hover:text-greyscale-900 hover:border-greyscale-300 hover:bg-greyscale-50 justify-between rounded-xl border-greyscale-200 {{ class }}"
+    class="md:h-[--corpus-search-height] btn-outline border-0 md:border hover:text-greyscale-900 hover:border-greyscale-300 hover:bg-greyscale-50 justify-between md:rounded-xl md:border-greyscale-200 px-0 py-[10px] pt-0 md:p-2 {{ class }}"
```
elisa-a-v:
```suggestion
    class="md:h-[--corpus-search-height] btn-outline border-0 md:border hover:text-greyscale-900 hover:border-greyscale-300 hover:bg-greyscale-50 justify-between md:rounded-xl md:border-greyscale-200 px-0 pb-2.5 pt-0 md:p-2 {{ class }}"
```

cl/search/templates/cotton/corpus_search/scope.html
```diff
@@ -56,20 +56,33 @@
...
+      class="flex flex-col items-start">
+       <span
+          x-text="selectedScope"
+          class="font-semibold md:font-normal"
+        ></span>
+        <span
+          x-text="selectedScopeDescription"
+          class="text-xs text-greyscale-500 md:hidden"
+        ></span>
+    </div>
```
elisa-a-v: `<div>` elements aren't allowed inside `<button>` tags cause they are block elements, but buttons only allow inline elements.

Could we replace the `div` with `span`?

Oh and this is very minor but could we also fix whitespace in these lines? Some of these indentation levels have 3 spaces, let's stick to 2 spaces everywhere please.

cl/search/templates/v2_homepage.html
```diff
@@ -36,8 +36,9 @@ <h1 class="w-full text-center lg:text-display-sm-cooper text-xl" x-text="tagline
     <div class="p-10 max-md:px-4">
       {#   MOBILE SEARCH   #}
       <div class="md:hidden bg-white rounded-2xl flex flex-col">
-        <c-corpus-search class="p-4 w-full !flex-col" only>
+        <c-corpus-search class="px-4 py-[10px] w-full !flex-col" only>
```
elisa-a-v:
```suggestion
        <c-corpus-search class="px-4 py-2.5 w-full !flex-col" only>
```

cl/search/templates/cotton/corpus_search/scope.html
bnjdev: Of course, that can be done. I implemented it that way because the button behaves like a block element on mobile devices. However, we can use spans as you suggested.

elisa-a-v:
@bnjdev I think something happened with the commit history 👀 

image

mlissner:
I always forget what causes that, but we can squash-merge, right, @elisa-a-v?

elisa-a-v:
Can't really review the changes like this though 😬

bnjdev:
> @bnjdev I think something happened with the commit history 👀
> 
> image

@elisa-a-v, the changes are here: https://github.com/freelawproject/courtlistener/pull/6264/commits/42f915b15b62a59038440a5c84a45d6635597f0d

cl/search/templates/cotton/corpus_search/scope.html
```diff
@@ -94,20 +106,20 @@
...
     x-transition:leave-end="opacity-0"
-    class="z-10 absolute mt-2 border border-greyscale-200 shadow-lg rounded-lg bg-white py-2 {{ menu_class }}"
+    class="z-10 absolute mt-0 sm:mt-2 border border-greyscale-200 shadow-lg rounded-lg bg-white py-2 {{ menu_class }}"
     role="menu"
     aria-label="Select the scope of your search"
   >
     <template x-for="scope in searchScopes">
       <li
         x-bind:key="scope.label"
-        class="{{ item_class }} px-4 flex hover:bg-greyscale-50"
```
elisa-a-v, line 104: I don't think `item_class` should have been removed from here, this changed the way it looks in desktop. I'm gonna ask you to please go over [this](https://github.com/freelawproject/courtlistener/wiki/New-Frontend-Architecture#component-attributes) part of the wiki in case you haven't.

### Before this PR
Image

### After this PR
Image

cl/search/templates/cotton/corpus_search/scope.html
```diff
@@ -123,4 +135,4 @@
   </menu>
 
 {% endif %}
-</div>
+</div>
```
elisa-a-v, line 138: Looks like the final newline at the end of the file was removed here. Could you restore it? I know it seems minor, but it’s best practice and avoids unnecessary diff noise.

mlissner:
I think the question is: The design is not accessible, but we implemented it so it was, and now we're tweaking it to match the design, and it's worse, therefore, should we keep the tweak that aligns with the design? If I have that right, yes, we should do that, and when we fix the design, we can fix the implementation.

elisa-a-v:
> The design is not accessible, but we implemented it so it was, and now we're tweaking it to match the design, and it's worse, therefore, should we keep the tweak that aligns with the design? If I have that right, yes, we should do that, and when we fix the design, we can fix the implementation.

I'm not entirely sure you got that right? There's two shades for the input in the design, one is a bit darker and it's the one in the desktop header. That _is_ in Figma, and it's the shade we were using before this PR everywhere.

Now, this PR matches the lighter shade that's in Figma for the mobile inputs and the homepage desktop, but it also changed the desktop header, so now instead of being the darker shade everywhere, it's the ligher shade everywhere.

I think we should just keep the darker shade that we had before, which is also in Figma, instead of having different colors throughout, none of which are compliant.

elisa-a-v:
Replaced by https://github.com/freelawproject/courtlistener/pull/6310

PR #6272, opened 2025-09-05
feat(new_header): enable advanced search in desktop header

### LINKED ISSUE #6072: Implement advanced search in desktop header
### Describe the issue.

View Figma for updated design with trigger button active state and overlay.

### Related Issue

_No response_

### Screenshots

Image

To Do

elisa-a-v:
## Fixes

Closes #6072

## Summary

This PR enables advanced search in the desktop header search bar. The filters are in a collapsible panel, with a trigger button inside the keyword input.

This PR also:
- Caps the width of the header and footer content, so they don't grow indefinitely which looks weird. This was required for the header to be aligned with the filters.
- Keeps the "Filters" button in the mobile header search UI visible even when the filters are open. This is completely unrelated but it was just removing one line so I thought it didn't really need its own PR.
- Enhances keyboard navigation and focus management in the relative date input:
    - Pressing <kbd>down</kbd> arrow while focusing the input or the trigger button moves focus inside the dropdown. If the value entered matches an option in the dropdown, said option is focused, otherwise the first option is focused.
    - Closing the dropdown menu returns focus to the input.
    - If the dropdown is open, pressing <kbd>esc</kbd> closes the dropdown; otherwise it clears the input value.
    - These behaviors are consistent with the expected keyboard interactions for the [editable combobox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role) and [listbox](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role) roles.
- Enables the `available_only` field in the RECAP Archive filters.
- Introduces a new composable for general purpose utilities, which has a function that triggers a given callback when the window passes the mobile breakpoint.

### Note about edge case

I noticed an edge case that I could not figure out how to solve, and it's that:
- If the panel has `overflow-auto`, it can scroll when the screen is smaller, but in larger screens the menus get cut.
- If the panel has `overflow-visible`, the menus are fully visible, but the panel isn't scrollable anymore, so some filters are lost.

This is what I mean by the menus getting cut:

<details>
<summary>Screenshot of partially blocked menus</summary>

image

</details>

I think it's safe to assume most windows that are wider are also taller, right? That means they're less likely to require scrolling, so I just made it so wider screens have `overflow-visible` and smaller ones `overflow-auto`. However, it is still possible in some screen resolutions for some content to get lost.

For example this is a screenshot of an entire window, and here we can't scroll because it's still desktop-sized, so some filters are out of view.

<details>
<summary>Screenshot of a small desktop-sized window, can't scroll</summary>

Image

</details>

Not sure how many devices have that resolution, for example iPads never need scrolling the filters:

<details>
<summary>iPad screenshots</summary>

image

image

</details>

So good enough? 🤷🏽 we could open an issue if we think this needs to be addressed, but I wouldn't block this PR for it.

### Question

We don't have a design for the visible active state in the Filters trigger button, so I improvised. Please note the screenshots showcasing the filters below were taken before adding the active state, same as the demo, but it looks like this now:

image

Does this work or do we want to tweak it?

## Deployment

**This PR should:**

  - [ ] `skip-web-deploy`
      
      
  - [x] `skip-celery-deploy`
      
      
  - [x] `skip-cronjob-deploy
      
     
  - [x] `skip-daemon-deploy`

<details closed>

<details open>
<summary><h4>Before / After</h4></summary>

Image

Image

</details>

<details open>
<summary><h4>Case Law Filters</h4></summary>

Image

</details>

<details open>
<summary><h4>RECAP Archive Filters</h4></summary>

Image

Image

Image

Image

</details>

<details open>
<summary><h4>Oral Arguments Filters</h4></summary>

Image

</details>

<details open>
<summary><h4>Case Law Filters</h4></summary>

Image

</details>

</details>

## Demo

https://github.com/user-attachments/assets/26e5c894-bf01-4fc6-a612-5f7f08a990ff


mlissner:
Looks amazing. I'm not sure I follow all of the details and corner cases completely, but I think it'll make sense once live, so let's chat through them in a meeting after launching (unless Eduardo has ideas first, of course).

Thanks!

mlissner (approved):
Looks good, thanks!

cl/assets/tailwind/input.css
```diff
@@ -11,9 +11,11 @@
     --mobile-main-padding: theme(spacing.4);
     --padded-full-width: calc(100vw - (2 * var(--mobile-main-padding)));
     --corpus-search-height: theme(spacing.12);
+    --desktop-header-height: theme(spacing.26);
+    --desktop-header-filters-height: calc(100svh - var(--desktop-header-height));
     --mobile-dialog-header-height: theme(spacing.18);
     --mobile-dialog-panel-height: calc(100vh - var(--mobile-dialog-header-height));
-    --content-max-width: 1400px;
+    --content-max-width: 1200px;
```
ERosendo: The tweak seems to have impacted the search component. Specifically, I've noticed that the date fields are smaller now, which causes the date to be cropped when the calendar component is used. (This issue is only present in the extra-large layout).  Here's a screenshot for your reference:

image

elisa-a-v:
Thanks for the comments!

I updated the max width to 1280px (pretty standard nowadays), so dates don't get cropped anymore. At first I thought we could have the same issue in smaller screens, but the number of columns is reduced before the inputs get too narrow so we're good.

As for the scrolling bug in Safari, I noticed it's happening even before this PR, and it's due to the live updates that the `x-collapse` directive performs, but I'm not sure why it only breaks in Safari.

After trying to fix it for a bit, I decided to simply get rid of the directive cause making it work proved to be much harder than I expected. This just means we don't have a smooth animation anymore, which is a bit of a bummer, but it feels like a reasonable compromise.

We could see if maybe a volunteer wants to use CSS transitions to add the animation back. Definitely doable, but getting it just right without breaking the scrollable filters is non-trivial.

PR #6275, opened 2025-09-06
fix(recap_into_opinions): remove flat=True in values_list()

quevon24:
## Summary

This PR fixes a values_list() method call in `recap_into_opinions` command, it removes the flat=True param because we are passing two fields. and this is causing a TypeError exception.

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/corpus_importer/management/commands/recap_into_opinions.py
+++ b/cl/corpus_importer/management/commands/recap_into_opinions.py
@@ -78,7 +78,7 @@ def import_opinions_from_recap(
                 is_free_on_pacer=True,
             )
             .only("id", "sha1")
-            .values_list("id", "sha1", flat=True)
+            .values_list("id", "sha1")
             .order_by("id")
         )
 


flooie:
@quevon24 is this assigned to someone?

flooie:
@quevon24 I cant get this test to pass

grossir (approved):
Just tested this on a shell, it has the expected shape

Setting all skips since it's a command that will be ran from a maintenance shell

```python
In [2]: from cl.search.models import RECAPDocument

In [3]: list(RECAPDocument.objects.filter(id=391262220).only("id", "sha1").order_by('id').values_list("id", "sha1"))
Out[3]: [(391262220, 'b92a2960cb1f764bccc3ac217dfee296b2247274')]

In
```

PR #6286, opened 2025-09-09
fix(proxy): Override send() to enforce HTTP protocol for all requests

### LINKED ISSUE #5567: ParserError when fetching appellate attachment pages
Yesterday, while debugging a problem reported by a user, I discovered that we weren't properly handling ParserError exceptions when fetching attachment pages. To address this, I opened the following PR: https://github.com/freelawproject/courtlistener/pull/5561. 

I used document **436559721** during my debugging process, as it was included in the error report submitted via email.

The request that led to my discovery regarding the unhandled exception can be found here: https://www.courtlistener.com/admin/recap/pacerfetchqueue/1586644/change/. The corresponding Sentry error report is available at: https://freelawproject.sentry.io/issues/6498118854/?project=5257254.

I retried the same fetch request for the attachment page and found that my second attempt was successful. This led me to believe that retrying the request could potentially resolve this issue, and I've included logic for this in the PR. The fetch queue that successfully retrieved the page is located here: https://www.courtlistener.com/admin/recap/pacerfetchqueue/1586728/change/

After the PR was merged, the user reported that the issue with appellate attachment pages persisted. I initially thought their credentials might be the problem. However, after attempting to fetch the same attachment page they mentioned in their second error report using different credentials, I noticed that the fetch was failing and the error message was updated after all retries were exhausted.

Interestingly, when I attempted to request the attachment page from my local development environment, the request was successful.

we should investigate it further to understand the underlying cause of this persistent issue. I'm creating this issue to ensure it's included in an upcoming sprint cycle and properly prioritized.

### LINKED ISSUE #5813: `PacerLoginException` on Appellate Courts
When debugging a user's question on https://github.com/freelawproject/courtlistener/discussions/5804,  I decided to broaden my investigation to other appellate courts. I discovered that fetch requests for `ca5`, `ca7`, and `ca11` are failing with a "PacerLoginException while getting pacer_case_id" error.

I don't believe this is a credential issue. I successfully purchased docket reports from other appellate courts using the same cached session, confirming my credentials were valid and active. Interestingly, these same failing requests succeed in my local development environment.

My hypothesis is that **Juriscraper** is encountering an unexpected response from the problematic courts, which triggers an unnecessary re-login attempt.

I'm creating this issue so we can investigate the interaction between our server and these courts.

ERosendo:
## Fixes

This PR fixes #5567

## Summary

This PR updates `ProxyPacerSession` to override the `send()` method,  ensuring that all requests (including those triggered by redirects) are processed through `_change_protocol`. This replaces HTTPS with HTTP and relies on the `X-WhSentry-TLS` header to instruct Webhook Sentry to establish TLS with the target server.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

## Additional notes

I used case [25-12872](https://www.courtlistener.com/docket/71200951/yolanda-delgado-v-the-evangelical-lutheran-good-samaritan-society/) in my local environment to test login, docket purchase, attachment purchase, and PDF purchase.

### DIFF
--- a/cl/lib/pacer_session.py
+++ b/cl/lib/pacer_session.py
@@ -68,6 +68,11 @@ class ProxyPacerSession(PacerSession):
         }
         self.headers["X-WhSentry-TLS"] = "true"
 
+    def send(self, request, **kwargs):
+        """Send a given PreparedRequest."""
+        request.url = self._change_protocol(request.url)
+        return super().send(request, **kwargs)
+
     def _pick_proxy_connection(self) -> str:
         """
         Picks a proxy connection string from available options.


mlissner:
Very nice, thanks.

PR #6290, opened 2025-09-09
feat(opinions): use llm to get correct case name for recap_into_opinions

quevon24:
## Summary

This PR adds the classify_opinion_case_name_task, which uses the LLM to extract the correct case name from the opinion content when possible. The task uses prompt templates stored in .txt files for the system and user prompts. Having the prompts in plain text files makes it easy to update and keep them separate from the code.

In the future, this design enables us to add placeholders inside these .txt prompt templates and dynamically replace them with context specific data before sending the prompt to the LLM. This will simplify prompt customization and enhance maintainability without modifying code logic.

Also included are retry logic for common OpenAI errors, safe database updates, and Sentry logging for better tracking of issues.

A few tests were added to cover normal cases, retries on rate limits.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


cl/corpus_importer/prompts/case_name_extract_user.txt
```diff
@@ -0,0 +1,51 @@
...
+
+Your tasks:
+1. Decide if this is a judicial opinion, memorandum, or order, or some other type of document.
+2. Extract:
+   - case_name: a normalized short style (e.g., "Smith v. Jones").
+   - case_name_full: the full caption exactly as it appears on the page, normalized:
+       * remove new lines
+       * remove excess whitespace
+       * convert all versus indicators (vs., versus, -against-) to "v."
+   - case_name_short: one or two words, taken from the non-governmental party if possible (e.g., defendant’s surname or company name).  Case Name Short should enver be shortened to a single Intial. like G.
```
grossir: typo: enver -> never

grossir: typo: Intial ->  Initial

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3062,128 @@ def recap_document_into_opinions(
...
+    :param recap_document_id: RECAPDocument id
+    """
+
+    system_prompt_path = (
+        "cl/corpus_importer/prompts/case_name_extract_system.txt"
+    )
+    user_prompt_path = "cl/corpus_importer/prompts/case_name_extract_user.txt"
+
+    try:
+        obj = OpinionCluster.objects.get(pk=pk)
```
grossir: since you are going to access the docket and the sub opinions, you can save a few DB queries by doing

```python
OpinionCluster.objects.filter(pk=pk).select_related("docket").prefetch_related("sub_opinions")
```

This can be further optimized by using a `Prefetch(queryset=Opinion.objects.only("plain_text"))` since you only care about the `plain_text`

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3062,128 @@ def recap_document_into_opinions(
...
+    case_name_full = (
+        llm_response.get("case_name_full")
+        if llm_case_name_match
+        else docket.case_name_full
+    )
+    case_name_short = (
+        llm_response.get("case_name_short")
+        if llm_case_name_match
+        else docket.case_name_short
+    )
```
grossir: this code means that the OpinionCluster names will be always overwritten
- if the LLM managed to extract names, by the extracted names
- by the docket's names

But since this is targeting OpinionClusters originating from RecapDocuments, the cluster's names are already the same as the docket's names; check a few lines before this task is called in `cl.corpus_importer.tasks.recap_document_into_opinions`

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3062,128 @@ def recap_document_into_opinions(
...
+    error_detected = llm_response.get("error", None)
+    needs_ocr = llm_response.get("needs_ocr", False)
+
+    if needs_ocr:
+        llm_response["recap_document_id"] = recap_document_id
+        logger.warning("Probably recap document needs OCR", llm_response)
+        return
```
grossir: early returns should go before all the name assignment logic. Taking the previous comment into account, I think this could be simplified by doing something like this

```python
    if llm_response.get("needs_ocr", False):
        llm_response["recap_document_id"] = recap_document_id
        logger.warning("Probably recap document needs OCR", llm_response)
        return

    if llm_response.get("error", None)
        return

    if not llm_response.get("case_name_match", False):
        # The name of the docket case and the opinion do not necessarily match, we log it for manual review
        llm_response["recap_document_id"] = recap_document_id
        logger.warning("Case name did not match", llm_response)
        return
    
    case_name = llm_response.get("case_name") or obj.case_name
    case_name_full = llm_response.get("case_name_full") or obj.case_name_full
    case_name_short = llm_response.get("case_name_short") or obj.case_name_short

    # Check if any of the values changed before updating the cluster
    changed = (
        obj.case_name_short != case_name_short
        or obj.case_name_full != case_name_full
        or obj.case_name != case_name
    )
```

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3062,128 @@ def recap_document_into_opinions(
...
+        llm_response["recap_document_id"] = recap_document_id
+        logger.warning("Case name did not match", llm_response)
+
+    # Check if any of the values changed before updating the cluster
+    changed = (
+        obj.case_name_short != case_name_short
+        or obj.case_name_full != case_name_full
+        or obj.case_name != case_name
+    )
+
```
grossir, line 3181: I think the check we demand to the LLM on the prompt should also be applied here

```python
if case_name_short and case_name and case_name_short not in case_name:
    logger.warning()
    
```

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3062,128 @@ def recap_document_into_opinions(
...
+    if needs_ocr:
+        llm_response["recap_document_id"] = recap_document_id
+        logger.warning("Probably recap document needs OCR", llm_response)
+        return
+
+    if not llm_case_name_match:
+        # The name of the docket case and the opinion do not necessarily match, we log it for manual review
+        llm_response["recap_document_id"] = recap_document_id
+        logger.warning("Case name did not match", llm_response)
```
grossir: shouldn't we use `logger.error` so it goes to Sentry? Since this is a task that will run everyday, instead of a one time command where we usually can easily keep the lgos

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3062,128 @@ def recap_document_into_opinions(
...
+    autoretry_for=(
+        RateLimitError,
+        APIConnectionError,
+        ConflictError,
+        InternalServerError,
+        APIError,
+    ),
+    retry_kwargs={"max_retries": 5},
+)
+def classify_opinion_case_name_task(self, pk: int, recap_document_id: int):
```
grossir: not sure if we need the `recap_document_id`

Since these are dockets generated from recap, it is reusing the Docket that contains the DocketEntry that links to the original RecapDocument. On top of that, the RecapDocument content itself was copied into the related Opinion; so we basically have the data we need to debug in the easy to reach related Opinion and Docket

mlissner: Perhaps:

```suggestion
def classify_case_name_by_llm(self, pk: int, recap_document_id: int):
```

mlissner:
Since this is our first PR quite like this, I think it'd be smart to have @rachlllg at least take a look. Do you mind giving a skim, Rachel?

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3062,128 @@ def recap_document_into_opinions(
...
+
+    :param self: The Celery task instance
+    :param pk: Primary key of the OpinionCluster object to process
+    :param recap_document_id: RECAPDocument id
+    """
+
+    system_prompt_path = (
+        "cl/corpus_importer/prompts/case_name_extract_system.txt"
+    )
+    user_prompt_path = "cl/corpus_importer/prompts/case_name_extract_user.txt"
```
mlissner: We should have a better way to import prompts. Notionally, it could be something like:

```
sys_prompt, user_prompt = get_prompts("case_name_extractor")
```

Or maybe:
```
sys_prompt = get_prompt("case_name_extractor", type="system")
user_prompt = get_prompt("case_name_extractor", type="user")
```

Not sure which is right or something else, but both are better than big long paths in our code. 

And, just stylistically, we should always use the `Path` module for this kind of thing too. If you ever see something like `some/path/to/a/file` it's a bug.

cl/lib/llm.py
```diff
@@ -0,0 +1,68 @@
...
+    - Only catches JSON decoding errors, not all exceptions.
+
+    :param s: JSON string, possibly wrapped with markdown code blocks.
+    :return: Parsed dictionary or None if parsing fails.
+    """
+    s_clean = s.replace("``````", "").replace("``````", "").strip()
+    try:
+        return json.loads(s_clean)
+    except json.JSONDecodeError:
+        return None
```
mlissner: I thought there was a way to force JSON from the LLM? Maybe @rachlllg has ideas here too?

cl/lib/llm.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+def call_llm(
```
mlissner, line 5: I think I'd make it clear in the function name that this is calling OpenAI's LLM. I also wonder if there's a library we should be using here to abstract this a bit.

cl/corpus_importer/tasks.py
rachlllg: I'm of the opinion that storing the prompts in .py files is better than .txt files, then you can import the prompt as a variable with no need to specify the path, and allow you to pass other variables into the prompt through format string (eg., f"The docket number you are reviewing is {docket_number}", allowing the docket_number to be dynamic). We can have one .py file for all the system prompts and one .py file for all the user prompts and just import whichever prompt is needed for the specific use case.

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3062,128 @@ def recap_document_into_opinions(
...
+        if llm_case_name_match
+        else docket.case_name_full
+    )
+    case_name_short = (
+        llm_response.get("case_name_short")
+        if llm_case_name_match
+        else docket.case_name_short
+    )
+    error_detected = llm_response.get("error", None)
+    needs_ocr = llm_response.get("needs_ocr", False)
```
rachlllg: More conservative to set needs_ocr to True?

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3062,128 @@ def recap_document_into_opinions(
...
+    if not error_detected or not needs_ocr and changed:
+        with transaction.atomic():
+            obj.case_name_short = case_name_short
+            obj.case_name_full = case_name_full
+            obj.case_name = case_name
+            obj.save()
+            logger.info(
+                "Case names successfully updated https://www.courtlistener.com/opinion/%s/decision/",
+                obj.id,
+            )
```
rachlllg: I may have missed this, but where is the LLM response stored if the case names do not match? I do see they are returned to the logger.warning, but in case there are many that don't match, would we be able to easily retrieve these records for analysis?

cl/lib/llm.py
rachlllg: I agree better to make it clear this is specific for OpenAI, in case we want to add other platforms which require different syntax. I'm not aware of any libraries that do this, it's fairly simple to call the api this way and is what the documentation recommends.

cl/lib/llm.py
rachlllg: Yes, almost all of the LLMs allow you to force json outputs, openai makes it very easy to do (see my comment below), other models (like claude) requires tool use. There is no need to clean the returned string, the returned json object has already been validated and can be directly loaded.

cl/lib/llm.py
```diff
@@ -0,0 +1,68 @@
...
+                {"type": "text", "text": prompt},
+                {"type": "text", "text": user_text},
+            ],
+        },
+    ]
+    # Use retries from the calling celery task
+    with OpenAI(max_retries=0) as client:
+        resp = client.chat.completions.create(
+            model=model,
+            response_format={"type": "json_object"},
```
rachlllg: I recommend change this to `{ "type": "json_schema", "json_schema": {...} }` and specify a json_schema with the specific output format you want. See documentation here: https://platform.openai.com/docs/guides/structured-outputs#structured-outputs-vs-json-mode.

cl/lib/llm.py
```diff
@@ -0,0 +1,68 @@
...
+            messages=messages,
```
rachlllg: what about `max_completion_tokens` and `temperature`?

`max_completion_tokens` allows you to better manage the cost, sometimes the model gets stuck in a loop and keep outputting nonsense until the `max_completion_tokens` is reached. By default 4o_mini has [16,384 max output tokens](https://platform.openai.com/docs/models/gpt-4o-mini), we probably want to limit this to somewhere around 1000 so we don't waste output tokens.

`temperature` defines how "creative" the model can be, by default it's set to 1 which is quite creative. For our tasks, we probably don't want the model to be creative at all, so I would set it to 0.1. See documentation https://platform.openai.com/docs/api-reference/chat/create#chat-create-temperature

cl/lib/llm.py
```diff
@@ -0,0 +1,68 @@
...
+        )
+    return resp.choices[0].message.content
```
rachlllg: In addition to the raw response, you might also consider logging the number of input tokens and output tokens to better manage cost, see documentation: https://platform.openai.com/docs/api-reference/responses/object#responses/object-usage, I would log the input tokens and output tokens separately as they are priced differently.

cl/lib/llm.py
flooie: can we set the temperature to 0 instead of 0.1?

rachlllg: I can't remember if it allows you to set to 0, but if it does, then yes 0 would be the best.

quevon24:
This is why I like code reviews, thanks everyone for the feedback, I've already been working on some changes, when I have them ready I'll bother you again with this 😅

cl/lib/llm.py
quevon24: I've been researching this and found some libraries to do this:

- Instructor (https://github.com/567-labs/instructor)
- LLM (https://github.com/simonw/llm)
- LiteLLM (https://github.com/BerriAI/litellm)
- LlamaIndex (https://github.com/run-llama/llama_index)

I'll investigate a little more but it seems that `instructor` is the best option due to its ease of use, it allows the use of different major providers (OpenAI, Anthropic, Gemini, Ollama, Groq), it supports sync and async, it is an active project, it integrates a library called pydantic for parsing and validating json.

rachlllg: LlamaIndex is what most people use in the field for this kind of thing, if we want to include this functionality. For most other models, Bedrock makes it easy to integrate & switch between models, the only downside is it won't have gpt models.

flooie: LlamaIndex looks like overkill if we are talking about simplifying our API calls across models.  

Have you tested any of these out yet? 

Instructor looks very nice but Im curious if it lives up to its promises.

rachlllg: I've only used llamaindex and [langchain](https://github.com/langchain-ai/langchain)

cl/corpus_importer/tasks.py
quevon24: This is in case that for some reason the case_name_match value of the llm is incorrect?

cl/corpus_importer/tasks.py
quevon24: The LLM response is sent along with the logger message to sentry

quevon24:
These are the changes I made:

- Refactored the existing `call_llm` function to use the Instructor library, this enables support for multiple llm providers beyond openai.
- Added parameters for `temperature` and `max_completion_tokens` with defaults
- Allowed passing a Pydantic response_model to parse and validate responses, returning structured typed models directly.
- Moved prompt templates from .txt files into Python variables stored in dedicated modules: `cl/corpus_importer/prompts/system.py` and `cl/corpus_importer/prompts/user.py`
- Defined CaseNameExtractionResponse Pydantic model with detailed fields matching expected LLM json output
- Updated task classify_case_name_by_llm to use instructor implementation in call_llm function
- Updated tests

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3066,144 @@ def recap_document_into_opinions(
...
+        obj = (
+            OpinionCluster.objects.select_related("docket")
+            .prefetch_related("sub_opinions")
+            .get(pk=pk)
+        )
+    except OpinionCluster.DoesNotExist:
+        return
+
+    sub_opinion = obj.sub_opinions.all().first()
+    if not sub_opinion or not sub_opinion.plain_text:
```
rachlllg, line 3128: Remind me why we are using the `plain_text` field instead of `with_best_text()`? I recall not all opinions have `plain_text`, is this functionality designed for only those with `plain_text`?

cl/lib/llm.py
```diff
@@ -0,0 +1,46 @@
...
+        },
+    ]
+
+    response = client.chat.completions.create(
+        messages=messages,
+        temperature=temperature,
+        max_tokens=max_completion_tokens,
+        response_model=response_model,
+    )
+    return response
```
rachlllg, line 55: Does Instructor allow us to log the input & output tokens for managing cost?

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3066,144 @@ def recap_document_into_opinions(
...
 
     # Return input task data to preserve the chain in scrape_pacer_free_opinion
     return task_data
+
+
+@app.task(
+    bind=True,
+    max_retries=5,
+    retry_backoff=True,
+    autoretry_for=(
```
rachlllg, line 3088: Since we are now using Instructor to call the LLMs, these errors are still specific to openai, would they still work? What if we switch to a different provider (say Anthropic)? Does Instructor provide its own error catching methods that we should use instead?

quevon24: Right now our Celery task is catching openai exceptions because the default model is openai/gpt-4o-mini and under the hood instructor just calls the OpenAI client. So those exceptions will still be raised and autoretry_for works as expected.

If we switch to another provider:

- instructor will raise that provider’s native exceptions (e.g., anthropic’s error types).
- It doesn’t unify or abstract error handling itself, it just delegates to the underlying SDK.

So if we ever use Anthropic, Mistral, etc., we’d need to update autoretry_for to include their exception classes. A generic fallback except Exception (with logging) is still in place, so we’d at least capture and track unexpected provider errors.

cl/lib/llm.py
quevon24: Yes, with OpenAI models, instructor preserves the .usage field so you can log prompt_tokens, completion_tokens, and total_tokens directly.

For other providers:

- Some (Anthropic, Groq) expose usage via .usage or headers.
- Others (Gemini, Ollama) don’t, so you’d need to count tokens manually with a tokenizer.

cl/corpus_importer/tasks.py
quevon24: This implementation is designed for recpa documents that are opinions. These come in pdf files, so it's necessary to extract the text (plain text or OCR). In this specific case, the extracted text is always stored in the plain_text field. You can see it here: https://github.com/freelawproject/courtlistener/blob/main/cl/corpus_importer/tasks.py#L3044

cl/lib/llm.py
rachlllg: ok, I should have clarified, so for this PR, we don't think it's necessary to log and track the usage?

quevon24: I'm not sure, should we consider logging and tracking the usage? @mlissner @flooie

mlissner: I think it's fine, generally. I guess long term this will be come difficult to figure out whether it's this or some other tool driving up our AI costs, but I don't know that tossing it in a log does us much good either, since we'd probably never look at that log. So I say let's just get this merged and we can worry about auditing our usage later.

cl/corpus_importer/tests.py
```diff
@@ -3915,3 +3922,149 @@ def test_can_get_recent_records_from_flat_models(self):
...
+            error=None,
+        )
+
+    @mock.patch.dict("os.environ", {"OPENAI_API_KEY": "123"}, clear=True)
+    @mock.patch("cl.corpus_importer.tasks.call_llm")
+    def test_classify_opinion_case_name_task_updates_fields(
+        self, mock_call_llm
+    ):
+        """Verifies that the classify_opinion_case_name_task updates OpinionCluster fields correctly"""
+        # mock_response = json.dumps(self.valid_response)
```
grossir: delete commented code

cl/corpus_importer/tasks.py
grossir: yes; LLMs can fail; and this is an easy check

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3066,144 @@ def recap_document_into_opinions(
...
+                "raw_response": llm_response.model_dump(),
+                "fingerprint": ["llm-casenames-llm-error"],
+            },
+        )
+        return
+
+    if not llm_response.case_name_match:
+        # The name of the docket case and the opinion do not necessarily match, we log it for manual review
+        llm_response["recap_document_id"] = recap_document_id
+        logger.error("Case name did not match", llm_response.model_dump())
```
grossir: I think you should prepend "LLM" so it's easier to scan Sentry issues. Something like "LLM case name did not match"

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3066,144 @@ def recap_document_into_opinions(
...
+
+    if not isinstance(llm_response, CaseNameExtractionResponse):
+        # Added this to avoid mypy errors
+        logger.error("Invalid llm_response type: %s", type(llm_response))
+        return
+
+    if llm_response.needs_ocr:
+        llm_response["recap_document_id"] = recap_document_id
+        logger.error(
+            "Probably recap document needs OCR", llm_response.model_dump()
```
grossir: Maybe drop the "Probably"? Something like "RecapDocument may need OCR" will make it easier to read the Sentry errors

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3066,144 @@ def recap_document_into_opinions(
...
+        )
+        return
+    except Exception as e:
+        # Only expect to get instructr exceptions here to track them
+        capture_exception(e)
+        raise
+
+    if not isinstance(llm_response, CaseNameExtractionResponse):
+        # Added this to avoid mypy errors
+        logger.error("Invalid llm_response type: %s", type(llm_response))
```
grossir: send some more context, probably the whole "llm_response" in the extra argument to the `logger.error`?

quevon24: It's a good idea, the problem is that when using instructor and pydantic is that you can't access the raw response, this would only fail if not all the values ​​of the CaseNameExtractionResponse model are returned and I think it would be strange, also if the response is wrong you would get a ValidationError in L3139, more than anything this was added to avoid mypy errors that cannot infer the type of the llm_response variable

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3066,128 @@ def recap_document_into_opinions(
...
+        if new not in (None, "") and old != new
+    }
+
+    if changed_fields:
+        with transaction.atomic():
+            for field_name, new_case_name in changed_fields.items():
+                # Update only necessary fields
+                setattr(obj, field_name, new_case_name)
+                # Check for overlap between cluster case name and extracted case name
+                overlap = winnow_case_name(new_case_name) & winnow_case_name(
```
grossir: you could compute the `winnow_case_name(cluster_name)` only once outside of the loop

cl/corpus_importer/tasks.py
```diff
@@ -3053,10 +3066,128 @@ def recap_document_into_opinions(
...
+    ]
+
+    # Check which values changed before updating the cluster
+    changed_fields = {
+        name: new
+        for name, old, new in case_names_pairs
+        if new not in (None, "") and old != new
+    }
+
+    if changed_fields:
```
grossir: could reverse and save an identation level

```python
if not changed_fields:
      return
```

grossir (approved):
Looks good, just a couple minor comments

Issue #6292, opened 2025-09-11
Add docket_number_raw to Docket


flooie:
### Describe the issue.

There’s interest in adding a new field to the Docket class: docket_number_raw.
This field would store all of the docket number values exactly as we’ve extracted them from our various sources.

With this change, the existing docket_number would effectively become a docket_number_cleaned field — a normalized value we could then use for linking cases across courts.

This would require a small model change we could wrap into the citation change. 

@rachlllg 

@quevon24 

### What type of issue is this?

Research

mlissner:
Thinking aloud, this would give us:

 - docket_number_raw: The raw value
 - docket_number: A cleaned value
 - docket_number_core: A different cleaned value

And then we have some `federal_dn_*` fields that contain bits and pieces of federal docket numbers. 

I think this all makes sense to me. I imagine adding this new field will mean making a copy of everything we've got in the current field into the new "raw" field, and then doing the cleanup.

flooie:
Yes I think that is correct mike.

quevon24:
> Thinking aloud, this would give us:
> 
> * docket_number_raw: The raw value
> * docket_number: A cleaned value
> * docket_number_core: A different cleaned value
> 
> And then we have some `federal_dn_*` fields that contain bits and pieces of federal docket numbers.
> 
> I think this all makes sense to me. I imagine adding this new field will mean making a copy of everything we've got in the current field into the new "raw" field, and then doing the cleanup.

Sounds good to implement it in the Citation model PR, just one question, Should we copy the value of `docket_number` to `docket_number_raw` within the migration or do it separately (directly on the db or using django shell)?

mlissner:
A change of that size to the DB will need to be done separately via a script and via an incremental approach (rather than one HUGE SQL query). Replicating huge queries is very difficult (we recently did this and regretted it).

Luis-manzur:
I haven't run/test this yet, but this is a way to do it in a migration without locking the db

```
def populate_docket_number_raw(apps, schema_editor):
    """
        Populates the new docket_number_raw field for all Docket records in batches.

        Batching is used to avoid locking the entire table and to reduce memory usage,
        which is important for large datasets. This approach ensures the migration
        remains efficient and minimizes database contention.
    """
    Docket = apps.get_model('search', 'Docket')
    batch_size = 10000
    last_pk = None
    qs = Docket.objects.order_by('pk')
    while True:
        batch = qs
        if last_pk is not None:
[...]

PR #6306, opened 2025-09-16
6292 add docket number raw to docket

Luis-manzur:
## Summary

This pull request introduces changes to the database schema and model definitions related to docket numbers in the `search` app. The main focus is on adding a new field, `docket_number_raw`, to the `docket` model to store the exact docket number values as extracted from various sources. Additionally, the migration updates database triggers to account for the new field.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

1. To deploy...

<details closed>
<details open>

</details>
<details open>

</details>


Luis-manzur:
```

batch_size = 10000
last_pk = None
qs = Docket.objects.order_by('pk')
while True:
    batch = qs
    if last_pk is not None:
        batch = batch.filter(pk__gt=last_pk)
    batch = batch[:batch_size]
    items = list(batch)
    if not items:
        break
    Docket.objects.filter(pk__in=[obj.pk for obj in items]).update(
        docket_number_raw=models.F('docket_number')
    )
    last_pk = items[-1].pk

```

this is the suggested script to set default values to docket_number_raw

cl/search/models.py
```diff
@@ -602,6 +602,14 @@ class Docket(AbstractDateTimeModel, DocketSources):
         blank=True,
         db_index=True,
     )
+    docket_number_raw = models.CharField(
+        help_text=(
+            "This field would store all of the docket number "
+            "values exactly as we’ve extracted them from our various sources."
```
grossir: I think this could be made shorter

"The raw docket number value as found on the source, with no cleaning or transformations applied"

cl/search/migrations/0042_add_docket_number_raw_field.sql
grossir, line 1: name of these files needs to be edited to something more precise?

grossir, line 1: missing `_customers.sql` file?
https://github.com/freelawproject/courtlistener/wiki/Database-migrations#steps-to-making-new-migrations

cl/search/models.py
```diff
@@ -602,6 +602,14 @@ class Docket(AbstractDateTimeModel, DocketSources):
...
+        ),
+        blank=True,
+        null=True,
```
grossir: this is CharField, so it shouldn't be null. Otherwise the queries get more complicated (instead of just using `WHERE docket_number_raw <> ''` you would have to use  `WHERE docket_number_raw <> '' and docket_number_raw IS NOT NULL`

grossir: I think this change will need to be paired with `default=''`

Luis-manzur:
> Is there anything to do in this PR about this [comment](https://github.com/freelawproject/courtlistener/issues/6292#issuecomment-3282715387)?

I think not, once the PR is approved and migrations applied, we would need to enter on the server and run the script to populate the new field.

grossir (approved):
Looks good!

PR #6324, opened 2025-09-23
build(deps): Adds `django-s3-express-cache` to dependencies

### LINKED ISSUE #6321: Add `django-s3-express-cache` to CourtListener
### Describe the issue.

This is just a little one: 

1. Add the cache as a dependency
2. Add it to the settings

### Related Issue

_No response_

### Screenshots

_No response_

To Do

ERosendo:
## Fixes

This PR fixes #6321 

## Summary

This PR introduces support for using django-s3-express-cache as a cache backend.

### Changes

- Adds [django-s3-express-cache](https://github.com/freelawproject/django-s3-express-cache) to project dependencies.

- Updated `settings.py` to include a `s3_cache` configuration:.
   - While integrating the config, I noticed that Django attempts to create an instance of the `s3_cache` during startup, this can lead to problems for contributors without S3 credentials. To avoid this, the cache is now only configured when not in `DEVELOPMENT` mode.

This PR depends on https://github.com/freelawproject/infrastructure/pull/489

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`


cl/settings/django.py
```diff
@@ -61,6 +61,14 @@
         "OPTIONS": {"MAX_ENTRIES": 2.5e5},
     },
 }
+
+DEVELOPMENT = env.bool("DEVELOPMENT", default=True)
+if not DEVELOPMENT:
+    CACHES["s3_cache"] = {
```
mlissner: Is it better to do:

```suggestion
    CACHES["s3"] = {
```

That matches the documentation and I think fits better alongside `db`?

ERosendo: you're right!

mlissner:
Ready to merge when infrastructure is ready.

mlissner:
Moving this back to you, Eduardo. I think you know where this stands?

ERosendo:
@mlissner This is ready to merge. I tested access to the bucket from one of our pods, and it’s working as expected. I’m not skipping deployments, since every pod should get the new dependency. I used the following code to test permission and connection: 

```python
import boto3
s3_express_client = boto3.client("s3")
s3_express_client.create_session(Bucket='com-courtlistener-cache--usw2-az1--x-s3')
```

mlissner:
Merged. The cache lives! Now to see if it works! 

Next time we do something like this, it wouldn't be crazy to do a little micro-view at `/monitoring/s3-cache/`, say, that we could just ping to exercise the basic functions. A thought!

PR #6332, opened 2025-09-23
fix(merge_opinion_versions): make text similarity threshold more flexible

### LINKED ISSUE #6138: Optionally make opinion versioning by URL more flexible
See[ `ga` opinion versions](https://github.com/freelawproject/courtlistener/issues/5245#issuecomment-3160919456) where we fail to merge some obvious versions due to the text differences

We could add an OR to the merging criteria. Currently the criteria is
- same url
AND
- passing the text similarity threshold

We could change the second clause to

- passing the text similarity threshold  
OR (
- passing a lower text similarity threshold  AND  same case name, same date filed, same ?
)

grossir:
Fixes #6138

- refactored `is_similar_text` to `get_text_similarity`, that now takes into account 2 thresholds. A strict one of 0.9 and a loose one of 0.7
- if the texts are similar according to the loose one, the merging of opinions will only happen if all metadata fields are the same for the opinion and the opinion cluster
- improve the `explain_version_differences` utility, so it only prints verbose output when asked for it; and so it can handle other content fields than `plain_text`


cl/scrapers/management/commands/merge_opinion_versions.py
```diff
@@ -70,17 +82,29 @@ def explain_version_differences(opinions: list[Opinion | int]) -> None:
         prev_key = ""
         diff = ""
         print(f"Difference between {op1} {op2}")
-        for index, i in enumerate(differ.compare(text1, text2)):
-            if i[0] in "-+?":
-                if prev_index == index - 1 and prev_key == i[0]:
-                    diff += i[-1]
-                else:
+        base_cl_url = "https://www.courtlistener.com/opinion/"
+        print(f"{base_cl_url}{id_to_op[op1].cluster_id}/x/")
```
quevon24, line 86: Shouldn't these prints go inside verbose condition?

quevon24 (approved):
All good, just a small comment related to verbose.

cl/scrapers/management/commands/merge_opinion_versions.py
grossir: I think they are OK there. This `print` prints a couple lines per pair of opinions, useful for debugging; The verbose part will print every difference, which may amount to a few hundred lines, or more, depending on the opinion pairs

PR #6359, opened 2025-10-02
fix(cl_scrape_opinions): do not update site hash if an opinion fails downloading

### LINKED ISSUE #6357: cl_scrape_opinions: do not save site hash when a document failed downloading
### Describe the issue.

A document may fail downloading due to a `BadContentError`, which is the parent class of `UnexpectedContentTypeError` . This happens when we hit an error page (most recently, happened due to bad URLs in `ariz`, reported by an user). We correctly avoid ingesting the error page, but we are saving the site hash. 

This makes following scrapes ignore the site if there are no new cases, even if it hasn't ingested the failing documents. For the `ariz` case mentioned above, I had to manually use `--fullcrawl` to trigger the scrape. 

This is probably causing ocasional missing opinions, and we can fix it by not saving the site hash if there was a `BadContentError`

https://github.com/freelawproject/courtlistener/blob/e3bb67332e501aa194159c10b659a33b405ad409/cl/scrapers/management/commands/cl_scrape_opinions.py#L282-L294

grossir:
Fixes #6357

During a scrape, if an opinion fails downloading with BadContentError or InvalidDocumentError, do not update the site hash to ensure a retry on the next scrape

### DIFF
--- a/cl/scrapers/management/commands/cl_scrape_opinions.py
+++ b/cl/scrapers/management/commands/cl_scrape_opinions.py
@@ -271,6 +271,9 @@ class Command(ScraperCommand):
 
         logger.debug("#%s %s found.", len(site), self.scrape_target_descr)
 
+        # do not update the site hash on a backscrape or manual full crawl
+        update_site_hash = not full_crawl
+
         added = 0
         for i, item in enumerate(site):
             try:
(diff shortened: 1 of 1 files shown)


Luis-manzur (approved):
nie work!

PR #6374, opened 2025-10-06
docs(recap_fetch_api): inform users of new PACER policy

### LINKED ISSUE #6277: Update RECAP Fetch API documentation to mention password rotation
### Describe the issue.

The judiciary made a new policy that requires that all PACER accounts rotate their passwords every 180 days. We need to document this on the RECAP Fetch API and give guidance to people about what to do.

We will deal with this by:
 - using two accounts to access PACER, selecting one of them randomly for each request
 - have a system for an account to be marked as dead if it is unable to authenticate
 - when that account is marked as dead, we will start throwing warning emails until it is fixed

The advice we can give to fetch API users is that they will need to rotate their passwords at least every 180 days, and that a second account will allow them to seamlessly swap accounts.

### Related Issue

_No response_

### Screenshots

_No response_

To Do

elisa-a-v:
## Fixes

Closes #6277 

## Summary

A section about PACER password rotation is added in the RECAP Fetch API docs.

<details>
<summary>Screenshot</summary>

image

</details>

### Questions

- The section was added between "Monitoring your request" and "Security of Fetch API". Is that the right place for this notice?
- I don't fully understand how the suggestion that users create two accounts would help with this, so I haven't added anything about that yet. Could you explain how that is helpful so I can better document it?
- The rest of the context provided in the issue talks about what _we_ will do internally (using multiple accounts selected at random, etc), but AFAIK those features are not implemented yet, so I didn't document that anywhere (and I assume the right place for that would be somewhere else like the wiki maybe?). Am I getting this right, or should I add that somewhere?

## Deployment

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/api/templates/recap-api-docs-vlatest.html
+++ b/cl/api/templates/recap-api-docs-vlatest.html
@@ -35,6 +35,7 @@
       <li><a href="#pacer-fetch">PACER Fetch API</a></li>
       <ul>
         <li><a href="#monitoring">Monitoring</a></li>
+        <li><a href="#password-rotation">Password Rotation Requirement</a></li>
         <li><a href="#security">Security</a></li>
         <li><a href="#known-issues">Known Issues</a></li>
         <li><a href="#pdf">Purchasing PDFs</a></li>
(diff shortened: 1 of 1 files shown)


cl/api/templates/recap-api-docs-vlatest.html
```diff
@@ -127,6 +128,19 @@ <h3 id="monitoring">Monitoring Your Request</h3>
...
+    Because the RECAP Fetch API uses your PACER credentials to log in and retrieve documents, this policy affects all Fetch API users.
+  </p>
+  <p>This means:</p>
+  <ul>
+    <li>You'll need to <strong>update your PACER password at least once every 180 days</strong>.
+    </li>
+    <li>If your password expires, the Fetch API will no longer be able to log in on your behalf until you update it.
+    </li>
+  </ul>
+
```
mlissner, line 144:
```suggestion
  <p>To use this API without downtime during password change events, we recommend using two PACER accounts, and rotating between them in your code.</p>

```

mlissner:
> I don't fully understand how the suggestion that users create two accounts would help with this, so I haven't added anything about that yet. Could you explain how that is helpful so I can better document it?

Imagine that you have an automated system scraping things from PACER and that you have to change it's password every 30 days. 

If you change the password late, your scraper will go down while your password is expired. That should be avoided.

If you change it on the 29th day, your process will be:

1. Use the password to log into PACER.
2. Change the password in PACER.
3. YOUR SCRAPER IS NOW FAILING.
4. Update the password in your scraping code.

So there's a period of time that's bad during step three.

If you have two accounts, it becomes:

1. Set your scraper using account A.
2. The day before account A expires, change the password for account B, so it's fresh.
3. Swap account A for account B in your scraper.
4. The day before account B expires, refresh the password for account A, so it's fresh.
5. Swap account B for account A in your scraper.

And so forth. No down time.

> The rest of the context provided in the issue talks about what we will do internally..., but AFAIK those features are not implemented yet, so I didn't document that anywhere 

Yeah, that's good.

elisa-a-v:
Got it, thank you for explaining. I've committed your suggestion 🫡 thanks for providing one!

mlissner:
Thank you, Elisa. Looks great. Merging.

PR #6387, opened 2025-10-09
feat(search): Skips versioned opinions in bulk embeddings indexing

### LINKED ISSUE #6328: Remove versioned opinions from ES indexing when indexing embeddings
### Describe the issue.

Versioned opinions are now removed from ES index, we are now generating embeddings for all case laws so we just need to make sure any versioned opinions are not indexed to ES when we are indexing the embeddings (we only need the main version indexed, not the sub versions)

To Do

ERosendo:
## Fixes

This PR fixes #6328 

## Summary

This PR updates the `index_embeddings` Celery task to skip opinions that have a `main_version` when performing bulk embeddings indexing.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`


ERosendo:
I think it's fine to skip the deployment for now, since we won't be using this command immediately.

ERosendo:
> One suggestion, could we also add the same condition to the generate_opinion_embeddings command? That way, if we use it again to generate embeddings, it won’t create embeddings for versioned opinions.

Thanks @albertisfu. This is a great idea. I'll tweak the command.

PR #6396, opened 2025-10-14
test(new_homepage): introduce baseline test suite

### LINKED ISSUE #5721: Home Page - Test new template
In #5643 the middleware that swaps templates was modified to prevent swapping when testing: a new home page stub was added, so tests were failing because none of the expected content was there, which is why we temporarily disabled swapping for tests.

We need to make sure that **both** home pages are properly tested.

elisa-a-v:
Closes #5721 

## Summary

This PR introduces a baseline test suite for the new homepage that covers:
- View routing.
- Transitional behavior (our "contract" while other related new pages are still pending, like search results)
- Basic structure
- Header variants
- Stats block rendering

Also updates the waffle flag default back to `False`. Since the waffle setting was changed, some unrelated tests started failing because they relied on automatically created active flags. @ERosendo and I reviewed all of them and found that only one of the flags was still being used (so the failing test now overrides it) and we cleaned up the rest.

The flags that were removed and can be safely deleted from prod after merge are:
- `enable_semantic_search`
- `ui_flag_for_o_es`
- `recap-alerts-active`

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`


mlissner (approved):
LGTM, but let's see what somebody else thinks! Glad to see this landing. 

When we deploy this, let's make sure we tell everybody about the waffle changing its default (even though we're changing to a more safe version).

PR #6399, opened 2025-10-15
feat(embedding): Skip embedding computation for docs below threshold

### LINKED ISSUE #6397: Unexpected data type for embeddings: <Response [422 Unprocessable Entity]>
Sentry Issue: [COURTLISTENER-B4C](https://freelawproject.sentry.io/issues/6944437072/?referrer=github_integration)

```
Unexpected data type for embeddings: <Response [422 Unprocessable Entity]> (<class 'httpx.Response'>)
```

@albertisfu noticed this error report this morning. It seems the issue is related to the opinion’s content — it has only occurred once so far:

- [Opinion 11169846](https://www.courtlistener.com/admin/search/opinion/11169846/change/)

The record’s content is just `<div></div>`. Alberto suggested refining the tasks that send requests to Inception so We can catch cases like this using the token count and return early.

ERosendo:
## Fixes

This PR fixes #6397 

## Summary

This PR adds a token count check to the `compute_single_opinion_embeddings` task to prevent embedding generation for opinions that fall below the token threshold.

Key changes: 

- Introduces a conditional check that skips embedding computation when opinion.token_count is below `settings.MIN_OPINION_SIZE`.

- Ensures unnecessary microservice calls are avoided for trivial or low-content opinions.

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/search/tasks.py
+++ b/cl/search/tasks.py
@@ -1823,6 +1823,9 @@ def compute_single_opinion_embeddings(self, pk: int) -> None:
     if not opinion:
         return None
 
+    if opinion.token_count < settings.MIN_OPINION_SIZE:
+        return None
+
     embeddings = asyncio.run(
         microservice(
             service="inception-text",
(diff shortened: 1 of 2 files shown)


ApacheAde:
Data breach.

PR #6413, opened 2025-10-16
Texas Second Court of Appeals model expansion

holtchesley:
# Adding Texas Second Court of Appeals Dockets to CourtListener

## Overview

This PR takes a first stab at adding support for state level dockets to CourtListener by way of the Texas Second Court of Appeals.
The primary goals of this endeavor are to gather the kinds of information that should be considered when expanding CL's data model to balance
1. Minimizing the number of models that need to be maintained while making sure they are not overly complicated.
2. Provide clear extension points within those models so that different state information and behavior can be stored and handled.
3. Maintaining acceptable performance of the CourtListener platform.

## New Models

To those ends, the existing data models are reused and lightly augmented where possible, and two new Models added where there is not an existing model that covers a similar concept.

1) AppellateBrief

Texas accounts for briefs that were expected but might not have been provided. This differs from DocketEntry, which should correspond currently to a document. As such, I've factored out this model to capture that. As this effor expands, I can see this model as evolving in one of two ways, either expanding from just covering about Appellate briefs (if we find other courts declare other documents that they expect and do not receive), or merging into DocketEntry and expanding the semantics of that model to include a missed expectation (if the class of documents across many courts gets much bigger).

2) CalendarEntry

Texas keeps a timeline of events and upcoming deadlines. These do not appear to have documents associated with them at all. This distinctly seperate from DocketEntry, and seems initially unlikely to evolve towards it.

## New Fields

There are additionally a few new fields and relations that have been added to capture information available from TAMES.

1) Docket: transfer_to, transfer_date

This captures the results of a successful motion to transfer venue, which shouldn't be limited to just Texas. This relation allows for multiple transfers in a linked list style structure. Of note, if we do stumble on many cases with multiple venue transfers, this may need to be revisited for performance considerations, as recursive CTEs aren't usually very snappy. Coalescing the entire list on updates is one possible solution to this, or having a one-to-many relation for all eventual-transfers. Django's lazy loading may also make this not a big issue if it's sparesely populated and long chains aren't often queried.

2) Docket: is_original_proceeding

This makes explicit if the case is known to be original (True), an appeal (False), or unknown (None/null), as TAMES gives us that information.

3) OriginatingCourtInformation: sentence, county

TAMES surfaces information about the sentence/punishment for appealed cases, as well as the county they are from. These fields are captured in the OriginatingCourtInformation model. Further expansion to more states may reveal a better name for the county field (perhaps jurisdiction_unit?).

4) AbstractPDF: source_url

TAMES makes publicly available the documents that we'll be using for DocketEntries and AppellateBriefs, and asks for attribution. Capturing the source_url lets us easily do that.

5) TAMES added as a SOURCE

While technically this isn't a field on a model, it is important that this be added to track the provenance of cases. This class seems like a good place to coordinate dynamic dispatch of behavior that varies depending on source.

## Future Work

This PR forms a start to exapanding the existing data model to cover state level cases. The next steps I would take to expand upon this are:
1) This same thing for a few more states that might capture a bit more diversity of procedure. Louisiana, California, New York, and Alaska come to mind.
2) Scrape cases from Texas and determine if the performance of transfer_to inference will be an issue.
3) Do an initial implementation for searching through CalendarEntry and AppellateBriefs, and determine if there are additional indexes warranted.
4) Reconsider the AppellateBrief model after looking at a few more examples of state courts.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/lib/models.py
+++ b/cl/lib/models.py
@@ -120,7 +120,7 @@ class AbstractPDF(models.Model):
         help_text="The URL this PDF was stored at before import, if known and public",
         max_length=1000,
         null=True,
-        blank=True
+        blank=True,
     )
 
     class Meta:
--- a/cl/search/models.py
+++ b/cl/search/models.py
@@ -238,9 +238,10 @@ class SOURCES:
             RECAP,
             "recap",
         ),
-        (   TAMES,
+        (
+            TAMES,
             "Texas courts public court records at search.txcourts.gov",
-         ),
+        ),
     )
 
     # use a frozenset since the order of characters is arbitrary
@@ -368,15 +369,14 @@ class OriginatingCourtInformation(AbstractDateTimeModel):
     sentence = models.TextField(
         help_text="The punishment or sentence determined by the originating court",
         blank=True,
-        null=True
+        null=True,
     )
     county = models.CharField(
         help_text="The name of the county of the originating court, if available",
         max_length=100,
         blank=True,
-        null=True
+        null=True,
     )
-
 
     @property
     def administrative_link(self):
@@ -457,7 +457,7 @@ class Docket(AbstractDateTimeModel, DocketSources):
     is_original_proceeding = models.BooleanField(
         help_text="Whether this is an original proceeding rather than an appeal, if known",
         default=None,
-        null=True
+        null=True,
     )
     appeal_from_str = models.TextField(
         help_text=(
@@ -1715,6 +1715,7 @@ class RECAPDocumentTags(RECAPDocument.tags.through):
     class Meta:
         proxy = True
 
+
 @pghistory.track()
 class AppellateBrief(AbstractDateTimeModel, AbstractPDF):
     """The model class for tracking Appellate Briefs"""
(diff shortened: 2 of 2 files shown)


cl/lib/models.py
```diff
@@ -116,6 +116,12 @@ class AbstractPDF(models.Model):
         null=True,
         blank=True,
     )
+    source_url = models.CharField(
+        help_text="The URL this PDF was stored at before import, if known and public",
+        max_length=1000,
+        null=True,
+        blank=True
```
holtchesley: null, blank=True should make most of these migrations a bit easier.

cl/search/models.py
```diff
@@ -423,6 +439,26 @@ class Docket(AbstractDateTimeModel, DocketSources):
         null=True,
         related_name="child_dockets",
     )
+    transfer_date = models.DateField(
+        help_text="The date the case was transfered to another court if applicable",
+        db_index=True,
```
holtchesley, line 444: This can be revisited, but dates have been indexed as they seem likely to be a fairly common part of queries.

cl/search/models.py
```diff
@@ -1678,6 +1715,121 @@ class RECAPDocumentTags(RECAPDocument.tags.through):
...
+    )
+
+    def __str__(self) -> str:
+        return f"{self.entry_type} of {self.reason_set} on {self.entry_date} for Docket {self.docket_id}"
+
+    class Meta:
+        verbose_name_plural = "Calendar Entries"
+        ordering = ["entry_date"]
+        indexes = [
+            models.Index(fields=["docket_id", "entry_date"])
```
holtchesley: The most likely query pattern /ORM access pattern I could see immediately.

Issue #6318, opened 2025-09-19
Implement batch docket number cleaning with LLM


rachlllg:
### Describe the issue.

Alberto provided really great guidance and suggestions on this process.

After the real-time docket number cleaning process https://github.com/freelawproject/courtlistener/issues/6317, we will:
1. using a daemon job to fetch the s3 cache of `docket_id` that needs to be cleaned with LLM and invoke the LLM API call to clean `docket_number_raw` and replace `docket_number` with the cleaned version
2. the `docket_id` is cleared from the cache if the docket_number cleaning was successful
3. if the cleaning was not successful (either because the LLM API was down, or the LLM failed to return a cleaned docket number), the `docket_id` is added to the next batch for cleaning
4. the daemon job will be invoked when a set number of docket_ids have been accumulated or when a certain time has passed
5. the daemon job need to also ensure the same `docket_id` is not processed by multiple jobs concurrently (ie, if it's already being processed by one of the earlier jobs that hasn't yet finished when the new job started)
6. The celery task is incorporated as part of the daemon job for distributed processing

Notes:
- Daemon job is preferred over a cron job as it allows more flexible handling (vs always invoked on a timed schedule)

This is the issue to create the LLM script with daemon job & celery task + tests. It's sized as a M but it may take me a bit more time as I get up to speed.

### What type of issue is this?

To Do

mlissner:
Nice. Daemons are nice. Better to have something running all the time than intermittently, I agree. 

Why use the S3 cache for the docket IDs? It'll be slower, but more importantly if you use redis, you can do fun things like LPUSH: https://redis.io/docs/latest/commands/lpush/ and RPOP: https://redis.io/docs/latest/commands/rpop/

rachlllg:
> Why use the S3 cache for the docket IDs?

Mostly for fear of running out of space because the ids could build up in case the openai api is down for days on end, or when we ingest a huge batch of dockets.

PR #6415, opened 2025-10-17
6318 llm clean docket number raw

rachlllg:
## Fixes
Fixes https://github.com/freelawproject/courtlistener/issues/6318

## Summary
This PR introduces a new Daemon job for using LLM to clean docket number in batches. 
- These are the docket numbers that could not be cleaned by the post save signal using regex https://github.com/freelawproject/courtlistener/issues/6317. 
- `docket_id` are retrieved from the redis cache, `docket_number_raw` are retrieved from DB along with `court_id`. 
- LLM calls are made for the respective court mapping.
- It uses a cascading model call approach by using two small models to reach consensus first, then use two full models for those that the two small models disagreed, then use a final tie-breaker model with revised prompt to break the tie.
- `docket_number` is updated in DB directly, with timestamp used to handle potential race conditions, where the `docket_number` is not updated (and `docket_id` not removed from redis cache) if the docket has changed since the Daemon job started (by comparing against the `date_modified` field in DB). 
- Dockets are batched by court_mapping (Fed Appellate, Fed Trial, etc), and each court batch is handled by one Celery task. Mike suggested using an existing batch job Celery queue, I assigned batch2 for now.
- ThreadPool is used to parallelize model calls to increase throughput.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [X] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/search/docket_number_cleaner.py
+++ b/cl/search/docket_number_cleaner.py
@@ -255,7 +255,7 @@ def clean_docket_number_raw(
 def create_llm_court_batches(
     llm_batch: list[int],
     r: Redis,
-) -> dict[str, list[dict[int, str]]]:
+) -> defaultdict[str | None, list[dict[int, str]]]:
     """
     Groups docket numbers by their associated court type for a given batch of docket_ids.
     Remove docket_ids that no longer exist, if any, from the Redis set.
@@ -265,7 +265,6 @@ def create_llm_court_batches(
     :return: A dictionary where each key is a court mapping (as determined by `court_map`),
               and each value is a list of dictionaries of {docket_id: docket_number_raw}.
     """
-    court_batches = {}
     existing_docket_ids = set()
     dockets = Docket.objects.filter(pk__in=llm_batch).values_list(
         "id", "court_id", "docket_number_raw"
@@ -314,16 +313,16 @@ def extract_with_llm(
                 "fingerprint": ["llm-docket-number-validation-error"],
             },
         )
-        return
+        return None
     except Exception as e:
         # Only expect to get instructor exceptions here to track them
         capture_exception(e)
-        return
+        return None
 
     if not isinstance(llm_response, CleanDocketNumber):
         # Added this to avoid mypy errors
         logger.error("LLM - Invalid response type: %s", type(llm_response))
-        return
+        return None
 
     records = {}
     for item in llm_response.docket_numbers:
@@ -361,7 +360,7 @@ def process_llm_batches(
         for i in range(0, len(llm_batches), batch_size)
     ]
     with ThreadPoolExecutor(
-        max_workers=settings.DOCKET_NUMBER_CLEANING_MAX_WORKERS
+        max_workers=settings.DOCKET_NUMBER_CLEANING_MAX_WORKERS  # type: ignore
     ) as executor:
         future_to_batch = {
             executor.submit(
@@ -434,8 +433,8 @@ def call_models_and_compare_results(
         system_prompt=system_prompts.get(court_mapping, ""),
         model_id=model_one,
         retry=0,
-        max_retries=settings.DOCKET_NUMBER_CLEANING_LLM_MAX_RETRIES,
-        batch_size=settings.DOCKET_NUMBER_CLEANING_LLM_BATCH_SIZE,
+        max_retries=settings.DOCKET_NUMBER_CLEANING_LLM_MAX_RETRIES,  # type: ignore
+        batch_size=settings.DOCKET_NUMBER_CLEANING_LLM_BATCH_SIZE,  # type: ignore
         all_cleaned=dict(),
     )
     model_two_results = process_llm_batches(
@@ -443,8 +442,8 @@ def call_models_and_compare_results(
         system_prompt=system_prompts.get(court_mapping, ""),
         model_id=model_two,
         retry=0,
-        max_retries=settings.DOCKET_NUMBER_CLEANING_LLM_MAX_RETRIES,
-        batch_size=settings.DOCKET_NUMBER_CLEANING_LLM_BATCH_SIZE,
+        max_retries=settings.DOCKET_NUMBER_CLEANING_LLM_MAX_RETRIES,  # type: ignore
+        batch_size=settings.DOCKET_NUMBER_CLEANING_LLM_BATCH_SIZE,  # type: ignore
         all_cleaned=dict(),
     )
 
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -77,4 +77,6 @@ jobs:
           cl/search/management/commands/cl_index_parent_and_child_docs.py \
           cl/search/management/commands/sweep_indexer.py \
           cl/search/management/commands/pacer_bulk_fetch.py \
-          cl/search/management/commands/generate_opinion_embeddings.py
+          cl/search/management/commands/generate_opinion_embeddings.py \
+          cl/search/management/commands/llm_clean_docket_number_daemon.py \
+          cl/search/docket_number_cleaner.py


cl/search/docket_number_cleaner.py
```diff
@@ -176,3 +244,208 @@ def clean_docket_number_raw(
...
+    """
+    Groups docket numbers by their associated court type for a given batch of docket_ids.
+
+    :param llm_batch: A list of docket_ids to be processed.
+    :return: A dictionary where each key is a court mapping (as determined by `court_map`),
+              and each value is a list of dictionaries of {docket_id: docket_number_raw}.
+    """
+    court_batches = {}
+    for docket_id in llm_batch:
+        docket_obj = get_docket(docket_id)
```
albertisfu: Instead of doing a for loop and fetching the docket from the database on each iteration, I think we can reduce the database load by performing a single query to retrieve the dockets with the selected required fields, and then iterate over those objects only.

So, we can have a query like:
```
dockets = (
        Docket.objects
        .filter(pk__in=llm_batch)
        .values_list("id", "court_id", "docket_number_raw")
    )
```


And then iterate over this list of dockets to build `court_batches`:

` for docket_id, court_id, raw in dockets:`

In case llm_batch is expected to be large, on the order of more than 2,000 IDs, you can use `dockets.iterator()` to prevent loading all the dockets into memory at once.

cl/search/docket_number_cleaner.py
```diff
@@ -176,3 +244,208 @@ def clean_docket_number_raw(
...
+        if not docket_obj:
+            continue
+        court_id, docket_number_raw = (
+            docket_obj.court_id,
+            docket_obj.docket_number_raw,
+        )
+        court_mapping = court_map.get(court_id, None)
+        if court_mapping not in court_batches:
+            court_batches[court_mapping] = []
+        court_batches[court_mapping].append({docket_id: docket_number_raw})
```
albertisfu: Then, if you use `defaultdict`, you can simplify this block of code. You could have something like:

```
 court_batches = defaultdict(list)
 for docket_id, court_id, raw in dockets:
      court_batches[court_map.get(court_id)].append({docket_id: raw})
```

Also, my question here is whether it’s expected to have a `court_batch` whose key is `None` in case `court_map.get(court_id)` returns `None`. Or, in those cases, should we omit appending the docket IDs to `court_batches`?

cl/search/docket_number_cleaner.py
```diff
@@ -30,6 +47,57 @@
...
+) -> None:
+    """
+    Update the docket_number field of a Docket object & remove it from the LLM cleaning Redis set if it hasn't been modified since the start timestamp.
+
+    :param docket_id: The primary key ID of the Docket to update.
+    :param docket_number: The new docket number to set.
+    :param start_timestamp: The timestamp marking the start of the daemon job.
+    """
+    docket = get_docket(docket_id)
+    r = get_redis_interface("CACHE")
```
albertisfu: I suggest moving `r = get_redis_interface("CACHE")` outside this method and passing it as a method argument instead. This way, the same Redis connection will be used for all calls within the for loop where `update_docket_number` is invoked.

cl/search/docket_number_cleaner.py
```diff
@@ -30,6 +47,57 @@
...
+
+    :param docket_id: The primary key ID of the Docket to retrieve.
+    :return: The Docket object if found, otherwise None.
+    """
+    try:
+        docket_obj = Docket.objects.filter(id=docket_id).first()
+        return docket_obj
+    except Docket.DoesNotExist:
+        logger.error(f"Docket with id {docket_id} does not exist.")
+        return None
```
albertisfu: A couple of suggestions here: if we're using `objects.filter().first()`, the `Docket.DoesNotExist` exception won't be triggered since it returns None instead. We should use `.get()` instead.

Also, I recommend using `.only()` so that only the required fields are selected, making the query a bit more performant.


```suggestion
    try:
        return (
            Docket.objects.only("id", "docket_number", "date_modified")
            .get(pk=docket_id)
        )
    except Docket.DoesNotExist:
        return None
```

cl/search/docket_number_cleaner.py
```diff
@@ -30,6 +47,57 @@
...
+    :param docket_number: The new docket number to set.
+    :param start_timestamp: The timestamp marking the start of the daemon job.
+    """
+    docket = get_docket(docket_id)
+    r = get_redis_interface("CACHE")
+    if docket:
+        if docket.date_modified < start_timestamp:
+            docket.docket_number = docket_number
+            docket.save(update_fields=["docket_number"])
+            r.srem(get_redis_key(), str(docket_id))
```
albertisfu: Correct me if I’m wrong, but I think we can still have a race condition between the time we fetch the docket from the database via `get_docket` and the time we call `docket.save()`.

Something like this:

21:00 — Daemon starts  
21:01 — get_docket ID 1000, date_modified: 20:30  
21:02 — A signal modifies docket_number  
21:03 — We compare 20:30 < 21:00, so we proceed to update `docket_number` with an older value than the one modified by the signal at 21:02

Does that makes sense?

If so, I think we can solve this by using a database transaction and creating a database lock with `select_for_update`, which prevents other processes from modifying the same row during this operation. It might look something like this:


```
from functools import partial

with transaction.atomic():
        docket = (
            Docket.objects
            .select_for_update()
            .only("id", "docket_number", "date_modified")
            .filter(id=docket_id)
            .first()
        )
        if not docket:
            return

        if docket.date_modified >= start_timestamp:
            return

        docket.docket_number = docket_number
        docket.save(update_fields=["docket_number"]) 
     
        transaction.on_commit(partial(r.srem, get_redis_key(), str(docket_id)))
```

Additionally, the Redis operation will only occur once the transaction has been successfully committed to the database.

cl/search/docket_number_cleaner.py
albertisfu: Sorry now I think we can just remove this method if the suggestions in other comments makes sense for you.

cl/search/docket_number_cleaner.py
```diff
@@ -176,3 +244,208 @@ def clean_docket_number_raw(
...
+        court_mapping = court_map.get(court_id, None)
+        if court_mapping not in court_batches:
+            court_batches[court_mapping] = []
+        court_batches[court_mapping].append({docket_id: docket_number_raw})
+    return court_batches
+
+
+def extract_with_llm(
+    batch: list[dict[int, str]], system_prompt: str, model_id: str
+) -> dict[int, str]:
```
albertisfu: It seems we need to update the return type annotation for this method, since we can also expect None values in case the LLM call fails, right?

Is the downstream method prepared to handle `None` values as well?

cl/search/docket_number_cleaner.py
```diff
@@ -1,7 +1,24 @@
...
+from cl.lib.llm import call_llm
+from cl.lib.redis_utils import get_redis_interface
 from cl.lib.string_utils import normalize_dashes
-from cl.search.models import Court
+from cl.search.llm_models import CleanDocketNumber
+from cl.search.llm_prompts import F_PROMPT, F_TIE_BREAKER
+from cl.search.models import Court, Docket
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
```
albertisfu: We don't need this line to setup the logger. We can just remove it.
```suggestion
```

cl/search/docket_number_cleaner.py
```diff
@@ -30,6 +47,57 @@
...
+    try:
+        docket_obj = Docket.objects.filter(id=docket_id).first()
+        return docket_obj
+    except Docket.DoesNotExist:
+        logger.error(f"Docket with id {docket_id} does not exist.")
+        return None
+
+
+def update_docket_number(
+    docket_id: int, docket_number: str, start_timestamp: timezone.datetime
```
albertisfu:
```suggestion
    docket_id: int, docket_number: str, start_timestamp: datetime.datetime
```

cl/search/tasks.py
```diff
@@ -2126,3 +2140,73 @@ def index_embeddings(
...
+    1. Running two mini models to find consensus on cleaned docket numbers.
+    2. For non-consensus results, running two full models for further comparison.
+    3. For remaining ties, using a tie-breaker model to finalize the cleaned docket numbers.
+    4. Updating the docket numbers in the database with the cleaned results.
+
+    :param court_batch: A batch of court records with docket IDs and numbers.
+    :param court_mapping: Identifier for the court mapping used in model prompts.
+    :param start_timestamp: Timestamp marking the start of the daemon job.
+    """
+    if court_mapping:
```
albertisfu: You can use early returns here to avoid nested ifs:

```suggestion
    if not court_mapping:
        return
```

cl/search/tasks.py
```diff
@@ -2126,3 +2140,73 @@ def index_embeddings(
...
+        # First pass with two mini models to find consensus
+        next_model_batches = call_models_and_compare_results(
+            court_batch=court_batch,
+            court_mapping=court_mapping,
+            model_one="openai/gpt-4o-mini",
+            model_two="openai/gpt-4.1-mini",
+            start_timestamp=start_timestamp,
+        )
+        # Next pass with two full models for non-consensus
+        if next_model_batches:
```
albertisfu: Similarly

```suggestion
        if not next_model_batches:
             return
```

cl/search/tasks.py
```diff
@@ -2126,3 +2140,73 @@ def index_embeddings(
...
+            tie_breaker_batches = call_models_and_compare_results(
+                court_batch=next_model_batches,
+                court_mapping=court_mapping,
+                model_one="openai/gpt-4o",
+                model_two="openai/gpt-4.1",
+                start_timestamp=start_timestamp,
+            )
+            # Third pass with tie-breaker model
+            if tie_breaker_batches:
```
albertisfu: Also:
```suggestion
            if tie_breaker_batches:
               return
```

cl/search/management/commands/llm_clean_docket_number_daemon.py
```diff
@@ -0,0 +1,120 @@
...
+            default=5,
+            help="The celery throttle min items.",
+        )
+
+    def process_llm_batch(
+        self,
+        celery_queue: str,
+        court_batch: list[dict[int, str]],
+        court_mapping: str,
+        start_timestamp: timezone.datetime,
```
albertisfu:
```suggestion
        start_timestamp: datetime.datetime,
```

albertisfu (changes requested):
Thank you @rachlllg this looks great.

I left some suggestions and comments that I think could help improve the performance of the database queries and address a potential race condition.

I also added a couple of minor comments.

Let me know what you think.

ApacheAde:
///dicey.made.constitution

ApacheAde:
IMG_2833

cl/search/docket_number_cleaner.py
rachlllg: Previously I had it return the elements with `None` key, but I think there is no need for that so I now refactored to only return the records with valid key.

cl/search/docket_number_cleaner.py
rachlllg: Good point, this method is called by `process_llm_batches()` and `process_llm_batches()` is prepared to handle `None` values, I also added a test `test_process_llm_batches_handles_none()` to confirm this .

cl/search/tasks.py
rachlllg: Since `create_llm_court_batches()` has been refactored to no longer return `court_batches` with `None` keys, this is no longer needed so I removed it

cl/search/docket_number_cleaner.py
rachlllg: Ah I see, you are referring to the possibility that immediately after the docket was retrieved by the `update_docket_number()` method, and right before the update was made to the DB, within this fraction of a second, the signal modifies the object and then the `update_docket_number()` method modifies the same object incorrectly. 

This makes sense, thank you!

rachlllg:
Thank you @albertisfu I updated per your comments.

cl/search/docket_number_cleaner.py
```diff
@@ -30,6 +49,52 @@
...
+
+def update_docket_number(
+    docket_id: int, docket_number: str, start_timestamp: datetime, r: Redis
+) -> None:
+    """
+    Update the docket_number field of a Docket object & remove it from the LLM cleaning Redis set if it hasn't been modified since the start timestamp.
+
+    :param docket_id: The primary key ID of the Docket to update.
+    :param docket_number: The new docket number to set.
+    :param start_timestamp: The timestamp marking the start of the daemon job.
```
albertisfu, line 78: Missing the ` r: Redis` from the docstrings

cl/search/docket_number_cleaner.py
```diff
@@ -30,6 +49,52 @@
...
+    """
+    with transaction.atomic():
+        docket = (
+            Docket.objects.select_for_update()
+            .only("id", "docket_number", "date_modified")
+            .filter(id=docket_id)
+            .first()
+        )
+        if not docket:
```
albertisfu, line 88: In case the docket is missing (for example, if it was removed before the daemon runs), should we remove the `docket_id` from the Redis set so we don’t end up with stale IDs lingering forever?

cl/search/docket_number_cleaner.py
```diff
@@ -30,6 +49,52 @@
...
+            return
+
+        if docket.date_modified >= start_timestamp:
+            return
+
+        docket.docket_number = docket_number
+        docket.save(update_fields=["docket_number"])
```
albertisfu: When testing the race condition protection added here, I found that `docket.date_modified` is not updated when using `update_fields`.

You need to explicitly include `date_modified` in the `update_fields` list:
```suggestion
        docket.save(update_fields=["docket_number", "date_modified"])
```

So we need to tweak this here and in other places where `docket_number` or `docket_number_raw` is updated using `update_fields`, such as in `clean_docket_number_raw_and_update_redis_cache` triggered by the signal. Otherwise, when performing the comparison:

`if docket.date_modified >= start_timestamp:`

it won’t reliably prevent overwriting `docket_number` with an outdated value.

cl/search/docket_number_cleaner.py
```diff
@@ -176,3 +241,206 @@ def clean_docket_number_raw(
...
+    Groups docket numbers by their associated court type for a given batch of docket_ids.
+
+    :param llm_batch: A list of docket_ids to be processed.
+    :return: A dictionary where each key is a court mapping (as determined by `court_map`),
+              and each value is a list of dictionaries of {docket_id: docket_number_raw}.
+    """
+    court_batches = {}
+    dockets = Docket.objects.filter(pk__in=llm_batch).values_list(
+        "id", "court_id", "docket_number_raw"
+    )
```
albertisfu, line 271: One thing I noticed about this query is that it only matches dockets that exist in our database. It might not be very common, but if a docket is removed before the daemon runs, its corresponding `docket_id` will remain in the Redis set indefinitely.

Maybe we could compare the IDs returned by this query and directly remove from the Redis set the ones that didn’t match any docket?

cl/search/docket_number_cleaner.py
```diff
@@ -176,3 +241,206 @@ def clean_docket_number_raw(
...
+        logger.error(
+            "LLM - Response validation error for batch=%s",
+            f"{batch}",
+            extra={
+                "validation_errors": e.errors(),
+                "fingerprint": ["llm-docket-number-validation-error"],
+            },
+        )
+        return
+    except Exception as e:
```
albertisfu, line 317: Since this is an API call, do we expect network errors, timeouts, or retriable HTTP errors (like 500) returned by the library?

I think it would be a good idea to catch these errors and return `None` instead of raising an exception. 

Otherwise, if one of the threads fails within `process_llm_batches`, all the work done by the successful threads might be lost.

cl/search/docket_number_cleaner.py
```diff
@@ -176,3 +241,206 @@ def clean_docket_number_raw(
...
+            executor.submit(
+                extract_with_llm, batch, system_prompt, model_id
+            ): batch
+            for batch in batches
+        }
+        for future in as_completed(future_to_batch):
+            parsed_output = future.result()
+            all_cleaned.update(parsed_output or {})
+    # Check for any still-unprocessed records (e.g., if LLM failed to extract)
+    processed_ids = [k for k in all_cleaned.keys()]
```
albertisfu: We can make this a SET for a faster membership check.

```suggestion
    processed_ids = set(all_cleaned.keys())
```

cl/search/docket_number_cleaner.py
```diff
@@ -176,3 +241,206 @@ def clean_docket_number_raw(
...
+    return all_cleaned
+
+
+def call_models_and_compare_results(
+    court_batch: list[dict[int, str]],
+    court_mapping: str,
+    model_one: str,
+    model_two: str,
+    start_timestamp: datetime,
+    r: Redis,
```
albertisfu, line 415: These two parameters are missing from the docstrings.

cl/search/docket_number_cleaner.py
```diff
@@ -176,3 +241,206 @@ def clean_docket_number_raw(
...
+            update_docket_number(
+                docket_id, docket_number_one, start_timestamp, r
+            )
+        else:
+            # Find the docket_number_raw from court_batch using docket_id
+            docket_number_raw = None
+            for batch in court_batch:
+                if docket_id in batch:
+                    docket_number_raw = batch[docket_id]
+                    break
```
albertisfu: I think we can refactor this like:


```suggestion
            docket_number_raw = next(
                (batch[docket_id] for batch in court_batch if
                 docket_id in batch),
                None,
            )

```

cl/search/docket_number_cleaner.py
```diff
@@ -176,3 +241,206 @@ def clean_docket_number_raw(
...
+            if docket_number_raw is not None:
```
albertisfu, line 468: Can it happen that `docket_number_raw` is an empty string from the source? If so, those empty strings will be sent to the LLM even though they aren’t valid candidates for cleaning?

Maybe that’s not too problematic unless that empty docket number is the only one in the batch, as it could trigger unnecessary API calls.

cl/search/management/commands/llm_clean_docket_number_daemon.py
```diff
@@ -0,0 +1,121 @@
...
+from cl.lib.celery_utils import CeleryThrottle
+from cl.lib.command_utils import VerboseCommand, logger
+from cl.lib.redis_utils import get_redis_interface
+from cl.search.docket_number_cleaner import (
+    create_llm_court_batches,
+    get_redis_key,
+)
+from cl.search.tasks import clean_docket_number_by_court
+
+logger = logging.getLogger(__name__)
```
albertisfu: I think we can remove this one since it's already being imported from `command_utils`

cl/search/tasks.py
```diff
@@ -2134,3 +2149,75 @@ def index_embeddings(
...
+        OpenAIApiError,
+    ),
+    retry_kwargs={"max_retries": 5},
+)
+def clean_docket_number_by_court(
+    self: Task,
+    court_batch: list[dict[int, str]],
+    court_mapping: str,
+    start_timestamp: datetime,
+    r: Redis = get_redis_interface("CACHE"),
```
albertisfu: It's not advisable to use a method as default value of a celery task since it's not serializable and it's evaluated at import time.

So I suggested instead using:
```suggestion
    r: Redis | None= None,
```

As I noticed you're passing the redis interface in tests.

So within the task you can instead do:

```
   if r is None:
```

cl/search/docket_number_cleaner.py
rachlllg: Yes, I hadn't realized this could a possibility until I inspected some of the docket_numbers and see that indeed they could be an empty string from the source. I opened https://github.com/freelawproject/courtlistener/pull/6475 to fix this from the signal so we don't add those dockets with empty docket_number_raws to the job.

cl/search/docket_number_cleaner.py
rachlllg: I see now, when using update_fields, it only updates the field specified. But when using .save(), it updated the date_modified also. I opened https://github.com/freelawproject/courtlistener/pull/6475 to fix this on the signal side, also updated the tests for the daemon to properly mimic updates using update_fields.

cl/search/docket_number_cleaner.py
rachlllg: Yes, we should remove them to avoid them lingering. I didn't use transaction.on_commit() here for removing them from redis, because from what I understand, transaction.on_commit() is only needed because we were performing DB updates right? Since we are not performing DB updates here, I can directly remove the id from the redis cache?

rachlllg:
Thank you @albertisfu I learned a lot from the comments, they have been updated. Something is wrong with the ci action, I tried rerunning it but it seems to get stuck, I will try again next week.

cl/search/docket_number_cleaner.py
```diff
@@ -176,3 +243,222 @@ def clean_docket_number_raw(
...
+    """
+    Groups docket numbers by their associated court type for a given batch of docket_ids.
+    Remove docket_ids that no longer exist, if any, from the Redis set.
+
+    :param llm_batch: A list of docket_ids to be processed.
+    :param r: Redis client instance for interacting with Redis.
+    :return: A dictionary where each key is a court mapping (as determined by `court_map`),
+              and each value is a list of dictionaries of {docket_id: docket_number_raw}.
+    """
+    court_batches = {}
```
albertisfu: It seems this is unused now.

cl/search/docket_number_cleaner.py
```diff
@@ -176,3 +243,222 @@ def clean_docket_number_raw(
                 return docket_number_raw, docket_id
 
     return docket_number_raw, None
+
+
+def create_llm_court_batches(
+    llm_batch: list[int],
+    r: Redis,
+) -> dict[str, list[dict[int, str]]]:
```
albertisfu: It seems that the return value here should be a `defaultdict` instead of a regular `dict`, or you can convert `court_batches` to a `dict` before returning it.

cl/search/docket_number_cleaner.py
```diff
@@ -176,3 +243,222 @@ def clean_docket_number_raw(
...
+    except ValidationError as e:
+        logger.error(
+            "LLM - Response validation error for batch=%s",
+            f"{batch}",
+            extra={
+                "validation_errors": e.errors(),
+                "fingerprint": ["llm-docket-number-validation-error"],
+            },
+        )
+        return
```
albertisfu: To make mypy happy, we need to explicitly return `None` here and in other places in this file where we are not explicitly returning anything.

cl/search/docket_number_cleaner.py
```diff
@@ -176,3 +243,222 @@ def clean_docket_number_raw(
...
+    :param all_cleaned: Accumulator dictionary for all processed records.
+    :return: A dictionary mapping record IDs to their cleaned/extracted values, including any unprocessed records after max retries.
+    """
+    logger.info(f"---Processing {len(llm_batches)} records with {model_id}---")
+    batches = [
+        llm_batches[i : i + batch_size]
+        for i in range(0, len(llm_batches), batch_size)
+    ]
+    with ThreadPoolExecutor(
+        max_workers=settings.DOCKET_NUMBER_CLEANING_MAX_WORKERS
```
albertisfu: To make `mypy` happy here, and in similar lines where you call `settings`, you can add a comment like: `# type: ignore`

rachlllg:
Thank you @albertisfu! I updated for mypy & fixed the commits.

albertisfu:
Thank you for all your hard work on this, @rachlllg. This is now ready to be merged.

I think the next step is to deploy the daemon and set `DOCKET_NUMBER_CLEANING_API_KEY` when we’re ready, right?

albertisfu (approved):

rachlllg:
> I think the next step is to deploy the daemon and set DOCKET_NUMBER_CLEANING_API_KEY when we’re ready, right?

Yes! 🚀 Here is the issue to turn on the different pieces https://github.com/freelawproject/courtlistener/issues/6468, tabled for Taylor Swift sprint 😜

PR #6429, opened 2025-10-21
feat(populate_docket_number_raw): new command

### LINKED ISSUE #6363: Create a django command for copying the content from docket_number to docket_number_raw
### Describe the issue.

Create a django command for copying the content from docket_number to docket_number_raw for all existing content in DB

To Do

grossir:
Fixes #6363

A command to copy Docket.docket_number into Docket.docket_number_raw

- does the update in 100k batches
- forces the user to break down the update into around 6 calls, by  putting a hard 10M limit. A vacuum should be ran around that number of updates (or an autovacuum should trigger)
- raises an error if the first Docket.number_raw of a batch already had a non empty value

### DIFF
--- a/cl/search/management/commands/populate_docket_number_raw.py
+++ b/cl/search/management/commands/populate_docket_number_raw.py
@@ -1,5 +1,7 @@
 import time
 
+import pgtrigger
+from django.db import IntegrityError, transaction
 from django.db.models import F
 
 from cl.lib.command_utils import CommandError, VerboseCommand, logger
@@ -61,7 +63,16 @@ class Command(VerboseCommand):
                 batch_start,
                 batch_end,
             )
-            qs.update(docket_number_raw=F("docket_number"))
+
+            with pgtrigger.ignore("search.Docket:update_update"):
+                with transaction.atomic():
+                    try:
+                        qs.update(docket_number_raw=F("docket_number"))
+                    except IntegrityError:
+                        logger.error(
+                            "Docket.docket_number_raw population failed",
+                            exc_info=True,
+                        )
 
             # prevent affecting all tests with this sleep
             if not TESTING:
--- a/cl/search/tests/tests.py
+++ b/cl/search/tests/tests.py
@@ -3714,3 +3714,16 @@ class PopulateDocketNumberRawCommandTest(TestCase):
             self.dn,
             "docket number raw was not updated",
         )
+
+        self.assertTrue(
+            self.docket.events.all().last() is None,
+            "pghistory event saving should be disabled by `populate_docket_number_raw`",
+        )
+
+        # see that regular pghistory trigger behavior is still working
+        self.docket.docket_number = "xxxx"
+        self.docket.save()
+        self.assertTrue(
+            self.docket.events.all().last() is not None,
+            "Event saving trigger is not working",
+        )


mlissner (commented):
Do you know — is there a way to disable adding rows to the history tables for this whole thing? It'd be nice not to double our storage. :)

grossir:
Since events are tracked using triggers, we would have to disable triggers for the command run. Can be done in Python via pgtrigger, it [seems](https://github.com/AmbitionEng/django-pghistory/issues/116)

mlissner:
It'll be a judgement call whether to do this on a particular bulk project, but this one feels big and safe, so I think it's probably a good idea. I do wonder if the trigger is disabled globally, but even if so, it's probably worth doing, I think.

grossir:
The trigger disabling is local. All `pgtrigger.ignore` logic is in a single file. This particular [line](https://github.com/AmbitionEng/django-pgtrigger/blob/38f9d1edecfedbe03560e2a49690d14c314d7f68/pgtrigger/runtime.py#L113C24-L113C34) prepends `"f"SELECT set_config('pgtrigger.ignore', %s, true); {sql}"` to the sql you will execute. `set_config` [signature](https://www.postgresql.org/docs/18/functions-admin.html) shows that it's passing `is_local=true`


I have added this to the PR, and I am writing some tests to double check that trigger disabling is working as expected

To get the trigger "URI"
```
./manage.py pgtrigger ls
INSTALLED ENABLED search.Docket:update_update
```

cl/search/management/commands/populate_docket_number_raw.py
```diff
@@ -0,0 +1,80 @@
...
+from cl.search.models import Docket
+from cl.settings import TESTING
+
+
+class Command(VerboseCommand):
+    help = "Populate `Docket.docket_number_raw` using `docket_number`"
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "start-id", help="Value of Docket.id to start from", type=int
```
quevon24, line 17: We regularly add them with double hyphens for arguments

```suggestion
            "--start-id", help="Value of Docket.id to start from", type=int
```

cl/search/management/commands/populate_docket_number_raw.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+        parser.add_argument(
+            "end-id", help="Last value of Docket.id to update", type=int
```
quevon24, line 20:
```suggestion
            "--end-id", help="Last value of Docket.id to update", type=int
```

quevon24 (approved):
Everything looks good, just two suggestions

cl/search/management/commands/populate_docket_number_raw.py
grossir: `--` [means](https://docs.python.org/3/library/argparse.html#name-or-flags) they are optional arguments. In this case we want both arguments everytime, so I think we shouldn't use `--`

grossir: Seeing the new Sentry error, I think you were right about this... 
If it's not a `--` it won't populate the `options` dict. But I don't know why this didn't error in the tests...

quevon24: Good to know, I have always used the parameter required=True to make it required.

quevon24: > Seeing the new Sentry error, I think you were right about this... If it's not a `--` it won't populate the `options` dict. But I don't know why this didn't error in the tests...

i think that happened  because you are calling handle() instead of using call_command it is a bit tricky to test commands

PR #6436, opened 2025-10-23
6389 new banner component

EstiShay:
## Summary
This PR adds a new banner component to the v2 homepage.

Obstacles:
1. Need to view Figma for: ensure color & size matching and add hover/on-click behavior
2. Need to know what the dropdown button options are supposed to do in order to create accurate sample data.
3. Need actual logo svgs added to repo.
4. Need the actual location of the image in the issue, then update homepage with an image that works.
5. Assigned buttons `inline-block` so they will appear side-by-side, but this caused the toggle caret to appear below the button text instead of beside it. Have not yet debugged.
6. Dropdown menu options have been connected to `expansionPanel` and assigned `x-show="isExpanded"` but toggling still does not make them appear and disappear. Right now, add or delete `x-cloak` to view or hide them.

<details closed>
<details open>
image
</details>
<details open>
image
image
</details>


cl/assets/templates/cotton/banner.html
```diff
@@ -0,0 +1,72 @@
+{% load static svg_tags component_tags %}
+{% require_script "js/alpine/components/expansion_panel.js" %}
+{% require_script "js/alpine/plugins/collapse@3.14.8" defer=True %}
+
```
elisa-a-v, line 4: Components that expect certain attributes should declare them using a `<c-vars />` tag at the top, you can read more about it [here](https://github.com/freelawproject/courtlistener/wiki/New-Frontend-Architecture#Component-Attributes).

cl/assets/templates/cotton/banner.html
```diff
@@ -0,0 +1,72 @@
...
+    <!-- Title -->
+    <h2 class="font-semibold m-2">{{ title_text }}</h2>
+
+    <!-- Paragraph -->
+    <p class="text-gray-700 m-2">{{ paragraph_text }}</p>
+
+    <!-- Buttons -->
+     {% for b in buttons %}
+        {% if b.dropdown %}
```
elisa-a-v: > Need to know what the dropdown button options are supposed to do in order to create accurate sample data.

The expected behavior for the buttons is as follows:
- When it has no children, it's a link, so it should use an `<a>` tag.
- When it does have children:
    - The button is not an `<a>` tag anymore but instead an actual `<button role="button">` that toggles the dropdown.
    - The items in the dropdown are all links, so they should use `<a>`.

Given this, it might be better to check for `children` in the `button` instead of a `dropdown` flag, very similar to what we're doing in the `navigation_menu.html` component. The difference would be that in the case of the nav menu they're all links, even those that have children.

The figma has only two banners in the new homepage (see screenshot below), and only one of them has a dropdown. You can get the links for the items in the dropdown in [this page](https://free.law/recap).

Image

cl/simple_pages/templates/v2_components.html
```diff
@@ -184,6 +185,82 @@ <h4 class="mt-3">Demo</h4>
...
 
+  {# BANNER #}
+  <section class="max-w-full w-full border-t-2 border-greyscale-200" x-intersect.margin.-100px="show" id="banner">
+    <h2 class="mt-6 mb-3">Banner</h2>
+    <h4 class="mt-3">Demo</h4>
+    <c-banner
+      eyebrow_text="RECAP Suite"
+      title_text="Turning PACER Around Since 2009"
+      paragraph_text="If you use PACER, consider installing the RECAP extension in your browser. If you are a researcher, you can search millions of federal filings in the RECAP Archive. Journalists and court watchers will find docket alerts invaluable. If you get notifications from PACER, check out our NEF notification handler."
+    ></c-banner>
```
elisa-a-v, line 219: Demoing the buttons is probably a good idea, and the code snippet below has them so we may want to match this one to that.

cl/simple_pages/templates/v2_components.html
```diff
@@ -184,6 +185,82 @@ <h4 class="mt-3">Demo</h4>
...
+                'dest': '/recap/edge',
+                'icon': 'logos/edge'
+              }
+            ]
+          }, {
+              'style': 'btn-white',
+              'name': 'Learn more'
+          }
+        ]"
+        image_source="{%  static 'png/recap_search_container.png' %}"
```
elisa-a-v, line 283: <details>
<summary> Looks like you forgot to load static in this file 👀 </summary>

Image

</details>

cl/assets/static-global/js/alpine/components/expansion_panel.js
```diff
@@ -1,6 +1,9 @@
 document.addEventListener('alpine:init', () => {
   Alpine.data('expansionPanel', () => ({
     isExpanded: false,
+    get supportCaretClass() {
```
elisa-a-v: I think we may be able to get rid of this and use the one below (`iconClass`) instead.

cl/assets/tailwind/input.css
```diff
@@ -113,6 +113,9 @@
   .btn-grey {
     @apply btn bg-greyscale-100 hover:bg-greyscale-200 focus-visible:ring-4 focus-visible:ring-greyscale-200 disabled:text-greyscale-400 disabled:bg-greyscale-50;
   }
+  .btn-white {
+    @apply btn text-greyscale-700 bg-white hover:bg-greyscale-100 border outline-none border-greyscale-300 focus-visible:ring-4 focus-visible:ring-greyscale-200 disabled:border-greyscale-200 disabled:text-greyscale-400;
+  }
```
elisa-a-v: The only difference between this one and `btn-outline` is that this has a set bg color, but looking at the [Figma buttons](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=1-1183&m=dev) I think the `btn-outline` was actually missing the `bg-white` class 😬  probably best if we tweak and use that one instead of adding a new one, would you remove this new class and use `btn-outline` instead please?

cl/assets/templates/cotton/banner.html
```diff
@@ -0,0 +1,76 @@
...
+                        class="transition-transform duration-200 flex justify-center items-center w-4 h-4"
+                        aria-hidden="true"
+                        x-cloak
+                    >{% svg 'chevron' class="w-full h-full text-white" %}
+                    </span>
+                </button>
+
+             {#  support menu  #}
+                <menu
+                    id="banner-button-menu"
```
elisa-a-v: Okay this is a fun one. Since this is a reusable component that could be used more than once in the same template, we should never use hardcoded `id` in them. To help with this, [Alpine has a pretty nifty directive](https://alpinejs.dev/directives/id) `x-id` that generates new id scopes for groups of ids.

For an example, our corpus search component uses it extensively, as well as the data selector sub-component. Let me know if you need further guidance on this, it took me a bit to fully wrap my head around how it works.

cl/assets/templates/cotton/banner.html
```diff
@@ -0,0 +1,76 @@
...
+                    aria-controls="banner-button-menu"
+                    aria-haspopup="menu"
+                    aria-label="Toggle menu"
+                    >
+                    {{ b.name }}
+                    <span
+                        x-bind:class="supportCaretClass"
+                        class="transition-transform duration-200 flex justify-center items-center w-4 h-4"
+                        aria-hidden="true"
+                        x-cloak
```
elisa-a-v: Any reason to add `x-cloak` here? I _think_ we should be able to safely remove it, since it's meant to [prevent "blips" on uninitialized templates before Alpine loads](https://alpinejs.dev/directives/cloak) 🤔

cl/assets/templates/cotton/banner.html
```diff
@@ -0,0 +1,76 @@
...
+                    >{% svg 'chevron' class="w-full h-full text-white" %}
+                    </span>
+                </button>
+
+             {#  support menu  #}
```
elisa-a-v: Let's either clean this comment, or fix it so it's accurate please.

cl/assets/templates/cotton/banner.html
```diff
@@ -0,0 +1,76 @@
...
+      <!-- Title -->
+      <h2 class="font-semibold m-2">{{ title_text }}</h2>
+
+      <!-- Paragraph -->
+      <p class="text-gray-700 m-2">{{ paragraph_text }}</p>
+
+      <!-- Buttons -->
+      <div class="inline-flex gap-x-1">
+         {% for b in buttons %}
+            {% if b.dropdown %}
```
elisa-a-v: How about checking `options` directly instead of an additional keyword? I think this would be a bit less redundant and less prone to error (e.g. adding `options` without passing the `dropdown` keyword, or viceversa), unless I'm missing something.

cl/assets/templates/cotton/banner.html
```diff
@@ -0,0 +1,76 @@
...
+                    role="menu"
+                    x-show="isExpanded"
+                    x-cloak
+                    x-bind:area-expanded="isExpanded"
+                >
+                    {% for opt in b.options%}
+                        <li
+                            class="text-md text-greyscale-800 flex hover:bg-greyscale-50"
+                        >
+                            <a href="{{ opt.dest }}" target="_blank" class="px-4 py-1.5 flex justify-start gap-2 w-full">
```
elisa-a-v: This is me nitpicking here but could we use `href` intead of `dest` please? We should try to be consistent whenever possible, so sticking to standard keyword names seems best.

Also, when adding `target="_blank"` to an anchor tag, we should always add `rel="noopener noreferrer"`. You can read more about it [here](https://stackoverflow.com/questions/79600136/when-adding-target-blank-to-a-html-link-is-it-still-necessary-to-add-rel-no). It's not a huge concern anymore cause modern browsers handle it gracefully but it's a very easy measure that could protect users in older devices.

cl/search/templates/v2_homepage.html
```diff
@@ -267,5 +267,61 @@ <h2 id="stats-title" class="sr-only">CourtListener Statistics</h2>
       </div>
     </dl>
   </section>
+{# BANNER #}
+  <div class="flex-row">
```
elisa-a-v:
```suggestion
  <div class="flex md:flex-row flex-col">
```

`flex-row` only sets `flex-direction: row` but you also need to set `display: flex` for that to take effect, which is why you also need the `flex` class in there. My suggestion uses `md` as a breakpoint for the row/col layout but you can check if another one works best.

EstiShay:
> * The dropdown doesn't look like the design when open:
> 
> Image

Which browser showed this?

elisa-a-v:
> Which browser showed this?

Firefox, Chrome and Safari

elisa-a-v:
Adding `absolute` to the `<menu>` and `relative` to the parent seems to fix it:

image

EstiShay:
Let me know if the image still isn't centered for you. I haven't been able to reproduce that.

mlissner:
Looks great. I think banners then stats, but if we want to merge, look, and tweak after we've lived with it a few days, that's all good with me.

Issue #6364, opened 2025-10-02
Identify areas where docket_number as params need to be changed to use docket_number_raw


rachlllg:
### Describe the issue.

In the current codebase, `docket_number` is used as params in numerous areas, need to review and identify which of these should be changed to use `docket_number_raw` instead

### What type of issue is this?

To Do

grossir:
- For everything user-facing, we should keep using `Docket.docket_number` which will be a cleaned version
- For the API, should we enable filtering by `docket_number_raw`? (we already allow docket_number and docket_number_core filters)
- Anything to update on the web search interface? If we need to change anything, we would have to update the `DocketDocument`
- For use cases where given a "scraped" docket number we try to find the matching docket, we should use "Docket.docket_number_raw"

-----------

Taking some notes while finishing sibling issues https://github.com/freelawproject/courtlistener/issues/6362

Change `existing_docket_number = clean_docket_number(docket.docket_number)` to use `docket.docket_number_raw` in

https://github.com/freelawproject/courtlistener/blob/27f02a12be869761ec8344ac717680327165703e/cl/recap/mergers.py#L76-L100


In the same file `get_appellate_docket_by_docket_number`


--------

https://github.com/freelawproject/courtlistener/blob/4234759e5362aa5a0fe2d65c830d9f11d38132e1/cl/recap/tasks.py#L3356-L3359

rachlllg:
> For the API, should we enable filtering by docket_number_raw? 

> Anything to update on the web search interface?

I think not. The plan is to not index `docket_number_raw` to ES, it'll just be a field to store the raw content just for reference (in DB only), filtering & web search will still support `docket_number` & `docket_number_core` as they do today.

Another thing to consider, copied from @albertisfu 's  notes in https://github.com/freelawproject/courtlistener/pull/6475:

> This might be a side effect once the cleaning method is enabled, but I wanted to bring it up so we can analyze whether it’s something we should handle before enabling it.
> 
> We have the `docket_number_core` field, which we usually use to look up RECAP dockets. Currently, this is updated within the docket’s save method when this condition is met:
> 
> ```
>  if self.docket_number and not self.docket_number_core:
>             self.docket_number_core = make_docket_number_core(
>                 self.docket_number
>             )
> ```
> 
> So if dockets already have a `docket_number_core`, then after we apply the new cleanup method, this field won’t benefit from the cleaned number. I think this could impact RECAP dockets that are also matched by a case law source.

[...]

PR #6445, opened 2025-10-24
fix(docket_number_raw): duplicate docket_number ingestion and updates for docket_number_raw

grossir:
Solves #6361

Whenever a scraper ingests or updates a Docket, now it populates Docket.docket_number_raw with the value that also goes into Docket.docket_number

- opinion scrapers
- oral argument scrapers
- free opinion on pacer
- recap extension (and email) through `lib.pacer.update_docket_metadata`

### DIFF
--- a/cl/api/templates/case-law-api-docs-vlatest.html
+++ b/cl/api/templates/case-law-api-docs-vlatest.html
@@ -148,6 +148,9 @@
   </p>
   <p>As with all other APIs, you can look up the field descriptions, filtering, ordering, and rendering options by making an <code>OPTIONS</code> request:
   </p>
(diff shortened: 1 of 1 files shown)


cl/recap/mergers.py
```diff
@@ -384,6 +384,7 @@ async def update_docket_metadata(
     d = update_case_names(d, docket_data["case_name"])
     await mark_ia_upload_needed(d, save_docket=False)
     d.docket_number = docket_data["docket_number"] or d.docket_number
```
grossir, line 386: Originally I had deleted this line, but had to add it back, since the change was breaking tests. Will have to be deleted here https://github.com/freelawproject/courtlistener/issues/6364

PR #4771, opened 2024-12-03
feat(assets): Adds EOY banner

ERosendo:
This PR fixes #4769 by updating the content of the banner to show our end-of-year message. 

It also tweaks the dismissible template to use a two-row layout, which should make it easier to read.

# Small screens
image

-----
# Medium screens
image

------
# Large screens
image

PR #4759, opened 2024-11-30
feat(assets): Adds GivingTuesday banner

ERosendo:
This PR fixes #4750 by adding a banner for GivingTuesday.

Instead of creating a new variable, we can use the existing `FUNDRAISING_MODE` setting. By simply toggling this variable, we can easily display the banner.

image

PR #6502, opened 2025-11-05
feat(profile_pages): introduce banner to ask for feedback about wireframes

### LINKED ISSUE #6474: Invite participants for wireframe testing
We’d like to collect feedback on the wireframe for the [unified view of alerts/notes/tags](https://github.com/freelawproject/courtlistener/issues/6430). 

To do this, we plan to invite users to participate via the [profile pages](https://www.courtlistener.com/profile/search-alerts/) in CourtListener. A banner should be viewable when any tab is selected (alerts, notes, account, etc.).

- "We are redesigning your view of alerts, notes, and tags, and we need your input! Sign up [here](https://docs.google.com/forms/d/e/1FAIpQLSfGgwJerlvbCWo9S_wwJfUPlsZxHE_WNm0djrwDHX1bxo-LsQ/viewform?usp=dialog) to provide your feedback."
- Link them to the [google form](https://docs.google.com/forms/d/e/1FAIpQLSfGgwJerlvbCWo9S_wwJfUPlsZxHE_WNm0djrwDHX1bxo-LsQ/viewform?usp=dialog).
- This will likely be active for a few weeks. We can manually remove it when we have enough participants.

Image

elisa-a-v:
## Fixes

Closes #6474 

## Summary

We need users to test the wireframes for the new unified view of alerts/notes/tags, so we now display a dismissible nav banner in all profile pages inviting users to sign up to provide feedback.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

<details>

image

image

</details>

### DIFF
--- a/cl/api/templates/includes/toc_sidebar.html
+++ b/cl/api/templates/includes/toc_sidebar.html
@@ -51,6 +51,7 @@
       <li><a href="#case-law-apis">Case Law</a></li>
       <li><a href="#pacer-apis">PACER</a></li>
       <li><a href="#recap-apis">RECAP</a></li>
+      <li><a href="#pray-and-pay-api">Pray and Pay</a></li>
       <li><a href="#search-api">Search</a></li>
       <li><a href="#judge-apis">Judges</a></li>
       <li><a href="#disclosure-apis">Financial Disclosures</a></li>
--- a/cl/api/templates/recap-api-docs-vlatest.html
+++ b/cl/api/templates/recap-api-docs-vlatest.html
@@ -42,6 +42,14 @@
         <li><a href="#attachment-pages">Scraping Attachment Page</a></li>
         <li><a href="#dockets">Purchasing Dockets</a></li>
       </ul>
+      <li><a href="#pray-and-pay-api">Pray and Pay API</a></li>
+      <ul>
+        <li><a href="#creating-prayers">Creating Prayers</a></li>
+        <li><a href="#listing-prayers">Listing Prayers</a></li>
+        <li><a href="#deleting-prayers">Deleting Prayers</a></li>
+        <li><a href="#prayer-webhooks">Webhook Notifications</a></li>
+        <li><a href="#prayer-limitations">Limitations</a></li>
+      </ul>
       <li><a href="#recap-upload">RECAP Upload API</a></li>
       <ul>
         <li><a href="#global-params">Global Parameters</a></li>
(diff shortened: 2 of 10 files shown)


mlissner:
I think we've got the right idea, but I just saw this on Github Actions: 

image

And then I noticed this messy alignment between the button, the × button, and the text: 

image

We don't have to get to Github quality, but I think we can do better? We're headed into the end of year donation season. Maybe we can take this opportunity to make our banner better in general? I assume this is fairly easy, right?

elisa-a-v:
Yeah you're right, it does look quite messy. How about this?

image

image

image

mlissner:
Waaaay better, but maybe minus the green? It seems really loud to me.

elisa-a-v:
Oh! [We chose the green in the issue](https://github.com/freelawproject/courtlistener/issues/6474#issuecomment-3492722296) out of a handful of options; the banner has a feature where you can easily change the color, but the alternatives are limited cause it uses bootstrap classes.

<details>
<summary>Other color combos</summary>

image

image

image

This one wasn't listed in the issue but I just found out about it and it uses the primary color of the CL UI:
image

</details>

Would any of the other ones listed above work better? Using any those without tweaking anything is trivial and switching between them takes just a few seconds, but if none of those work, we could try another color combo; wouldn't be hard but it would take just a little bit longer.

elisa-a-v:
Now that I think about it, maybe the fact it's loud is actually a good thing? We want users to notice it, and they can dismiss it so it won't bother them for long 🤷🏽

mlissner:
Let's do the white one. They won't miss it!

elisa-a-v:
lol that's actually my least favorite 😂 but that's okay, I don't hate it and it is nice that the blue is consistent with our branding in the current UI, so I'm not pushing back.

It's updated with the white background now 🫡

cl/assets/templates/includes/dismissible_nav_banner.html
```diff
@@ -1,25 +1,21 @@
...
 
   Parameters:
     - text: Text of the banner.
     - link: The URL for the button.
     - cookie_name: Name of the cookie used to remember if the user has already
     dismissed the banner. This prevents them from seeing the same message
     repeatedly.
-    - single_row (optional): A Boolean flag that, when set to True for medium
-    and large resolutions, condenses the banner into a single line of text
-    with a button positioned to the right. Defaults to False.
```
ERosendo, line 13: Is there a reason to remove this keyword? We originally added it so the template could handle both short texts ( like the one we used for GivingTuesday - https://github.com/freelawproject/courtlistener/pull/4759) and longer banner content where the button needs to wrap to a second row (https://github.com/freelawproject/courtlistener/pull/4771). It looks like this PR makes the template less flexible by enforcing a single layout.

cl/assets/templates/includes/dismissible_nav_banner.html
```diff
@@ -35,29 +31,33 @@
...
-      <div class="{% if single_row %}col-xs-2{% else %}col-xs-12{% endif %} flex justify-content-center justify-content-sm-end {% if not single_row %}second-row-btn{% endif %}">
-        <a href="{{link}}"
-            class="btn btn-{% if variant %}{{ variant }}{% else %}danger{% endif %} btn-lg hidden-xs">{% if button_emoji %}{{button_emoji}}{% endif %}&nbsp;{% if button_text %}{{button_text}}{% else %}Learn More{% endif %}</a>
-        <a href="{{link}}"
-            class="btn btn-{% if variant %}{{ variant }}{% else %}danger{% endif %} btn-sm hidden-sm hidden-md hidden-lg">{% if button_emoji %}{{button_emoji}}&nbsp;{% endif %}{% if button_text %}{{button_text}}{% else %}Learn More{% endif %}</a>
-      </div>
-    </div>
+  <div
+    class="flex flex-column align-items-center flex-md-row alert-{% if variant %}{{ variant }}{% else %}danger{% endif %}"
+    style="border: 2px solid lightgray; padding-inline: 4rem; padding-block: 2rem; position: relative; gap: 1.5rem"
```
ERosendo, line 39: Noticed this uses inline styles. we usually keep styling in `override.css` instead of adding it directly to the HTML. 

Would it make sense to move this inline style into a reusable CSS class in override.css? It’s tweaking several properties, so having it as a class might make it easier to maintain.

cl/assets/templates/includes/dismissible_nav_banner.html
```diff
@@ -35,29 +31,33 @@
 {% endcomment %}
 
 {% if cookie_name  not in request.COOKIES %}
-<div class="navbar navbar-default subnav alert-{% if variant %}{{ variant }}{% else %}danger{% endif %} alert-dismissible" role="navigation">
```
ERosendo, line 38: We use the `alert-dismissible` class in  [base.js](https://github.com/freelawproject/courtlistener/blob/30f8a759b7a631ab065a539ca1eceb76cec36797/cl/assets/static-global/js/base.js#L226-L238) to handle banner dismissal and set the corresponding cookie when users click the close button. The banner isn’t closing because the new template doesn’t include this class.

cl/assets/templates/includes/dismissible_nav_banner.html
elisa-a-v: It was removed because I originally misunderstood the expected behavior of that keyword and thought the refactored banner achieved the same, but thanks for explaining that we still needed a way to force the layout to always be a column (button below text) , instead of using a row in larger screens.

Instead of restoring the exact same keyword though, the default behavior now is that the layout is a column in all screen sizes (thus achieving the `single_row=False` behavior), and using a new `responsive` flag changes the layout to a row in larger screen sizes (`single_row=True`). This feels clearer to reason about, especially since the `single_row=True` didn't use a single row in every screen size, which I think is what tripped me at first; saying it's responsive feels more accurate. What do you think?

cl/assets/templates/includes/dismissible_nav_banner.html
elisa-a-v: Thanks for the suggestion, although I’m not sure this one belongs in `override.css`. That file is already quite large and this style combo is very specific to this banner, I don’t expect it’ll be reused anywhere else.

Inline feels reasonable here because it keeps the tweak close to the component and avoids adding more one-off rules to a file we’re trying not to grow anymore (and that's on its way out anyway!)

That said, if this ever ends up used in another place, I’m happy to move it into a class, how does that sound?

cl/assets/templates/includes/dismissible_nav_banner.html
elisa-a-v: Oops 😅 thanks for catching that!! When trying to fix it I found that restoring the class messed with the layout too much so instead, I decided to extend the script so it now supports multiple selectors: now, elements that use either the `alert-dismissible` class OR a `data-dismissible-alert` attribute will behave almost the same, with the second being decoupled from styles.

cl/assets/templates/includes/dismissible_nav_banner.html
mlissner: I've been religious about no inline styles for a decade, but this UI is on its dying days, so I don't think it matters if it works.

elisa-a-v:
Thanks for all your feedback!

Close button has been fixed, and the previous layout flexibility of the banner has been restored (it can either be a column in all screen sizes, or a column in smaller ones and a row in larger ones). The explanations are in the responses to your comments and in the commit messages :)

Let me know if anything else comes up!

cl/assets/templates/includes/dismissible_nav_banner.html
ERosendo: Thanks!

cl/assets/templates/includes/dismissible_nav_banner.html
ERosendo: > What do you think?

That makes perfect sense. thanks for the explanation! The new approach sounds clearer and more intuitive.

Issue #6473, opened 2025-10-31
Admin buttons require super user role


mlissner:
### What happened?

In the interface, we have buttons like this: 

<img width="784" height="749" alt="Image" src="https://github.com/user-attachments/assets/e8fb73f3-1a6e-4821-9e0a-2c17d9b20498" />

These buttons are only available to super users. When I recently updated permissions to give lots of staff members view-only access, they couldn't see these buttons anymore even though they could see the data in admin if they browsed to it manually. That's not great! 

Let's update the template to show the buttons to anybody that has view-or-higher access to each object.

It'll be worth grepping the code for any other places that make sense too.

### Expected behavior

The buttons should be visible if the user can view, modify, or delete the objects from the admin site. Super user access should not be necessary to view the buttons.

### Actual behavior

The buttons are hidden unless you have super user access.

### Steps to reproduce

Give yourself view-only access.

Load a page.

Note that the buttons are gone.

Grant yourself super user access.

Load a page.

Note that the buttons are there.

### What is the severity of the bug?

Medium

quevon24:
Should this apply only to **Docket**, **Cluster**, **Opinions** buttons? Currently, these buttons are shown only if the user has the change permission.

There shouldn’t be any issue if the **Block Cluster and Docket** button or the **Blocked** button (which doesn’t perform any action) are also displayed, since the view that executes the action checks whether the user is a superuser. Still, I just wanted to confirm to be sure.

Also, one more question, do you assign permissions manually to each user, or do you manage them through groups?

mlissner:
> Should this apply only to Docket, Cluster, Opinions buttons? Currently, these buttons are shown only if the user has the change permission.

Those buttons should also be shown if people have view access. 

> There shouldn’t be any issue if the Block Cluster and Docket button or the Blocked button (which doesn’t perform any action) are also displayed, since the view that executes the action checks whether the user is a superuser. Still, I just wanted to confirm to be sure.

Can we make this more granular so they're shown and work if you have edit access on the appropriate models? 

> Also, one more question, do you assign permissions manually to each user, or do you manage them through groups?

Groups — mostly!

mlissner:
Hm, I could use some help here, getting @libbyem working properly. She needs the ability to block cases, but can't see the buttons at present. 

Her permissions are set up this way: 

<img width="1618" height="721" alt="Image" src="https://github.com/user-attachments/assets/7bc3f336-3cb7-4129-8bf6-4606f626ac07" />

[...]

PR #6511, opened 2025-11-06
feat(opinion_page): update buttons visibility for dockets and clusters template

quevon24:
## Fixes

Fixes: #6473

## Summary

This PR updates the Cluster and Docket page templates.

- **Cluster detail template**
  - Updated visibility for *Docket*, *Cluster*, and *Opinion* buttons  
    - Previously shown only to users with `change_*` permission  
    - Now visible to users with `view`, `change`, or `delete` permissions  
  - Adjusted *Block Cluster and Docket* and *Blocked* buttons  
    - Now visible to users with `change_docket` or `change_opinioncluster` permissions  
    - Still restricted by backend checks for actual block actions

- **Docket detail template**
  - Updated *Edit Docket* button visibility  
    - Now visible to users with `view`, `change`, or `delete` permissions  
  - Adjusted *Block This Docket* and *Blocked* buttons  
    - Now visible to users with `change_docket` permission  
    - Backend still enforces superuser-only execution

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/alerts/tests/tests.py
+++ b/cl/alerts/tests/tests.py
@@ -2006,6 +2006,7 @@ class PrayAndPayAlertsWebhooksTest(TestCase):
             "type=oa&docket_number=19-1010&order_by=score+desc&stat_Published=": False,
             "q=hello": False,
             "filter=value": False,
+            "&highlight=on&semantic=true": True,
         }
 
         for query_string, expected in test_cases.items():
--- a/cl/alerts/utils.py
+++ b/cl/alerts/utils.py
@@ -61,7 +61,7 @@ from cl.search.types import (
     SearchAlertHitType,
 )
 
-COMMON_QUERY_PARAMS = {"type", "order_by"}
+COMMON_QUERY_PARAMS = {"type", "order_by", "semantic", "highlight"}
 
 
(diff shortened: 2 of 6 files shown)


mlissner:
Description sounds right, but I'm out of time for doing reviews, so let's have somebody on your team do it even though it's a small one. @flooie, can you please help?

grossir (changes requested):
This looks good, but it's missing a few things
1. the view permission check
2. similar changes for recap documents view page

I think blocking will still not work due to this part requiring super user privilege

https://github.com/freelawproject/courtlistener/blob/b1047d17fd9b7d78aefc300735086154096be70a/cl/opinion_page/views.py#L1629


Same changes need to be applied for Recap Documents? They also block stuff

Image

quevon24:
@grossir I implemented this changes:

- Update permissions for RECAP document admin buttons
- Update block_item view to check permissions instead of superuser
- Add tests for button visibility and block permissions

PR #6526, opened 2025-11-12
fix(search): filter courts when searching RECAP

### LINKED ISSUE #5439: Jurisdiction picker on RECAP home page has fewer courts than in search
The jurisdiction picker on the RECAP home page doesn't have inactive courts, which makes sense, because those courts were abolished long before PACER was created.

Image

However, once a user has already made a search and is seeking to refine it, those dormant courts become options again. I don't see the practical utility in this, but it may make sense if the underlying search utility for both RECAP and Opinions is the same code, because then it may not be possible to change for RECAP without also affecting Opinions.

Image

Image

### LINKED ISSUE #6279: Court picker shows too many courts in RECAP results page
### What happened?

When you go to the RECAP search homepage, you see three courts on the "More" tab of the court picker. After you do a search though, when you're on the RECAP search results page, you see dozens of courts.

### Expected behavior

You should only see three courts on the More tab when doing RECAP searches.

### Actual behavior

You see dozens of courts when you're on the results page.

### Steps to reproduce

1. Go to the RECAP search homepage: 

    https://www.courtlistener.com/recap/

2. Pull up the court picker and go to the More tab.

4. Note that there are three courts shown.

5. Do your search.

6. On the results page, open the court picker again and go to the More tab.

7. Note that there are now dozens of courts (but only three have PACER data)

### Screenshots

https://github.com/user-attachments/assets/ccac8d0e-b4f1-44cc-b804-6b7f7566abef

Low

elisa-a-v:
Fixes #5439
Fixes #6279 

## Summary

The jurisdiction picker in search results displayed courts that aren't in RECAP when searching RECAP, so we now filter correctly in the results view.

This PR also introduces a custom manager in the `Court` model with a method to filter courts that are valid in RECAP.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


cl/search/models.py
```diff
@@ -1970,6 +1970,14 @@ def military_courts(self) -> models.QuerySet:
         return self.filter(jurisdictions__in=Court.MILITARY_JURISDICTIONS)
 
 
+class CourtQuerySet(models.QuerySet):
+    def recap_courts(self):
+        """QuerySet of valid PACER courts."""
+        return self.filter(
+            pacer_court_id__isnull=False, end_date__isnull=True
+        ).exclude(jurisdiction=Court.FEDERAL_BANKRUPTCY_PANEL)
```
mlissner: Did you see this: https://github.com/freelawproject/courtlistener/pull/6526/files#diff-9ac9fb3ae1e0f66d9ce76f66a7adeccd088b9ed7e5ea88e033dcca5fffcf3e9cR1915? I think it accomplishes the same goal, no?

elisa-a-v: I'm not sure, that link brings me to this same snippet here. Did you maybe mean a different spot?

mlissner: Arg, sorry. See line 1915.

mlissner: Arg, sorry. See line 1915.

elisa-a-v: Thank you! I checked the actual filters but I'm unsure what's the correct one. If I understand the issue correctly, we should probably use the same query for both the advanced view and the results view, and AFAIK the jurisdictions currently shown in the RECAP Archive `advanced` view are correct, right?

Those are fetched using this query:

```
        courts = courts_in_use = Court.objects.filter(in_use=True)
        if request.path == reverse("advanced_r"):
            obj_type = SEARCH_TYPES.RECAP
            courts_in_use = courts.filter(
                pacer_court_id__isnull=False, end_date__isnull=True
            ).exclude(jurisdiction=Court.FEDERAL_BANKRUPTCY_PANEL)
```

OTOH, the one in line 1915 is as follows:

```
    def all_pacer_courts(self) -> models.QuerySet:
        return self.filter(
            Q(
                jurisdiction__in=[
                    Court.FEDERAL_DISTRICT,
                    Court.FEDERAL_BANKRUPTCY,
                    Court.FEDERAL_APPELLATE,
                ]
            )
            | Q(pk__in=["cit", "jpml", "uscfc", "cavc"]),
            end_date__isnull=True,
        ).exclude(pk="scotus")
```

They're clearly not the same query, but given the name of the method it would make sense to me to use the latter for RECAP (instead of the former which is the one we're currently using in the advanced view), so I wonder what's best here. Thoughts?

mlissner: Hm, don't know. Can you diff the results?

elisa-a-v: Sure, I ran them against the dev DB and found that all courts in the `advanced` view are also included in the `all_pacer_courts` query, but the latter has 10 additional courts not included in the former: `{'circtedsc', 'circtsdfl', 'indcirct', 'usdistct', 'circtmissterr', 'circtdms', 'circtndfl', 'okd', 'circtny', 'cavc'}`

Adding the `in_use=True` filter to those, we're left with just one court that's not included in the view which is `cavc`

mlissner: Adding `in_use` makes a lot of sense and `cavc` is indeed a PACER court, so I think that's a plan. I just checked and making that tweak to the `all_pacer_courts` method shouldn't have any unintended consequences (and will have some good side effects, actually).

elisa-a-v:
Okay I've removed the new manager and refactored the code to use `all_pacer_courts` instead 🫡 it is indeed much simpler now, thanks for the feedback!

btw, do we want tests for this? IIRC we said no but can't find it written down anywhere.

mlissner:
> btw, do we want tests for this? IIRC we said no but can't find it written down anywhere.

I think we can live without for now.

Issue #6456, opened 2025-10-28
Track semantic vs. keyword in the user query table


mlissner:
### Describe the issue.

We have a table where we store user queries. It doesn't distinguish between semantic and keyword queries, but it should. Let's add a column for that.

### What type of issue is this?

To Do

mlissner:
I'm hoping it makes sense to put @nadahlberg on this one. DB migrations can be painful, but it will give some visibility into how we do this stuff that should be useful.

PR #6548, opened 2025-11-17
Track semantic search vs keyword queries

nadahlberg:
## Fixes

Fixes: #6456 

## Summary

This PR tracks semantic versus keyword queries by doing the following:

1. Adds a `query_mode` column to the `SearchQuery` model with choices: (1, "Keyword"), (2, "Semantic")
2. Adds `.py` and `.sql` migration files
3. Updates the `store_search_query` and `store_search_api_query` to set the `query_mode` value when creating a `SearchQuery` instance. True if the "semantic" param is not falsy.
4. Adds two assert statements to existing tests to see if the `query_mode` value is being set correctly.

@mlissner let me know if I should move / remove the test tweaks. I couldn't figure out a good place for testing all the possibilities in terms of API vs. Website and semantic vs. keyword. This PR currently adds asserts to:

- `test_semantic_search_opinion.SemanticSearchTests.test_can_perform_a_regular_semantic_query` to test an API search with a "Semantic" query mode.
- `test_es_opinion.OpinionsESSearchTest.test_can_perform_a_regular_text_query` to test a website search with a "Keyword" query mode.

Maybe we don't want these at all, but if we do, would there be a good place to test a "keyword" API query, as well as a "semantic" website query?

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

1. To deploy...

### DIFF
--- a/cl/lib/search_utils.py
+++ b/cl/lib/search_utils.py
@@ -26,6 +26,7 @@ from cl.lib.elasticsearch_utils import (
     convert_str_date_fields_to_date_objects,
     fetch_es_results,
     get_facet_dict_for_search_query,
+    has_semantic_params,
     limit_inner_hits,
     merge_courts_from_db,
     merge_unavailable_fields_on_parent_document,
@@ -249,15 +250,6 @@ async def add_depth_counts(
             return cited_cluster
     else:
         return None
-
-
-def has_semantic_params(params: QueryDict) -> bool:
-    """Check if query parameters are a valid semantic search query."""
-    return bool(
-        params.get("semantic")
-        and params.get("q", "")
-        and params.get("type", SEARCH_TYPES.OPINION) in [SEARCH_TYPES.OPINION]
-    )
 
 
 def store_search_query(request: HttpRequest, search_results: dict) -> None:
--- a/cl/lib/elasticsearch_utils.py
+++ b/cl/lib/elasticsearch_utils.py
@@ -2874,6 +2874,23 @@ def build_semantic_query(
     return keyword_query, semantic_query
 
 
+def has_semantic_params(get_params: QueryDict | CleanData) -> bool:
+    """Check if parameters represent a valid semantic search query.
+
+    :param get_params: A QueryDict or CleanData object containing the
+    search parameters.
+    :return: True if the user enabled semantic search, the search type supports
+    it, and the query has content to generate an embedding or an embedding
+    is provided.
+    """
+    return bool(
+        get_params.get("semantic", False)
+        and (get_params.get("q", "") or get_params.get("embedding", None))
+        and get_params.get("type", SEARCH_TYPES.OPINION)
+        in [SEARCH_TYPES.OPINION]
+    )
+
+
 def build_full_join_es_queries(
     cd: CleanData,
     child_query_fields: dict[str, list[str]],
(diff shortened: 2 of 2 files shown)


mlissner:
This looks pretty good. Thanks @nadahlberg. I'll assign to @ERosendo for review. While we wait on that, we need a few other things. First, we need an issue in the infra repo to:

1. Migrate the schema. This can be a small issue pointing here and listing any special notes that are needed for the migration. (They know the general process though.) 

2. Migrate the existing data. This can be a one-off script or just some Python code that the infra team can run to back-fill the existing rows.

We also need tweaks to the admin page to filter by this field. Can you add that too, please? I think that can go in this PR easily enough?

cl/lib/search_utils.py
```diff
@@ -292,6 +296,7 @@ def store_search_api_query(
     :param engine: The search engine used to execute the query.
     :return: None
     """
+    is_semantic = bool(request.GET.get("semantic"))
```
ERosendo: We should refine this logic here (and anywhere it’s repeated). Semantic search is only supported for `Opinion` searches, but with the current implementation we may incorrectly create `SearchQuery` records marked as semantic for other search types, such as `Recap`. For example, the following request produces a record with the semantic query mode:

```
GET /search/?type=rd&semantic=true&court=ca1
```

Additionally, simply including the `semantic` parameter in the query string is not enough to trigger semantic search. We have validation logic that checks for the required fields before enabling vector search in Elasticsearch. See the relevant code here:

https://github.com/freelawproject/courtlistener/blob/24f1b61bb0385239ab338798aa6e8f70ec7c7af2/cl/lib/elasticsearch_utils.py#L2904-L2912

As shown in that snippet, our validation also requires the presence of the q parameter. With that in mind, the refined condition should look like this:

```suggestion
    is_semantic = (
        bool(request.GET.get("semantic"))
        and request.GET.get("q", "")
        and request.GET.get("type", SEARCH_TYPES.OPINION) in [SEARCH_TYPES.OPINION]
    )
```

It might also be worth refactoring this into a small helper function so we can reuse the logic consistently across the codebase.

mlissner: > It might also be worth refactoring this into a small helper function so we can reuse the logic consistently across the codebase.

I agree with this. It's the kind of thing we'll need again and again.

nadahlberg:
Thanks @ERosendo! I added a helper that accepts a QueryDict, but would it make sense to have this also accept a CleanData object (and incorporate the `cd.get("embedding", None)` condition) so that it could also be used in `elasticsearch_utils.py` as well? Or would you say that's messy, mixing types, and that we should keep the elastic / search_utils separate?

i.e. would you prefer this:

```python
def has_semantic_params(params: QueryDict | CleanData) -> bool:
    """Check if query parameters are a valid semantic search query."""
    return bool(
        params.get("semantic")
        and (params.get("q", "") or params.get("embedding", None))
        # Or maybe `(params.get("embedding", None) and isinstance(params, CleanData))`
        and params.get("type", SEARCH_TYPES.OPINION) in [SEARCH_TYPES.OPINION]
    )
```

ERosendo:
@nadahlberg  I like your suggestion. I think centralizing the validation in a single helper makes sense, and for our use case both `QueryDict` and `CleanData` expose a similar `.get()` interface. Your refactor for `has_semantic_params` looks good to me as-is, we don’t need the `isinstance` check.

This also wouldn’t be the first helper with that type hint. Here’s an example:

https://github.com/freelawproject/courtlistener/blob/78676f7f25a3e5c7b86110c97dff9f22d27cbd2b/cl/lib/elasticsearch_utils.py#L3127-L3131

`CleanData` is essentially an alias for a dict with `string` keys and arbitrary values, and `QueryDict` inherits most dict-like behavior, so having the helper accept both feels reasonable to me. Let’s go with that approach.

This PR is pretty much ready. After adding the refactor, I think we’ll be good to merge.

nadahlberg:
@ERosendo that makes sense. Moved the helper to the elastic utils and incorporated in `build_full_join_es_queries`. Thanks for the help!

PR #6550, opened 2025-11-17
feat(admin): user admin tweaks

### LINKED ISSUE #6428: More admin tweaks
### Describe the issue.

I realized today that we don't have an easy way of seeing the tags for somebody. 

Let's add a button to the admin next to these ones: 

Image

That way, admins can click the button and see what tags the user has created.

### Related Issue

_No response_

### Screenshots

_No response_

To Do

elisa-a-v:
Closes #6428 

## Summary

This PR:
- Adds a new button at the top of the change user view that links to the list of Tags a user has created. The button isn't added if the user has no tags, but this can be easily changed if we want the button to be always visible.
- Refactors the user admin to reduce repeated code.
- Enables searching users by pk.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

<details closed>

### Filter users by pk

This is the dev DB so there's only two users and my user id is 2:

image

### Link to user tags in change view

image

</details>

### DIFF
--- a/cl/api/tests.py
+++ b/cl/api/tests.py
@@ -557,13 +557,14 @@ class ApiEventCreationTestCase(TestCase):
         request.user = self.user
         await sync_to_async(view)(request, **{"version": f"{api_version}"})
 
+    @mock.patch("cl.api.utils.create_or_update_zoho_account")
     @mock.patch(
         "cl.api.utils.get_logging_prefix",
         return_value="api:Test",
     )
     @mock.patch.object(LoggingMixin, "milestones", new=[1])
     async def test_are_v3_events_created_properly(
-        self, mock_logging_prefix
+        self, mock_logging_prefix, mock_zoho_task
     ) -> None:
         """Are event objects created as v3 API requests are made?"""
         await self.hit_the_api("v3")
@@ -579,14 +580,16 @@ class ApiEventCreationTestCase(TestCase):
             f"User '{self.user.username}' has placed their 1st API v3 request."
         )
         self.assertEqual(event_descriptions, expected_descriptions)
+        mock_zoho_task.delay.assert_not_called()
 
+    @mock.patch("cl.api.utils.create_or_update_zoho_account")
     @mock.patch(
         "cl.api.utils.get_logging_prefix",
         return_value="api:Test",
     )
     @mock.patch.object(LoggingMixin, "milestones", new=[1])
     async def test_are_v4_events_created_properly(
-        self, mock_logging_prefix
+        self, mock_logging_prefix, mock_zoho_task
     ) -> None:
         """Are event objects created as V4 API requests are made?"""
         await self.hit_the_api("v4")
@@ -602,6 +605,7 @@ class ApiEventCreationTestCase(TestCase):
             f"User '{self.user.username}' has placed their 1st API v4 request."
         )
         self.assertEqual(event_descriptions, expected_descriptions)
+        mock_zoho_task.delay.assert_called_once_with(self.user.pk, 1)
 
     # Set the api prefix so that other tests
     # run in parallel do not affect this one.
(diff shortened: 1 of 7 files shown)


cl/users/admin.py
```diff
@@ -76,37 +77,55 @@ class UserAdmin(admin.ModelAdmin, AdminTweaksMixin):
...
         user = self.get_object(request, object_id)
-
-        extra_context["proxy_events_url"] = build_admin_url(
-            UserProxyEvent,
-            {"pgh_obj": object_id},
+        profile_id = (
+            user.profile.pk if user and hasattr(user, "profile") else None
+        )
+        has_tags = (
+            user and hasattr(user, "user_tags") and user.user_tags.count()
```
mlissner: We'll always have a profile so no need to check that. I think for the tags it's fine to always show it because otherwise I'll get confused, "Wasn't there a button here before??" Maybe just keep these simple, and same for search queries?

cl/lib/admin.py
```diff
@@ -45,3 +45,46 @@ def build_admin_url(
...
+class AdminLinkConfig(TypedDict):
+    should_add: bool
+    label: str
+    model_class: type[Model]
+    query_params: dict[str, Any] | None
+
+
+class AdminLink(TypedDict):
+    href: str
+    label: str
```
mlissner, line 58: Nice.

cl/lib/admin.py
```diff
@@ -45,3 +45,46 @@ def build_admin_url(
...
+    - label: Label for the link.
+    - model_class: The Django model class for which the admin URL will be built.
+    - query_params: Query parameters to append to the URL (e.g. {"docket": 123}).
+    """
+    generated_links: list[AdminLink] = []
+
+    for link in links:
+        try:
+            if link["should_add"]:
+                generated_links.append(
```
mlissner: Early continue here, please.

cl/lib/admin.py
```diff
@@ -45,3 +45,46 @@ def build_admin_url(
...
+                    {
+                        "href": build_admin_url(
+                            link["model_class"],
+                            query_params=link["query_params"],
+                        ),
+                        "label": link["label"],
+                    }
+                )
+        except KeyError:
+            continue
```
mlissner: Seems better to fail loudly if this is misconfigured?

elisa-a-v:
Thanks for the feedback! `generate_admin_links` is much simpler now 👌🏽 

- Links are now always shown.
    - No checks for profile, alarms or tags in user admin.
    - No `should_add` key in `generate_admin_links`
- `try/except` has been removed in `generate_admin_links` so now we fail loudly if misconfigured.
- Also replaced the `app.get_model` call for `SearchQuery` with a pythonic `import` in `users/admin.py`. The models for events are still retrieved with `app.get_model` because they can't be imported directly from the user module since they're added by a third party library, but we now cast them to make type checkers happy.

cl/lib/admin.py
```diff
@@ -45,3 +45,41 @@ def build_admin_url(
...
+    label: str
+
+
+def generate_admin_links(
+    links: list[AdminLinkConfig],
+) -> list[AdminLink]:
+    """
+    Construct a list of links with links to given admin views.
+    Each link should have the following keys:
+    - should_add: Condition to check whether the link should be added.
```
mlissner:
```suggestion
```

mlissner (approved):
Nice. One last thing that I'll just commit myself.

mlissner:
Automerge is set. Thank you!

cl/lib/admin.py
elisa-a-v: oops, thanks for catching that!

PR #6554, opened 2025-11-18
feat(lib): Use S3 cache backend in `cache_page_ignore_params`

ERosendo:
## Fixes

Relates to #6418 but doesn't fully resolve it.

## Summary

This PR updates the `cache_page_ignore_params` decorator to support using the S3 cache backend in production.

It introduces two main changes:

- Use S3 cache in production: We now route cached responses to the S3-backed cache in production. For now, only the production environment has an S3 cache directory, so development continues using the default/local cache.

- Add a time-based prefix to cache keys: The decorator now computes a prefix based on the timeout value,

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/assets/templates/admin/user_change_form.html
+++ b/cl/assets/templates/admin/user_change_form.html
@@ -1,26 +1,12 @@
 {% extends "admin/change_form.html" %}
 
 {% block object-tools-items %}
-  {% if proxy_events_url %}
+  {% for link in custom_links %}
     <li>
-      <a class="historylink" href="{{ proxy_events_url }}">
-        UserProxy Events
+      <a class="historylink" href="{{ link.href }}">
+        {{ link.label }}
       </a>
     </li>
-  {% endif %}
-  {% if profile_events_url %}
-    <li>
-      <a class="historylink" href="{{ profile_events_url }}">
-        UserProfile Events
-      </a>
-    </li>
-  {% endif %}
-  {% if search_queries_url %}
-    <li>
-      <a class="historylink" href="{{ search_queries_url }}">
-        Search Queries
-      </a>
-    </li>
-  {% endif %}
+  {% endfor %}
   {{ block.super }}
 {% endblock object-tools-items %}
--- a/cl/lib/admin.py
+++ b/cl/lib/admin.py
@@ -1,4 +1,4 @@
-from typing import Any
+from typing import Any, TypedDict
 from urllib.parse import urlencode
 
 from django.contrib.contenttypes.admin import GenericTabularInline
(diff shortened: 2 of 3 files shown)


cl/lib/decorators.py
```diff
@@ -105,7 +107,14 @@ def decorator(view_func):
         async def _wrapped_view(request, *args, **kwargs):
             url_path = urlparse(request.build_absolute_uri()).path
             hash_key = md5(url_path.encode("ascii"), usedforsecurity=False)
-            cache_key = f"custom.views.decorator.cache:{hash_key.hexdigest()}"
+            # Use S3 cache in production only
+            cache = (
+                caches["s3"] if not settings.DEVELOPMENT else caches["default"]
+            )
```
mlissner: This is interesting, but I don't like how it uses S3 without saying so in the name or parameters. This decorator should be usable without reading its code and figuring out that it uses an exotic (and slow-ish) cache. 

Maybe we default to the regular cache, but allow the user to pass through a different one if they want via a parameter?

cl/lib/decorators.py
```diff
@@ -105,7 +107,14 @@ def decorator(view_func):
...
+            # Compute the time-based prefix
+            days = int(ceil(timeout / 60 * 60 * 24))
+            time_based_prefix = f"{days}-day" if days == 1 else f"{days}-days"
```
mlissner: Makes me think we should rename our lifecycle rule to `1-days`, plural rather than have this workaround, or maybe add a second lifecycle rule for the special case and let people do what they feel?

mlissner (changes requested):
A couple thought for you!

cl/lib/decorators.py
ERosendo: > Makes me think we should rename our lifecycle rule to 1-days, plural rather than have this workaround, or maybe add a second lifecycle rule for the special case and let people do what they feel?

I think we can add the second rule. The one-rule-per-day setup is just a recommendation, users of the S3 cache backend are free to choose a different approach. We suggest a daily rule simply because it's the smallest interval.

ERosendo: Added second rule via https://github.com/freelawproject/infrastructure/pull/547

cl/lib/decorators.py
ERosendo: In commit [1f53ca5](https://github.com/freelawproject/courtlistener/pull/6554/commits/1f53ca5741c19dfad147c3b6e312dfeaa8e15c03), I added a new `cache_alias` parameter. By default, it uses the `default` cache, but we can pass a different cache alias if needed. The decorator will log an error if an invalid alias is provided and also has logic to compute time_based prefixes when the s3 cache backend is used.

cl/lib/decorators.py
```diff
@@ -105,7 +109,38 @@ def decorator(view_func):
...
+            # handle S3 cache keys differently
+            if should_use_time_based_prefix:
+                days = int(ceil(timeout / (60 * 60 * 24)))
+                # Compute the time-based prefix
+                time_based_prefix = f"{days}-days"
+                cache_key = f"{time_based_prefix}:custom.views.decorator.cache.{hash_key.hexdigest()}"
+            else:
+                cache_key = (
+                    f"custom.views.decorator.cache:{hash_key.hexdigest()}"
+                )
```
mlissner: Maybe:

```python
            cache_key = (
                f"custom.views.decorator.cache:{hash_key.hexdigest()}"
            )
            if should_use_time_based_prefix:
                days = int(ceil(timeout / (60 * 60 * 24)))
                cache_key = f"{days}-days:{cache_key}"
```

ERosendo: Thanks, this looks much better. I’ll go ahead and apply the tweak.

Issue #6373, opened 2025-10-06
Add option to `delete_duplicates` to compare extracted text


grossir:
- Add option to compare `plain_text` or `html` after standard clean ups

This will mostly help delete duplicate PDFs. For timestamped PDFs it is not straightforward to implement a `cleanup_content`. Even deleting the "CreationDate" metadata attribute in the PDF binary won't solve the issue. However, content is mostly the same, except for the timestamp. 

We are seeing this with `coloctapp` PDFs; will also happen with `neb`

Standard cleanups should be datestrings as the ones seen in timestamps, whitespace, etc

<details>
<summary>Examples
</summary>

https://www.courtlistener.com/opinion/10297352/the-people-of-the-state-of-colorado-v-tommy-rae-mickey/pdf/
https://www.courtlistener.com/opinion/10297292/the-people-of-the-state-of-colorado-v-tommy-rae-mickey/pdf/

```python
from cl.scrapers.management.commands.merge_opinion_versions import explain_version_differences
from cl.search.models import Opinion

explain_version_differences([10763940, 10763880, 10763811])

# text is mostly the same except for timestamps
Difference between 10763940 10763880
https://www.courtlistener.com/opinion/10297352/x/
https://www.courtlistener.com/opinion/10297292/x/
SequenceMatcher.ratio(text1, text2) 0.998011434253045
SequenceMatcher.ratio(text2, text1) 0.998011434253045
Difference between 10763940 10763811
https://www.courtlistener.com/opinion/10297352/x/
https://www.courtlistener.com/opinion/10297223/x/
SequenceMatcher.ratio(text1, text2) 0.998011434253045
SequenceMatcher.ratio(text2, text1) 0.998011434253045
Difference between 10763880 10763811
https://www.courtlistener.com/opinion/10297292/x/
https://www.courtlistener.com/opinion/10297223/x/
SequenceMatcher.ratio(text1, text2) 0.998011434253045
SequenceMatcher.ratio(text2, text1) 0.9975142928163062

x = Opinion.objects.get(id=10763940)
y = Opinion.objects.get(id=10763880)

content = x.local_path.read()
content2 = y.local_path.read()

In [14]: content[:1000]
[...]

PR #6557, opened 2025-11-19
feat(delete_duplicates): add option to compare extracted text

grossir:
Solves #6373

Add an option to the `delete_duplicates` command to compare extracted text.

Only supports neb, nebctapp, coloctapp for now, which are known cases where the only difference between duplicates are timestamps

Refactor code to reuse opinion grouping by download_url

Add tests for new functionality


cl/scrapers/management/commands/merge_opinion_versions.py
```diff
@@ -326,6 +326,45 @@ def get_query_from_url(url: str, url_filter: str) -> Q:
...
+        )
+        .order_by()
+        # compute the number of  distinct hashes to prevent colliding with
+        # actual duplicates, which will be handled in another command
+        .filter(number_of_rows__gte=2, number_of_hashes__gte=2)
+    )
+    return qs
+
+
+def get_same_url_opinions(opinion_group: dict, seen_urls: set):
```
quevon24: Could you please add the docstrings here?

grossir: Thanks @quevon24 I added the docstrings and type annotations

Issue #6404, opened 2025-10-15
Delete web or error pages ingested as opinions


grossir:
Parent issue to group opinions needing deletion because they are not opinions, but web pages or error pages or error documents

grossir:
A round 60 more with  "Page not found"


https://www.courtlistener.com/?q=%22page%20not%20found%22&type=o&order_by=score%20desc&stat_Published=on&stat_Unpublished=on&stat_Separate=on&stat_Relating-to=on&stat_Unknown=on


- mass / massappct
     - "Page not found (error 404)", 4 [results](https://www.courtlistener.com/?q=%22Page%20not%20found%20(error%20404)%22&type=o&order_by=dateFiled%20asc&stat_Published=on&stat_Unpublished=on&stat_Separate=on&stat_Relating-to=on&stat_Unknown=on)
     - "State of Mississippi Judiciary.+Page not found." single [result](https://www.courtlistener.com/opinion/4661586/william-kevin-hadley-v-fedex-ground-package-system-inc/?q=%22page+not+found%22&type=o&order_by=score+desc&stat_Published=on&stat_Unpublished=on&stat_Separate=on&stat_Relating-to=on&stat_Unknown=on)

<img width="621" height="653" alt="Image" src="https://github.com/user-attachments/assets/ce16ce61-ee4a-4cac-ac04-c88e4de99ea5" />

- idaho
    - "Page not found | Supreme Court"; 55 [results](https://www.courtlistener.com/opinion/4661586/william-kevin-hadley-v-fedex-ground-package-system-inc/?q=%22page+not+found%22&type=o&order_by=score+desc&stat_Published=on&stat_Unpublished=on&stat_Separate=on&stat_Relating-to=on&stat_Unknown=on)

<img width="697" height="781" alt="Image" src="https://github.com/user-attachments/assets/0e36e86c-6541-4544-bbf7-f349c4ef17a6" />


[...]

PR #6572, opened 2025-11-25
Add documentation for Opinion Cluster redirection

quevon24:
## Summary

This PR adds documentation explaining how OpinionClusters are redirected using the `ClusterRedirection` model. It describes why clusters may be removed (duplicate detection, newer versions, sealed cases, etc.) and clarifies that the redirect records exist internally but are not yet exposed in the REST API. This helps replication clients understand why older `cluster_id` values may resolve to newer clusters.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/.github/scripts/frontend_checks.py
+++ b/.github/scripts/frontend_checks.py
@@ -10,6 +10,7 @@ Input file must be in git --name-status format (STATUS\\tPATH per line).
 from __future__ import annotations
 
 import argparse
+import fnmatch
 import re
 import sys
 from dataclasses import dataclass
@@ -956,6 +957,13 @@ def main() -> int:
         default=None,
         help="Write markdown summary to this file (only if findings exist)",
     )
+    parser.add_argument(
+        "--skip-files",
+        nargs="+",
+        default=[],
+        metavar="GLOB",
+        help="Glob patterns for files to skip (e.g. 'cl/assets/templates/v2_foo/*')",
+    )
     args = parser.parse_args()
 
     repo_root = Path(args.repo_root).resolve()
@@ -982,7 +990,10 @@ def main() -> int:
 
     # Filter to relevant files
     relevant = [
-        f for f in changed_files if f.endswith(".html") or is_input_css(f)
+        f
+        for f in changed_files
+        if (f.endswith(".html") or is_input_css(f))
+        and not any(fnmatch.fnmatch(f, g) for g in args.skip_files)
     ]
 
     if not relevant:
--- a/.github/workflows/frontend-lint.yml
+++ b/.github/workflows/frontend-lint.yml
@@ -38,7 +38,8 @@ jobs:
             --repo-root . \
             --format github \
             --changed-files changed_files.txt \
-            --summary-file summary.md
+            --summary-file summary.md \
+            --skip-files 'cl/*/test_assets/*.html'
 
       - name: Minimize previous bot comments
         if: always()
--- a/uv.lock
+++ b/uv.lock
@@ -1855,7 +1855,7 @@ wheels = [
 
 [[package]]
 name = "juriscraper"
-version = "3.0.6"
+version = "3.0.7"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "certifi" },
@@ -1874,9 +1874,9 @@ dependencies = [
     { name = "selenium" },
     { name = "tldextract" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/03/50/3a742fcb16afbf312072c111853658412e6fd528911a769051f21cdf0a76/juriscraper-3.0.6.tar.gz", hash = "sha256:ee27738012505164e3f1ad49a92391bb227858ba0fef09e3d2ea82cc37b21987", size = 390176, upload-time = "2026-04-02T21:52:26.415Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/65/39/ef0170d3d171734ba3ce769aa07a0245d2f6a8f5ac047f8f8b5087d36a6e/juriscraper-3.0.7.tar.gz", hash = "sha256:0c1631b28cba386af373c289fea6ae09cd597994de774d09c555a38c59a299a0", size = 391922, upload-time = "2026-04-06T22:14:29.14Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/21/ec/996541790ae06dea6e15eb270a424ebda5c631d5ff4256583af9466c5a68/juriscraper-3.0.6-py3-none-any.whl", hash = "sha256:21acf74a35ff05c17c8413c4cca501b2bf1cb9b8b766aed1342b0a52c063a5ff", size = 619164, upload-time = "2026-04-02T21:52:24.56Z" },
+    { url = "https://files.pythonhosted.org/packages/83/5e/f385cc9f8ba97abcd7623bec3127e4c78c8f032741f9b718245371b11cd0/juriscraper-3.0.7-py3-none-any.whl", hash = "sha256:a9bbd185f7c6cecea5b562728e5406fc2f9d089e1d9427451462307390d4fc5e", size = 620811, upload-time = "2026-04-06T22:14:27.115Z" },
 ]
 
 [[package]]


mlissner:
I'd like to take a look at this one once it's been reviewed. Can you please flag me before merging?

cl/api/templates/case-law-api-docs-vlatest.html
```diff
@@ -202,6 +202,42 @@ <h3 id="opinion-endpoint">Opinions <small> — <code>{% url "opinion-list" versi
...
+
+
+    <p>
+        <strong>Note:</strong> Cluster redirection data is not yet exposed via the
+        REST API. Replication clients should be aware that an old
+        <code>cluster_id</code> may internally redirect to a different, newer
+        cluster.
+    </p>
+
+
```
grossir, line 227: We should add a note about deleted clusters with no remaining references.

- non opinion documents ingested into the opinion and opinion cluster tables, such as error pages, opinion listings, etc

Related to #6404

cl/api/templates/case-law-api-docs-vlatest.html
```diff
@@ -202,6 +202,42 @@ <h3 id="opinion-endpoint">Opinions <small> — <code>{% url "opinion-list" versi
...
+        active one.
+    </p>
+
+    <p>
+        Clusters may be removed or consolidated for several reasons, including:
+    </p>
+
+    <ul>
+        <li>Duplicate clusters being merged</li>
+        <li>Clusters replaced by a newer version of the same opinion</li>
```
grossir: wording is incorrect here; a cluster is not replaced by a newer version of the opinion. Something like the following should be correct


> Sometimes a new version of an existing opinion is published, and incorrectly indentified as a different cluster. When our system corrects this, the two opinions are consolidated into a single cluster, and the oldest opinion is marked as being a version of the newer one. The other clusters are deleted

cl/api/templates/case-law-api-docs-vlatest.html
```diff
@@ -202,6 +202,42 @@ <h3 id="opinion-endpoint">Opinions <small> — <code>{% url "opinion-list" versi
...
+        <li>Clusters removed or sealed by court order</li>
+    </ul>
```
grossir: Add a new entry about cluster consolidation

> Some courts publish a file for each opinion in a cluster: the lead opinion, a dissent, a concurrence. When this happens, the opinions may be spread into different cluster objects. When our system corrects this, we keep a single cluster and delete the others.

mlissner:
Does this need to be applied to the new templates?

quevon24:
I updated the documentation text and applied changes to the v2 template.

image


image

cl/api/templates/case-law-api-docs-vlatest.html
```diff
@@ -215,6 +215,63 @@ <h3 id="opinion-endpoint">Opinions <small> — <code>{% url "opinion-list" versi
...
+            Opinions from the same case (lead opinion, dissent, concurrence)
+            spread across different clusters because courts published them as
+            separate files. When corrected, we keep a single cluster and
+            delete the others.
+        </li>
+        <li>Clusters removed or sealed by court order.</li>
+        <li>
+            Non-opinion documents (such as error pages or opinion listing
+            pages) incorrectly ingested as opinion clusters.
+        </li>
```
grossir: these deleted clusters shouldn't be included in the list, because they don't have redirections...

cl/api/templates/case-law-api-docs-vlatest.html
```diff
@@ -215,6 +215,63 @@ <h3 id="opinion-endpoint">Opinions <small> — <code>{% url "opinion-list" versi
...
+    </ul>
+
+    <p>
+        When a cluster is deleted and redirected, its <code>cluster_id</code> is
+        stored as a <code>deleted_cluster_id</code>, and a redirection record
+        points to the surviving cluster. This ensures no dead links and allows
+        clients to maintain persistent references even across merges or data
+        corrections.
```
grossir: Maybe there should be a explicit mention here about SEALED clusters returning a 410; and the other types returning a 301, and actual redirection

(maybe the model itself could be named better :/)

cl/api/templates/v2_case-law-api-docs-vlatest.html
```diff
@@ -17,6 +17,7 @@
       {'href': '#docket-endpoint', 'text': 'Dockets'},
       {'href': '#cluster-endpoint', 'text': 'Clusters'},
       {'href': '#opinion-endpoint', 'text': 'Opinions'},
+      {'href': '#cluster-redirections', 'text': 'Cluster Redirections'},
```
grossir, line 20: is there a navigation panel on the old template? If there is, there is no entry for the new Cluster Redirection section

cl/api/templates/case-law-api-docs-vlatest.html
```diff
@@ -215,6 +215,63 @@ <h3 id="opinion-endpoint">Opinions <small> — <code>{% url "opinion-list" versi
     </li>
   </ul>
 
+    <h4 id="cluster-redirections">Opinion Cluster Redirections (Coming Soon to API)</h4>
```
grossir: all these new tags are floating around. Should they be grouped into a container tag? (Section maybe?)

grossir: about "Coming Soon to API" we should probably have a unified style / component? I had some different style here https://github.com/freelawproject/courtlistener/pull/6897

What do you think @elisa-a-v ?

quevon24:
@grossir I addressed all your comments except the "Coming Soon to API" style, which is pending input from @elisa-a-v

cl/api/templates/case-law-api-docs-vlatest.html
elisa-a-v: Yes that's a great idea! I notice for [legacy templates](https://github.com/freelawproject/courtlistener/pull/6897/changes#diff-ddcb57b1ba21d99275526d4d5f57831d9bd0fd8109f4b1511574579d65afc0a7R77) you simply used `label label-info` which is very easy to replicate so we can use that here too.

In the case of [new templates](https://github.com/freelawproject/courtlistener/pull/6897/changes#diff-f304d0889ae8c59de5fecc8972954c41839f113bed90b6c41d190db56a70d3f8R115) we could turn `text-xs bg-primary-100 text-primary-700 px-2 py-0.5 rounded-full` into a utility class in `cl/assets/tailwind/input.css` in the components layer, maybe call it `chip` or something similar. Please note I didn't include the `ml-1` because that's placement-related and is better to keep decoupled from the styling itself.

Appreciate the attention to this 🙌🏽

quevon24:
Thanks @elisa-a-v

I implemented the "Coming Soon" badge style as suggested. The old template uses `label label-info` and the new template uses a new `chip` utility class in input.css. Below you can see the changes:

image

image

mlissner:
Sorry, I was curious what documentation we were doing here. I'm not sure what's here quite helps, and I think it's a bit confusing because it's talking about front end functionality in the API documentation. 

If I'm a replication user, I wouldn't know how to use this information because it doesn't explain what table to look at (plus, I wouldn't find this b/c I probably wouldn't be looking at the API documentation). If I'm an API user, this doesn't really help me either because there's no API. 

Hm. I think what we want to do is encourage people to ignore items that we've concluded are dups, but I'm not sure this does that. If this resonates, could y'all take a look at how we might change this to make it actionable by API and/or replication users? Maybe it goes on a different page or maybe it needs to wait until we have an API for it? Something else?

grossir:
you are right

I think we saw the need for some docs after we deleting duplicated and versioned clusters. We got some questions from users (If I remember correctly, replication users) wondering what happened with some of the clusters they had saved

This should go in the "Replication" documentation, since these users are the only ones that can access those tables right now.

The way API users will see this for now are redirections, so we should add a smaller note about redirections and the possible reasons, perhaps linked to the replication section that contains a more in-depth explanation

mlissner:
Sounds good to me. Is it hard to add the redirect information to the current APIs? Maybe it works as a nested key on OpinionClusters?

quevon24:
> Sounds good to me. Is it hard to add the redirect information to the current APIs? Maybe it works as a nested key on OpinionClusters?

What do you think about this, a small note in the api help:

image

And move the section to replication help:

image

If you think it looks good or more or less good, I can push this change and just adjust what's necessary.

And related to add the redirection information to the API. I think it shouldn't be hard. The `ClusterRedirection` model already has a ForeignKey back to `OpinionCluster` (via merged_clusters), and the `OpinionClusterViewSet` already handles redirections for deleted clusters. We'd just need to create a small serializer for `ClusterRedirection` and add it as a nested field on `OpinionClusterSerializer`. We could implement it in this PR or create a new one just for that.

mlissner:
Thanks for the work on this and for trying to satisfy me. Sorry I'm difficult. 

Now that I'm looking in the replication docs, I'm remembering that they don't have technical guidance otherwise, so it looks quite out of place there. Maybe it's worth making a separate page for this that's general purpose and explains about the redirection table and system in a way that makes sense to both API and replication users? That could then be linked under the Cluster documentation in the API, and replication users could probably find it that way?

quevon24:
> Thanks for the work on this and for trying to satisfy me. Sorry I'm difficult.
> 
> Now that I'm looking in the replication docs, I'm remembering that they don't have technical guidance otherwise, so it looks quite out of place there. Maybe it's worth making a separate page for this that's general purpose and explains about the redirection table and system in a way that makes sense to both API and replication users? That could then be linked under the Cluster documentation in the API, and replication users could probably find it that way?

I think the cluster redirection documentation can go in the `/help/` section as a standalone page at `/help/cluster-redirections/`. That way it's a general purpose explanation of the redirection system, and both the Case Law API docs and the Replication docs can just link to it. This avoids duplicating content and keeps the replication page focused on what it's about.

It will look something like this:

image

image

image

On a separate note, I've also implemented exposing the redirection data on the Clusters API endpoint as a nested merged_clusters field. The response would look like this:

```
{
      "id": 123456,
      "absolute_url": "/opinion/123456/go/",
      "citations": [...],
      "merged_clusters": [
          {
              "deleted_cluster_id": 1234567,
              "reason": 2,
              "date_created": "2026-02-16T12:00:00Z"
          },
          {
              "deleted_cluster_id": 7654321,
              "reason": 1,
              "date_created": "2026-02-16T08:30:00Z"
          }
      ],
      ...
  }
```

Should I include the API changes in this same PR? That way I can also mention the new `merged_clusters` field in the help page. Otherwise I can do it in a new PR, but that would require updating the documentation again after this one merges.

mlissner:
That seems better. I can't speak to the specific API implementation, but I am worried about how that would work if there are hundreds of duplicates of a particular case. Is that a risk?

quevon24:
I added the cluster redirection documentation as a new help page here: `/help/cluster-redirections/`

I'll create a sub-issue to integrate the cluster redirection in the API.

cl/simple_pages/urls.py
```diff
@@ -68,6 +69,11 @@
     path("help/pray-and-pay/", prayer_help, name="pray_and_pay_help"),  # type: ignore[arg-type]
     path("help/relative-dates/", relative_dates, name="relative_dates"),  # type: ignore[arg-type]
     path("help/mcp/", mcp_help, name="mcp_help"),  # type: ignore[arg-type]
+    path(
+        "help/cluster-redirections/",
```
grossir, line 73: it seems like this doesn't exist. All the links leading to this give the error below

Image

cl/assets/tailwind/input.css
```diff
@@ -127,6 +127,9 @@
   .screenshot {
     @apply rounded-2xl block max-w-full;
   }
+  .chip {
+    @apply text-xs bg-primary-100 text-primary-700 px-2 py-0.5 rounded-full;
+  }
```
grossir: added but unused?

elisa-a-v:
Oops, I meant to check in before merging main into this branch, but jumped the gun. I wanted to pull in the new frontend checks to surface any issues before @quevon24 addressed @grossir's comments, but should have given a heads up first, my bad!

cl/simple_pages/urls.py
quevon24: Yes, it was my mistake, I didn't add the file to the commit

elisa-a-v:
Thanks for that last commit! For the record, the update makes it explicit that certain classes are specific for the `<a>` tags and not the entire descendants, so any new text outside the anchors but inside the list is styled differently:

```html
        <ol class="list-decimal [&_a]:text-primary-600 marker:text-greyscale-800 marker:font-normal text-sm font-normal space-y-2">
          <li><a href="{% url "alert_help" %}">Help with search and docket alerts</a> &mdash; this is outside the &lt;a&gt; tag!</li>
```

image

#7101 fixes the false positive by checking for tailwind descendant selectors which is what we're using here, so once that is merged, the checks should pass!

mlissner:
A thought here. The docs say that we don't delete things, but once in a while I do. Is it worth:
1. Adding a quick page to the wiki that says how to set up these redirections via admin? 
2. Maybe removing the delete option from the admin so people like me can't make mistakes anymore? 

Number 2 is probably a pain, but number 1 is probably worth doing if it takes 5-10 minutes?

The wiki has a section for case law: https://wiki.free.law/c/courtlistener/dev-guide/case-law

Just thinking aloud about how to do this well....

quevon24:
> A thought here. The docs say that we don't delete things, but once in a while I do. Is it worth:
> 
> 1. Adding a quick page to the wiki that says how to set up these redirections via admin?
> 2. Maybe removing the delete option from the admin so people like me can't make mistakes anymore?
> 
> Number 2 is probably a pain, but number 1 is probably worth doing if it takes 5-10 minutes?
> 
> The wiki has a section for case law: https://wiki.free.law/c/courtlistener/dev-guide/case-law
> 
> Just thinking aloud about how to do this well....

For 1, happy to add a quick wiki page to the case law section. 

One option is to add it manually from the admin panel. You only need the ID of the cluster to be deleted, the ID of the cluster to which it redirects, and the reason for the redirection.
The other option is to find the cluster in admin, use the "Seal clusters" action, confirm there are no blockers, and the redirection gets created automatically

For 2, it's actually pretty simple. We just need to override `has_delete_permission` on the  `ClusterRedirectionAdmin` admin class to return False. I can add that to this PR if you want or create a new one just for this change.

mlissner:
> I can add that to this PR if you want or create a new one just for this change.

Sure, your call. Glad to hear it's easy!

quevon24:
The new wiki page is here:

https://wiki.free.law/c/courtlistener/dev-guide/case-law/setting-up-cluster-redirections-via-the-admin

grossir:
@mlissner does this need a final check from you?

mlissner:
No, I think if you're all happy with it, we should just go for it and tweak later if needed.

PR #6583, opened 2025-11-27
feat(clone_from_cl): update cloning command

quevon24:
When I was working on the issue of unextracted content from opinions, I needed to make some changes to the command to download the opinion file (from `local_path` field), as well as adjust some things so I wouldn't have to delete and clone objects every time I modified them and needed to test again. I applied those changes to the rest of the code.

## Summary

- By default, the tool now uses `update_or_create` to **update existing objects** (Docket, Opinion Cluster, Person, etc.) with the latest data from the CourtListener API.
- Implemented the `--no-update` parameter consistently across all primary and recursive cloning functions (including opinions, docket entries, and positions) to **skip updating** any object that already exists in the local database.
- Added support for cloning `Audio` records directly via the command line (e.g., `--type audio.Audio`).
- The internal code structure was refactored for improved consistency and clarity
- Added the option to clone opinion files from `local_path` field

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1755,7 +1755,7 @@ wheels = [
 
 [[package]]
 name = "juriscraper"
-version = "2.6.96"
+version = "2.6.97"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "certifi" },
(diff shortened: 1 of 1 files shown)


grossir (approved):
Looks good!

Issue #6582, opened 2025-11-26
Remove banner inviting research participants


s-taube:
We have enough user research participants, so we can remove the [banner on the Profile page](https://github.com/freelawproject/courtlistener/issues/6474#issuecomment-3492639451) in CourtListener.

s-taube:
CC @mlissner

PR #6602, opened 2025-11-30
fix(profile): remove user research participants banner

mlissner:
## Fixes
Fixes: #6582

## Summary
This PR removes the user research participants banner from the profile pages. We have enough participants for the wireframe testing of the new unified view of alerts/notes/tags, so the banner is no longer needed.

The banner was added in commit 26fe24b44 and subsequently modified for styling. This change simply removes the include line from the profile nav template, reverting it to its previous state while keeping the general improvements made to the dismissible banner component (which are still used by other banners like the testimonial banner).

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

1. Standard web deployment - no special steps required

### DIFF
--- a/cl/stats/tests.py
+++ b/cl/stats/tests.py
@@ -2,7 +2,6 @@ from datetime import datetime
 
 import pytest
 import time_machine
-from asgiref.sync import async_to_sync
 from django.core import mail
 from django.core.management import call_command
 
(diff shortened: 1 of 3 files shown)


mlissner:
This is my second PR that used Claude Code. It's interesting and I think it works nicely! For this one I had it review the git history for the changes you made so that it would do a good job. Curious what you think!

mlissner:
> so I guess I could give Claude Code a try 

It was kind of fun. With little bugs like this you mostly sit back and press enter every 30s to accept everything it wants to do until it's done.

Issue #6539, opened 2025-11-17
Research model changes for Opinion in Scanning Project


flooie:
### Describe the issue.

Add storage and metadata fields to support AI-extracted text. This includes creating (or reusing) extracted_by_ocr - and deciding if we should just use this flag to avoid creating a new plain_text_ai or html_ai fields to Opinion.  

Decide between:

Option A: Add a new dedicated field
- ai_text = models.TextField(blank=True, null=True)

Option B: Reuse an existing field and add a flag
- Use plain_text and html but updated the extracted by ocr flag to indicate extracted by AI.

### What type of issue is this?

Research

quevon24:
I have these questions:

- Is it possible that some of the data we want to add to the system already exists? If so, it would be necessary to merge data, making the `plain_text` and `html` fields impractical in some cases. We generally use `plain_text` when extracting text from a PDF, and `html` is used when the opinion comes from an HTML file (from a scraper, dataset, etc) and we don't want to delete existing data (right?). If this is the case, it would be ideal to have a separate field to store the opinion (`ai_text` or something similar).

- Will the information extracted with LLM be plain text or formatted in HTML? If the text is first extracted using LLM and then formatted using LLM, then at least two fields will be necessary, something like `raw_ai_text` and `html_ai_text`

- Regarding whether or not to add a flag to identify if the opinion was extracted with AI, it largely depends on how the previous fields are handled. If any of the previously proposed fields are not empty, it means that it contains content extracted with AI. However, I also don't think it's a bad idea to add a new boolean field like extracted_by_ai, as it's a bit simpler for filtering purposes.

- Regarding any additional fields for the Opinion model, we would first need some examples to see if it is necessary to add anything else, ideally to apply all the changes in the same migration (there shouldn't be much of a problem since they will be new fields with some default value)

grossir:
I think looking at a more normalized shape of the Opinion data may help deciding which shape we should use for this feature. However, the fully normalized shape may be **very slow** to implement, since it's a big change and we would need to wait for replication users. The migration strategy would be the hardest part probably

The basic concept is making a difference between an Opinion and the "document" that contains it. 
The data we are going to get represents the "official" and maybe "final" representation of the opinion document, but the Opinion itself in its attributes should not change.
This is the same problem we saw when implementing opinion versions, which are different versions of the documents, with more or less text. So changing the schema would also make the versioning problem very simple to solve and to understand

The model below would help us add any variation of content from any type of source 
[...]

Issue #6395, opened 2025-10-13
Duplicate indexes


grossir:
Following some comments on Slack about this Haki Benita's [article](https://hakibenita.com/django-foreign-keys#identifying-duplicate-indexes)  I noticed we have some duplicated indexes that we should clean up for space and speed savings

Summary

<google-sheets-html-origin>

table | memory saved | redundant index columns | other index columns | redundant index name | other index name
-- | -- | -- | -- | -- | --
search_opinion | 439MB | (cluster_id) | (cluster_id, ordering_key) | search_opinion_a97b1c12 | "unique_opinion_ordering_key"
search_citation | 388MB | (cluster_id) | (cluster_id, volume, reporter, page) | search_citation_a97b1c12 | "search_citation_cluster_id_7a668830aad411f5_uniq"
search_citation | 168MB | (volume, reporter) | (volume, reporter, page) | search_citation_volume_251bc1d270a8abee_idx | "search_citation_volume_ae340b5b02e8912_idx"
search_opinionscited | 1137MB | (citing_opinion_id) | (citing_opinion_id, cited_opinion_id) | search_opinionscited_5b8b69a0 | "search_opinionscited_citing_opinion_id_7165e96b2aed974f_uniq"
citations_unmatchedcitation | 281MB | (citing_opinion_id) | (citing_opinion_id, volume, reporter, page) | citations_unmatchedcitation_citing_opinion_id_bb038174 | "citations_unmatchedcitat_citing_opinion_id_volume_ca9f46d3_uniq"
favorites_prayer | 480kB | (date_created) | (date_created, user_id, status) | favorites_prayer_date_created_07aee6d0 | "favorites_prayer_date_created_user_id_status_880d7280_idx"
favorites_prayer | 224kB | (user_id) | (user_id, recap_document_id) | favorites_prayer_user_id_0626a4dc | "unique_prayer_for_user_document"
favorites_prayer | 536kB | (recap_document_id) | (recap_document_id, status)  (recap_document_id, user_id) | favorites_prayer_recap_document_id_2d28d777 | "favorites_prayer_recap_document_id_status_82e2dbbb_idx"  "favorites_prayer_recap_document_id_user_id_c5d30108_idx"
favorites_dockettag | 480kB | (docket_id) | (docket_id, tag_id) | favorites_dockettag_docket_id_b227f907 | "favorites_dockettag_docket_id_tag_id_80109530_uniq"

</google-sheets-html-origin>

-----------

### `search` app

[...]

PR #6619, opened 2025-12-02
feat(search): add OpinionContent model

quevon24:
## Summary

This PR implements the new OpinionContent model to store different versions of Opinions

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/tests/cases.py
+++ b/cl/tests/cases.py
@@ -11,7 +11,7 @@ from django.test import SimpleTestCase
 from django.urls import reverse
 from django.utils.dateformat import format
 from django.utils.html import strip_tags
-from django.utils.timezone import now
+from django.utils.timezone import localtime, now
 from django_elasticsearch_dsl.registries import registry
 from lxml import etree, html
 from lxml.html import HtmlElement
@@ -720,13 +720,16 @@ class SearchAlertsAssertions:
     def _assert_date_updated(self, date_to_compare, html_content, txt_content):
         """Confirm that date_updated is properly set in the alert email."""
 
+        # Use localtime() to normalize DST, matching the email template which
+        # applies |localtime before |date formatting.
+        localized_date = localtime(date_to_compare)
         self.assertIn(
-            f"Date Updated: {format(date_to_compare, 'F jS, Y h:i a T')}",
+            f"Date Updated: {format(localized_date, 'F jS, Y h:i a T')}",
             html_content,
         )
 
         self.assertIn(
-            f"Date Updated: {format(date_to_compare, 'F jS, Y h:i a T')}",
+            f"Date Updated: {format(localized_date, 'F jS, Y h:i a T')}",
             txt_content,
         )
 


quevon24:
@grossir Do you think we should also create the opinion type field in OpinionContent model? If we are planning to have opinioncontent objects for each source, some sources have combined opinions and others have divided opinions

grossir:
At first glance, I don't think we should since the `type` is an attribute of the Opinion, not of its content...

But this is actually a problem; which is a different one than versioning and duplication
We may have a cluster that has:
- the combined opinion (probably from scraping)
- the separated opinions (probably from Harvard, and now from scanning)
     - lead
     - concurrence
     - dissent
     
The combined opinion is not actually a version... nor is it a duplicate... but in a sense, it's acting like both. It is sort of a duplicate of the whole group; and it is actually duplicating the OpinionsCited related objects, for example... 

So, how should we handle it? Maybe we should just tag it as a version (which now would be a boolean field), which would make it invisible. But we should still keep it. What do you think?

quevon24:
> At first glance, I don't think we should since the `type` is an attribute of the Opinion, not of its content...
> 
> But this is actually a problem; which is a different one than versioning and duplication We may have a cluster that has:
> 
> * the combined opinion (probably from scraping)
> * the separated opinions (probably from Harvard, and now from scanning)
>   
>   * lead
>   * concurrence
>   * dissent
> 
> The combined opinion is not actually a version... nor is it a duplicate... but in a sense, it's acting like both. It is sort of a duplicate of the whole group; and it is actually duplicating the OpinionsCited related objects, for example...
> 
> So, how should we handle it? Maybe we should just tag it as a version (which now would be a boolean field), which would make it invisible. But we should still keep it. What do you think?

I think they could be marked as a version.

But this needs to be resolved: the Opinion object will have the content of one of the OpinionContent objects. I think it would be necessary to implement something to choose the best text and mark the others as versions. You can see the priority here: https://github.com/freelawproject/courtlistener/blob/main/cl/search/models.py#L3116

grossir:
>  I think it would be necessary to implement something to choose the best text and mark the others as versions

Yes, sure. We usually choose the best formatted content right? So, scraped content gets the last place. I think Bill is implementing the LLM extraction that uses the Harvard XML schema, so that will be of good quality too.

So, the best opinion content would get pulled to be processed and produce `html_with_citations`. Which I think should live in the `Opinion` object, since it's the processed content

quevon24:
@grossir Do you think we might have to merge opinions from different sources (combined sources)? For example, Scanning + Harvard, Scanning + Columbia, etc.

grossir:
Each of those would be a OpinionContent row

If we have an Opinion with Opinion.xml_harvard, Opinion.html_columbia and Opinion.html populated, that single Opinion would produce 3 OpinionContent rows; each row would have its individual "source" identification. So the OpinionCluster.source would remain as it is, showing that we have information from different sources for the cluster; but the OpinionContent itself would let us know where each individual content came from

quevon24:
> Each of those would be a OpinionContent row
> 
> If we have an Opinion with Opinion.xml_harvard, Opinion.html_columbia and Opinion.html populated, that single Opinion would produce 3 OpinionContent rows; each row would have its individual "source" identification. So the OpinionCluster.source would remain as it is, showing that we have information from different sources for the cluster; but the OpinionContent itself would let us know where each individual content came from. 

@grossir Yes, you're absolutely right about that, we do have the source field in OpinionContent or we can simply check which content field isn't empty to know its source. And what do you think about having a constraint to prevent OpinionContent objects where more than one content field is filled (perhaps directly in the model's clean method)?

But in this case specifically for OpinionCluster, I think it would still be necessary to add the combined sources for "Scanning", mainly to facilitate queries to find a cluster with a specific source and avoid having to do something like: 

```
OpinionCluster.objects.filter(sub_opinions__opinioncontent__source=7) # two joins
```

by simply just doing this:

```
OpinionCluster.objects.filter(source="S") # S for scan
```

or this

```
OpinionCluster.objects.filter(source="US") # Harvard + Scan
OpinionCluster.objects.filter(source="CUS") # Court + Harvard + Scan
```

However, it's important to note that when doing this, we need to be careful when updating this opinioncluster source value in case any scanned content needs to be linked to an existing cluster.

Do you think it's necessary to avoid disrupting the current operation of OpinionCluster?

grossir:
About adding a new "Scanning" source to OpinionCluster.source and Docket.source; I think we definetly should

About this
> what do you think about having a constraint to prevent OpinionContent objects where more than one content field is filled (perhaps directly in the model's clean method)?

I am not sure I get it; the OpinionContent model should have a single field to be filled with extracted content; so what you describe should not happen? Or what do you mean?

quevon24:
> About adding a new "Scanning" source to OpinionCluster.source and Docket.source; I think we definetly should
> 
> About this
> 
> > what do you think about having a constraint to prevent OpinionContent objects where more than one content field is filled (perhaps directly in the model's clean method)?
> 
> I am not sure I get it; the OpinionContent model should have a single field to be filled with extracted content; so what you describe should not happen? Or what do you mean?

Sorry, I made a mistake. I thought I had already moved the fields html_lawbox, html_columbia, html_anon_2020, etc., to the new model, so I thought I had to force only one of them to be filled. But I just remembered that only the `content` field should be used.  Besides, removing those fields is something more for the future.

quevon24:
@grossir I just added the new sources for the clusters, you might want to take a look at them.

I also think it would be a good idea to do something similar to Docker Sources and put the cluster sources in a separate file. What do you think?

cl/search/models.py
grossir, line 1: I think we should probably move the Opinion Cluster sources definition / enumeration to another file... it will make search/models harder to read that it already is

cl/search/models.py
```diff
@@ -2578,6 +2788,12 @@ class OpinionCluster(AbstractDateTimeModel):
         storage=IncrementingAWSMediaStorage(),
         blank=True,
     )
+    filepath_pdf_scan = models.FileField(
+        help_text="The case PDF from the Scanning Project",
+        upload_to=make_upload_path,
+        storage=IncrementingAWSMediaStorage(),
+        blank=True,
+    )
```
grossir, line 2465: 2 questions about this
- how is Bill extracting the opinions? If a cluster is to be split into its parts then this field shouldn't be here
- during the transition period this will be reflected into  `OpinionContent.local_path` right?

There should be a comment about this

You are going to add something like a `Opinion.xml_scanning` to hold the content right? while we wait for transition to fully migrate to `OpinionContent`?  This is related to the first question, on how different opinion types inside a cluster will be handled

cl/search/docket_sources.py
```diff
@@ -143,6 +143,7 @@ class DocketSources:
         126
     )
     RECAP_AND_SCRAPER_AND_COLUMBIA_AND_IDB_AND_HARVARD_AND_DIRECT_INPUT_AND_ANON_2020 = 127
+    SCANNING_PROJECT = 128
```
grossir, line 146: we are missing all the combinations here?

mlissner:
I'm still hanging back waiting for y'all to ping me for a look. Let me know when it's a good time!

cl/search/docket_sources.py
quevon24: I haven't finished working on the combinations for the dockets yet.

Regarding the clusters, some sources don't have as many combinations, for example: `RECAP` and `BRAD_HEATH_ARCHIVE`.

quevon24:
> I'm still hanging back waiting for y'all to ping me for a look. Let me know when it's a good time!

Okay, thanks Mike. Once we have all the new models we can see if we can separate them into a new app, it would simply be a matter of copying and pasting some things and regenerating the new migrations.

mlissner:
Interesting. A new Django app for this model? That feels....complicated? I know the current app is quite big, but an app for a single model feels weird — or did you have other things in mind for the app too?

cl/search/models.py
quevon24: From what I understand, he's only going to separate the pages of each case so he doesn't have to process the entire scan with hundreds of pages. By simply dividing it like this, I think the cluster should have the entire document in `filepath_pdf_scan`, and if it's a split opinion, we would just need to create the Opinion object along with OpinionContent for each split opinion. Similar to what was done with Harvard. 

During the transition period, I believe the solution would be to copy the `local_path` value from Opinion to OpinionContent when creating the objects for old opinions (those that do not come from scans)

The other option would be to omit the `filepath_pdf_scan` field and use the `local_path` field from OpinionContent. If there are multiple types of opinions, the `local_path` value would be repeated.

Regarding the xml content from scans, it can be saved directly in the new `content` field and used directly. With Harvard, we stored the xml, and that's what's used to generate the `html_with_citations`. You can take a look at this cluster and see that the content uses the Harvard structure: https://www.courtlistener.com/opinion/4423362/go/

quevon24:
> Interesting. A new Django app for this model? That feels....complicated? I know the current app is quite big, but an app for a single model feels weird — or did you have other things in mind for the app too?

I'm thinking specifically about things related to AI. The idea was to store the prompts along with other things needed to execute them, so it would be a good idea to have a separate app for that, where we could have models, helpers, utilities, and other things specific to that purpose. I mention this in the main issue, here: https://github.com/freelawproject/courtlistener/issues/6539#issuecomment-3607868871 (I have already proposed some models for storing prompts and making them usable with any model (Generic foreign keys))

mlissner:
Oh, right, that makes a lot more sense!

cl/search/models.py
quevon24: @grossir I mentioned to Bill that we could use the `filepath_pdf_scan` field to store the entire document, and we can use the `local_path` of OpinionContent to store the raw XML response from the LLM as a file. He mentioned that the XML can return other fields such as `headmatter`, and this is stored in the OpinionCluster. By storing the XML in `local_path`, we can later process the file to extract the values ​​of those fields, which can be stored in cluster and in content only the opinion xml in `<opinion>`.  What do you think?

cl/search/models.py
```diff
@@ -86,395 +87,395 @@ def get_status_value_reverse(cls, name):
...
-
-        unique_parts = frozenset(source1 + source2)
-        if cls.parts_to_source_mapper.get(unique_parts):
-            return cls.parts_to_source_mapper.get(unique_parts)
-
-        # Unexpected case
-        if len(source1) > len(source2):
-            return source1
-        return source2
+# class ClusterSources:
```
grossir: delete commented code

quevon24:
Regarding what we talked about in the sprint planning meeting, this is what I took notes:

- `OpinionContent` will not be used at the moment to show information to the user, so it is necessary to make the `opinion` field optional in the model
- The `Opinion` model must maintain important fields that are not content fields (xml_harvard, html_anon_2020, html_columbia, etc.) except for `html_with_citations`. Why this? When you have an `OpinionContent` object that is the main version, then its information would be copied to `Opinion` object (and this is what will be shown to the user) (This can help not break elasticsearch by having to go deeper to index information from `OpinionContent` by keeping the indexed fields in Opinion and OpinionContent)
- Later when we have all the information in OpinionContent we can proceed to create its other objects (`Docket`, `OpinionCluster`, `Opinion` and `Citation`)
- And finally, make the `opinion` field required in `OpinionContent`.

What do you think of this? @grossir @flooie

grossir:
@quevon24 

> OpinionContent will not be used at the moment to show information to the user, so it is necessary to make the opinion field optional in the model

I think doing this would just make us unable to process anything from scanning or to evaluate / analyze our results. Because we would just have the OpinionContent, but no associated metadata (opinion -> cluster -> docket); which I think is not our goal.

If we don't want to show anything yet we could just not index it into ES. 

But we could also index it in ES and show it. It would be a little weird because it would have no raw text field, given that we wouldn't populate any of the text fields except `html_with_citations`. We would need a small change to have that field as a last option for the `OpinionDocument.text`. The same `Opinion.html_with_citations` will allow us to display the result in the frontend. We can ingest the scanning results with minimal changes on our regular models; just adding the sources to Docket and OpinionCluster. We would be in a "duplication" stage, where we would fit the data as we could into the `Opinion` model, while having mostly the same in `OpinionContent`

quevon24:
I added a new field named `filepath_xml_scan` in OpinionCluster so we can store the raw xml response of the case as a file containing all the metadata and the opinion(s). e.g., We will get something like this:

```
<document>
  <casebody firstpage="100" lastpage="105">
    <headmatter>
      <court>Sample Court of Example Jurisdiction.</court>
      <docketnumber>No. 12345-EXAMPLE</docketnumber>
      <citation case_name="Sample Corporation v. Test Respondent" volume="999" reporter="F.3d" page="100" court="Example Ct." year="2023">999 F.3d 100</citation>
      <parties>
        <party role="appellant">SAMPLE CORPORATION; and INSURANCE COMPANY INC., Appellants,</party>
        <separator>v.</separator>
        <party role="respondent">TEST RESPONDENT, Respondent.</party>
      </parties>
      <decisiondate>Filed January 1, 2023</decisiondate>
    </headmatter>
    <pagenumber pgmap="100">*100</pagenumber>
    <opinion type="majority">
      <p id="p100-1" pgmap="100">
        Sample Corporation and Insurance Company Inc., (appellants) appeal from a lower court decision regarding a standard legal matter. Example Judicial District, Sample County; Judge Jane Smith, Presiding.
      </p>
    </opinion>
  </casebody>
</document>
```

Some of those values ​​will be stored in the OpinionCluster fields. 

For the content I also created a new field named `xml_scanning` as @grossir suggested for the Opinion model. This field will be used while we transition to the OpinionContent model. This field will store everything that is in the <opinion> tag (this tag included).

cl/search/models.py
```diff
@@ -2790,6 +2417,12 @@ class OpinionCluster(AbstractDateTimeModel):
         storage=IncrementingAWSMediaStorage(),
         blank=True,
     )
+    filepath_xml_scan = models.FileField(
+        help_text="The XML obtained from LLM containing all available metadata and opinion(s).",
+        upload_to=make_upload_path,
+        storage=IncrementingAWSMediaStorage(),
+        blank=True,
+    )
```
grossir, line 2459: Looking at this I finally understand what you (or Bill?) meant yesterday with creating something like a `OpinionClusterContent` model. 

We currently have these 2 "content" fields in the cluster, that are exclusive of Harvard imports
- `filepath_json_harvard`: not used anywhere (only ingested); but I guess it's the origin of the data now living in the tables
- `filepath_pdf_harvard`: used as the PDF displayed in the frontend

I think we will be having a very similar scenario here
There will be at least 2 files related for each opinion
- redacted PDF of the opinion
- LLM output in XML

For this case, I have the following questions;
- aren't we missing the PDF filepath; or will that one be in the Opinion? I think it should live in the opinion (and maybe the `filepath_pdf_harvard` should be in the opinion too...)
- do we need to save the LLM response filepath in the cluster? Will it be used for anything?
        - if it only has audit / data origin tracking purposes, we could create a separate table and store it there. The `filepath_json_harvard` should be migrated to that table, eventually

-------------

From a data modelling standpoint, we should try to avoid these kind of fields that will be somewhat sparse. These "filepath" fields don't hurt as much as the big text fields in the opinion model, since they are smaller. 

They are still confusing (see some users confused [here](https://flp-talk.slack.com/archives/C07PC93V676/p1760456854860099?thread_ts=1760345380.155319&cid=C07PC93V676)), since the content is expected to be in the opinion level; not on the cluster level. In an ideal scenario, we should have these in another table

As with the Harvard fields, these are corpus specific fields, that shouldn't be in the general cluster, but should be reachable via a join. One would know they exist thanks to the OpinionCluster.source mentioning that corpus

cl/search/migrations/0050_opinioncontent_remove_opinioncluster_update_update_and_more.sql
```diff
@@ -0,0 +1,65 @@
...
+--
+-- (no-op)
+--
+-- Add field opinion to opinioncontent
+--
+ALTER TABLE "search_opinioncontent" ADD COLUMN "opinion_id" integer NOT NULL CONSTRAINT "search_opinioncontent_opinion_id_c2bb5a87_fk_search_opinion_id" REFERENCES "search_opinion"("id") DEFERRABLE INITIALLY DEFERRED; SET CONSTRAINTS "search_opinioncontent_opinion_id_c2bb5a87_fk_search_opinion_id" IMMEDIATE;
+CREATE INDEX "search_opinioncontent_date_created_07b21785" ON "search_opinioncontent" ("date_created");
+CREATE INDEX "search_opinioncontent_date_modified_04c81b86" ON "search_opinioncontent" ("date_modified");
+CREATE INDEX "search_opinioncontent_sha1_b21c7eaa" ON "search_opinioncontent" ("sha1");
+CREATE INDEX "search_opinioncontent_sha1_b21c7eaa_like" ON "search_opinioncontent" ("sha1" varchar_pattern_ops);
```
grossir: i have never used this kind of pattern; we should drop this `varchar_pattern_ops` / like index for sha1 https://github.com/freelawproject/courtlistener/issues/6395

cl/search/models.py
grossir: the opinion specific XML content will go into `Opinion.xml_scan` or something like that, right? 

The XML part that matches the headmatter should be put into the `OpinionCluster.headmatter` at creation time, right? So why would we need to store it in local_path?

cl/search/migrations/0050_opinioncontent_remove_opinioncluster_update_update_and_more.sql
quevon24: I believe this is related to Django version 5, in previous versions, I don't recall seeing it generate an index with `varchar_pattern_ops`. Do you think we need the index for this field?

cl/search/models.py
quevon24: Specifically for Harvard opinions, a single PDF/JSON file sometimes contains multiple opinions. If we move those fields to the `OpinionContent` level, the path would simply be repeated for each opinion row. In this case, I added `filepath_xml_scan` and `filepath_pdf_scan` to follow that same pattern.

Regarding moving the Harvard PDF/JSON fields out of the cluster, that sounds feasible (also the new pdf and xml scan fields). They could live in `OpinionContent` since that model serves as the "raw" content layer. This keeps the `Opinion` model focused solely on the final, processed version for the end user, while the content model maintains the connection to the source files.

There is also a logic to why the Harvard JSON is currently in the `OpinionCluster`: it contains metadata like court, citation, and date filed that applies to the entire case. While the text inside the `<opinion>` tags is extracted into the `Opinion` model, the file as a whole represents the metadata for the cluster.

Finally, I believe it's important to preserve the LLM response XML in case we need to re-import data or handle errors. Having that XML allows us to reread the existing output instead of running the LLM against the PDF again. It’s also useful for auditing, if a prompt needs improvement, we can review the exact XML response to understand where the logic failed.

What do you think @flooie @grossir?

cl/search/migrations/0050_opinioncontent_remove_opinioncluster_update_update_and_more.sql
grossir: we need an index on sha1 for duplicate lookups; but we don't really need an index for `sha1 LIKE 'aaaa%'` we basically never use those queries on the `Opinion.sha1` field

So, I think we should not create this index

`CREATE INDEX "search_opinioncontent_sha1_b21c7eaa_like" ON "search_opinioncontent" ("sha1" varchar_pattern_ops);` 

I suggested the same (well, deleting the existing sha1 `varchar_pattern_ops` index) on the  issue I linked above

quevon24: @grossir There are also varchar_pattern_ops indexes for `download_url` and `local_path`. Do you think we need them too?

As a note, to prevent indexes from being created with `varchar_pattern_ops`, they must be defined in the Meta class of the model instead of defining it with the db_index=True option in the field.

cl/search/models.py
quevon24: Yes, it will go in the xml_scanning field in Opinion model.

And yes, the headmatter will be stored in the cluster at creation time. The idea behind storing the original file is to have all the metadata, not just the opinion. I think I mentioned that in another comment.

cl/search/models.py
flooie: I agree with @quevon24 here with regard to with the filepath would be.  it cant live on the Opinion because the file is for the cluster.  

But why would we move the harvard pdf and json to the opinioncontent?  The files represent the clusters.  The xml will also container the headmatter section also on the cluster.

quevon24:
@flooie @grossir I've reviewed the PR again. This basic OpinionContent model gives us what we need to start ingesting LLM-extracted content from scans. When we build the full ingestion pipeline (Docket + OpinionCluster + Opinion + Citation + OpinionContent), we'll need to disable ES indexing during creation so we can audit the data before it's visible to users or at minimum, ensure it's only accessible if you know the PK directly. 

On the broader roadmap: the new ai Django app with the LLM-related models (config, prompts, etc.) is already in progress in a separate PR (https://github.com/freelawproject/courtlistener/pull/6654). And as documented in the model's docstring, we'll add the remaining fields (sha1, download_url, local_path, etc.) when we're ready to copy existing data from Opinion to OpinionContent.

cl/search/models.py
```diff
@@ -3445,6 +3280,72 @@ class Meta:
...
+    content = models.TextField(
+        help_text="The text content of the opinion",
+        blank=True,
+    )
+    source = models.SmallIntegerField(
+        help_text="Source of the opinions content",
+        choices=SOURCES,
+    )
+    extraction_type = models.SmallIntegerField(
+        help_text="Source of the opinions content",
```
grossir: this help_text is repeated from `source`

cl/search/models.py
```diff
@@ -3445,6 +3280,72 @@ class Meta:
...
+    FLP_XML = 7
+    SOURCES = (
+        (SCRAPERS, "Got from juriscraper"),
+        (XML_HARVARD, "Got from CAP"),
+        (HTML_ANON_2020, "Got from 2020 Anon archive"),
+        (COLUMBIA_ARCHIVE, "Got from Columbia import"),
+        (LAWBOX, "Got from Lawbox"),
+        (RECAP, "Free opinions on RECAP"),
+        (FLP_XML, "FLP XML"),
+    )
```
grossir, line 3342: I think this part needs some editions
- dropping the XML_ or HTML_ prefixes. We don't focus on the format, but on the source
- FLP_XML should be something like FLP_SCANNING: that way we focus on the source entity and project
- similarly, HARVARD should probably be HARVARD_CAP
- docstrings can be improved, the harvard one could be "Got from the Harvard Case Law Access Project"
- "FLP XML" as a description can be improved too

cl/search/models.py
```diff
@@ -3445,6 +3280,72 @@ class Meta:
         proxy = True
 
 
+class OpinionContent(AbstractDateTimeModel):
```
grossir, line 3315: I think this class is missing multiple content related fields
- filepath
- sha1
- page_count
- is_main_version

etc

I wrote longer about this on the related issue #6539

cl/search/cluster_sources.py
```diff
@@ -0,0 +1,389 @@
...
+        (COLUMBIA_M_ANON_2020, "columbia merged with 2020 anonymous database"),
+        (
+            HARVARD_CASELAW,
+            "Harvard, Library Innovation Lab Case Law Access Project",
+        ),
+        (COURT_M_HARVARD, "court website merged with Harvard"),
+        (DIRECT_COURT_INPUT, "direct court input"),
+        (ANON_2020, "2020 anonymous database"),
+        (ANON_2020_M_HARVARD, "2020 anonymous database merged with Harvard"),
+        (COURT_M_HARVARD, "court website merged with Harvard"),
```
grossir: see line 14, and these 2. We have at least 3 `COURT_M_HARVARD`

Check the whole file again, let's catch any other duplicates

cl/search/models.py
```diff
@@ -3445,6 +3280,72 @@ class Meta:
...
+    extraction_type = models.SmallIntegerField(
+        help_text="Source of the opinions content",
+        choices=EXTRACTION_METHOD,
+    )
+
+    def __str__(self) -> str:
+        try:
+            return f"{getattr(self, 'pk', None)} - {self.opinion.cluster.case_name}"
+        except AttributeError:
+            return f"Orphan opinion with ID: {self.pk}"
```
grossir: I think it will never be the case that the OpinionContent doesn't have a related Opinion. The key is required, has a on delete cascade condition. So we can remove these try-except clause

cl/search/models.py
```diff
@@ -2599,6 +2419,18 @@ class OpinionCluster(AbstractDateTimeModel):
...
+    filepath_xml_scan = models.FileField(
+        help_text="The XML obtained from LLM containing all available metadata and opinion(s).",
+        upload_to=make_upload_path,
+        storage=IncrementingAWSMediaStorage(),
+        blank=True,
+    )
+    filepath_pdf_scan = models.FileField(
+        help_text="The case PDF from the Scanning Project",
+        upload_to=make_upload_path,
+        storage=IncrementingAWSMediaStorage(),
```
grossir: for now let's put the redacted opinion PDF in private storage. We publish the files (will probably require moving data to a different bucket)  once we are 100% sure of redaction quality

grossir (commented):
The only "big" comment I have is about using private storage for the redacted PDFs
The rest are code quality comments.

cl/search/models.py
quevon24: These fields (sha1, page_count, download_url, local_path, is_main_version) were actually present in an earlier commit (https://github.com/freelawproject/courtlistener/commit/12a70985d), based on your proposal in https://github.com/freelawproject/courtlistener/issues/6539#issuecomment-3559061836. I ended up removing them after a meeting with @flooie where we discussed that we probably don't need them yet at this stage.

The thing is, these fields won't really affect the scanning project pipeline until we have the final redacted documents with no extra pages. Only then would we actually generate the sha1, store the local_path, compute the page_count, etc. So having them now or adding them later shouldn't block any current work.

That said, I'm fine either way. I can add them back now if you'd prefer to have the full model shape in place from the start, or we can add them when we're ready to populate them. What do you think?

cl/search/models.py
quevon24: Yes, I changed it to use S3PrivateUUIDStorage so the redacted PDFs stay private until we're confident about the redaction quality

When we're ready to publish them, the plan would be:

1. Copy the files from the private bucket to the public one, keeping the same relative paths
2. Swap the storage class on the field to IncrementingAWSMediaStorage (this is a no-op migration since Django doesn't track storage in the DB schema)
3. Optionally rename the files in the public bucket from UUID names to something more meaningful and update the DB paths accordingly (using a management command or django shell script)

Since Django only stores the relative path in the filepath_pdf_scan field, switching storage classes won't break anything as long as the files exist at the same relative path in the new bucket.

cl/search/models.py
grossir: @quevon24 

I think we should add them:

- one less migration to think about in the future. 
- And less "fixing" scripts in the future to populate the columns
- non-controversial fields: it's nothing new, just copying Opinion fields
- they will make things easier in some senses: 
       - local_path links directly to the original file, instead of having to do joins
       - sha1 may help double check ingestion for duplicates
 
These will be specially useful for content that will be merged into already existing content. Right now, if we merge into an existing opinion, we cannot save the local_path or sha1, because that opinion will have those fields one from the scraped source.


Apart from that, I think its important that those conversations are recorded in the issues / PRs so we can have easier understanding

grossir (commented):
I think this is mostly ready. My only [comment](https://github.com/freelawproject/courtlistener/pull/6619#discussion_r2795592582) is about adding the whole OpinionContent fields, not only a part of them

Othewise, I think this is mostly ready on the important aspects which are the models.  We can probably improve secondary aspects later.

@quevon24

cl/search/cluster_sources.py
```diff
@@ -0,0 +1,407 @@
...
+    COLUMBIA_M_LAWBOX_M_HARVARD = "ZLU"
+    COLUMBIA_M_DIRECT_COURT_INPUT_M_HARVARD = "ZDU"
+    COLUMBIA_M_LAWBOX_M_RESOURCE_M_HARVARD = "ZLRU"
+    COLUMBIA_M_LAWBOX_M_COURT_RESOURCE_M_HARVARD = "ZLCRU"
+    COLUMBIA_M_COURT_M_HARVARD = "ZCU"
+    COLUMBIA_M_MANUAL_INPUT_M_HARVARD = "ZMU"
+    COLUMBIA_M_PUBLIC_RESOURCE_M_HARVARD = "ZRU"
+    COLUMBIA_M_LAWBOX_M_COURT_M_HARVARD = "ZLCU"
+    RECAP = "G"
+    SCANNING_PROJECT = "S"
```
grossir, line 48: missing the combination of these 2. That combination may happen, since we will scan federal reporters

quevon24: added

quevon24:
Can you take a look at it, @mlissner?

cl/search/models.py
```diff
@@ -3323,6 +3155,9 @@ class Opinion(AbstractDateTimeModel):
     xml_harvard = models.TextField(
         help_text="XML of Harvard CaseLaw Access Project opinion", blank=True
     )
+    xml_scanning = models.TextField(
```
mlissner: Should this be `xml_scan`? `scanning` implies the verb, I think? Like "is the XML currently scanning?"

mlissner (commented):
This looks pretty good to me, generally speaking. Some thoughts for you: 

1. I sort of expected that the `OpinionContent` table would mean we didn't need the xml_scanning field. I guess we're adding it for expediency, since we're not ready to use `OpinionContent` yet? 

2. Please do a quick look through the code base for places where we might exclude the big text fields. I think we've optimized a number of queries to exclude those fields, and we should be careful not to accidentally include these new fields.

3. Have we checked to see if these new fields will show up in API responses? Do we want them to? 

4. What about bulk data? We probably need to add some of these fields to the bulk data export, right? 

5. Will any of these fields have values we wouldn't share with our replication customers? They're going to get these values from the outset, so we want to be careful about this. 

Finally, I haven't read the SQL files to make sure they're synced and I only skimmed the rest. I leave the more careful reviews to y'all!

mlissner:
And, sorry, one other thought: What about search? Do we need to add these fields to Elastic so that their content is searchable?

grossir:
Some thoughts on @mlissner questions @quevon24 

> 5.  Will any of these fields have values we wouldn't share with our replication customers? They're going to get these values from the outset, so we want to be careful about this.

I think the only thing we don't want to share are the early redacted scans. But those will be in a private bucket, so even if the path is exposed in the DB, replication users won't be able to access them

> And, sorry, one other thought: What about search? Do we need to add these fields to Elastic so that their content is searchable?

This should be straight forward, its just adding a couple of lines [here](https://github.com/freelawproject/courtlistener/blob/a5f96af77a2407bbdf8b008fe8d071da8e9d8141/cl/search/documents.py#L2061-L2081). But we wanted to focus on the models themselves to get this done as quickly as possible. We also need to update citation extraction to use this new field, the frontent, bulk data and a bunch of places that would use the new content field


> 3. Have we checked to see if these new fields will show up in API responses? Do we want them to?

By default, the Opinion.xml_scanning would be exposed by the API
https://github.com/freelawproject/courtlistener/blob/a5f96af77a2407bbdf8b008fe8d071da8e9d8141/cl/search/api_serializers.py#L246-L248

cl/search/models.py
quevon24: Good catch on the naming! In this context, "Scanning" is part of the noun phrase "Scanning Project". I think the current name works, but I'm open to suggestions if you think it could be clearer.

quevon24:
> This looks pretty good to me, generally speaking. Some thoughts for you:
> 
> 1. I sort of expected that the `OpinionContent` table would mean we didn't need the xml_scanning field. I guess we're adding it for expediency, since we're not ready to use `OpinionContent` yet?
> 2. Please do a quick look through the code base for places where we might exclude the big text fields. I think we've optimized a number of queries to exclude those fields, and we should be careful not to accidentally include these new fields.
> 3. Have we checked to see if these new fields will show up in API responses? Do we want them to?
> 4. What about bulk data? We probably need to add some of these fields to the bulk data export, right?
> 5. Will any of these fields have values we wouldn't share with our replication customers? They're going to get these values from the outset, so we want to be careful about this.
> 
> Finally, I haven't read the SQL files to make sure they're synced and I only skimmed the rest. I leave the more careful reviews to y'all!

1. That's right, we're not ready yet, this is the first step to have the model and be able to run tests.

2. Yes, I'll review the Opinion queries to see if it's necessary to exclude the xml_scanning field, although for now it will be empty.

3. It will be displayed as Gianfranco mentioned earlier. For now, it will be empty, but we can exclude it and display that field again when we're ready.

4. Yes, you're right, we need to add the xml_scanning field to the bulk data to avoid errors when importing the search_opinion table. I'll do that in this PR. And for now, it's not necessary to add the `search_opinioncontent` table to the script.

5. As gianfranco mentioned, we don't want them to see the first redacted scans, but since they will be stored in a private bucket, even if they have the path they won't be able to see the files.

cl/search/models.py
mlissner: `xml_scan` seems much better to me

mlissner:
All sounds good. Thanks!

quevon24:
I made these small adjustments:

- Rename `xml_scanning` to `xml_scan`
- Add `xml_scan` to bulk data script
- Update migrations

PR #6617, opened 2025-12-02
feat(stats): Adds switch to disable statistics increment globally

ERosendo:
## Summary

This PR introduces a global switch to control the execution of the `tally_stat` function. Previously, the feature flag check was performed at each view. Now, the check is centralized inside `tally_stat` itself, simplifying the code and allowing for a single global toggle.

## Deployment

**This PR should:**

- [ ] `skip-deploy` (skips everything below)
    
    
    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

1. To deploy, we need to make sure that the global switch named `increment-stats` has been created and is disabled

PR #6624, opened 2025-12-03
Refactor(stats): Updates tally_stat function to use Redis

### LINKED ISSUE #6405: Reduce contention in `stats_stat` table updates
### Describe the issue.

We’ve noticed high database connection counts, as flagged in the #keep-it-up channel. Investigating the RDS performance dashboard revealed that the following query is contributing to contention

```sql
SELECT set_config('pghistory.context_id', '2a0b3c76-5ab4-4f80-87c7-f790a2f4fda9', true),
       set_config('pghistory.context_metadata', '{"user": null, "url": "/"}', true);
UPDATE "stats_stat"
SET "name" = 'search.homepage_loaded',
    "date_logged" = '2025-10-15'::date,
    "count" = ("stats_stat"."count" + 1)
WHERE "stats_stat"."id" = 32041;

```

The issue happens because multiple connections are trying to update the same row in the stats_stat table at the same time. This table tracks global events like searches and homepage loads, so it's common for many connections to attempt to increment the same counter simultaneously. As a result, transactions get blocked while waiting for locks, leading to a buildup of commits waiting on those locks and impacting overall database performance.

## Proposed Solution:

To address this, we could introduce a ledger-style table where each event is stored as a new row, and keeping the `stats_stat` table as an aggregated view of the ledger. By doing so, each connection can insert a new row without conflicting with others, eliminating contention. Additionally, we can use a [BRIN](https://www.postgresql.org/docs/current/brin.html) index on the ledger table to speed up date-based aggregations.

## Tradeoffs / Considerations:

One tradeoff of this approach is that we would need a periodic task to update or rebuild the aggregated `stats_stat` table. Despite this, the approach should significantly reduce contention and make tracking high-traffic global events more scalable.

---

I’m opening this issue to discuss this approach further and to review other possible solutions, including the potential integration of a dedicated analytics tool, to handle high-traffic event tracking more efficiently.

### Related Issue

_No response_

### Screenshots

_No response_

Research

ERosendo:
## Fixes

This PR fixes #6405 

## Summary

This PR changes the implementation of the `tally_stat` function so that it now uses Redis for incrementing counters instead of writing directly to the database on each event. The main change is to switch from row updates (which previously caused locking and contention) to an in‑memory increment, deferring persistence.

Additionally, several stats that are no longer needed have been removed.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

This PR depends on https://github.com/freelawproject/courtlistener/pull/6617

### DIFF
--- a/cl/stats/utils.py
+++ b/cl/stats/utils.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+from datetime import datetime, timedelta
 
 import redis
 from django.db import OperationalError, connections
@@ -61,13 +62,27 @@ def tally_stat(name, inc=1, date_logged=None) -> int:
         return
 
     r = get_redis_interface("STATS")
+    current_dt = now()
     if date_logged is None:
-        date_logged = now().date()
+        date_logged = current_dt.date()
 
     key = f"{name}.{date_logged.isoformat()}"
+
+    # Compute expiration:
+    # Keys live for 10 full days after the date they represent. For example,
+    # a key for June 1 will expire at June 12 at 00:00:00.
+    midnight_today = datetime.combine(
+        date_logged, datetime.min.time(), tzinfo=now().tzinfo
+    )
+    expire_at_date = midnight_today + timedelta(days=11)
+
+    # Convert to seconds-from-now for Redis EXPIRE
+    ttl_seconds = int((expire_at_date - now()).total_seconds())
+
+    # Increment and apply expiration atomically
     pipe = r.pipeline()
     pipe.incrby(key, inc)
-    pipe.expire(key, 60 * 60 * 24 * 10)  # 10 days
+    pipe.expire(key, ttl_seconds)
     value, _ = pipe.execute()
 
     return value
--- a/cl/stats/tests.py
+++ b/cl/stats/tests.py
@@ -1,9 +1,10 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 
 import pytest
 import time_machine
 from django.core import mail
 from django.core.management import call_command
+from django.utils.timezone import now
 from waffle.testutils import override_switch
 
 from cl.lib.redis_utils import get_redis_interface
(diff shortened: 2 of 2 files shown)


cl/search/views.py
```diff
@@ -170,9 +169,6 @@ def show_results(request: HttpRequest) -> HttpResponse:
     # This is a GET request: Either a search or the homepage
     if len(request.GET) == 0:
         # No parameters --> Homepage.
-        if not is_bot(request):
```
mlissner, line 173: @elisa-a-v, this is the PR implementing the changes we were just talking about. I just wanted to highlight for you that when/if we put stats back in with prometheus, we'll probably want to use the `is_bot` function again.

cl/alerts/tasks.py
```diff
@@ -360,7 +359,7 @@ def send_alert_and_webhook(
     connection.send_messages(messages)
 
     # Work completed. Tally, log, and clean up
-    async_to_sync(tally_stat)("alerts.docket.alerts.sent", inc=len(messages))
+    tally_stat("alerts.docket.alerts.sent", inc=len(messages))
```
mlissner: All the alert-related names can just be `alerts.sent` now, right? In prometheus, we can distinguish between docket alerts and search alerts, but this is for the homepage, where we just have them combined.

ERosendo:
@mlissner  I’ve fixed failing tests. This PR is now ready for review.

cl/stats/utils.py
```diff
@@ -53,24 +50,27 @@ def get_milestone_range(start, end):
...
-        date_logged=date_logged,
-        defaults={"count": F("count") + inc},
-        create_defaults={"count": inc},
-    )
+        date_logged = now().date()
+
+    key = f"{name}.{date_logged.isoformat()}"
+    pipe = r.pipeline()
+    pipe.incrby(key, inc)
+    pipe.expire(key, 60 * 60 * 24 * 10)  # 10 days
```
mlissner: I think we need to do this a bit differently. Instead of calculating the number of seconds in ten days, we should find the time that's midnight ten days ago (or nine or 11 days, I'm not sure).

But the distinction is that `60 * 60 * 24 * 10` sets a different expiration date if you tally the stat in the morning vs. `(date.today() - (days=10))`, which would be midnight on that day. 

The end result should be that after ten days running this, we should have ten keys in redis right up until midnight, at which point one deletes, and another begins. 

I think we use midnight EST if we want to be fancy?

cl/alerts/tasks.py
ERosendo: > All the alert-related names can just be alerts.sent now, right? 

Yeah, that's right. This PR updates several other keys so that all alert-related events are aggregated under the `alerts.sent` key

cl/stats/utils.py
ERosendo: After reviewing `get_homepage_stats`, I noticed that it sums values where the key's date is greater than or equal to the cutoff date (10 days ago): 

https://github.com/freelawproject/courtlistener/blob/3c0243e77ebdb048bce5edf1e96ba9b8779a87e8/cl/search/utils.py#L75-L83

For example, if today is the 11th, the method will sum values from the 1st (10 days ago) through the 11th.

I refactored the `tally_stat` method in [0dc881c](https://github.com/freelawproject/courtlistener/pull/6624/commits/0dc881c2c9e6e7b7dff09a866b694663f4e8ad87)

With the new expiration logic, a key for the 1st will expire on the 12th. This ensures it is included in the full 10-day rolling window but removed once it falls outside the sum window.

PR #6647, opened 2025-12-08
6535 Implement docket entries merger for SCOTUS data

albertisfu:
## Fixes

Fixes: #6535

## Summary

This PR adds support for merging docket entries from SCOTUS dockets. As with regular dockets, each `docket_entry` created also results in a corresponding RECAPDocument instance.

We also merge attachments. Similar to appellate dockets in SCOTUS, if there is only a single attachment, it is considered the main `PACER_DOCUMENT`. If there are multiple attachments, then all related RECAPDocuments are treated as `ATTACHMENTs`.

Because `attachment_number` is not available in SCOTUS data, it is assigned sequentially during the merge process based on the order of the attachments.

SCOTUS documents do not have a `pacer_doc_id`. However, we assign a blank value to attachments to prevent missing-key errors during the merge process. Instead of relying on `pacer_doc_id` as the unique identifier, we use `document_number` together with `attachment_number` as the identifiers.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/recap/mergers.py
+++ b/cl/recap/mergers.py
@@ -949,11 +949,8 @@ async def add_docket_entries(
         # or throw an error.
         params = {"docket_entry": de}
         short_description = docket_entry.get("short_description")
-        if (
-            not docket_entry["document_number"]
-            and short_description
-            or is_scotus
-            and short_description
+        if short_description and (
+            not docket_entry["document_number"] or is_scotus
         ):
             params["description"] = docket_entry["short_description"]
 
@@ -992,9 +989,8 @@ async def add_docket_entries(
             if is_scotus:
                 # SCOTUS documents don't have a pacer_doc_id, so use the
                 # document_number instead to match the document.
-                doc_number = docket_entry.get("document_number")
                 get_params["document_number"] = (
-                    doc_number if doc_number else ""
+                    docket_entry.get("document_number") or ""
                 )
             if de_created is False:
                 # Try to match the RD regardless of the document_type.
@@ -2301,8 +2297,6 @@ def merge_scotus_docket(report_data: dict[str, Any]) -> Docket | None:
 
     # Docket entries merger:
     enrich_scotus_attachments(report_data["docket_entries"])
-    items_returned, rds_created, content_updated = async_to_sync(
-        add_docket_entries
-    )(d, report_data["docket_entries"])
-   
+    async_to_sync(add_docket_entries)(d, report_data["docket_entries"])
+
     return d


albertisfu:
Applied an additional change here. While working on https://github.com/freelawproject/courtlistener/issues/6536, I noticed that unnumbered RECAPDocuments were being removed and re-created each time the docket was re-ingested. The issue was that `RECAPDocument.document_number` is stored as an empty string, but the lookup was being performed using a `None` value.

I also found that attachments were not being matched correctly during re-ingestion. Attachments do not have a `pacer_doc_id` or `pacer_case_id` to use when looking up the corresponding RECAPDocuments for a case. To address this, instead of using `pacer_doc_id`, we now rely on `short_description`, which is the only metadata that reliably differentiates attachments belonging to the same entry.
And instead of using `pacer_case_id`, we now use `docket_number_core`, which is unique to each SCOTUS case and has the advantage of being also an indexed field, making the lookup fast. This prevents incorrect matches with documents that potentially share the same document number across different cases.

cl/recap/mergers.py
```diff
@@ -2133,7 +2167,30 @@ def process_case_query_report(
...
+        for idx, attachment in enumerate(attachments, start=1):
+            if idx == 1:
+                main_doc_short_desc = attachment["description"]
+            attachment["attachment_number"] = idx
+            attachment["pacer_doc_id"] = ""
+
+        if not attachments:
+            entry["attachments"] = None
+
+        entry["short_description"] = main_doc_short_desc
```
mlissner, line 2188: Does this belong here or in Juriscraper? Feels a bit like the kind of cleanup we should be doing there?

cl/recap/mergers.py
```diff
@@ -1003,6 +1012,8 @@ async def add_docket_entries(
             if rd is None:
                 try:
                     params["pacer_doc_id"] = docket_entry["pacer_doc_id"]
+                    if is_scotus:
+                        del params["document_number"]
```
mlissner: Weird enough to get a comment?

cl/recap/mergers.py
albertisfu: Yeah, I thought about that as well. However, my reasoning for putting it here is that these tweaks are “artificial” and specifically tailored for Courtlistener. For example, does not make sense to display a `pacer_doc_id` because `SCOTUS` content is not in PACER.

Similarly, `short_description` does not exist on SCOTUS entries, but here we are taking the first document’s description and treating it as the entry `short_description` as a way to identify the main document during ingestion.

The only field that might belong in Juriscraper is the `attachment_number`, which does not exist in the source data but is something we assign sequentially.

Overall, I'm thinking to not alter the SCOTUS parser with artificial fields so that Juriscraper users are not confused by data that does not actually exist in SCOTUS. What do you think?

mlissner: Sounds good, thanks. :)

cl/recap/mergers.py
albertisfu: Sure. I Refactored this and added a comment.

cl/recap/mergers.py
```diff
@@ -946,8 +948,12 @@ async def add_docket_entries(
...
         # or throw an error.
         params = {"docket_entry": de}
-        if not docket_entry["document_number"] and docket_entry.get(
-            "short_description"
+        short_description = docket_entry.get("short_description")
+        if (
+            not docket_entry["document_number"]
+            and short_description
+            or is_scotus
+            and short_description
```
ERosendo: I think we can simplify this if condition by factoring out `short_description`:

```python

if short_description and (not docket_entry["document_number"] or is_scotus):
```

cl/recap/mergers.py
```diff
@@ -984,6 +989,13 @@ async def add_docket_entries(
             get_params = deepcopy(params)
             if de_created is False and not appellate_court_id_exists:
                 get_params["pacer_doc_id"] = docket_entry["pacer_doc_id"]
+            if is_scotus:
+                # SCOTUS documents don't have a pacer_doc_id, so use the
+                # document_number instead to match the document.
+                doc_number = docket_entry.get("document_number")
+                get_params["document_number"] = (
+                    doc_number if doc_number else ""
+                )
```
ERosendo: I think we can drop the temporary variable:


```suggestion
                get_params["document_number"] = docket_entry.get("document_number") or ""
```

cl/recap/mergers.py
```diff
@@ -2143,96 +2201,111 @@ def merge_scotus_docket(report_data: dict[str, Any]) -> Docket | None:
...
-    qp_url = report_data.get("questions_presented")
-    scotus_data.questions_presented_url = (
-        qp_url if qp_url else scotus_data.questions_presented_url
-    )
-    scotus_data.save()
+    # Docket entries merger:
+    enrich_scotus_attachments(report_data["docket_entries"])
+    items_returned, rds_created, content_updated = async_to_sync(
+        add_docket_entries
+    )(d, report_data["docket_entries"])
```
ERosendo: It looks like `items_returned`, `rds_created`, and `content_updated` aren’t used after this call. Can we drop the assignment if they’re not needed?

ERosendo (approved):
Looks good to me. I’ve highlighted a few small comments to address before merging.

cl/recap/mergers.py
albertisfu: Yeah this is simpler. Change applied, thanks!

cl/recap/mergers.py
albertisfu: Change applied!

cl/recap/mergers.py
albertisfu: Yeah, I’m expeting to use these values in a follow up PR; for now, I’ve removed them and will reintroduce them if necessary.

albertisfu:
Thank you @ERosendo I've applied your suggested changes and this is ready for another review!

ERosendo (approved):
LGTM 🚀

PR #6654, opened 2025-12-10
feat(ai): add a new app with llm models to track llm runs, store prompts, etc

quevon24:
## Summary

This PR introduces a new application named `ai` to manage all logic related to Large Language Model (LLM) configuration, execution, and auditing.

This app centralizes the following capabilities:

- Model Storage: It hosts all LLM-related models (Prompt, LLMConfig, LLMPromptSet, LLMRun), separating configuration and audit data from core business logic.
- Utility Layer: It serves as the primary location for helpers and utilities needed to execute and audit tasks efficiently.
- Clean Architecture: This separation ensures the application remains modular, allowing for independent updates to LLM strategy or configuration without impacting core features.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/lib/model_helpers.py
+++ b/cl/lib/model_helpers.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import re
 from collections.abc import Callable
@@ -8,6 +9,9 @@ from uuid import uuid4
 from django.core.exceptions import ValidationError
 from django.utils.text import get_valid_filename, slugify
 from django.utils.timezone import now
+from juriscraper.state.texas.common import (
+    DOCKET_NUMBER_REGEXES as TEXAS_DN_REGEXES,
+)
 
 from cl.custom_filters.templatetags.text_filters import oxford_join
 from cl.lib.recap_utils import get_bucket_name
@@ -16,6 +20,8 @@ from cl.lib.string_utils import normalize_dashes, trunc
 dist_d_num_regex = r"(?:\d:)?(\d\d)-[a-zA-Z]{1,5}-(\d+)"
 appellate_bankr_d_num_regex = r"(\d\d)-(\d+)"
 scotus_d_a_num_regex = r"(\d{2})a(\d{1,5})"
+
+logger = logging.getLogger(__name__)
 
 
 def is_docket_number(value: str) -> bool:
@@ -61,10 +67,47 @@ def clean_docket_number(docket_number: str | None) -> str:
     if len(bankr_m) == 1:
         return bankr_m[0]
 
-    # Match SCOTUS docket numbers.
+    return ""
+
+
+def clean_scotus_docket_number(docket_number: str | None) -> str:
+    """Clean a SCOTUS docket number, prioritizing the NN-NNNN format over the
+    NNA format.
+
+    SCOTUS docket numbers can appear in formats like:
+        No. 01-8148         -> 01-8148
+        No. 01A576          -> 01a576
+        No. 01A576 (01-8099) -> 01-8099  (prioritize NN-NNNN)
+        No. 01-8148 (01A587) -> 01-8148  (prioritize NN-NNNN)
+
+    If multiple numbers of the same type are found (e.g., "01A576 01A578"),
+    a ValueError is raised.
+
+    :param docket_number: The docket number to clean.
+    :return: The cleaned docket number or an empty string.
+    """
+    if not docket_number:
+        return ""
+
+    docket_number = normalize_dashes(docket_number)
+    docket_number = docket_number.lower()
+
+    scotus_m = re.findall(r"(?<![^ ,(])\d\d-\d+", docket_number)
     scotus_a_m = re.findall(r"\b\d{2}a\d{1,5}\b", docket_number)
+
+    if len(scotus_m) == 1:
+        return scotus_m[0]
+    if len(scotus_m) > 1:
+        logger.error(
+            "Multiple NN-NNNN docket numbers found in: %s", docket_number
+        )
+        return ""
+
     if len(scotus_a_m) == 1:
         return scotus_a_m[0]
+    if len(scotus_a_m) > 1:
+        logger.error("Multiple NNA docket numbers found in: %s", docket_number)
+        return ""
 
     return ""
 
@@ -113,8 +156,86 @@ def make_docket_number_core(docket_number: str | None) -> str:
     return ""
 
 
+def is_texas_court(court_id: str) -> bool:
+    """Check if the given court_id belongs to a Texas state court.
+
+    :param court_id: The CourtListener court_id to check.
+    :return: True if the court is a Texas state court.
+    """
+    return (
+        court_id == "tex"
+        or court_id == "texcrimapp"
+        or court_id.startswith("txctapp")
+        or court_id.startswith("texdistct")
+        or court_id.startswith("texcrimdistct")
+        or court_id.startswith("texctyct")
+    )
+
+
+def clean_texas_docket_number(docket_number: str | None) -> str:
+    """Clean a Texas docket number by extracting the valid docket number
+    from potentially dirty input using Juriscraper's regex patterns.
+
+    Converts inputs like:
+
+        Case Number: 04-97-00972-CV -> 04-97-00972-CV
+        04-97-00972-CV -> 04-97-00972-CV
+
+    :param docket_number: The docket number to clean.
+    :return: The cleaned docket number or empty string if no valid match.
+    """
+    if not docket_number:
+        return ""
+
+    docket_number = normalize_dashes(docket_number)
+
+    # Try fullmatch on the entire string first (clean input)
+    for regex in TEXAS_DN_REGEXES:
+        if regex.fullmatch(docket_number):
+            return docket_number
+
+    # Try fullmatch on each whitespace-separated token (dirty input
+    # like "Case Number: 04-97-00972-CV"). We use fullmatch rather
+    # than search because these regexes were designed for fullmatch
+    # and can produce false positives with partial matching.
+    for token in docket_number.split():
+        for regex in TEXAS_DN_REGEXES:
+            if regex.fullmatch(token):
+                return token
+
+    return ""
+
+
+def make_texas_docket_number_core(docket_number: str | None) -> str:
+    """Normalize Texas docket numbers.
+
+    First cleans the docket number using Juriscraper's DOCKET_NUMBER_REGEXES
+    to extract the actual docket number, then normalizes by stripping all
+    non-alphanumeric characters and lowercasing.
+
+    There is overlap between valid Texas docket numbers and valid Federal
+    docket numbers, but they need to be normalized differently so we need a
+    separate method.
+
+    :param docket_number: The docket number to normalize.
+    :return: The normalized docket number, or empty string if no valid
+    docket number is found.
+    """
+    if docket_number is None:
+        return ""
+
+    cleaned = clean_texas_docket_number(docket_number)
+    if not cleaned:
+        return ""
+
+    not_alphanum_regex = re.compile(r"[^a-z0-9]")
+    return not_alphanum_regex.sub("", cleaned.lower())
+
+
 def make_scotus_docket_number_core(docket_number: str | None) -> str:
-    """Normalize SCOTUS docket numbers like 16A985.
+    """Normalize SCOTUS docket numbers like 16A985 or 01-8148.
+
+    Prioritizes the NN-NNNN format over the NNA format when both are present.
 
     :param docket_number: The docket number to condense.
     :return: empty string if no change possible, or the condensed version if it
@@ -124,14 +245,14 @@ def make_scotus_docket_number_core(docket_number: str | None) -> str:
     if not docket_number:
         return ""
 
-    cleaned_docket_number = clean_docket_number(docket_number)
+    cleaned_docket_number = clean_scotus_docket_number(docket_number)
 
     if bankr_n_core := make_appellate_bankr_number_core(cleaned_docket_number):
         return bankr_n_core
 
-    scouts_a_m = re.search(scotus_d_a_num_regex, cleaned_docket_number)
-    if scouts_a_m:
-        year, serial = scouts_a_m.groups()
+    scotus_a_m = re.search(scotus_d_a_num_regex, cleaned_docket_number)
+    if scotus_a_m:
+        year, serial = scotus_a_m.groups()
         return f"{year}A{int(serial):05d}"
 
     return ""
--- a/cl/api/templates/migration-guide.html
+++ b/cl/api/templates/migration-guide.html
@@ -2,6 +2,19 @@
 {% load static %}
 {% load extras %}
 
+{% comment %}
+╔═════════════════════════════════════════════════════════════════════════╗
+║                               ATTENTION!                                ║
+║ This template has a new version behind the use_new_design waffle flag.  ║
+║                                                                         ║
+║ When modifying this template, please also update the new version at:    ║
+║ cl/api/templates/v2_migration-guide.html                                ║
+║                                                                         ║
+║ Once the new design is fully implemented, all legacy templates          ║
+║ (including this one) and the waffle flag will be removed.               ║
+╚═════════════════════════════════════════════════════════════════════════╝
+{% endcomment %}
+
 {% block title %}V4 API Migration Guide &ndash; CourtListener.com{% endblock %}
 {% block description %}
 {% endblock %}
@@ -41,16 +54,16 @@
   </p>
 
   <h2 id="support">Support</h2>
-  <p>Questions about this migration can be sent <a href="https://github.com/freelawproject/courtlistener/discussions" target="_blank">to our GitHub Discussions forum</a> or to our <a href="{% url "contact" %}">contact form</a>.
+  <p>Questions about this migration can be sent <a href="https://github.com/freelawproject/courtlistener/discussions" target="_blank" rel="noopener noreferrer">to our GitHub Discussions forum</a> or to our <a href="{% url "contact" %}">contact form</a>.
   </p>
   <p>We prefer that questions be posted in the forum so they can help others. If you are a private organization posting to that forum, we will avoid sharing details about your organization.
   </p>
   <p>
     <a href="https://github.com/freelawproject/courtlistener/discussions"
-        target="_blank"
+        target="_blank" rel="noopener noreferrer"
         class="btn btn-default">Ask in GitHub Discussions</a>
     <a href="{% url "contact" %}"
-       target="_blank"
+       target="_blank" rel="noopener noreferrer"
        class="btn btn-default">Send us a Private Message</a>
   </p>
 
@@ -112,7 +125,7 @@
     <li>Timezone handling is more consistent.</li>
     <li>We've added a number of new searchable fields.</li>
   </ul>
-  <p>For more details, please<a href="https://free.law/2024/01/18/new-recap-archive-search-is-live/" target="_blank"> see our blog</a>.</p>
+  <p>For more details, please<a href="https://free.law/2024/01/18/new-recap-archive-search-is-live/" target="_blank" rel="noopener noreferrer"> see our blog</a>.</p>
   <h2 id="breaking-changes">Breaking Changes to v3 of the Search API</h2>
   <p>We cannot continue running Solr forever, but we can do our best to support v3 of the API. To do this, on <strong>November 25, 2024</strong>,
     v3 of the Search API will be upgraded to use ElasticSearch. We expect this to support most uses, but it will cause some breaking changes, as outlined in this section.
--- a/cl/api/templates/v2_migration-guide.html
+++ b/cl/api/templates/v2_migration-guide.html
@@ -1,0 +1,503 @@
+{% extends 'new_base.html' %}
+{% load extras static %}
+
+{% block title %}V4 API Migration Guide &ndash; CourtListener.com{% endblock %}
+{% block description %}
+{% endblock %}
+
+{% block content %}
+<c-layout-with-navigation
+  data-first-active="overview"
+  :nav_items="[
+    {'href': '#overview', 'text': 'Top'},
+    {'href': '#support', 'text': 'Support'},
+    {'href': '#timeline-changes', 'text': 'Timeline'},
+    {'href': '#what-if-do-nothing', 'text': 'Can I Do Nothing?'},
+    {'href': '#what-to-expect', 'text': 'New Features', 'children': [
+      {'href': '#cursor-pagination', 'text': 'Deep Pagination'},
+      {'href': '#elasticsearch', 'text': 'Enhanced Search'},
+    ]},
+    {'href': '#breaking-changes', 'text': 'Breaking Changes to v3', 'children': [
+      {'href': '#recap-breaking-changes', 'text': 'RECAP (type=r)'},
+      {'href': '#opinions-breaking-changes', 'text': 'Opinions (type=o)'},
+      {'href': '#oa-breaking-changes', 'text': 'Oral Arguments (type=oa)'},
+      {'href': '#people-breaking-changes', 'text': 'People (type=p)'},
+    ]},
+    {'href': '#migrate-db-apis', 'text': 'Migrate Database APIs', 'children': [
+      {'href': '#total-count-results', 'text': 'Result counts'},
+      {'href': '#invalid-cursor', 'text': 'Invalid cursors'},
+    ]},
+    {'href': '#enhancements-v4', 'text': 'Search Enhancements', 'children': [
+      {'href': '#enhancements-cursor-pagination', 'text': 'Deep Pagination'},
+      {'href': '#enhancements-results-sorting', 'text': 'Better Ordering'},
+      {'href': '#enhancements-highlighting', 'text': 'Highlighting'},
+      {'href': '#enhancements-empty-fields', 'text': 'Empty fields'},
+    ]},
+    {'href': '#v4-backwards-incompatible-changes', 'text': 'Search Incompatibilities', 'children': [
+      {'href': '#backwards-estimated-counts', 'text': 'Query counts'},
+      {'href': '#backwards-highlighting', 'text': 'Highlighting'},
+      {'href': '#backwards-nested-keys', 'text': 'Nested Keys'},
+      {'href': '#backwards-type-r', 'text': 'Nested Docket Search'},
+      {'href': '#backwards-type-rd', 'text': 'Document Search'},
+      {'href': '#backwards-type-d', 'text': 'Docket Search'},
+      {'href': '#backwards-removed-fields', 'text': 'Removed Fields'},
+      {'href': '#backwards-changed-fields', 'text': 'Changed Field Values'},
+      {'href': '#backwards-date-times', 'text': 'Dates and Times'},
+      {'href': '#backwards-no-random-sorting', 'text': 'Random Sorting'},
+      {'href': '#backwards-stemming-synonyms', 'text': 'Stemming and Synonyms'},
+      {'href': '#backwards-get-params', 'text': 'Changes to GET parameters'},
+      {'href': '#backwards-400-error', 'text': 'Error Code: 400'},
+      {'href': '#backwards-500-error', 'text': 'Error Code: 500'},
+      {'href': '#backwards-404-error', 'text': 'Error Code: 404'},
+    ]},
+  ]"
+>
+  <c-layout-with-navigation.section id="overview">
+    <h1 id="about">V4 API Migration Guide</h1>
+    <p class="overview lead">After several years of planning and development, we have released v4 of our APIs.</p>
+    <p>This upgrade responds to feedback we have received over the years and should be much better for our users — faster, more featureful, more scalable, and more accurate.</p>
+    <p>Unfortunately, we couldn't make these new APIs completely backwards compatible, so this guide explains what's new.</p>
+  </c-layout-with-navigation.section>
+
+  <c-layout-with-navigation.section id="support">
+    <h2>Support</h2>
+    <p>Questions about this migration can be sent <a href="https://github.com/freelawproject/courtlistener/discussions" class="underline" target="_blank" rel="noopener noreferrer">to our GitHub Discussions forum</a> or to our <a href="{% url "contact" %}" class="text-primary-600">contact form</a>.</p>
+    <p>We prefer that questions be posted in the forum so they can help others. If you are a private organization posting to that forum, we will avoid sharing details about your organization.</p>
+    <div class="flex items-start gap-4">
+      <a href="https://github.com/freelawproject/courtlistener/discussions"
+         target="_blank"
+         rel="noopener noreferrer"
+         class="btn-primary">Ask in GitHub Discussions</a>
+      <a href="{% url "contact" %}"
+         target="_blank"
+         rel="noopener noreferrer"
+         class="btn-primary">Send us a Private Message</a>
+    </div>
+  </c-layout-with-navigation.section>
+
+  <c-layout-with-navigation.section id="timeline-changes">
+    <h2>Timeline for Changes</h2>
+    <p>v4 of the API is available now and is the default version for anybody creating new systems. Before its full release, a number of organizations beta tested it.</p>
+    <p>All of our APIs except for our search API are powered by our database. We do not have plans at present to deprecate any of these APIs, but we'd like to remove them someday and urge you to migrate to v4 as soon as possible so we can do that.</p>
+    <p>That said, the v3 Search API is currently powered by Solr while v4 is powered by ElasticSearch. In <strong>nine weeks</strong> we aim to switch v3 so it uses ElasticSearch too. This will change v3 in small backwards incompatible ways, but will allow us to continue supporting it even after turning off our Solr server.</p>
+    <p>If you are a v3 Search API user, you will soon get an email from us to communicate and discuss timelines.</p>
+  </c-layout-with-navigation.section>
+
+  <c-layout-with-navigation.section id="what-if-do-nothing">
+    <h2>What If I Do Nothing?</h2>
+    <p>You might be fine. Most of the database and search APIs are only changing slightly, and v3 will be supported for some period of time. But you should read this guide to see if any changes are needed to your application.</p>
+    <p>The remainder of this guide is in three sections:</p>
+    <ul>
+      <li>New features you can expect</li>
+      <li>How to migrate database APIs</li>
+      <li>How to migrate search APIs</li>
+    </ul>
+    <p>We're very excited to be releasing v4 of our APIs. We hope you will review these changes so we can all have a smooth transition.</p>
+  </c-layout-with-navigation.section>
+
+  <c-layout-with-navigation.section id="what-to-expect">
+    <h2>What New Features Can I Expect?</h2>
+  </c-layout-with-navigation.section>
+
+  <c-layout-with-navigation.section id="cursor-pagination">
+    <h3>Cursor-Based Pagination</h3>
+    <p>Our database-powered APIs now support cursor-based pagination. This allows you to crawl very deeply in the API. In v3, any page past 100 was blocked.</p>
+  </c-layout-with-navigation.section>
+
+  <c-layout-with-navigation.section id="elasticsearch">
+    <h3>ElasticSearch</h3>
+    <p>v4 of the Search API is powered by ElasticSearch instead of Solr. This is a huge upgrade to our API and search engine.</p>
+    <p>Some improvements include:</p>
+    <ul>
+      <li>In v4, all PACER cases are now searchable. In v3 you only got results if a case had a docket entry.</li>
+      <li>You can search for PACER filings based on what decisions they cite.</li>
+      <li>You can now search for exact words like "Deposit" and not get back results for things like "Deposition."</li>
+      <li>We've added about 800 legal acronyms like "IRS" to make sure those bring back results.</li>
+      <li>Better relevancy for edge cases:
+        <ul>
+          <li>Small words like "of," "to," and "the" are now searchable.</li>
+          <li>Camelcase words like "McDonalds" are more searchable.</li>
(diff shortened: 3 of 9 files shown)


quevon24:
I designed this workflow to be used for dynamic, version-controlled LLM tasks, like the one below.

For example, to extract structured text from a document scan and create a new OpinionContent object, we need to do something like this:

**Create the config**

```python

# Create config
config_ocr = LLMConfig.objects.create(
    config_name="ocr_extractor_v1",
    provider="google",  # Example provider change
    model_name="gemini-2.5-flash",
    parameters={"temperature": 0.0},  # Low temp for precision
)

# Define prompts
prompt_system_ocr = Prompt.objects.create(
    role=Prompt.SYSTEM,
    text="You are an expert OCR validation tool. Analyze the file content and return the extracted data strictly in the requested XML format. Only output the XML block.",
    position=1,
)
# 2. User Prompt (The Template for input data)
prompt_user_ocr = Prompt.objects.create(
    role=Prompt.USER, text="Apply ocr to the attached file", position=2
)
# Define Prompt Set
prompt_set_ocr = LLMPromptSet.objects.create(
    name="pdf_ocr_extraction_task",
    version=1,
    description="Set of instructins for structured data extraction from scanned documents.",
)
# Link the Prompts to the PromptSet
prompt_set_ocr.prompts.set([prompt_system_ocr, prompt_user_ocr])

``` 

Run the LLM and link created object or save bad response for debugging purposes
```python
# get prompts
prompts = prompt_set_ocr.prompts.all().order_by("position")

# get input content
file_path = "scanned_pdf_document.pdf"

# Map DB roles to API roles
role_map = {Prompt.SYSTEM: "system", Prompt.USER: "user"}

# build the llm conversation
messages_payload = []

# add system prompt
messages_payload.append({"role": role_map.get(p.role), "content": content})
# attach file if neccesary
messages_payload.append(
    {
        "role": role_map.get(p.role),
        "content": [
            {"type": "text", "text": user_prompt_text},
            {
                "type": "file",
                "data": base64_file_data,
                "mime_type": mime_type,
                "file_name": os.path.basename(file_path),
            },
        ],
    }
)

# call a function to do the llm call and get response (this is a mock function to exemplify funcitonality)
response = call_llm(**config_ocr.__dict__)

with transaction.atomic():
    # Store settings used for this run
    llm_run = LLMRun.objects.create(
        llm_config=config_ocr, prompt_set=prompt_set_ocr, output=raw_response
    )

    # do some validation
    validated_xml_data = parse_xml_for_db(raw_response)

    # create final object
    opinion_content_instance = OpinionContent.objects.create(
        content=validated_xml_data,
        source="SCANNING",
    )

    # Link LLMRun to created object using GFK
    llm_run.content_object = opinion_content_instance
    llm_run.failed = False
    llm_run.save()


```

cl/ai/models.py
```diff
@@ -0,0 +1,153 @@
...
+        default=dict,
+        blank=True,
+        help_text="JSON dictionary for specific model settings like {'temperature': 0.5, 'max_tokens': 1000}",
+    )
+
+    class Meta:
+        constraints = []  # TODO: Maybe add a constrait avoiding repeated configurations?
+
+    def __str__(self):
+        return f"{self.provider} / {self.model_name}"
```
grossir: shouldn't this be the `config_name`; since the provider and model _name combinations are not unique?

cl/ai/models.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def save(self, *args, **kwargs):
+        # Convert the entire name to lowercase
+        self.name = self.config_name.lower()
+        self.name = re.sub(r"[_\s-]+", "_", self.name).strip("_")
```
grossir: self.name is not a defined attribute of the model?

quevon24:
Thanks @grossir. I made a few changes in the models. I implemented the admin models. And a second migration to create the configuration for `classify_case_name_by_llm` task.

Once you've applied the migrations, you'll be able to go to the admin panel to view data in: LLMConfig, Prompts, and LLMPromptSet

for LLMRun, it will look something like this:

image

image

quevon24:
I made some minor changes to the pull request. Originally, I created a migration that automatically created the necessary objects to prevent the task from failing. Now, if the objects haven't been manually created, we'll get a message in Sentry.

 It now also store the raw response from the LLM in LLMRun:

image

cl/corpus_importer/tasks.py
```diff
@@ -3144,15 +3174,25 @@ def classify_case_name_by_llm(self, cluster_pk: int, recap_document_id: int):
                 "fingerprint": ["llm-casenames-validation-error"],
             },
         )
+        # Save error as plain text
+        llm_run.output = str(e)
+        # Save error as json
```
grossir: flag `llm_run.success = False`?

quevon24:
> I think this is mostly ready
> 
> Just a question from our most recent meeting: do we need to add anything to the LLMRun model to flag batch requests? A boolean? Or would that be implicit in the config name or something like that. This is beyond the actual implementation of waiting for batch responses, which we should do in another PR

That's a good question. I actually looked into adding a flag to `LLMRun`, but I think a separate model for batch jobs is a better way to go.

The main reason is that batch jobs have a lot of data that doesn't really belong in `LLMRun`. Things like the provider's job ID, the S3 file paths for the input and output files, the expiration date are all about the batch process itself, not the individual result.

If we added those fields to `LLMRun`, they would just sit there empty for every single call request we ever make. By keeping them separate, we keep the database much cleaner, `LLMRun` stays focused on the actual answer from the AI, while a new model, like LLMBatchJob (proposed here: https://github.com/freelawproject/courtlistener/issues/6714) handles the files and the status in the provider's queue.

It also makes the Admin a lot nicer to use. We can have a dedicated page to check the progress of a job or download the raw JSONL files without cluttering the list of individual results from `LLMRun`.

-----------
That sounds like a good idea. I think that once we have this core infrastructure settled, we should look at moving those other parts of the code, like the transcription and docket cleaning into this new system. It would be much better to have a consistent way to manage prompts and logs across the entire repo instead of having different patterns for different tasks.

Regarding Whisper, I also like the idea of saving more of the raw output. One of the main goals of these new models was to get better traceability and see exactly what's happening under the hood, so saving the full technical response for audio transcription fits perfectly with that.

cl/ai/models.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+class LLMConfig(AbstractDateTimeModel):
+    """Stores the specific LLM model identifier and generation parameters used for execution"""
+
+    PROVIDER_CHOICES = (
+        ("openai", "OpenAI"),
+        ("anthropic", "Anthropic"),
+        ("google", "Google Gemini"),
+    )
```
mlissner: I'd use an integer field for this and as a general practice.

cl/ai/models.py
```diff
@@ -0,0 +1,194 @@
...
+        self.name = slugify(self.name)
+        super().save(*args, **kwargs)
+
+    @property
+    def full_model_name(self) -> str:
+        """Returns the provider/model string for instructor (e.g., openai/gpt-4o-mini)."""
+        return f"{self.provider}/{self.model_name}"
+
+
+class LLMPromptSet(AbstractDateTimeModel):
```
mlissner: I'm only passing through, but this model and LLMTask feel over-engineered to me. Do we need a task model to hold prompt collections that hold prompts? 

Again, I'm only passing through, but it feels like a lot?

mlissner:
I tossed out a couple comments while skimming this. Have we gotten a review from @nadahlberg on this one too? I know he's pushing for `litellm`, which could impact this, maybe?

quevon24:
> I tossed out a couple comments while skimming this. Have we gotten a review from @nadahlberg on this one too? I know he's pushing for `litellm`, which could impact this, maybe?

I've reviewed the options for batch processing with Gemini. Here's the current landscape:

**Current Setup:**

- Using Instructor library for single LLM calls with structured output (Pydantic), this has been working well for our use case (extract case names from recap_into_opinions)
- Draft of scanning project currently uses Google's Gemini API via the `google-genai` library for batch processing

**Key Finding:**
If we adopt LiteLLM for Gemini batch processing, we can only use it through VertexAI, not the native Gemini API. VertexAI is Google's enterprise ML platform that:

- Requires full Google Cloud project setup (vs simple API key)
- Has delayed access to latest models
- Adds infrastructure overhead we don't currently need

**Recommendation:**
Keep using Instructor for single calls (proven, works well) and implement a custom wrapper specifically for Google GenAI batch processing. This gives us:

- Direct access to latest Gemini models immediately
- Simple API key authentication (no Google Cloud project required)

This approach maintains what's working while adding efficient batch processing without unnecessary abstraction layers

cl/ai/models.py
quevon24: I definitely hear the concern about the number of models, it looks like a lot of glue for what seems like a simple API call. The reason I went with this structure was to decouple the AI's behavior from the application code to avoid a few specific bottlenecks:

- **Versioning & Hot-fixes:** By having these models, we can tweak a prompt in the Admin to fix a hallucination or edge case and have it live in seconds. If we hardcode these, a 2-minute text change requires a full PR and deployment cycle.
- **Model Independence:** LLMConfig lets us swap the engine (e.g., GPT-4o to Gemini) or tweak parameters like temperature without touching business logic.
- **Traceability:** LLMRun acts as our audit log to monitor raw responses for debugging.

Can we lean it down? If the current structure feels over-engineered, we can definitely simplify it:

1. Remove LLMPromptSet: If we're confident we only ever need a 'System' and 'User' prompt pair, we can move those text fields directly onto the LLMTask model and get rid of the PromptSet and Prompt models entirely.
2. Flatten LLMTask: We could manage versions by adding a version field and an is_active boolean directly to the LLMTask. In the code, we’d just fetch the latest active record for a given name (e.g., 'recap_into_opinions_extract_casename').

What do you think? I'm all ears for your thoughts on which direction we should take

This is the flattened model draft:

```python
class LLMTask(AbstractDateTimeModel):
    name = models.CharField(
        max_length=255,
        unique=True,
        help_text="The static key used in Python code (e.g. 'recap-opinion-casename-extraction'). This will be hardcoded in the code so never rename it.",
    )
    description = models.TextField(
        blank=True,
        help_text="Description of the task",
    )
    current_config = models.ForeignKey(
        LLMConfig,
        on_delete=models.PROTECT,
        help_text="The configuration currently live for this task",
    )
    # New fields
    version = models.PositiveIntegerField(default=1)
    is_active = models.BooleanField(default=False)
    system_prompt = models.TextField()
    user_prompt_template = models.TextField()

    class Meta:
        unique_together = ('name', 'version')
        ordering = ['-version']

    def __str__(self):
        return (
            f"{self.name} -> {self.current_config}"
        )

    @classmethod
    def get_active(self, task_name):
        return self.objects.filter(name=task_name, is_active=True).first()

    def save(self, *args, **kwargs):
        self.name = slugify(self.name)
        super().save(*args, **kwargs)
```

nadahlberg:
**Batch Processing with Gemini:** I think it makes sense to stick with the Google GenAI package for this. Batch jobs don’t yet seem to be widely standardized across providers in packages like `litellm`

**Instructor vs LiteLLM**

Here would be my pushback for `litellm`. `instructor` is a library that specializes in single-step structured output extraction. Even on the front page of their docs they recommend using other, more minimal frameworks for agents (they say [“Instructor for extraction, PydanticAI for agents”](https://python.useinstructor.com/)).

I would argue that `litellm` can do what `instructor` does but is less opinionated and more flexible, allowing for fully agentic workflows that involve tool calling (here is a nice [article from Anthropic](https://www.anthropic.com/engineering/building-effective-agents) that does a much better job spelling out the difference between workflows that chain together structured output calls versus agents).

**Single-step structured outputs**

If you want to give the model some text and a json format, and get json in response, both libraries support `pydantic` and implement this similarly:

```python
import json
import instructor
from litellm import completion
from pydantic import BaseModel

class ExtractUser(BaseModel):
    name: str
    age: int
   
# Using instructor
client = instructor.from_provider("google/gemini-2.5-flash-lite")
user = client.create(
    messages=[{"role": "user", "content": "John Doe is 30 years old."}],
    response_model=ExtractUser,
)
print(user)
>> name='John Doe' age=30

# Using litellm
response = completion(
    model="gemini/gemini-2.5-flash-lite",
    messages=[{"role": "user", "content": "John Doe is 30 years old."}],
    response_format=ExtractUser,
)
user = ExtractUser(**json.loads(response.choices[0].message.content))
print(user)
>> name='John Doe' age=30
```

`instructor` automatically converts the response to an instance of the `pydantic` model and has some additional pydantic-specific hooks, whereas `litellm` returns the full completion response object.

Also in `litellm` the `response_format` arg can also be provided in json. One advantage is that this could be included in the `LLMConfig`, allowing users to specify different json formats from the admin site directly:

```python
response_format={
	"type": "json_object", 
	"json_schema": ExtractUser.model_json_schema()
},
```

**Tool Calling**

Using the `response_format` (or `response_model`) args makes sense when you know you want exactly one type of json output per input. However “tool use” is useful when any of the following apply:

- You have multiple json schemas and want to allow the model to decide which one to use.
- You may want to call a tool zero-to-many times for a given input, or in a specific order, decided by the model.
- Your tools “do things” that return information to the model, and you want the model to be able to use these tools, get feedback, and continue its response.

A few examples:

- An agent that has a tool for doing semantic search on the Case Law archive, and other tools for hitting various RECAP endpoints. Depending on the user query it can decide which search endpoints make sense to hit, if any. It can also tweak it’s query and search again if not satisfied with the results.
- An agent that extracts provisions from a document. It has an extraction tool, but it also has a tool for doing semantic search on a dataset of previous extractions. For easy examples it can just use the extraction tool, but for tricky ones it can search previous annotations to make sure it’s annotating things consistently. By making this agentic, you save tokens by only using the search tool when deemed necessary by the model.

With `litellm` we can pass `pydantic` model schemas to be used as tools. Here’s a somewhat trivial example using the `ExtractUser`  model from earlier:

```python
response = completion(
    model="gemini/gemini-2.5-flash-lite",
    messages=[
        {"role": "system", "content": "Use your ExtractUser tool to extract details."},
        {"role": "user", "content": "John Doe is 30 years old."}
    ],
    tools=[{
        "type": "function",
        "function": {
            "name": ExtractUser.__name__,
            "description": ExtractUser.__doc__,
            "parameters": ExtractUser.model_json_schema(),
        },
    }],
)
for tool_call in response.choices[0].message.tool_calls:
    user = ExtractUser(**json.loads(tool_call.function.arguments))
    print(user)
```

The response includes the `tool_calls` in the appropriate format so that, if this were something like a search tool, we could make the API call and append a `tool_response` message to the messages and have the model continue working.

We have an `Agent` abstraction in the [classifier-experiments repo](https://github.com/freelawproject/classifier-experiments/blob/main/clx/llm/agent.py) that simplifies the tool calling loop with `litellm` so that the tool action / response can be defined directly in the `pydantic` model by implementing `__call__`. It’s worth checking out! You could use this for both single-step structured outputs as above and for more elaborate agentic workflows, and it includes hooks for all the potential non-serializable steps of an agentic workflow (the stuff that couldn’t just be dumped into an `LLMConfig` row).

Here are two examples. First structured outputs:
```python
from clx import Agent, Tool

class ExtractUser(Tool):
    name: str
    age: int

agent = Agent(
    model="gemini/gemini-2.5-flash-lite",
    response_format=ExtractUser,
)

response = agent.step("John Doe is 30 years old.")
user = ExtractUser(**json.loads(response["content"]))
>> name='John Doe' age=30
```

 And here's an example of an agent with a mock weather API tool:
```python
class WeatherTool(Tool):
    location: str

    def __call__(self, agent: Agent) -> str:
        """Tool call."""
        return "The weather in " + self.location + " is 72 degrees Fahrenheit."

agent = Agent(
    model="gemini/gemini-2.5-flash-lite",
    tools=[WeatherTool],
)

response = agent.run("What's the weather in North Carolina and San Diego?")
print(response["content"])
>> "The weather in North Carolina is 72 degrees Fahrenheit. The weather in San Diego is 72 degrees Fahrenheit."
```
If we inspected `agent.messages` we'd see two tool calls and two tool response messages, before the agent's final response message to the user.

My case for `litellm` would be that it gives us the same thing as `instructor` but is more flexible allowing us to standardize “agentic” stuff as well. I don't think this would affect the data models themselves, but rather what we put into things like the `LLMConfig` -- whether to use `litellm` versus `instructor` conventions for model names and config arguments.

cl/ai/models.py
```diff
@@ -0,0 +1,331 @@
...
+    status = models.SmallIntegerField(
+        help_text="The current status of the task",
+        choices=TaskStatus.NAMES,
+        default=TaskStatus.UNPROCESSED,
+    )
+
+    llm_key = models.CharField(
+        max_length=255,
+        help_text="The LLM key for this task",
+    )
```
grossir, line 216: I am not sure what this is referring to. Can you improve the attribute name and the help_text?

cl/ai/models.py
```diff
@@ -0,0 +1,331 @@
...
+
+    # Input file (generic - could be PDF, text, etc.)
+    input_file = models.FileField(
+        help_text="The input file to be processed (PDF, text, etc.)",
+        upload_to=make_input_file_path,
+        storage=IncrementingAWSMediaStorage(),
+        max_length=1000,
+        blank=True,
+    )
```
grossir, line 239: Can we assign an already existing path to this field? Can you add a test for that?
 
If we are going to LLM-extract a scraped document, then it is already saved in S3. So we should just put that path here, not duplicate it. But I am not sure if that will fail or not

cl/ai/models.py
```diff
@@ -0,0 +1,331 @@
...
+
+    response_file = models.FileField(
+        help_text="Full response from the LLM provider stored as a file",
+        upload_to=make_response_file_path,
+        storage=IncrementingAWSMediaStorage(),
```
grossir: I think this is a public bucket. Do we want the raw response to be in a public bucket?

cl/ai/models.py
```diff
@@ -0,0 +1,331 @@
...
+
+
+class TaskStatus:
+    """LLM Task status choices"""
+
+    UNPROCESSED = 0
+    IN_PROGRESS = 1
+    SUCCEEDED = 2
+    FAILED = 3
+    FINISHED = 4
```
grossir: what's the difference between SUCCEEDED and FINISHED?

cl/ai/migrations/0001_initial.sql
```diff
@@ -0,0 +1,31 @@
...
+ALTER TABLE "ai_llmrequest_prompts" ADD CONSTRAINT "ai_llmrequest_prompts_prompt_id_62f3058d_fk_ai_prompt_id" FOREIGN KEY ("prompt_id") REFERENCES "ai_prompt" ("id") DEFERRABLE INITIALLY DEFERRED;
+CREATE INDEX "ai_llmrequest_prompts_llmrequest_id_2221e73e" ON "ai_llmrequest_prompts" ("llmrequest_id");
+CREATE INDEX "ai_llmrequest_prompts_prompt_id_62f3058d" ON "ai_llmrequest_prompts" ("prompt_id");
+ALTER TABLE "ai_llmtask" ADD CONSTRAINT "ai_llmtask_content_type_id_8d14a327_fk_django_content_type_id" FOREIGN KEY ("content_type_id") REFERENCES "django_content_type" ("id") DEFERRABLE INITIALLY DEFERRED;
+ALTER TABLE "ai_llmtask" ADD CONSTRAINT "ai_llmtask_request_id_80ee2b0f_fk_ai_llmrequest_id" FOREIGN KEY ("request_id") REFERENCES "ai_llmrequest" ("id") DEFERRABLE INITIALLY DEFERRED;
+CREATE INDEX "ai_llmtask_date_created_1c75ad5f" ON "ai_llmtask" ("date_created");
+CREATE INDEX "ai_llmtask_date_modified_d8083f72" ON "ai_llmtask" ("date_modified");
+CREATE INDEX "ai_llmtask_content_type_id_8d14a327" ON "ai_llmtask" ("content_type_id");
```
grossir: we could delete this index, since it's contained in the index in line 30

cl/ai/models.py
```diff
@@ -0,0 +1,331 @@
...
+    extra_config_params = models.JSONField(
+        help_text="Additional input data or parameters for the LLM",
+        default=dict,
+        blank=True,
+        null=True,
+    )
+    batch_response_file = models.FileField(
+        help_text="The batch response file from the LLM provider (e.g., a JSONL file)",
+        upload_to=make_response_file_path,
+        storage=IncrementingAWSMediaStorage(),
```
grossir: same question here. Do we want the response in a public bucket?

Also, what's the bucket policy? I think the response should be in a somewhat "frozen" policy, since we intend to use it once for extraction and then rarely touch it...

grossir (commented):
Looks good, mostly storage questions

cl/ai/models.py
flooie: I think SUCCEEDED as it successfully returned a response we want from an LLM call.  Finished means it was processed.  so if we successfully got a good response for case name extraction, until its processed it's not finished.  it will allow us to handle failures in the last step of the process.

cl/ai/models.py
flooie: yes I agree.  I think the response could have value down the line, and also if we encounter unexpected issues when it scales up.  

Until then thoughI think you are right, we dont want this just sitting in a public bucket.

cl/ai/models.py
flooie: when we push a batch, we generate a key that is used with the batch return.  send 100 opinions in a batch, we need to be able to identify which response goes to which file.

cl/ai/models.py
flooie: yes, that makes sense, we should be doing that when a scan is processed by blackletter

cl/ai/models.py
flooie: no probably not

cl/ai/models.py
quevon24: I understand that it is possible, but the `input_file` bucket would have to point to the same bucket where the scraped document is located (otherwise it would generate a non-existent url)

Something like this:

```python
obj = LLMTask()
obj.input_file.name = 'path-to-scrape/file.pdf' # S3 object key
obj.save() 
```

cl/ai/models.py
quevon24: So we need to implement a new `S3Storage` in `storage.py`

We will use the bucket specified here: `settings.AWS_PRIVATE_STORAGE_BUCKET_NAME`? or would it be a different bucket?

cl/ai/models.py
quevon24: Improving the help_text, as Gianfranco suggests, is not a bad idea.

quevon24:
The last few commits implement the Google Gemini batch API integration. Added the wrapper class with support for caching and batch operations, plus a response validator that handles both batch and single API responses. Also refactored the models to use `IntegerChoices` and moved some helpers to a shared location. Tests are all passing now.

cl/ai/migrations/0001_initial.sql
quevon24: I think we should keep both indexes. The single `content_type` index is useful when we're filtering by type alone, and the composite `(content_type, object_id)`  index is needed for lookups on specific objects. It's a common pattern with GenericForeignKey to have both since they optimize different query patterns.

grossir: The `(content_type, object_id)` index can be used as a standalone `(content_type)` index, that why I say its redundant. From postgres [docs](https://www.postgresql.org/docs/current/indexes-multicolumn.html)

> A multicolumn B-tree index can be used with query conditions that involve any subset of the index's columns, but the index is most efficient when there are constraints on the leading (leftmost) columns. The exact rule is that equality constraints on leading columns, plus any inequality constraints on the first column that does not have an equality constraint, will always be used to limit the portion of the index that is scanned

cl/ai/llm_providers/google.py
```diff
@@ -0,0 +1,290 @@
...
+            json.loads(line)
+            for line in jsonl_content.splitlines()
+            if line.strip()
+        ]
+
+        for result in results:
+            key = result.get("key")
+            text = _ResponseValidator.get_text(result)
+
+            status = "SUCCEEDED" if text else "FAILED"
```
grossir: should this use the `TaskStatus` from models.py?

cl/ai/models.py
grossir: I think we should just use that one. The organization into "folders" (they are not really that) is created by a series of slashes in the path

cl/ai/models.py
grossir: yes, that was why I was thinking this may create a problem, and wanted to see a test for it. But I am not sure if we can test S3 integration

quevon24: Good question, I'm not sure. I'll look into it a bit, and I'll also check existing tests to see if they've tested anything like that.

cl/ai/migrations/0001_initial.sql
quevon24: Thanks for the info, now I understand your suggestion better.

cl/ai/llm_providers/google.py
quevon24: The goal of doing it this way is to decouple the LLM logic from Django, although in this case it would only be importing a value, so I do see it as possible.

quevon24:
> Looks good, I think we are only missing solving the remaining open questions on S3 storage (buckets, policies)
> 
> Logic in `llm_providers/google` doesn't really affect the models? It looks good, but I think it would be better placed on a PR that also implements a specific case. It can stay here too, and be changed afterwards

No, it doesn't affect any model. I wrote the wrapper so it can be easily reused without being tied to a specific model. And I also don't see a problem moving this to a separate PR since it doesn't depend on any models.

quevon24:
I implemented tests to cover the batch management commands (`send_gemini_batches` and `check_gemini_batch_status`).

I updated both management commands:                                                                                                     
  - Changed them to inherit from `VerboseCommand` for better consistency with the codebase                                                
  - Replaced `stdout.write()` and `stderr.write()` calls with `logger.info()` to eliminate duplicate output                               
  - Updated environment variable loading to use `environ.FileAwareEnv()` for better `.env` file support
 
I made a few improvements to decouple the wrapper better:                                                                               
  - Replaced `print()` statements with proper logging                                                                                     
  - Added a `ProcessedResult` TypedDict for better type safety                                                                            
  - Improved error messages to include job state context                                                                                  
  - Added an early return for empty JSONL content

cl/ai/migrations/0001_initial.sql
grossir: I really think we should delete this, and just prevent having to make future changes. More info here https://github.com/freelawproject/courtlistener/issues/6395

cl/ai/models.py
```diff
@@ -0,0 +1,291 @@
...
+
+class LLMProvider(models.IntegerChoices):
+    """LLM Provider choices"""
+
+    GEMINI = 1, "Google Gemini"
+    OPENAI = 2, "OpenAI"
+    ANTHROPIC = 3, "Anthropic"
+
+
+class Task(models.IntegerChoices):
```
grossir: I think this should have a more specific name, like `LLMTask`

Because Task is a celery class used in a lot of places, this may create some confusion

Image

cl/ai/models.py
```diff
@@ -0,0 +1,291 @@
...
+
+
+class TaskStatus(models.IntegerChoices):
+    """LLM Task status choices"""
+
+    UNPROCESSED = 0, "Unprocessed"
+    IN_PROGRESS = 1, "In Progress"
+    SUCCEEDED = 2, "Succeeded"
+    FAILED = 3, "Failed"
+    FINISHED = 4, "Finished"
```
grossir: I think we should change the names of Succeeded and Finished. I know you and @flooie explained that one means that the LLM returned a successful response; the other means that we are done processing this batch, but as it is, but the text description should be better

Maybe 
SUCCEEDED = 2, "LLM provider returned a successful response"
FINISHED = 4, "We are done processing this LLM response">

cl/ai/models.py
```diff
@@ -0,0 +1,291 @@
...
+    status = models.SmallIntegerField(
+        help_text="The current status of the task",
+        choices=TaskStatus,
+        default=TaskStatus.UNPROCESSED,
+    )
+
+    llm_key = models.CharField(
+        max_length=255,
+        help_text="A unique identifier for this task, used to map results back from a batch job.",
+    )
```
grossir, line 216: Maybe this is better, more explicit

"A unique identifier from the LLM  provider system that identifies this this task, used to map results back from a batch job."

cl/ai/models.py
```diff
@@ -0,0 +1,291 @@
...
+from django.contrib.contenttypes.models import ContentType
+from django.db import models
+
+from cl.lib.model_helpers import (
+    make_llm_request_response_file_path,
+    make_llm_task_input_file_path,
+    make_llm_task_response_file_path,
+)
+from cl.lib.models import AbstractDateTimeModel
+from cl.lib.storage import IncrementingAWSMediaStorage
```
grossir: You probably need to implement a new class in `cl.lib.storage` 

Something like this is all we need for it to be private. We don't have to create anything manually on S3, just use the private bucket

```python
class S3LLMResponseStorage(S3Storage):
    """Uses S3 GlacierInstantRetrieval storage class with private ACL"""

    default_acl = "private"
    bucket_name = settings.AWS_PRIVATE_STORAGE_BUCKET_NAME
    file_overwrite = False
```

I see you have created path makers in `cl.lib.model_helpers`. However, I think 

With @flooie  we have been discussing this, and we agree about using llm_responses

The LLMRequest id is unique, and holds a link to all the metadata about this request, so probably the path template should be the "filename"

cl/lib/model_helpers.py
```diff
@@ -147,6 +147,39 @@ def make_path(root: str, filename: str) -> str:
...
 
+def make_llm_task_input_file_path(instance, filename: str) -> str:
+    """Create a date-based path for LLMTask input files.
+
+    Ensures the filename is unique by prepending the task's primary key.
+
+    Example: llm-tasks/2026/01/15/101-scan_file.pdf
+    """
+    unique_filename = f"{instance.pk}-{filename}"
+    return make_path("llm-tasks", unique_filename)
```
grossir: I think we could just straight use the instance.pk as filename .... the variable "filename" doesn't really add much context

cl/ai/management/commands/check_gemini_batch_status.py
```diff
@@ -0,0 +1,211 @@
...
+        logger.info(
+            f"Found {len(pending_requests)} pending batch requests to check."
+        )
+
+        if not pending_requests:
+            return
+
+        wrapper = GoogleGenAIBatchWrapper(api_key=batch_api_key)
+        for request in pending_requests:
+            logger.info(
```
grossir: extract the whole loop into another function. There is way too much indentation going on inside. Also, you could make an atomic transaction per request, if needed. How do we resume processing the responses if a single part of the batch has something unexpected and makes everything fail?

cl/ai/models.py
```diff
@@ -0,0 +1,292 @@
...
+        choices=LLMTaskStatusChoices,
+        default=LLMTaskStatusChoices.UNPROCESSED,
+    )
+
+    llm_key = models.CharField(
+        max_length=255,
+        help_text="A unique identifier from the LLM provider for this task, used to map results back from a batch job.",
+    )
+
+    task = models.SmallIntegerField(
```
grossir: this would be better names as "task_type"?

The column would be `ai_llmtask.task` which is not too informative

cl/ai/models.py
```diff
@@ -0,0 +1,292 @@
...
+    api_model_name = models.CharField(
+        help_text="Specific model version (e.g., gemini-3-pro-preview, claude-sonnet-4.5)",
+        max_length=100,
+        blank=True,
+    )
+    status = models.SmallIntegerField(
+        help_text="The current status of the request",
+        choices=LLMTaskStatusChoices,
+        default=LLMTaskStatusChoices.UNPROCESSED,
+    )
```
grossir, line 130: the LLMRequest and the LLMTask model are both using `LLMTaskStatusChoices` .  LLMTask references LLMRequest

Shouldn't there be 2 separate enums? `LLMRequestStatusChoices` and  `LLMTaskStatusChoices`  ? 

I guess that when processing on batch, some of the tasks may fail due to X reasons and the rest succeed. In the end, the batch request  itself was actually processed. So some tasks get marked as "failed" and others as "succeeded"; the batch itself is marked as "succeeded". It can be confusing...

Maybe it's worth to separate them for clarity

cl/ai/tests.py
```diff
@@ -0,0 +1,51 @@
...
+        llm_request.prompts.add(prompt)
+        self.assertEqual(LLMRequest.objects.count(), 1)
+        self.assertEqual(llm_request.prompts.count(), 1)
+
+        llm_task = LLMTask.objects.create(
+            request=llm_request,
+            task=LLMTaskChoices.CASENAME,
+            content_object=self.docket,
+            llm_key="test-key-1",
+        )
```
grossir, line 48: can you try "uploading" a file to `LLMTask.input_file` . ?

### First use case

The doubt I have is the usage of f"{instance.pk}" on the `make_llm_task_input_file_path` and sibling functions. If the object is saved to the DB before creating the path, it will have a PK. If not, this the PK will be None

So, you could include this in the test to make sure this won't happen.... Would need to mock the actual upload, what we really want to test is the case name

I don't know how regularly we will upload a "new" file (see the next use case) So maybe we can just leave it as an open issue?

### Second use case

For the planned development we will use already existing public files
- scraped opinions
- scraped recap documents
- redacted files

So, will this `input_file` field take an already existing publick bucket path as valid?

quevon24: Good catch on the instance.pk concern. I tested it and confirmed: the current upload_to helpers using `f"{instance.pk}"` only work correctly when setting an existing file path directly `(e.g., task.input_file.name = "some/existing/key.pdf")`, which is the pattern we use in send_gemini_batches.py (See: https://github.com/freelawproject/courtlistener/pull/6907). In that flow, upload_to is never called, so instance.pk is irrelevant.

However, if we ever upload a file directly via FileField.save(), Django calls upload_to before the instance is saved to the DB, so instance.pk is None and the path becomes something like `llm-tasks/2026/02/12/None.pdf`.

I can switch the path helpers from instance.pk to `uuid4().hex` so filenames are always unique regardless of save order. For example: `llm-tasks/2026/02/12/a3b2c1d4e5f6789012345678abcdef01.pdf`. That way we're safe for both use cases (existing paths and direct uploads). What do you think?

Regarding `input_file` taking an existing public bucket path: yes, it works. Django's FileField is just a varchar column. Setting `task.input_file.name = "some/existing/path.pdf"` and saving simply stores that string in the DB with no validation that the file exists at that path. It doesn't trigger `upload_to` or any storage check.

grossir: Great. Can you handle the fallback with the UUID4? Just in case

cl/ai/models.py
```diff
@@ -0,0 +1,302 @@
...
+        help_text="Maximum number of retries for failed tasks",
+        default=3,
+    )
+    total_cost_estimate = models.DecimalField(
+        help_text="Estimated API cost before running",
+        max_digits=10,
+        decimal_places=4,
+        null=True,
+        blank=True,
+    )
```
flooie, line 159: What is this for?  is this something we have in our tests ?

quevon24: Those fields aren't used in the current implementation yet. I added `total_cost_estimate` and `total_cost_actual` to track api costs per request, so we have the data model ready when we need to monitor spending across providers (Gemini, OpenAI, etc.). Since this is an initial migration, it's easier to include them now than to add them later with a separate migration. We could also track costs directly on the provider panels, though I'm not sure if they support tracking per batch job.

grossir (approved):

quevon24:
Can you take a look at it, @mlissner?

cl/ai/views.py
mlissner, line 1: I *think* if we don't have views we don't need this file? It used to be that you needed certain files for Django to recognize that a directory was an app, but I think that's not necessary anymore, or that there's a workaround? We should empty files if we can though.

cl/lib/model_helpers.py
```diff
@@ -146,6 +147,39 @@ def make_path(root: str, filename: str) -> str:
     )
 
 
+def make_llm_task_input_file_path(instance, filename: str) -> str:
+    """Falls back to uuid4 when instance.pk is None (unsaved instance)
+
+    Example: llm-tasks/2026/01/15/101.pdf
+    Example: llm-tasks/2026/01/15/a3b2c1d4e5f6.pdf
+    """
+    ext = os.path.splitext(filename)[1]
```
mlissner: This and the other locations should use Pathlib instead of os.path.

cl/lib/model_helpers.py
```diff
@@ -146,6 +147,39 @@ def make_path(root: str, filename: str) -> str:
...
+    name = instance.pk or uuid4().hex
+    return make_path("llm-tasks", f"{name}{ext}")
+
+
+def make_llm_request_response_file_path(instance, filename: str) -> str:
```
mlissner: I might be mistaken, but I *think* you can probably simplify these three functions with https://docs.python.org/3/library/functools.html#functools.partial?

quevon24:
I applied these changes:

- Remove the views.py file (is enough to have the `_init_.py `and to be listed in `INSTALLED_APPS` to be recognized as an app)
- Simplified the file path helpers for LLM models

chaco-fl:
Customers RDS Updated 
Screenshot 2026-02-27 at 11 27 25


Adding the new table to each replication.

chaco-fl:
I have updated all users grants to have access to the new tables
Screenshot 2026-03-02 at 15 31 04


I updated all the current publications
Screenshot 2026-03-02 at 15 31 09


I also refresh each on them

Screenshot 2026-03-02 at 15 37 26


Cl replica was updated

chaco-fl:
I'm ready to updated prod and then run some validations

PR #6655, opened 2025-12-10
fix(registration): handle race condition

bhsauro:
## Fixes
Fixes: #6406 - `IntegrityError` that occurs during user registration when two requests attempt to create the same username simultaneously (race condition). 

## Summary
The registration flow was updated to defensively catch an `IntegrityError` and determine whether the user already exists. If the user exists, the error is treated as a successful registration and the request is redirected to the success page. If the user does not exist, the form is re-rendered with an error message. Logging was added in the latter case for observability.

A known compromise is that the handler catches a general `IntegrityError` rather than checking a specific database constraint. This allows the registration flow to gracefully handle both the expected username-collision case and rare integrity errors affecting other fields, instead of allowing the request to crash. All cases are logged to ensure unexpected failures are visible and can be refined later.

This change was manually tested by forcing an `IntegrityError` (disabled captcha validation):
- `IntegrityError` was caught and logged, and the redirect to success page works
- Unable to test edge case (requires `IntegrityError` on non-username field)

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

No special deployment steps required.

### DIFF
--- a/cl/users/views.py
+++ b/cl/users/views.py
@@ -546,7 +546,7 @@ def register(request: HttpRequest) -> HttpResponse:
                     up.save()
 
                 except IntegrityError as e:
-                    # Redirect to success if user already exists 
+                    # Redirect to success if user already exists
                     try:
                         user = User.objects.get(username=cd["username"])
                         get_str = f"?next={urlencode(redirect_to)}&email={urlencode(user.email)}"
@@ -562,7 +562,10 @@ def register(request: HttpRequest) -> HttpResponse:
                             exc_info=True,
                         )
 
-                        form.add_error("username", "An error occurred during registration. Please try again.")
+                        form.add_error(
+                            "username",
+                            "An error occurred during registration. Please try again.",
+                        )
                         return TemplateResponse(
                             request,
                             "register/register.html",


bhsauro:
@albertisfu fyi

cl/users/views.py
```diff
@@ -516,33 +517,72 @@ def register(request: HttpRequest) -> HttpResponse:
...
+                    up.save()
+
+                except IntegrityError as e:
+                    # Catch race condition on username (rare occurrence)
+                    logger.warning(
+                        "IntegrityError during registration for username=%s: %s",
+                        cd["username"],
+                        str(e),
+                        exc_info=True,
                     )
```
albertisfu: I think we can remove this warning. If the user exists, they will be redirected to the success page, which is all we need. If the user does not exist, we will log an error with more details.

cl/users/views.py
```diff
@@ -516,33 +517,72 @@ def register(request: HttpRequest) -> HttpResponse:
...
+                            reverse("register_success") + get_str
+                        )
+
+                    # Else, display generic error message and rerender form
+                    except User.DoesNotExist:
+                        logger.error(
+                            "Unexpected IntegrityError during registration: user %s does not exist after IntegrityError. "
+                            "This may indicate a constraint violation on a different field.",
+                            cd["username"],
+                        )
```
albertisfu: In this error log, we do not need the `username`. Instead, we can display `str(e)` and add `exc_info=True`.

We can also remove the line, `“This may indicate a constraint violation on a different field.”` since the issue in this case might be something else.

cl/users/views.py
```diff
@@ -516,33 +517,72 @@ def register(request: HttpRequest) -> HttpResponse:
...
+                        form.add_error(
+                            None,
```
albertisfu: Here we need to add a field to display the error; otherwise, it will not be shown. I think we can use the `username` field.

```suggestion
                            "username",
```

PR #6657, opened 2025-12-10
fix(corpus_search): prevent duplicate IDs in form fields

### LINKED ISSUE #6198: Ensure unique form field IDs to fix label focus, JS behavior, a11y, and spec compliance in new corpus search component
### Describe the issue.

Our current implementation of the corpus search in new templates is producing duplicate id attributes in the DOM, which is a clear spec violation.

### Current implementation

- We render the entire form more than once on the page, so every single field is present at least twice in the document.
- Within each form, we have multiple fieldsets (for different search scopes), and some fields are present in more than one scope (for example `case_name` and `docket_number` are present in all scopes but judges).
- Within a form, only one fieldset is active at a time; inactive ones are disabled and hidden, but the elements are still present in the DOM.

### Requirement

We need a way to ensure each input in the document has a unique ID, and that every label is correctly associated with its input.

### Why this matters

Markup is incorrect:
- Duplicate IDs are explicitly invalid HTML.

It hurts UX:
- Labels with `for` attributes don’t consistently focus their corresponding input.
- Screen readers cannot reliably map labels to inputs. This impacts a11y in searches across **all** pages, so one of CourtListener's core features could be rendered virtually unusable in some situations for a lot of people.

Maintainability is also hurt:
- `document.getElementById()` returns only the first match, breaking JS that expects a unique element.
- CSS selectors like `#field_id` apply unpredictably across duplicates. With Tailwind this is probably not too bad though.
- Tests may target the wrong element.

### Possible solutions

- Django's Forms have an `auto_id` but I think we can only tweak the id by having multiple form instances. We currently use the same one everywhere, but we could introduce per-scope forms and setting unique ids that way.
- `widget_tweaks` provides a way to override attributes in the template. This sounds like it could be the simplest? But it looks like we don't actually have it installed even though it's being loaded in a couple of templates.
- Using some sort of scope namespacing. Maybe updating the IDs whenever the scope changes, so only the inputs in the active fieldset have `id` and `for` set. We're already using JS to handle the fieldsets on scope change.
- Using the very same DOM nodes everywhere and just moving them around. Semantically this makes sense to me, but sounds cumbersome to implement.
- Something else?

### Related Issue

_No response_

### Screenshots

_No response_

To Do

elisa-a-v:
Fixes #6198 

## Summary

This PR fixes the duplicate ID bug by implementing server-side unique ID generation for corpus search form fields, with comprehensive test coverage.

## Implementation Approach

The solution uses custom Django template filters to dynamically generate unique IDs for form fields:
- `get_item`: Accesses form fields by name (dictionary-style access in templates)
- `render_field_with_id`: Renders a field with a custom ID, removing Django's default ID

IDs follow the pattern: `{form_prefix}_{namespace}_{field_name}` (e.g., `desktop_opinions_case_name`, `mobile_recap_docket_number`)

### Why not Alpine's x-id?

We opted for server-side generation because:
- IDs need to be present in the initial HTML for proper label/input associations (critical for screen readers)
- Server-side approach ensures predictable, testable IDs
- No JS dependency for core accessibility functionality

### Why not [widget_tweaks](https://pypi.org/project/django-widget-tweaks/)?

Initially attempted to use django-widget-tweaks' `render_field` to override IDs in templates, but encountered conflicts with django-cotton (it didn't work properly inside `with` blocks, and the `attr` filter couldn't accept dynamic values built by concatenating strings passed as component attributes.)

Since we need this to work within our cotton components, custom template filters were implemented instead.

## Test Coverage

**Unit tests** (`cl/custom_filters/tests.py`):
- Test the template filters in isolation with both simple forms and the actual `CorpusSearchForm`
- Verify custom IDs are applied correctly and default Django IDs are removed
- Cover all field types (text inputs, selects, checkboxes)

**Integration tests** (`cl/search/tests/test_v2_pages.py`):
- Test the rendered HTML on the actual search page
- Verify no duplicate IDs exist anywhere in the DOM
- Validate all label `for` attributes have matching input IDs (a11y check)

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


cl/assets/templates/new_base.html
```diff
@@ -1,4 +1,4 @@
-{% load static %}{% load humanize %}{% load widget_tweaks %}{% load extras %}{% load waffle_tags %}
+{% load static %}{% load humanize %}{% load extras %}{% load waffle_tags %}
```
mlissner:
```suggestion
{% load static humanize extras waffle_tags %}
```

mlissner (commented):
I have to admit this is over my head at this point, but I did see one nitpicky thing while looking through. Let's get @flp-esti to take a look!

cl/search/templates/cotton/corpus_search/filters/index.html
```diff
@@ -1,16 +1,18 @@
-{% load widget_tweaks svg_tags %}
```
ghost, line 1: Is the removed `widget_tweaks` just dead code? Just wondering because I only see additions to this file, no other code deletions, unless I missed something.

elisa-a-v: Yeah it actually is! `widget_tweaks` is a tag provided by the `django-widget-tweaks` app, which we don't even have installed 😅 so that wasn't doing anything at all!

I didn't remove it from all the other templates where it's been loaded because they're all legacy templates and it felt like it would just add noise to this PR which has a different purpose.

flp-esti (approved):

elisa-a-v:
Ignore those updates, we were testing workflow stuff with @mitch-mitchel

Issue #6660, opened 2025-12-11
Add admin feature to link IP addresses to users


mlissner:
### Describe the issue.

We currently collect IP addresses via AWS Athena logs, but they're not associated with a user. We also collect IP addresses as part of API access. 

When there's an outage, we need a trick for linking a bunch of IP addresses back to the user that is associated with them. 

Probably the best way to do this is a little UI at `/monitoring/ip-lookup/` or something like that where you can paste in a bunch of IPs and get back a bunch of usernames. Bonus points if the usernames are linked to the admin panel itself, and triple points if we link back to zoho! 

It looks like the best URL you can do with Zoho is something like this: 

https://crm.zoho.com/crm/freelawproject/search?searchword=XXX&isRelevance=false

You can't do any fielded searches, but if you just search for the user's ID it might work fine. 

### What type of issue is this?

To Do

mlissner:
I realized that if this is a website, it won't do us much good during an outage. Wah wah. 

I guess it could be a management command?

PR #6684, opened 2025-12-21
feat(stats): Add management command to look up users by IP address

mlissner:
## Fixes

Fixes: #6660

## Summary

This PR adds an `ip_lookup` management command that searches Redis API logs to find users associated with IP addresses. This is designed to work during outages since it runs via `docker exec` and only needs Redis access.

**Features:**
- Accept IPs as command arguments or via stdin (for piping from files)
- Search configurable number of days (default 14, matches Redis expiry)
- Deduplicated results (shows most recent match per IP)
- Reports IPs not found in the search period
- Output includes Django admin and Zoho CRM links

**Usage:**
```bash
# Pass IPs as arguments
docker exec cl-django python manage.py ip_lookup 192.168.1.1 10.0.0.2

# Pipe from a file
cat ips.txt | docker exec -i cl-django python manage.py ip_lookup --stdin

# Search more days
docker exec cl-django python manage.py ip_lookup --days 30 192.168.1.1
```

## Deployment

**This PR should:**

- [X] `skip-deploy` (skips everything below)

**Wiki Updates**

Once this is deployed, I need to update the wiki to explain how to use it. 

**Maintenance Pods**

Maintenance pods will need to be re-created to deploy it. I'll just do mine since I don't know how the others are in use.


mlissner:
That's fine. The maintenance pods aren't ever updated anyway.

PR #6695, opened 2025-12-24
feat(lib): Adds SQL commenter middleware to append context to queries

ERosendo:
## Fixes

This PR fixes https://github.com/freelawproject/infrastructure/issues/565

## Summary

This PR introduces a middleware component that annotates every database query with contextual comments. These comments embed useful metadata (such as the current user ID and request path) directly into the SQL sent to the database.

Before, queries executed by Django would look like:

```sql
SELECT ... FROM table WHERE ...
```

After this PR, they will look like:

```sql
SELECT ... FROM table WHERE ... /* user_id=123, url=/cases/abc123 */
```

The goal of this PR is to regain some of the execution context we previously got from the `set_config` queries added by pghistory. That context was very helpful for debugging database activity during outages.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/corpus_importer/management/commands/import_tn.py
+++ b/cl/corpus_importer/management/commands/import_tn.py
@@ -121,24 +121,23 @@ def import_tn_corpus(
             )
 
         # oci - originating_court_information is always `None`` for this import
-        docket, opinion, cluster, citations, oci = make_objects(
+        docket, opinions, cluster, citations, oci = make_objects(
             make_item(case),
             courts[case["court"]],
-            sha1_hash,
-            pdf_data,
+            [(make_item(case), pdf_data, sha1_hash)],
         )
 
         save_everything(
             items={
                 "docket": docket,
-                "opinion": opinion,
+                "opinions": opinions,
                 "cluster": cluster,
                 "citations": citations,
             }
         )
 
         extract_opinion_content.delay(
-            opinion.pk,
+            opinions[0].pk,
             ocr_available=ocr_available,
         )
         logging.info(
--- a/cl/opinion_page/forms.py
+++ b/cl/opinion_page/forms.py
@@ -487,11 +487,16 @@ class BaseCourtUploadForm(forms.Form):
         sha1_hash = sha1(force_bytes(self.cleaned_data.get("pdf_upload")))
         court = Court.objects.get(pk=self.cleaned_data.get("court_str"))
 
-        docket, opinion, cluster, citations, _ = make_objects(
+        docket, opinions, cluster, citations, _ = make_objects(
             self.cleaned_data.get("item"),
             court,
-            sha1_hash,
-            self.cleaned_data.get("pdf_upload"),
+            [
+                (
+                    self.cleaned_data.get("item"),
+                    self.cleaned_data.get("pdf_upload"),
+                    sha1_hash,
+                )
+            ],
         )
 
         if not citations:
@@ -502,7 +507,7 @@ class BaseCourtUploadForm(forms.Form):
         save_everything(
             items={
                 "docket": docket,
-                "opinion": opinion,
+                "opinions": opinions,
                 "cluster": cluster,
                 "citations": citations,
             }
@@ -517,7 +522,7 @@ class BaseCourtUploadForm(forms.Form):
             docket.originating_court_information = originating_court
             docket.save()
 
-        extract_opinion_content.delay(opinion.pk, ocr_available=True)
+        extract_opinion_content.delay(opinions[0].pk, ocr_available=True)
 
         logging.info(
             "Successfully added object cluster: %s for %s",
--- a/cl/scrapers/management/commands/cl_scrape_opinions.py
+++ b/cl/scrapers/management/commands/cl_scrape_opinions.py
@@ -55,11 +55,10 @@ cnt = CaseNameTweaker()
 def make_objects(
     item: dict[str, str | Any],
     court: Court,
-    sha1_hash: str,
-    content: bytes,
+    opinions_content: list[tuple[dict, bytes, str]],
 ) -> tuple[
     Docket,
-    Opinion,
+    list[Opinion],
     OpinionCluster,
     list[Citation],
     OriginatingCourtInformation,
@@ -75,7 +74,7 @@ def make_objects(
     plural names, like in
     `item.get("disposition") or item.get("dispositions", "")`
 
-    Returns the created objects.
+    :return: the created objects.
     """
     blocked = item["blocked_statuses"]
     if blocked:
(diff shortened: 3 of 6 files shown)


ERosendo:
@mlissner thanks for those tests. They look good to me.

cl/lib/sqlcommenter.py
```diff
@@ -0,0 +1,74 @@
...
+
+def add_sql_comment(sql: str, **meta: Any) -> str:
+    """
+    Append a SQL comment containing key=value pairs from meta.
+    """
+    if not meta:
+        return sql
+
+    comment_parts = [
+        f"{key}={value}" for key, value in meta.items() if value is not None
```
albertisfu: I performed the following test:

`/api/rest/v4/coverage/*scotus/`

I wanted to see what would happen if the URL already included the `/*` comment syntax.

I got this error:

```
unterminated /* comment at or near "/* user_id=8, url=/api/rest/v4/coverage/*scotus/ */"
LINE 1: ...t" WHERE "search_court"."id" = '*scotus' LIMIT 21 /* user_id...
```

Separately from this error, I’m also wondering whether this could be a security issue, specifically, if someone could use it to terminate the comment early and attempt SQL injection?

Can we escape `/*` and `*/` in the `meta.items` values (for example, by replacing them with `/ *` and `* /`)?

cl/lib/sqlcommenter.py
```diff
@@ -0,0 +1,74 @@
...
+        annotation.
+        """
+        user = (
+            self.request.user._meta.pk.get_db_prep_value(
+                self.request.user.pk, context["connection"]
+            )
+            if hasattr(self.request, "user")
+            and hasattr(self.request.user, "_meta")
+            else None
+        )
```
albertisfu, line 87: Would it be possible to do something simpler, like this?

```suggestion
        user = (
        self.request.user.pk
        if hasattr(self.request, "user") and self.request.user.is_authenticated
        else None
    )
```

cl/lib/sqlcommenter.py
```diff
@@ -0,0 +1,74 @@
...
+        return {"user_id": user, "url": self.request.path}
+
+    def __call__(self, execute, sql, params, many, context):
+        context_info = self.get_context(context)
```
albertisfu: If there are multiple queries within the same request, `self.get_context(context)` is executed for each query, even though the `user` and `request.path` are always the same, correct?

Can we cache the return value of the `get_context` method so it is computed only once per request, regardless of the number of database queries?

cl/lib/sqlcommenter.py
mlissner: I had claude compare sqlcommenter to this PR. Lots of suggestions and it includes an exploit: https://gist.github.com/mlissner/7f3ea13e0c38f75fe63fc71a86a36948

cl/lib/sqlcommenter.py
ERosendo: @albertisfu Yeah, I think the output should be equivalent. As mentioned in the docstring, this method mirrors the behavior of the `get_context` method in [`pghistory.middleware.HistoryMiddleware`](https://github.com/AmbitionEng/django-pghistory/blob/d5b516fc882cd412042f18b17d71e5e2c4dbf047/pghistory/middleware.py#L52-L58)

cl/lib/sqlcommenter.py
ERosendo: > If there are multiple queries within the same request, self.get_context(context) is executed for each query, even though the user and request.path are always the same, correct?

You’re correct, `get_context` is executed for each query, even if `user` and `request` don’t change.

> Can we cache the return value of the get_context method so it is computed only once per request, regardless of the number of database queries?

I don’t think we’re doing anything particularly expensive in `get_context`. The request object is already in memory while the request is active, and we’re not hitting the database each time we access `user.pk` (i think). Do you think this code could actually slow down responses?

I’ll give `@cached_property` a try.

cl/lib/sqlcommenter.py
ERosendo: Thanks both for your comments and catching this issue. I’ve introduced a new helper, `escape_sql_comment_value`, to properly sanitize the values we add to the SQL comment, following the suggestions from the gist.

While reviewing Alberto’s example, I noticed that a URL like that would resolve to an invalid view, but it can still trigger the issue because some middleware executes small database queries before URL resolution completes. I also noticed something I initially overlooked when reviewing the [sqlcommenter](https://github.com/google/sqlcommenter/) implementation: they annotate queries using `request.resolver_match.route `instead of `request.path`.

The key difference is that `resolver_match` is only set after URL resolution has completed. That means it’s always available inside views, but not in middleware that runs before URL resolving takes place. I decided to use this as a validation and now only include the request path when `resolver_match` is present (`self.request.path if resolver_match else None`).

This has two implications:
1. Our code only add the request path to queries triggered inside views (it will still add the user_id to middleware queries)
2. If the URL does not resolve to a valid view, we won’t attempt to include an invalid path in the SQL comment.

I chose to continue using `request.path` (rather than switching to `resolver_match.route`) because request.path reflects the actual URL the user requested (e.g. `users/3/`), whereas `resolver_match.route` returns the URL pattern (e.g. `users/<id>/`).

Let me know what you think.

cl/lib/sqlcommenter.py
albertisfu: I think this was adding only minimal overhead, but thank you for applying `cached_property`.

cl/lib/sqlcommenter.py
```diff
@@ -0,0 +1,101 @@
...
+        user = (
+            self.request.user.pk
+            if hasattr(self.request, "user")
+            and self.request.user.is_authenticated
+            else None
+        )
+        resolver_match = self.request.resolver_match
+        return {
+            "user_id": user,
+            "url": self.request.path if resolver_match else None,
```
albertisfu: I was thinking on that URLs can be huge like:

`/docket/2675000/r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33vr-s-v-east-brunswick-township-school-district-lorem-3399r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33vr-s-v-east-brunswick-township-school-district-lorem-3399r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33vr-s-v-east-brunswick-township-school-district-lorem-3399r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33vr-s-v-east-brunswick-township-school-district-lorem-3399r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33vr-s-v-east-brunswick-township-school-district-lorem-3399r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33-r-s-v-east-brunswick-township-school-district-lorem-33vr-s-v-east-brunswick-township-school-district-lorem-3399/
`

And even larger.

I was wondering whether this could cause a performance issue. In any case, since we now have the `url_name`, maybe we can truncate URLs to a limited length?

albertisfu (commented):
Thanks, @ERosendo. The changes look great. I also confirmed that the vulnerability issues are resolved by the new escape method that was introduced.

I just have one additional comment.

cl/lib/sqlcommenter.py
ERosendo: @albertisfu i don’t think this would cause a performance issue for the query itself. However, some tools (like RDS Performance Insights) have limits on how much of a query they display, so very long URLs could result in truncated queries when debugging.

 I agree that truncating the URL to a reasonable length could make sense. Do you have a specific character limit in mind?

ERosendo: We could also make the maximum path length configurable via a setting (with default, e.g. 255 characters), so we can adjust it if needed without changing code.

let me know what you think

albertisfu: Yeah I think a setting is a good idea. A default of 255 characters seems reasonable.

ERosendo: Great. I'll add the new setting 👍

PR #6709, opened 2025-12-31
6536 Ingest SCOTUS PDF documents

albertisfu:
## Fixes

Fixes: #6536

## Summary

This PR introduces a new `ingest_scotus_docket` method which will be used by our SCOTUS importer/scraper.

It triggers a Celery chain that includes the following tasks:

- **process_scotus_docket**: Merges the SCOTUS docket along with its docket entries and document instances into the database. If a Questions Presented file is available, it downloads it. Returns a list of RECAPDocuments that need to be downloaded in a subsequent task.
- **merge_scotus_documents**: Retrieves the RECAPDocument instances that need to be downloaded, then attempts to download each PDF using streaming and a temp file to reduce the memory footprint. Finally, the RECAPDocument metadata is updated and persisted in the database. Returns a list of PDFs that require extraction.

After `merge_scotus_documents` completes, the `extract_recap_pdf` task is executed to extract text from all the downloaded PDFs.

A new `download_pdf_in_stream` method was introduced to handle downloading PDF documents via streaming and storing them in a temporary file. This allows us to retry each PDF download independently if one of them raises an error, instead of retrying the entire Celery task.

`merge_scotus_docket` was modified to return a list of RECAPDocuments that need to be downloaded—specifically those that have a `document_url` but are not yet available.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

This PR requires a migration to be deployed, since a new `document_url` field that stores the original PDF source was added to the `RECAPDocument` model.
This field has a default empty string value to backfill existing rows, so it will require following similar steps as the migration applied in #6573.

However, because RECAPDocument is a large table, we may need to take additional precautions during the migration.

### DIFF
--- a/cl/corpus_importer/tasks.py
+++ b/cl/corpus_importer/tasks.py
@@ -15,6 +15,7 @@ from tempfile import NamedTemporaryFile
 from typing import IO, Any
 from urllib.parse import urljoin
 
+import botocore.exceptions
 import environ
 import eyecite
 import internetarchive as ia
@@ -55,6 +56,11 @@ from juriscraper.pacer import (
     ShowCaseDocApi,
 )
 from juriscraper.pacer.reports import BaseReport
+from juriscraper.scotus import (
+    SCOTUSDocketReport,
+    SCOTUSDocketReportHTM,
+    SCOTUSDocketReportHTML,
+)
 from juriscraper.state.texas import (
     TexasCaseEvent,
     TexasCaseParty,
@@ -138,6 +144,7 @@ from cl.lib.recap_utils import (
     get_document_filename,
 )
 from cl.lib.redis_utils import delete_redis_semaphore, get_redis_interface
+from cl.lib.storage import AWSMediaStorage
 from cl.lib.types import TaskData
 from cl.people_db.models import Attorney, Role
 from cl.recap.constants import CR_2017, CR_OLD, CV_2017, CV_2020, CV_OLD
@@ -3513,11 +3520,11 @@ def merge_scotus_docket_entry(
         de.date_filed = date_filed
         de.save()
 
-    # Merge attachments
-    attachments = input_docket_entry["attachments"]
-    for document in attachments:
-        merge_scotus_document(de, document)
-    return de_created, de.pk
+        # Merge attachments
+        attachments = input_docket_entry["attachments"]
+        for document in attachments:
+            merge_scotus_document(de, document)
+        return de_created, de.pk
 
 
 def add_scotus_docket_entries(
@@ -3762,6 +3769,48 @@ def download_scotus_document_pdf(self: Task, doc_pk: int) -> int | None:
         )
         doc.save()
         return doc_pk
+
+
+@app.task(
+    bind=True,
+    autoretry_for=(
+        botocore.exceptions.HTTPClientError,
+        botocore.exceptions.ConnectionError,
+    ),
+    max_retries=5,
+    retry_backoff=10,
+    ignore_result=True,
+)
+def download_and_parse_scotus_docket(
+    self: Task, bucket: str, s3_key: str
+) -> dict[str, Any] | None:
+    """Download a SCOTUS docket file from S3 and parse it.
+
+    :param self: The Celery task instance.
+    :param bucket: The S3 bucket name.
+    :param s3_key: The S3 key for the docket file.
+    :return: Parsed docket data dict.
+    """
+
+    logger.info("Downloading and parsing SCOTUS file: %s.", s3_key)
+    storage = AWSMediaStorage(bucket_name=bucket)
+    with storage.open(s3_key, "rb") as f:
+        content = f.read().decode("utf-8")
+
+    ext = s3_key.rsplit(".", 1)[-1].lower()
+    parser_map: dict[str, type] = {
+        "json": SCOTUSDocketReport,
+        "html": SCOTUSDocketReportHTML,
+        "htm": SCOTUSDocketReportHTM,
+    }
+    if not (parser_cls := parser_map.get(ext)):
+        logger.error("Unsupported file extension: %s for %s", ext, s3_key)
+        self.request.chain = None
+        return None
+
+    parser = parser_cls()
+    parser._parse_text(content)
+    return parser.data
 
 
 def ingest_scotus_docket(docket_data: dict[str, Any]) -> None:
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -82,3 +82,4 @@ jobs:
           cl/search/docket_number_cleaner.py \
           cl/search/management/commands/clean_docket_number_raw.py \
           cl/scrapers/management/commands/back_scrape_dockets.py \
+          cl/corpus_importer/management/commands/import_scotus_dockets.py
--- a/cl/corpus_importer/management/commands/import_scotus_dockets.py
+++ b/cl/corpus_importer/management/commands/import_scotus_dockets.py
@@ -1,0 +1,127 @@
+import csv
+import time
+
+from celery import chain
+from django.conf import settings
+
+from cl.corpus_importer.tasks import (
+    download_and_parse_scotus_docket,
+    process_scotus_docket,
+)
+from cl.lib.celery_utils import CeleryThrottle
+from cl.lib.command_utils import VerboseCommand, logger
+from cl.lib.indexing_utils import (
+    get_last_parent_document_id_processed,
+    log_last_document_indexed,
+)
+
+
+def compose_redis_key() -> str:
+    """Compose a Redis key for SCOTUS importer log.
+    :return: A Redis key as a string.
+    """
+    return "scotus_docket_import:log"
+
+
+class Command(VerboseCommand):
+    help = "Import SCOTUS dockets from S3 using an inventory CSV."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--inventory-file",
+            required=True,
+            help="Path to the inventory CSV relative to MEDIA_ROOT.",
+        )
+        parser.add_argument(
+            "--retrieval-queue",
+            default="celery",
+            help="Which celery queue to use for S3 retrieval",
+        )
+        parser.add_argument(
+            "--ingesting-queue",
+            default="celery",
+            help="Which celery queue to use for DB ingesting",
+        )
+        parser.add_argument(
+            "--throttle-min-items",
+            type=int,
+            default=5,
+            help="CeleryThrottle min_items parameter.",
+        )
+        parser.add_argument(
+            "--delay",
+            type=float,
+            default=1.0,
+            help="Seconds to sleep between scheduling tasks.",
+        )
+        parser.add_argument(
+            "--start-row",
+            type=int,
+            default=0,
+            help="Row number to start from (for manual resume).",
+        )
+        parser.add_argument(
+            "--inventory-rows",
+            type=int,
+            required=True,
+            help="Total number of rows in the inventory CSV. Used to "
+            "log progress percentage.",
+        )
+        parser.add_argument(
+            "--auto-resume",
+            action="store_true",
+            default=False,
+            help="Resume from last row stored in Redis.",
+        )
+
+    def handle(self, *args, **options):
+        super().handle(*args, **options)
+
+        retrieval_queue = options["retrieval_queue"]
+        ingesting_queue = options["ingesting_queue"]
+        delay = options["delay"]
+        inventory_rows = options["inventory_rows"]
+        inventory_path = settings.MEDIA_ROOT / options["inventory_file"]
+
+        start_row = options["start_row"]
+        if options["auto_resume"]:
+            start_row = get_last_parent_document_id_processed(
+                compose_redis_key()
+            )
+            logger.info("Auto-resuming from row %s.", start_row)
+
+        throttle = CeleryThrottle(
+            min_items=options["throttle_min_items"],
+            queue_name=ingesting_queue,
+        )
+
+        with open(inventory_path) as f:
+            reader = csv.reader(f)
+            for row_idx, row in enumerate(reader):
+                if row_idx < start_row:
+                    # Skip row if auto-resume is enabled
+                    continue
+
+                bucket = row[0].strip()
+                s3_key = row[1].strip()
+
+                throttle.maybe_wait()
+                chain(
+                    download_and_parse_scotus_docket.si(bucket, s3_key).set(
+                        queue=retrieval_queue
+                    ),
+                    process_scotus_docket.s().set(queue=ingesting_queue),
+                ).apply_async()
+                time.sleep(delay)
+
+                if row_idx % 100 == 0:
+                    progress = f" ({row_idx / inventory_rows:.1%})"
+                    logger.info(
+                        "Scheduled %s rows %s. Current row: %s.",
+                        row_idx,
+                        progress,
+                        s3_key,
+                    )
+                    log_last_document_indexed(row_idx, compose_redis_key())
+
+        logger.info("Finished scheduling all rows from inventory.")


cl/search/models.py
```diff
@@ -1361,6 +1361,12 @@ class RECAPDocument(
         max_length=64,
         blank=True,
     )
+    document_url = models.URLField(
+        help_text="The URL of the file from the original source",
+        max_length=1000,
+        blank=True,
+        default="",
+    )
```
mlissner: For PACER docs, we have the `pacer_url` class property. Could we accomplish that here without storing the URLs in the DB?

albertisfu: Yeah. The problem is that SCOTUS document URLs are complex. For instance:
```
http://www.supremecourt.gov/DocketPDF/24/24A1/316143/20240701161149036_2024.07.01%20Taylor%20Application%20for%20Extension%20of%20Time.pdf

http://www.supremecourt.gov/DocketPDF/24/24A100/319653/20240723152722701_Johnson%20cert%20extension.pdf

https://www.supremecourt.gov/DocketPDF/21/21-1397/247399/20221123131814988_42997%20pdf%20Ebner.pdf
```

The URL is common only up to `https://www.supremecourt.gov/DocketPDF/`.

We can form the URL using the docket year, the docket number, and the document number, but we still need the file name: `20221123131814988_42997%20pdf%20Ebner.pdf`.

One option is to store the file name in the `pacer_doc_id` field so we can form the URL, but I think that would be messy, and it’s limited to 64 characters.

So that's the reason I introduced this new field. 

What do you think?

albertisfu:
> Do you want somebody on your team to look at this or is it better to ask Eduardo?

I can assign Brennan to review this one.

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3282,120 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+        scotus_url = rd.document_url
+        rd_id = rd.pk
+        if rd.is_available:
             logger.warning(
-                "SCOTUS PDF download: Unable to download %s for docket %s. "
-                "Exception was: %s",
-                qp_url,
-                docket_id,
-                str(exc),
+                "The SCOTUS document for % is already available; skipping.",
```
Brennan-Chesley-FLP:
```suggestion
                "The SCOTUS document for %s is already available; skipping.",
```

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3282,120 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+                rd_id,
             )
             return None
```
Brennan-Chesley-FLP, line 3868: Should this be a return or a continue?

cl/corpus_importer/tasks.py
```diff
@@ -3179,13 +3182,60 @@ def classify_case_name_by_llm(self, cluster_pk: int, recap_document_id: int):
...
+            headers={"User-Agent": "Free Law Project"},
+        ) as response:
+            response.raise_for_status()
+            if not is_pdf(response):
+                logger.warning(
+                    "PDF download: Expected application/pdf for %s from %s; aborting.",
+                    identifier,
+                    url,
+                )
+                return False
```
Brennan-Chesley-FLP: What is the workflow we use right now for checking out these warnings?

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3282,120 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+                rd.document_number,
+                rd.attachment_number,
+            )
+            rd.filepath_local.save(file_name, File(tmp), save=False)
+            rd.file_size = rd.filepath_local.size
+            rd.is_available = True
+            rd.date_upload = rd.date_upload or now()
+
+            tmp.seek(0)
+            rd.sha1 = sha1(tmp.read())
```
Brennan-Chesley-FLP: We can chunk this read to avoid pulling it all into memory (or propagate it from download_to_file).

Brennan-Chesley-FLP:
Nice! Looks good on the whole. A couple of Qs and Ss.

cl/corpus_importer/tasks.py
albertisfu: Yeah. This was wrong. I've refactored this whole method.

cl/corpus_importer/tasks.py
albertisfu: Sure. I went with the `download_to_file` propagation approach.

cl/corpus_importer/tasks.py
albertisfu: Currently we can use Loki in Grafana, where we can filter and search logs.

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+def enrich_scotus_attachments(docket_entries: list[dict[str, Any]]) -> None:
+    """Add sequential attachment numbers and pacer_doc_id to SCOTUS attachments.
+
+    :param docket_entries: A list of docket entry dictionaries.
+    :return: None. The input list is changed in place.
+    """
+
+    for entry in docket_entries:
+        main_doc_short_desc = ""
+        attachments = entry.get("attachments") or []
```
MorganBennetDev:
```suggestion
        attachments = entry.setdefault("attachments", [])
        main_doc_short_desc = attachments[0]["description"] if len(attachments) > 0 else ""
        entry["pacer_doc_id"] = ""
```

We can use the `setdefault` method to concisely set `entry["attachments"]` if it is not set and return its value, which gets rid of the `if` setting `entry["attachments"]` at the end of the loop.

Additionally, if we move the `attachments` declaration before the `main_doc_short_desc` declaration, we can move the `if` out of the inner loop.

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+        entry["pacer_doc_id"] = ""
+
+        for idx, attachment in enumerate(attachments, start=1):
+            if idx == 1:
+                main_doc_short_desc = attachment["description"]
```
MorganBennetDev:
```suggestion
```

We can now delete this...

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+            attachment["attachment_number"] = idx
+            attachment["pacer_doc_id"] = ""
+
+        if not attachments:
+            entry["attachments"] = []
```
MorganBennetDev:
```suggestion
```

...and this.

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+                check_if_needed=False, model_name="search.SCOTUSDocument"
+            ),
+        ).apply_async()
+        return True, scotus_document.pk
+    return False, scotus_document.pk
+
+
+def scotus_docket_entry_sequence_numbers(
+    docket_entries: list[dict[str, Any]],
+) -> list[str]:
```
MorganBennetDev: Per [this review comment](https://github.com/freelawproject/courtlistener/pull/6761#discussion_r2717342700) this logic should be in `corpus_importer/utils.py`. I've already moved it there in my PR, so if you don't see any issues with the setup there, you can copy it to this PR.

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+                logger.error(
+                    "Multiple matching SCOTUSDocketEntries found for entry_number "
+                    "%s on Docket %s.",
+                    entry_number,
+                    docket.pk,
                 )
-            logger.info(
-                "SCOTUS PDF download: Stored Questions Presented PDF for docket %s.",
-                docket_id,
+                return False, None
```
MorganBennetDev, line 3474: Is there a DB performance benefit to using `get` versus using `count` and `first`? I'm not familiar with how Django actually maps these operations to queries, but AFAIK exceptions tend to be expensive to handle.

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+            if count == 0:
+                des_seq = SCOTUSDocketEntry.objects.filter(
+                    docket=docket,
+                    sequence_number=sequence_number,
+                )
+                if des_seq.count() == 0:
+                    de = SCOTUSDocketEntry(
+                        docket=docket,
+                        date_filed=date_filed,
+                        entry_number=entry_number,
```
MorganBennetDev: `entry_number` will be `None` in this branch so there's no need to pass it.

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
             )
-    except RequestException as exc:
-        if self.request.retries == self.max_retries:
+            count = des.count()
+            if count == 0:
+                des_seq = SCOTUSDocketEntry.objects.filter(
+                    docket=docket,
+                    sequence_number=sequence_number,
+                )
+                if des_seq.count() == 0:
```
MorganBennetDev: There is no `else` branch setting `de` or returning early so we can reach line 3469 without it being set, causing an `AttributeError` exception.

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+    for sequence_number, docket_entry in zip(
+        sequence_numbers, docket_entries, strict=True
+    ):
+        de_created, de_pk = merge_scotus_docket_entry(
+            docket,
+            sequence_number,
+            docket_entry,
+        )
+
+        if not de_pk:
```
MorganBennetDev: It doesn't seem that Postgres can generate primary keys equal to 0 ([StackOverflow](https://stackoverflow.com/questions/32728179/can-a-primary-key-in-postgres-have-zero-value)), but I'd rather just have

```suggestion
        if de_pk is None:
```

since that will only ever match when `merge_scotus_docket_entry` fails.

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+    """
+    with transaction.atomic():
+        court = Court.objects.get(pk="scotus")
+        docket_number = report_data["docket_number"]
+        if not docket_number:
+            raise ValueError(
+                "Docket number can't be missing in SCOTUS dockets."
+            )
+
+        case_name = report_data.get("case_name") or ""
```
MorganBennetDev:
```suggestion
        case_name = report_data.get("case_name", "")
```

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+        lower_court_name = report_data.get("lower_court")
+
+        d = async_to_sync(find_docket_object)(
+            court.pk,
+            None,
+            docket_number,
+            None,
+            None,
+            None,
+        )
```
MorganBennetDev, line 3595: Should we acquire a lock on `d` here or is that unnecessary?

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+        d.save()
+
+        # Merge ScotusDocketMetadata
+        defaults = {
+            "capital_case": bool(report_data.get("capital_case")),
+            "date_discretionary_court_decision": report_data.get(
+                "discretionary_court_decision"
+            ),
+        }
+        if links := report_data.get("links"):
```
MorganBennetDev: Are we checking for empty string, `None`, or both here?

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+) -> tuple[Docket, bool]:
+    """Merge SCOTUS docket data into a Docket and ScotusDocketMetadata.
+
+    This will create or update the Docket row for the SCOTUS and
+    then create or update the related ScotusDocketMetadata instance.
+
+    :param report_data: A dictionary containing parsed SCOTUS docket data.
+    :return: A three-tuple: the created or updated Docket instance, whether the
+    QP file should be downloaded, and the RECAPDocument instances that should
+    be downloaded.
```
MorganBennetDev: Currently returns a two-tuple.

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+
+        if qp_url := report_data.get("questions_presented"):
+            defaults["questions_presented_url"] = qp_url
+
+        scotus_metadata, _ = ScotusDocketMetadata.objects.update_or_create(
+            docket=d,
+            defaults=defaults,
+        )
+        download_qp = bool(
+            qp_url and not scotus_metadata.questions_presented_file
```
MorganBennetDev, line 3685: The `not scotus_metadata.questions_presented_file` condition should always evaluate to `True` if an object was created by `ScotusDocketMetada.objects.update_or_create` so we could use the second output of that method in this check instead. It will also succeed if `download_qp_scotus_pdf` fails and the merger is run again at a later time. Is that retry logic what this is enabling?

cl/corpus_importer/tasks.py
```diff
@@ -3232,54 +3292,430 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
-                )
-                return None
-            with NamedTemporaryFile(prefix="scotus_qp_", suffix=".pdf") as tmp:
-                # Download the PDF into a tmp file to avoid using too much memory
-                for chunk in response.iter_content(chunk_size=8 * 1024):
-                    if chunk:
-                        tmp.write(chunk)
+    with download_pdf_in_stream(qp_url, docket_id, "scotus_qp_") as result:
+        if result is None:
+            return None
```
MorganBennetDev, line 3345: Should we log something here?

cl/lib/microservice_utils.py
```diff
@@ -60,7 +60,7 @@ async def clean_up_recap_document_file(item: RECAPDocument) -> None:
 async def microservice(
```
MorganBennetDev, line 62: What is your opinion on just checking for instances of `AbstractPDF` here? I have an implementation in [#6761](https://github.com/freelawproject/courtlistener/pull/6761/changes#diff-cdbfdcd1ad38aeaa745d254f6108069c2c48a860dedf97b8cfe3c4b4ab9ba31e), but we can do whatever you think is best.

cl/corpus_importer/tasks.py
albertisfu: Sure, change applied

cl/corpus_importer/tasks.py
albertisfu: Well, if we use `get`, we only need a single DB operation to retrieve the instance or to know if it doesn’t exist or if there is more than one.

On the other hand, if we use `filter()` with `count()` and `first()`, we would need two DB operations.

I’ve refactored the code below, which was still using filters, to use the `get` approach.

cl/corpus_importer/tasks.py
albertisfu: removed

cl/corpus_importer/tasks.py
albertisfu: I've refactored this block of code. This case should be covered now.

cl/corpus_importer/tasks.py
albertisfu: done

cl/corpus_importer/tasks.py
albertisfu: Here we're checking for empty strings. I've applied a refactor to return always an empty string instead of `None`

cl/corpus_importer/tasks.py
albertisfu: updated

cl/corpus_importer/tasks.py
albertisfu: Do you mean that we can simply check whether the `scotus_metadata` instance was created to decide whether to download the file?

My concern is that if `download_qp_scotus_pdf` fails, the `scotus_metadata` instance would still be created, but `questions_presented_file` would be None. In that case, during subsequent merges, the `questions_presented_file` would not be downloaded.

Am I missing something?

cl/corpus_importer/tasks.py
albertisfu: Sure, added a log.

cl/lib/microservice_utils.py
albertisfu: Yeah it's a good idea. I've applied your implementation.

albertisfu:
Thanks @MorganBennetDev I've applied the suggested changes and left a couple of comments. Let me know what do you think.

cl/corpus_importer/tasks.py
MorganBennetDev: Sorry for the bad phrasing, but yes I was trying to ask if the retry is intended here. It seems so, so I'll resolve this.

albertisfu:
@MorganBennetDev I’ve applied a small tweak to assign the `SCOTUSDocument.file_size` without making an HTTP request. Previously, it was fetched from S3; now it’s taken directly from the local file.

In addition, I’ve enabled the admin interfaces for the `SCOTUSDocument` and `SCOTUSDocketEntry` models.

albertisfu:
Solved the merge conflicts and applied a small tweak regarding relatives `qp_url`.

cl/search/models.py
```diff
@@ -4160,3 +4160,25 @@ class Meta:
...
+    def needs_extraction(self):
+        """Does the item need extraction and does it have all the right
+        fields? Items needing OCR still need extraction.
+        """
+        return all(
+            [
+                self.ocr_status is None or self.ocr_status == self.OCR_NEEDED,
+                self.filepath_local,
+            ]
+        )
```
MorganBennetDev: Using `all` here doesn't read as clearly to me as `and` since it's such a short condition.

cl/corpus_importer/tasks.py
```diff
@@ -3319,57 +3326,445 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
         )
-        raise self.retry(exc=exc)
+        downloaded_file = File(tmp)
+        doc.filepath_local.save(filename, downloaded_file, save=False)
+        doc.file_size = downloaded_file.size
+        doc.sha1 = sha1_hash
+        response = async_to_sync(doc_page_count_service)(doc)
+        if response.is_success:
+            doc.page_count = int(response.text)
+        assert isinstance(doc.page_count, (int | type(None))), (
```
MorganBennetDev, line 3767: Can this condition ever occur?

cl/corpus_importer/tasks.py
```diff
@@ -3319,57 +3326,445 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+            None,
+            docket_number,
+            None,
+            None,
+            None,
+        )
+
+        d.source = Docket.SCRAPER
+        d.docket_number = docket_number
+        d.docket_number_raw = docket_number
```
MorganBennetDev, line 3605: We should also set `docket_number_core` here if it's not set since `find_docket_object` does not.

albertisfu: The `docket_number_core` assignment is handled in the Docket.save() method.

cl/corpus_importer/tasks.py
albertisfu: `response.is_success` condition?

Yeah, sometimes doctor can return an error.

cl/search/models.py
albertisfu: done!

albertisfu:
Thanks @MorganBennetDev I've addressed your suggestion and answer your questions.

cl/corpus_importer/tasks.py
MorganBennetDev: Sorry, the highlight was broader than I thought. I meant can the assertion ever fail.

albertisfu:
@MorganBennetDev pushed an additional tweak regarding the correct handling of adding the source. Previously, the `SCRAPER` source was always added, even for existing dockets, which is not correct. For dockets created by other sources, the SCRAPER source should only be appended to the existing source.

I think this is also a change we should take into account for the Texas merger.

cl/corpus_importer/tasks.py
```diff
@@ -3585,7 +3585,11 @@ def merge_scotus_docket(
             None,
         )
 
+        # Add SCRAPER source by default if it's a new Docket.
         d.source = Docket.SCRAPER
+        if d.pk:
+            # If the docket already exists, compound the SCRAPER source.
+            d.add_scraper_source()
```
MorganBennetDev: This seems to be the source of the test failures.

cl/corpus_importer/tasks.py
albertisfu: Yeah, this is just a sanity check in case Doctor starts doing weird things and returns a string instead of the count or None.

cl/corpus_importer/tasks.py
albertisfu: Yeah the assignment was incorrect. I've pushed the fix.

albertisfu:
Ok this is ready to go. Merging...

Issue #6534, opened 2025-11-17
Text Similarity Threshold Adjustment for Opinion Versioning


Luis-manzur:
### Describe the issue.

Implement and calibrate text similarity comparison for opinions that pass the metadata matching criteria (court, date, case name). Since document URLs change, the text content may differ slightly (corrections, formatting, OCR improvements), requiring a relaxed similarity threshold compared to same-URL versioning.

**New Requirements**
Relaxed threshold for different-URL versions
Justification: Updates may include:
Typo corrections
Formatting changes
Added metadata (headers, footers)
Redactions or corrections
Target: Catch legitimate updates while avoiding false positives

```# Configuration values to test
SIMILARITY_THRESHOLDS = {
    'same_url': {
        'min_similarity': 0.95,  # Very high - expect near-identical
        'description': 'Opinions with same download_url (current behavior)'
    },
    'different_url': {
        'min_similarity': 0.85,  # RELAXED - allow for updates
        'description': 'Opinions with different URLs (new behavior)'
    },
    'manual_review': {
        'min_similarity': 0.75,  # Flag for human review
        'description': 'Borderline cases needing manual verification'
    }
}```

### Related Issue

_No response_

### Screenshots

_No response_

### What type of issue is this?

Research

flooie:
This may be harder than you think.  Why dont you run some tests on the relaxing standard you want and bring back the results. It would be good to get a lot of these dupes merged.

Luis-manzur:
## Summary

The analysis of 200 real opinion pairs from production shows that **the current threshold system (0.9/0.7) is performing excellently** with 100% precision and 100% recall.

---

## Key Results

### Performance Metrics ✅

| Metric | Value | Target | Status |
|--------|-------|--------|--------|
| **Precision** | **100%** | >99% | ✅ Exceeds |
| **Recall** | **100%** | >95% | ✅ Exceeds |
| **F1 Score** | **1.000** | >0.97 | ✅ Exceeds |
| **False Positive Rate** | **0%** | <1% | ✅ Exceeds |

### Dataset Breakdown

- **200 pairs analyzed** (100 same-URL + 100 different-URL)
- **86 pairs should merge** (ground truth: yes)
- **93 pairs should not merge** (ground truth: no)
- **21 pairs need manual review** (ground truth: unsure)

### Confidence Distribution

- **83.5% high confidence** labels (167 pairs)
- **14% medium confidence** (28 pairs)
[...]

PR #6715, opened 2026-01-05
feat: Add length ratio validation to opinion merging #6534

Luis-manzur:
## Summary

 Add length ratio validation to opinion merging. It adds a pre-check layer  before text similarity calculations to against data quality issues such as incomplete text extraction, URL reuse by courts, or encoding issues.

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


cl/scrapers/management/commands/merge_opinion_versions.py
```diff
@@ -386,7 +391,32 @@ def clean_opinion_text(opinion: Opinion) -> str:
...
+    :param text2: second text to compare
+    :param min_ratio: minimum acceptable ratio of shorter to longer text.
+        Default 0.5 means texts must be within 2x length of each other.
+    :return: tuple of (passes_check, length_ratio)
+    """
+    len1, len2 = len(text1), len(text2)
+
+    # Handle edge cases
+    if len1 == 0 and len2 == 0:
+        return True, 1.0
```
grossir: I think this should return False. Depending on how we define this function it shouldn't validate "empty" texts? we risk grouping together stuff that's empty due to some extraction problem

cl/scrapers/management/commands/merge_opinion_versions.py
```diff
@@ -43,6 +43,11 @@
 
 MIN_SEQUENCE_SIMILARITY_STRICT = 0.9
 MIN_SEQUENCE_SIMILARITY_LOOSE = 0.7
+
+# Length ratio threshold to prevent merging texts with vastly different lengths
+# A ratio of 0.5 means texts must be within 2x length of each other
+MIN_LENGTH_RATIO = 0.5
```
grossir: I would expect a higher ratio to work, something like 0.70

From versions from different courts, the final versions usually have
- more headings: like the official citation on every page
- less disclaimers: no more "this is a slip opinion", "it will be published on X Reports"
- small edits
- maybe a few extra pages at the beggining or end: syllabus, footnotes, etc

grossir (commented):
Looks generally good, and it's a good idea. Will refactor this a bit and merge it

PR #6717, opened 2026-01-05
feat(cache): Move ES opinion cache keys to S3 Express (#6419)

mitch-mitchel:
## Fixes

  Fixes: #6419

## Summary

This PR moves Elasticsearch opinion cache keys to S3 Express in production while maintaining database cache in development/testing environments.

  Changes include:
  - Added cl/lib/s3_cache.py with two helper functions:
    - get_s3_cache(): Returns S3 cache in production, falls back to db_cache in dev/test
    - make_s3_cache_key(): Adds time-based prefix (e.g., "7-days:") to cache keys in production for better organization by TTL
  - Updated cl/opinion_page/utils.py to use these helpers for all ES-related caching (related clusters, cited clusters, citing clusters, case counts)
  - Added comprehensive tests for the new helper functions
 
## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

 1. No special deployment steps required. Standard deploy will pick up the changes.

<details closed>
<details open>

</details>
<details open>

</details>

### DIFF
--- a/cl/settings/third_party/rest_framework.py
+++ b/cl/settings/third_party/rest_framework.py
@@ -162,6 +162,7 @@ REST_FRAMEWORK = {
         "arivdc": "10/hour",
         "Mpits003": "1/hour",
         # Throttling up.
+        "hmturner08": "15000/hour",  # Tryandai needs big boost through 1/19
         "JonasHappel": "10000/hour",
         "YFIN": "430000/day",
         "mlissner": "1000000/hour",  # Needed for benchmarking (not greed)


mitch-mitchel:
came up when running CI tests - this was Claude suggestion to fix failing test

```
The test at line 2921 does:
  stored_email = EmailSent.objects.all()  # No ordering!

  Then at line 2935-2937 it compares:
  self.assertEqual(
      failed_email_enqueued[0].stored_email.pk, stored_email[0].pk
  )

  Without explicit ordering, EmailSent.objects.all() returns objects in undefined order. When tests run in isolation it might work, but the 56 vs 57 PK difference suggests database state from other tests is affecting the ordering.

  The fix would be to add .order_by('pk') to make ordering deterministic:
  stored_email = EmailSent.objects.all().order_by('pk')
  ```

**Comment by mitch-mitchel** [comment]
> came up when running CI tests - this was Claude suggestion to fix failing test
> 
> ```
> The test at line 2921 does:
>   stored_email = EmailSent.objects.all()  # No ordering!
> 
>   Then at line 2935-2937 it compares:
>   self.assertEqual(
>       failed_email_enqueued[0].stored_email.pk, stored_email[0].pk
>   )
> 
>   Without explicit ordering, EmailSent.objects.all() returns objects in undefined order. When tests run in isolation it might work, but the 56 vs 57 PK difference suggests database state from other tests is affecting the ordering.
> 
>   The fix would be to add .order_by('pk') to make ordering deterministic:
>   stored_email = EmailSent.objects.all().order_by('pk')
> ```

tested fix in latest commit, and seemed to fix the issue on CI @albertisfu

File: cl/lib/s3_cache.py [hunk 1]
```diff
@@ -0,0 +1,38 @@
...
+
+    :param base_key: The base cache key (e.g., "clusters-mlt-es:123")
+    :param timeout_seconds: Cache timeout in seconds
+    :return: Cache key with time-based prefix (production) or unchanged (dev/test)
+    """
+    is_dev_or_test = settings.DEVELOPMENT or settings.TESTING
+    if is_dev_or_test:
+        return base_key
+    days = int(ceil(timeout_seconds / (60 * 60 * 24)))
+    return f"{days}-days:{base_key}"
```
albertisfu, line 45: If we want to use `make_s3_cache_key` in other places as well where the timeout can be None (a persistent object). I think we can add support for those kinds of keys here.

According to the [documentation](https://github.com/freelawproject/django-s3-express-cache/blob/main/README.md#4-deleting-a-value) this is straightforward.

We should simply avoid using an `N-days` prefix and instead use "persistent".

So, if `timeout_seconds` is `None`, the prefix should be: `f"persistent:{base_key}"`

We can also add a test for this case.

cl/lib/s3_cache.py
```diff
@@ -0,0 +1,38 @@
+from math import ceil
+
+from django.conf import settings
+from django.core.cache import caches
+
+
+def get_s3_cache(fallback_cache: str = "db_cache"):
```
albertisfu:
```suggestion
def get_s3_cache(fallback_cache: str = "db_cache") -> BaseCache:
```

cl/lib/tests.py
```diff
@@ -2204,3 +2204,94 @@ def test_needs_ocr_exhibit_exception_true(self):
...
         )
+
+
+class TestS3CacheHelpers(SimpleTestCase):
+    """Tests for the S3 cache helper functions in cl/lib/s3_cache.py"""
+
+    @override_settings(DEVELOPMENT=True, TESTING=False)
+    def test_get_s3_cache_returns_fallback_in_development(self) -> None:
+        """In development mode, get_s3_cache should return the fallback cache."""
+        from cl.lib.s3_cache import get_s3_cache
```
albertisfu:
```suggestion
```

We don't usually add imports within each test method unless there is a circular import error. If possible, we should move these imports in tests to the top as module-level imports.

I think even if we are overriding settings, module-level imports should still work. 

This also applies to `from unittest.mock import MagicMock, patch` and `from cl.lib.s3_cache import make_s3_cache_key`.

albertisfu (changes requested):
Thanks @mitch-mitchel this looks good.
I just have a couple of comments and suggestions.


Also, is there a reason to skip the deployment for everything?

mitch-mitchel:
Changes included:
  - Add BaseCache return type hint to get_s3_cache()
  - Support persistent cache keys when timeout_seconds=None
  - Move s3_cache imports to module level in tests
  - Add tests for persistent cache key functionality

cl/lib/tests.py
mlissner: I'm going to add this to CLAUDE.md. I've seen Claude doing this to me too.

mitch-mitchel: I've added it to my personal global Claude.md file as well - Claude loves the inline imports

cl/lib/s3_cache.py
```diff
@@ -0,0 +1,43 @@
...
+    for cache entries that should never expire. In development/testing,
+    returns the key unchanged to maintain compatibility with existing
+    cache behavior.
+
+    :param base_key: The base cache key (e.g., "clusters-mlt-es:123")
+    :param timeout_seconds: Cache timeout in seconds, or None for persistent
+    :return: Cache key with time-based prefix (production) or unchanged (dev/test)
+    """
+    is_dev_or_test = settings.DEVELOPMENT or settings.TESTING
+    if is_dev_or_test:
```
albertisfu: When I reviewed these keys, I found that there are around 150K items cached in the DB. I’m thinking on the possibility that many of these keys correspond to popular cases, if that's the case they are likely being re-cached in S3 within a short period of time. I think we may be able to support those queries.

However, as a precaution, I’d suggest adding this feature behind a waffle switch so we can easily disable it if we see Elasticsearch becoming overwhelmed.

```suggestion
    if is_dev_or_test_or_cache_disabled:
```

Once this is rolled out, after a week or so, we can consider removing the switch.

mitch-mitchel:
@albertisfu please check my waffle-fu

cl/lib/s3_cache.py
```diff
@@ -0,0 +1,45 @@
...
```
albertisfu: It seems we’re missing the switch here as well. Otherwise, we’ll start generating S3-compatible keys and storing them in the `db_cache`.

```suggestion
    if is_dev_or_test or not switch_is_active("enable-s3-cache"):
```

albertisfu (commented):
It seems that some failing tests need to be updated with the switch.

cl/opinion_page/utils.py
```diff
@@ -252,8 +251,10 @@ async def es_get_related_clusters_with_cache(
     :param request:The user request
     :return:Related Cluster Data
     """
-    cache = caches["db_cache"]
-    mlt_cache_key = f"clusters-mlt-es:{cluster.pk}"
+    cache = get_s3_cache("db_cache")
+    mlt_cache_key = make_s3_cache_key(
+        f"clusters-mlt-es:{cluster.pk}", settings.RELATED_CACHE_TIMEOUT
+    )
```
albertisfu: @mitch-mitchel  one more thing, please: since `get_s3_cache and make_s3_cache_key` now access the switch, and `switch_is_active` is a sync method, and these methods are being called from an async context, this is now failing with the error: `You cannot call this from an async context - use a thread or sync_to_async.`

So we need to update these calls here and in all the other places, such as:

```suggestion
    cache = await sync_to_async(get_s3_cache)("db_cache")
    mlt_cache_key = await sync_to_async(make_s3_cache_key)(
        f"clusters-mlt-es:{cluster.pk}", settings.RELATED_CACHE_TIMEOUT
    )
```

I also suggest adding a test to cover this case. I think it should have DEVELOPMENT and TESTING disabled, and the switch enabled.

mitch-mitchel:
@albertisfu I would like to pair to review this, either before or after you merge? just want to talk through it directly at some point - no rush.

albertisfu:
Thanks @mitch-mitchel this is ready to be merged.

albertisfu:
This is now live and working.

Screenshot 2026-01-15 at 1 27 59 p m

No load issues in ES so far. I'll continue monitoring it.

mlissner:
Can we see the performance change in grafana or are we not tracking that yet?

albertisfu:
> Can we see the performance change in grafana or are we not tracking that yet?


The ES performance?

This is from the last hour, no noticeable change. I was worried that many keys might be re-created at the same time and could impact ES performance, but that doesn’t seem to be happening so far.


Screenshot 2026-01-15 at 1 39 11 p m


Or do you mean the performance of the views using those new cache keys?
If so, I don’t think we have those metrics in Prometheus.

We’d need something like response time per view.

mlissner:
> Or do you mean the performance of the views using those new cache keys?

Yeah, that's what I meant. I thought we could get that from @elisa-a-v's new work, but perhaps it's not quite that granular.

PR #6721, opened 2026-01-06
New Texas Models

### LINKED ISSUE #6719: Create New Models for Texas Dockets
### Describe the issue.

Create `TexasDocketEntry`, `TexasDocument`, and `CaseTransfer` models as discussed in #6589 to allow ingestion of Texas docket data.

Spreadsheets with model specifications:

- [TexasDocketEntry](https://docs.google.com/spreadsheets/d/1QkIa0rViCcsopzwUUDAANZ9ClF8iIB6ThKtkQkvA7qA/edit?usp=sharing)
- [TexasDocument](https://docs.google.com/spreadsheets/d/1RLgSjK7kO9_y30cHTEqt838qPRDBfbUGyYs-8mp-aR8/edit?usp=sharing)
    - Will have additional `filepath_local` index, which was not allowed due to data validation rules in the spreadsheet.
- [CaseTransfer](https://docs.google.com/spreadsheets/d/1OorS2BywmOQ5iEzWe-PC6nQpAVmNes9oBh_nBSLz2pM/edit?usp=sharing)

### Related Issue

_No response_

### Screenshots

_No response_

To Do

MorganBennetDev:
## Summary

This PR adds new models to allow full capture of data from Texas dockets as discussed in #6589 and summarized in #6719.

Adds the following models
- `CaseTransfer` to allow tracking of docket transfers beyond just appeals. This is needed to capture workload balancing transfers between different Texas Appellate courts;
- `TexasDocketEntry` to allow capturing fields unique to Texas docket entries such as descriptions and SC remarks on entries;
- `TexasDocument` to allow efficient capture and merging of `TexasDocketEntry` PDF attachments and their descriptions.

The `TexasDocketEntry` and `TexasDocument` models are contained in the `cl/state/texas` module to keep state-specific logic organized.

Additionally, this PR adds a `document_model` decorator to `cl/lib/model_helpers.py` to allow setting unset `Field.help_text` and `Field.db_comment` values to a value extracted from the model's docstring (if present).

Resolves #6719 

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/api/templates/bulk-data.html
+++ b/cl/api/templates/bulk-data.html
@@ -202,7 +202,7 @@
 
     <h2 id="odds-and-ends">Odds and Ends</h2>
     <h3 id="schedule">Generation Schedule</h3>
-    <p>As can be seen on the public <a href="https://www.google.com/calendar/embed?src=michaeljaylissner.com_fvcq09gchprghkghqa69be5hl0@group.calendar.google.com&ctz=America/Los_Angeles">CourtListener maintenance calendar</a>, bulk data files are regenerated on the last day of every month beginning at 3AM PST. Generation can take many hours, but in general is expected to conclude before the 1st of each month. Check the date in the filename to be sure you have the most recent data.
+    <p>As can be seen on the public <a href="https://www.google.com/calendar/embed?src=michaeljaylissner.com_fvcq09gchprghkghqa69be5hl0@group.calendar.google.com&ctz=America/Los_Angeles">CourtListener maintenance calendar</a>, bulk data files are regenerated quarterly on the last day of March, June, September, and December beginning at 3AM PST. Generation can take many hours, but in general is expected to conclude before the next day. Check the date in the filename to be sure you have the most recent data.
     </p>
 
     <h3 id="contributing">Adding Features and Fixing Bugs</h3>
--- a/cl/api/templates/v2_bulk-data.html
+++ b/cl/api/templates/v2_bulk-data.html
@@ -124,7 +124,7 @@
 
   <c-layout-with-navigation.section id="schedule">
     <h3>Generation Schedule</h3>
-    <p>As can be seen on the public <a class="underline" href="https://www.google.com/calendar/embed?src=michaeljaylissner.com_fvcq09gchprghkghqa69be5hl0@group.calendar.google.com&ctz=America/Los_Angeles">CourtListener maintenance calendar</a>, bulk data files are regenerated on the last day of every month beginning at 3AM PST. Generation can take many hours, but in general is expected to conclude before the 1st of each month. Check the date in the filename to be sure you have the most recent data.</p>
+    <p>As can be seen on the public <a class="underline" href="https://www.google.com/calendar/embed?src=michaeljaylissner.com_fvcq09gchprghkghqa69be5hl0@group.calendar.google.com&ctz=America/Los_Angeles">CourtListener maintenance calendar</a>, bulk data files are regenerated quarterly on the last day of March, June, September, and December beginning at 3AM PST. Generation can take many hours, but in general is expected to conclude before the next day. Check the date in the filename to be sure you have the most recent data.</p>
   </c-layout-with-navigation.section>
 
   <c-layout-with-navigation.section id="contributing">


cl/api/templates/bulk-data.html
```diff
@@ -210,6 +210,7 @@ <h3 id="contributing">Adding Features and Fixing Bugs</h3>
     </p>
 
     <h3 id="old">Release Notes</h3>
+    <p><strong>2026-01-07</strong>: Added <code>CaseTransfer</code>, <code>TexasDocketEntry</code>, and <code>TexasDocument</code> models to allow full capture of Texas court data.</p>
```
albertisfu: For now, I think we can omit adding the new fields and tables from `make_bulk_data.sh` until we actually have data.

Let’s remove it for now and create an issue to ensure we remember to add it later.

cl/lib/model_helpers.py
```diff
@@ -564,6 +565,52 @@ def linkify_orig_docket_number(agency: str, og_docket_number: str) -> str:
     return ""
 
 
+FIELD_DOCSTRING_EXTRACTION_RE = re.compile(
+    r":(?:var|ivar|cvar)\s+([a-z_]+):([^:]+)", re.IGNORECASE | re.MULTILINE
```
albertisfu: Should we also support fields with numbers, such as `field2`?

cl/lib/model_helpers.py
```diff
@@ -564,6 +565,52 @@ def linkify_orig_docket_number(agency: str, og_docket_number: str) -> str:
...
+            continue
+
+        doc = field_docs[field_name]
+        if isinstance(field.help_text, str) and len(field.help_text) == 0:
+            field.help_text = doc
+        if (
+            isinstance(field.db_comment, str)
+            and len(field.db_comment) == 0
+            or field.db_comment is None
+        ):
```
albertisfu: I think this can be simplified to:

```suggestion
        if not field.db_comment:
```

cl/lib/model_helpers.py
```diff
@@ -564,6 +565,52 @@ def linkify_orig_docket_number(agency: str, og_docket_number: str) -> str:
...
+            field.db_comment = doc
+        print(getattr(model, field_name))
```
albertisfu: stray `print` statement

cl/search/state/texas/models.py
```diff
@@ -0,0 +1,84 @@
...
+    appellate_brief = models.BooleanField(default=False)
+    description = models.TextField(blank=True)
+    remarks = models.TextField()
+    date_filed = models.DateField(
+        null=True,
+        blank=True,
+    )
+    entry_type = models.TextField()
+    sequence_number = models.CharField(
+        unique=True,
```
albertisfu: I don’t think this should be unique, right? That would prevent entries from other cases from having the same `sequence_number`, which is possible across different cases.

cl/search/models.py
```diff
@@ -3969,3 +3971,64 @@ def __str__(self) -> str:
...
+        on_delete=models.CASCADE,
+        related_name="case_transfer_destination_court",
+    )
+    destination_docket = models.ForeignKey(
+        "search.Docket",
+        on_delete=models.CASCADE,
+        related_name="case_transfer_destination_docket",
+    )
+    transfer_date = models.DateField()
+    transfer_time = models.TimeField()
```
albertisfu: Do we actually have time data to populate this?

albertisfu (changes requested):
Thanks @MorganBennetDev this looks great! I just have a couple of comments, please.

cl/lib/model_helpers.py
MorganBennetDev: Good idea. Will update.

cl/search/state/texas/models.py
MorganBennetDev: Good catch.

cl/search/models.py
MorganBennetDev: I put it there to support potential future courts/states that include time, but would probably be best to leave it off until we know we need time.

cl/search/state/texas/models.py
```diff
@@ -0,0 +1,83 @@
...
+    remarks = models.TextField()
+    date_filed = models.DateField(
+        null=True,
+        blank=True,
+    )
+    entry_type = models.TextField()
+    sequence_number = models.CharField(
+        max_length=16,
+    )
+
```
albertisfu, line 50: It seems we're missing the `disposition` field?

cl/search/models.py
```diff
@@ -3969,3 +3971,63 @@ def __str__(self) -> str:
...
+    )
+    destination_docket = models.ForeignKey(
+        "search.Docket",
+        on_delete=models.CASCADE,
+        related_name="case_transfer_destination_docket",
+    )
+    transfer_date = models.DateField()
+    transfer_remarks = models.TextField(
+        blank=True,
+    )
```
albertisfu: Do we have data in Texas to populate this field, or is it just a provision in case we need it in the future for other states? If it’s the latter, I think we might want to add it later once we confirm it’s needed.

cl/search/models.py
```diff
@@ -3969,3 +3971,63 @@ def __str__(self) -> str:
...
+    transfer_type = models.SmallIntegerField(
+        choices=transfer_type_choices.items(),
```
albertisfu, line 4027: Do we have this value in the Texas data?  or what logic do we expect to use to populate it? Can we have a default or `UNKNOWN` value in cases where the logic cannot determine the transfer type?

albertisfu (changes requested):
Thanks @MorganBennetDev 

It seems there is a conflict with the migrations, since there is another `0050` migration. We need to re-create the migration files, please.

I left a couple of additional comments.

cl/search/models.py
MorganBennetDev: We do not. This is another "_might be relevant_" field. Will remove for now.

cl/search/models.py
MorganBennetDev: For Texas, this will be set to `APPEAL` for COA to SC/CCA appeals and `WORKLOAD` for COA to COA transfers.

albertisfu (approved):
Thanks for the changes @MorganBennetDev this is ready to be merged.


@mlissner we need to apply these migrations. I created this issue for that: https://github.com/freelawproject/infrastructure/issues/585

The PR is ready to be merged, but I’m holding off for now because merging it would block other deployments. We can merge it shortly before we are ready to apply the migrations.

mlissner:
Great. Diego has another PR from me too that has DB changes, so I'll note that for him in yours and he can do 'em both.

The `document_model` decorator is interesting, @MorganBennetDev. I missed my chance to comment on it, but I worry that it uses too magic while separating the location of the field documentation from the fields themselves. These models are small now, but this will get worse as they grow and the docstring isn't on the same page as the field itself. But this is an interesting approach, so I'm open to seeing how we like it and we can easily swap it out later if we aren't into it or if it causes problems. Trade-offs! 

Thanks all!

PR #6741, opened 2026-01-08
feat(docket-scraper): add command for running BaseStateScrapers.backfill

Brennan-Chesley-FLP:
## Summary

This PR adds a new command to gather the case page html for Texas dockets via TAMES, and a foundation that can be built upon for the future docket scrapers. It uses [this scraper](https://github.com/freelawproject/juriscraper/pull/1756) in Juriscraper to navigate the search page. It saves case page HTML and headers(which is generally uninteresting) in S3, as well as the metadata from the search for comparison and validation. 

The search functionality has some load shedding via 403 responses that appears to occur when it's only search queries that are being made. To time out the requests in a way the the server is better able to handle while still being efficient for us, this PR grabs a page of search results, then all of the cases from it, then the next page, and so on. We should probably only invoke this with a rate limit of 0.2 or lower (5/req), though fetching the case pages is not currently rate limited so the actual throughput will be much higher. This PR does not save the search results pages, just case pages. 

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

This command must be invoked after deployment.

### DIFF
--- a/cl/api/management/commands/notify_bad_api_filter_users.py
+++ b/cl/api/management/commands/notify_bad_api_filter_users.py
@@ -1,0 +1,166 @@
+import sys
+from typing import Any
+
+from django.contrib.auth.models import User
+from django.core.mail import EmailMultiAlternatives, get_connection
+from django.template import loader
+
+from cl.api.utils import (
+    clear_bad_filter_params_for_user,
+    get_all_users_with_bad_filter_params,
+    get_bad_filter_params_for_user,
+)
+from cl.lib.command_utils import VerboseCommand, logger
+
+
+class Command(VerboseCommand):
+    """Send notification emails to users who have used invalid API filter
+    parameters.
+
+    This command should be run nightly to notify users about their invalid
+    API usage before we start blocking such requests.
+    """
+
+    help = (
+        "Sends notification emails to users who have used invalid API filter "
+        "parameters, warning them that such requests will soon be blocked."
+    )
+
+    def add_arguments(self, parser: Any) -> None:
+        parser.add_argument(
+            "--simulate",
+            action="store_true",
+            default=False,
+            help="Don't send any emails, just show what would be sent.",
+        )
+        parser.add_argument(
+            "--clear-after-send",
+            action="store_true",
+            default=False,
+            help="Clear Redis records after sending emails (use with caution to prevent repeat emails).",
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        super().handle(*args, **options)
+        self.options = options
+
+        if self.options["simulate"]:
+            sys.stdout.write(
+                "**********************************\n"
+                "* SIMULATE MODE - NO EMAILS SENT *\n"
+                "**********************************\n\n"
+            )
+
+        user_ids = get_all_users_with_bad_filter_params()
+        if not user_ids:
+            logger.info("No users with bad filter parameters found.")
+            return
+
+        logger.info(
+            f"Found {len(user_ids)} user(s) with bad filter parameters"
+        )
+
+        messages = []
+        users_to_clear: list[int] = []
+
+        for user_id in user_ids:
+            try:
+                user = User.objects.get(pk=user_id)
+            except User.DoesNotExist:
+                logger.warning(f"User {user_id} not found, skipping")
+                continue
+
+            bad_params = get_bad_filter_params_for_user(user_id)
+            if not bad_params:
+                continue
+
+            message = self._create_email(user, bad_params)
+            messages.append(message)
+            users_to_clear.append(user_id)
+
+            if self.options["simulate"]:
+                self._print_simulation(user, bad_params)
+
+        if not messages:
+            logger.info("No emails to send.")
+            return
+
+        if not self.options["simulate"]:
+            connection = get_connection()
+            connection.send_messages(messages)
+            logger.info(f"Sent {len(messages)} notification email(s).")
+
+            if self.options["clear_after_send"]:
+                for user_id in users_to_clear:
+                    clear_bad_filter_params_for_user(user_id)
+                logger.info(
+                    f"Cleared Redis records for {len(users_to_clear)} user(s)."
+                )
+        else:
+            sys.stdout.write(
+                f"\nSimulation complete. Would have sent {len(messages)} "
+                "email(s).\n"
+            )
+
+    def _create_email(
+        self,
+        user: User,
+        bad_params: list[dict[str, Any]],
+    ) -> EmailMultiAlternatives:
+        """Create an email message for a user with bad filter parameters.
+
+        :param user: The user to notify.
+        :param bad_params: List of bad parameter records.
+        :return: An EmailMultiAlternatives object ready to send.
+        """
+        txt_template = loader.get_template(
+            "emails/bad_api_filter_params_email.txt"
+        )
+
+        # Group bad params by endpoint for cleaner display
+        params_by_endpoint: dict[str, list[dict[str, Any]]] = {}
+        for record in bad_params:
+            endpoint = record["endpoint"]
+            if endpoint not in params_by_endpoint:
+                params_by_endpoint[endpoint] = []
+            params_by_endpoint[endpoint].append(record)
+
+        context = {
+            "name": user.first_name or user.username,
+            "params_by_endpoint": params_by_endpoint,
+            "total_count": sum(r["count"] for r in bad_params),
+        }
+
+        email_txt = txt_template.render(context)
+
+        return EmailMultiAlternatives(
+            subject="Action Required: Invalid API Filter Parameters Detected",
+            body=email_txt,
+            from_email="CourtListener <noreply@courtlistener.com>",
+            to=[user.email],
+            headers={"X-Entity-Ref-ID": f"bad_api_filter:{user.pk}"},
+        )
+
+    def _print_simulation(
+        self,
+        user: User,
+        bad_params: list[dict[str, Any]],
+    ) -> None:
+        """Print simulation output for a user.
+
+        :param user: The user who would receive the email.
+        :param bad_params: List of bad parameter records.
+        """
+        sys.stdout.write(f"\n{'=' * 60}\n")
+        sys.stdout.write(f"Would email: {user.email} (User ID: {user.pk})\n")
+        sys.stdout.write(f"{'=' * 60}\n")
+
+        for record in bad_params:
+            sys.stdout.write(
+                f"  Endpoint: {record['endpoint']}\n"
+                f"  Parameter: {record['param']}\n"
+                f"  Count: {record['count']}\n"
+                f"  First seen: {record['first_seen']}\n"
+                f"  Last seen: {record['last_seen']}\n"
+                f"  ---\n"
+            )
--- a/cl/api/templates/emails/bad_api_filter_params_email.txt
+++ b/cl/api/templates/emails/bad_api_filter_params_email.txt
@@ -1,0 +1,65 @@
+Hello{% if name %} {{ name }}{% endif %},
+
+We're writing to let you know that in the last 24 hours we detected API
+requests from your account that include invalid GET parameters our system
+doesn't recognize. In the coming days, we will begin returning HTTP 400 errors
+for such requests.
+
+This email is your advance notice and your opportunity to address this issue.
+
+We apologize for any inconvenience this may cause. This change is necessary to
+protect the integrity of our system and ensure that all API users receive
+accurate, filtered results. When invalid filter parameters are used, they are
+silently ignored, which can lead to unexpected results—often returning far more
+data than intended. Previously, this was not a common problem, but a plethora
+of vibe-coded solutions now regularly trigger very heavy usage of our system
+that we need to urgently address.
+
+
+YOUR AFFECTED API REQUESTS
+--------------------------
+{% for endpoint, params in params_by_endpoint.items %}
+Endpoint: {{ endpoint }}
+{% for param in params %}  - Parameter "{{ param.param }}" was used {{ param.count }} time(s)
+    First seen: {{ param.first_seen }}
+    Last seen: {{ param.last_seen }}
+{% endfor %}
+{% endfor %}
+Total requests with invalid parameters: {{ total_count }}
+
+
+WHAT YOU NEED TO DO
+-------------------
+Please review your API integration code and remove or correct the invalid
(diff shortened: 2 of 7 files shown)


cl/scrapers/management/commands/cl_back_scrape_dockets.py
```diff
@@ -0,0 +1,412 @@
...
+                response.raise_for_status()
+
+            logger.warning(
+                "Received 403 Forbidden for %s. Backing off for %d seconds (total: %d/%d)",
+                url,
+                backoff,
+                total_wait,
+                self.max_backoff_seconds,
+            )
+            time.sleep(backoff)
```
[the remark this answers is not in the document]
Brennan-Chesley-FLP: This sleep is on purpose. If we want to go with an off-the-shelf rate limiter like pyrate_limiter, we certainly can.

cl/scrapers/management/commands/cl_back_scrape_dockets.py
```diff
@@ -0,0 +1,414 @@
...
+
+            logger.info(
+                "Backfilling %d courts from %s to %s",
+                len(courts),
+                start_date,
+                end_date,
+            )
+
+            # Create a separate session for fetching case URLs
+            case_session = requests.Session()
```
albertisfu: Is there any reason not to use the `request_manager` here, so that case retrieval benefits from retries and the rate-limit configuration?

cl/scrapers/management/commands/cl_back_scrape_dockets.py
```diff
@@ -0,0 +1,414 @@
...
+    Args:
+        response: The HTTP response to save
+        scraper_class_name: Name of the scraper class (e.g., "TAMESScraper")
+        court_id: Court identifier extracted from response data
+        case_meta: Optional metadata dict from the scraper (e.g., case_number,
+            date_filed, etc.) to save alongside the response
+    """
+    storage = S3GlacierInstantRetrievalStorage()
+
+    now_str = datetime.now().strftime("%Y/%m/%d/%H_%M_%S_%f")
```
albertisfu: For case responses what if instead of using `now_str` we use the docket number? It’s supposed to be unique, right? That way, if we resume the scraper and get an overlap, we won’t end up with duplicated files.

cl/scrapers/management/commands/cl_back_scrape_dockets.py
```diff
@@ -0,0 +1,414 @@
...
+
+        # Create rate-limited request manager with response saving callback
+        def save_scraper_response(manager, response):
+            court_id = "unknown_court"
+            save_docket_response(response, scraper_class_name, court_id)
+
+        request_manager = RateLimitedRequestManager(
+            requests_per_second=options["rate"],
+            max_backoff_seconds=options["max_backoff"],
+            all_response_fn=save_scraper_response,
```
albertisfu: This is to store the search responses right? 
Can we store them in a different directory? e.g:

`base_name = f"responses/dockets/{scraper_class_name}/{searches}/{now_str}"`

Instead of `unknown_court`

cl/scrapers/management/commands/cl_back_scrape_dockets.py
albertisfu: Maybe we just need a different request_manager with no `all_response_fn` set, so we don’t duplicate the stored responses.

cl/scrapers/management/commands/cl_back_scrape_dockets.py
```diff
@@ -0,0 +1,414 @@
...
+
+    # Save case metadata if provided
+    if case_meta is not None:
+        meta_json = json.dumps(case_meta, indent=4, default=str)
+        storage.save(f"{base_name}_meta.json", ContentFile(meta_json))
+
+    try:
+        content: str | bytes = json.dumps(
+            json.loads(response.content), indent=4
+        )
```
albertisfu: Considering `ContentFile(content)` expect bytes.

Should we do:

```suggestion
        content: str | bytes = json.dumps(
            json.loads(response.content), indent=4
        ).encode('utf-8')
```

albertisfu: Considering `ContentFile(content)` expect bytes.

Should we do:

```suggestion
        content: str | bytes = json.dumps(
            json.loads(response.content), indent=4
        ).encode('utf-8')
```

cl/scrapers/management/commands/cl_back_scrape_dockets.py
```diff
@@ -0,0 +1,414 @@
...
+                    if not case_url:
+                        logger.warning("Case without case_url: %s", case)
+                        continue
+
+                    # Extract court_id from case data if available
+                    court_id = case.get("court_code") or "unknown_court"
+
+                    # Fetch and save the case page
+                    try:
+                        case_response = case_session.get(case_url, timeout=60)
```
albertisfu: What about HTTP errors not related to being blocked, like 500 or 404?

Should we use `raise_for_status()` so the error gets logged?

```suggestion
                        case_response = case_session.get(case_url, timeout=60)
                        case_response.raise_for_status()
```

cl/scrapers/management/commands/cl_back_scrape_dockets.py
```diff
@@ -0,0 +1,414 @@
...
+
+
+class Command(BaseCommand):
+    help = "Runs BaseStateScraper backfill operations for docket enumeration."
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "court_id",
+            help="The module path of the scraper to run (e.g., juriscraper.state.texas.tames)",
+        )
```
albertisfu, line 274: It seems this argument is missing the `--`.

Also, could we give it a different name, like `scraper`? 

And make it a required argument.

albertisfu (changes requested):
Thanks @Brennan-Chesley-FLP this looks good. Just a few comments.

In addition, can you add this file to lint.yml in the section:
`# Add files here as they're ready`

So this new module is checked by mypy.

albertisfu:
@Brennan-Chesley-FLP I've released a new version of Juriscraper containing the changes from https://github.com/freelawproject/juriscraper/pull/1756 required for this PR

cl/scrapers/management/commands/cl_back_scrape_dockets.py
Brennan-Chesley-FLP: Added a second RequestManager to handle the case downloads and a second command line flag to specify its rate limit.

cl/scrapers/management/commands/cl_back_scrape_dockets.py
Brennan-Chesley-FLP: Added this with a dash of S3-friendly name mangling.

cl/scrapers/management/commands/cl_back_scrape_dockets.py
Brennan-Chesley-FLP: Removed this to simplify this initial implementation. We're only grabbing html for now.

cl/scrapers/management/commands/cl_back_scrape_dockets.py
Brennan-Chesley-FLP: Without the `--` it's treated as an argument in the unix sense (no flag, required). Changed it to a `--scraper` option and noted it is required.

cl/scrapers/management/commands/cl_back_scrape_dockets.py
Brennan-Chesley-FLP: Raised this in the new RequestManager.

cl/scrapers/management/commands/back_scrape_dockets.py
```diff
@@ -0,0 +1,435 @@
...
+    content = response.content
+    extension = "html"
+
+    content_name = f"{base_name}.{extension}"
+    storage.save(content_name, ContentFile(content))
+
+
+def save_search_response_factory(scraper_class_name: str):
+    storage = S3GlacierInstantRetrievalStorage()
+    prefix = f"dockets/{scraper_class_name}/searches/"
```
albertisfu: It seems this should go to “responses” as well

```suggestion
    prefix = f"responses/dockets/{scraper_class_name}/searches/"
```

cl/scrapers/management/commands/back_scrape_dockets.py
```diff
@@ -0,0 +1,435 @@
...
+    # Save case metadata if provided
+    if case_meta is not None:
+        meta_json = json.dumps(case_meta, indent=4, default=str)
+        storage.save(f"{base_name}_meta.json", ContentFile(meta_json))
+
+    content = response.content
+    extension = "html"
+
+    content_name = f"{base_name}.{extension}"
+    storage.save(content_name, ContentFile(content))
```
albertisfu, line 239: Within this method, we have three save requests. I tested it, and it is taking around 3 seconds to complete for each docket saved. I think this can slow down the overall process and make it take even more time than what the rate limit allows.

Do you have an idea of how much time we can afford for this method to take so that the overall process still finishes in the expected time and is only capped by the rate limit?

If this is indeed slowing down the process, an alternative is to download files concurrently. I tested that approach, and the average time to complete was about 2 seconds.

If we want to be more aggressive, we could move this code to a Celery task, but we would need to store the content temporarily in Redis so we don’t pass the entire response to the task, which could cause memory issues, as we have experienced before when passing large arguments to tasks.

mlissner: Could we also stash those in our new S3 cache instead? I think we considered that last time, but I forget what we decided...

albertisfu: > Could we also stash those in our new S3 cache instead? I think we considered that last time, but I forget what we decided...

Instead of the Redis cache if we use a celery task? 

Not for storing the scraped HTML files, right?

mlissner: Right.

albertisfu: Sure. I’ll run a couple of tests to measure the S3 cache latencies to see if it’s fast enough to be worth using, since we’ll need to do something like:

- Store 3 files in the cache (S3 Express zone)
- Retrieve the files from the S3 Express zone
- Store the files in a regular S3 bucket

We need the first step to be fast enough.

albertisfu: Ok, here are the results from my tests:

- Average time to save the 3 files from a docket response to a regular S3 bucket: 3 seconds
- Average time to save the 3 files from a docket response to an S3 Express Zone bucket: 0.40 seconds

So it seems we can use the S3 cache to pass the content to a Celery task.

@Brennan-Chesley-FLP, do you think ~0.40 seconds is fast enough to avoid slowing down the scraper more than expected? I think so, since the rate limit we’re considering for fetching dockets is higher than that, right?

If so, the idea would be the following:

Within `save_docket_response`, store the files in the S3 cache.

For example:

```
from django.core.cache import caches

cache = caches["s3"]
cache.set(f"2-days:{base_name}.html", response.content, timeout=3600 * 48)
```

Then, in a Celery task, we pass only the cache keys as arguments. In the task, we retrieve the elements from the cache and then store the files in the regular S3 bucket and remove the elements from the S3 cache.

Brennan-Chesley-FLP: We can certainly offload or optimize the docket page downloads further. There doesn't appear to be the load shedding behavior for direct docket access that there is for the search feature of TAMES. There's a few optimizations we can implement here:
1. We can consolidate the _meta files so there's only 1 write to S3 for every 25 rows instead of every row. This would correlate the case meta info with the search results pages instead of the docket pages, but it's certainly not hard to wed them together again.
2. We can bundle the headers and html into a single file to reduce writes to S3. We could also bundle the metadata in this if wanted to keep it colocated and reduce S3 writes.
3. We can concurrently scrape the docket pages with a little bit of asyncio. I didn't see the same load shedding behavior for individual pages, so we can push our concurrency a bit here. A little Asyncio queue with a few workers and some blocking puts could manage this for us and we can cap the queue size if we have memory concerns. 
4. We can use Redis Queue/Stream and a celery task to download the docket pages directly to Glacier_IR. This will allow us to have a stop/done event in the queue if we'd like.
5. We can save the files to S3 Express and just run a bash one-liner, boto3, or the AWS console to move them. The celery task seems like extra complexity to me unless you imagine it doing extra work or sitting behind a queue. 

Operationally, do we have an idea of how frequently we would like to be running a full backfill (as opposed to a window of recent cases)? Do we have some goals or restrictions on how long this can run (besides, of course wanting it to run as quickly as it can without bumping into bot protection measures)? Alberto, when you ran the S3 timing test, was that writing from your own computer, or from the CL server/cluster?

albertisfu: > Operationally, do we have an idea of how frequently we would like to be running a full backfill (as opposed to a window of recent cases)? 

In an ideal scenario, we want to run the full backfill only once, unless we later encounter errors or gaps that need to be fixed.

> Do we have some goals or restrictions on how long this can run (besides, of course wanting it to run as quickly as it can without bumping into bot protection measures)?

I don’t think we have a specific target. The goal is simply to run it as fast as possible without getting blocked. For example, in an ideal scenario, with a rate limit of 1 case per second, if we expect to finish in about a week. But if some part of the process, like storing files, slows it down to three weeks, then we should definitely try to optimize it.

> Alberto, when you ran the S3 timing test, was that writing from your own computer, or from the CL server/cluster?

I ran it from my computer. I just repeated the same tests in the CL cluster, and saving the three files takes around` 0.5` seconds to a regular S3 bucket, either sequentially or concurrently using `ThreadPoolExecutor`. It seems that this time is dominated by saving the largest file.

This is much faster than what I measured locally. I’m not sure yet if this speed is sufficient for our needs, but it might mean we don’t need to optimize right away.

What if we go with the current approach, and once we start running the scraper, we measure the actual progress? Then we can see the real scraping rate and decide whether any optimization is necessary.

albertisfu: According to our conversation, we’ll merge this and run the scraper and monitor the scraping rate over the next few hours. If we find it too slow, we can optimize it.

Issue #6730, opened 2026-01-07
Merger for Texas `Docket`s


MorganBennetDev:
### Describe the issue.

Implement a merger for Texas Supreme Court, Courts of Appeal, and Court of Criminal Appeals dockets.

### What type of issue is this?

To Do

Issue #1795, opened 2021-10-22
Harvard Import Cleanup generating lots of edge case issues


flooie:
Starting with citation edge cases.

Florida, has 49 failing citation (edge cases) that come in ... 40 different flavors.  

```
1        239 La
2        No. 5D14-2134.
3        195 La 1
4        746 So. 2d.
5        779 So. 2d15.
6        Within the record provided on appeal, which does not contain a transcript there is nothing to sufficiently indicate that the judgment of the trial court was in error. We assume that evidence was presented supporting the trial judge’s finding of unconscionability, and we assume the plaintiff had the opportunity to present evidence as required by
7        95 Fla. 736a.
8        Judgment, of ouster awarded.
9        135 So, 839
10       146- So. 98
11       146 S. 660
12       146 S. 900
13       147 SO. 463
14       147 S. 254
15       149.So. 331
16       149 So, 574
17       149. So. 573
18       149 S. 582
19       ISO So. 133
20       152. So. 429
21       154 So, 206
22       1S4 So. 833
23       159 S. 46
24       163 S. 486
25       183 So, 15
26       180 Sol 625
27       Division A
28       1 So. (2nd Series) 183
29       8 So. (2nd)) 279
30       9. So. (2nd) 170
31       10 So. (2nd)) 431
32       17'So. (2nd) 78
33       17 So .(2nd) 125
34       20. So. (2nd) 484
35       20 So. (2nd 695
36       2i So. (2nd) 356
37       21- So. (2nd). 211
38       22 So, (2nd) 154
39       24th So. (2nd) 51
40       26 So. So. (2nd) 667
41       27 (2nd) 609
42       28 S. (2nd) 836
43       30 (2nd) 100
44       30 (2nd) 101
45       30 (2nd) 100
46       32 (So. 2nd) 593
47       32 So. 2nd) 738
48       37 St. (2nd) 538
49       37 S. (2nd) 369
```

I say the issues here are 

1. Poor OCR: see Nos. 19, 37, 26, 36.
2. Incomplete data: No. 1, 41, 43 - 45, 48.
3. Typos that could be fixed in eye cite but would make it worse: Nos. 38, 25 and lots of others.
4. Wrong data: Nos 6, 8, 27.

I'm not sure the best approach to these.  In some cases, I think tweaking the json we download could be the right approach and then re-importing.  In this case - the original stays on IA and our file can always reference back.  

[...]

PR #6761, opened 2026-01-10
Texas docket entry and attachment mergers

### LINKED ISSUE #6728: Merger for `TexasDocketEntry`s
### Describe the issue.

Implement a data merger for new `TexasDocketEntry` model.

### Related Issue

_No response_

### Screenshots

_No response_

To Do

### LINKED ISSUE #6729: Merger for `TexasDocument`s
### Describe the issue.

Implement a merger for `TexasDocument` models, which will download `TexasDocketEntry` attachments, determine if they have changed since the last scrape, and proceed accordingly.

### Related Issue

_No response_

### Screenshots

_No response_

Feature

MorganBennetDev:
## Summary

This PR implements asynchronous mergers for Texas docket entries and Texas documents, which will be used by the full docket merger in #6730. Operation should be idempotent.

Mergers are asynchronous to speed up operation by preventing a list of mergers waiting on just one merger to complete a network or DB request. The final top-level merger will wrap an asynchronous version of itself in the `async_to_sync` function and be called from a celery task.

Tests will be added on Monday 01/12.

Will be updated with finalized Texas models once #6721 is merged.

Resolves #6728, resolves #6729 

## Deployment

    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [ ] `skip-cronjob-deploy`
    - [ ] `skip-daemon-deploy`

Skip all deployments since this does not add any usable functionality.

### DIFF
--- a/cl/api/templates/v2_webhooks-getting-started.html
+++ b/cl/api/templates/v2_webhooks-getting-started.html
@@ -1,0 +1,161 @@
+{% extends 'new_base.html' %}
+{% load static extras %}
+
+{% block title %}Getting Started With Webhooks &ndash; CourtListener.com{% endblock %}
+{% block description %}
+{% endblock %}
+
+{% block content %}
+<c-layout-with-navigation
+  data-first-active="about"
+  :nav_items="[
+    {'href': '#about', 'text': 'Overview'},
+    {'href': '#receive-a-webhook', 'text': 'Receive a Webhook Event'},
+    {'href': '#setup-a-webhook', 'text': 'Set Up an Endpoint'},
+    {'href': '#testing', 'text': 'Testing an Endpoint'},
+  ]"
+  >
+  <c-layout-with-navigation.section id="about">
+    <h1>Getting Started With Webhooks</h1>
+    <p class="lead">To create your first webhook, there are a handful of steps you will need to take:</p>
+    <ol>
+      <li><p>Creating and receiving webhooks does not require any special access, but sooner or later you will want to have access to the RECAP API endpoints. <a class="text-primary-600" href="{% url 'contact' %}">Send us a note with your username and use case</a> to get that process started.</p></li>
+      <li><p>You need to <a class="text-primary-600" href="#receive-a-webhook">create a URL on your server</a> to receive events from CourtListener.</p></li>
+      <li><p>You need to <a class="text-primary-600" href="#setup-a-webhook">configure CourtListener to send events</a> to that URL.</p></li>
+    </ol>
+    <p>Once you have created the URL on your server and linked it up to ours, you can <a class="text-primary-600" href="#testing">send test events from CourtListener</a> and you can wait for events to be triggered. As you send test events or events are automatically triggered, they appear in the <a class="text-primary-600" href="{% url 'view_webhook_logs' 'logs' %}">logs section of your webhooks panel</a>.</p>
+    <p>Read on to learn more about setting up your first webhook.</p>
+  </c-layout-with-navigation.section>
+
+  <c-layout-with-navigation.section id="receive-a-webhook">
+    <h2>Receive a Webhook Event on Your Server</h2>
+    <p>Webhook events are nothing more than HTTP POST requests sent to your server. To receive these events, you will need to begin by creating a URL in your application.</p>
+    <p>Some requirements to consider for the URL that you create:</p>
+    <ul>
+      <li><a class="text-primary-600" href="{% url 'webhooks_docs' %}#security">Our webhook system does not support authentication</a>, so the URL you create should be long and random.</li>
+      <li>Our POST requests will come from one of two static IP addresses, <code>34.210.230.218</code> and <code>54.189.59.91</code>. We recommend only allowing traffic from these addresses.</li>
+      <li>Your URL must receive POST requests and should not accept any other HTTP verbs.</li>
+      <li>If you are using a web framework like Django or Ruby on Rails, the webhook URL will need to be exempted from the cross site request forgery (CSRF) protection system. If this is not done, most web frameworks will block POST requests from outside domains like ours.</li>
+      <li>To avoid <a class="text-primary-600" href="{% url 'webhooks_docs' %}#retries">timeout errors and event retries</a>, your application may need to process events asynchronously.</li>
+      <li>Your application must respond to events with a 2xx HTTP status code or else the event will be retried.</li>
+      <li>Your application should use the <a class="text-primary-600" href="{% url 'webhooks_docs' %}#headers">Idempotency-Key included in the event headers</a> to ensure that it only processes the event once.</li>
+    </ul>
+    <p>Below are examples of endpoints configured in Flask and Django</p>
+    <p><strong>Flask</strong></p>
+    <c-code>
+from flask import Flask, request, Response
+app = Flask(__name__)
+
+@app.route("/webhooks/a-long-random-url-here/", methods=["POST"])
+def respond():
+    # Do your event processing.
+    print(request.json)
+    return Response(status=200)
+    </c-code>
+    <p><strong>Django</strong></p>
+    <c-code>
+from django.views.decorators.csrf import csrf_exempt
+from django.http import HttpRequest, HttpResponse
+
+@csrf_exempt
+def testing_webhook(request: HttpRequest) -> HttpResponse:
+    if request.method == "POST":
+        print("Headers: ",request.headers)
+        body = request.body.decode("utf-8")
+        print("Body: ", body)
+
+    return HttpResponse("OK")
+    </c-code>
+  </c-layout-with-navigation.section>
+
+  <c-layout-with-navigation.section id="setup-a-webhook">
+    <h2>Set Up a Webhook Endpoint in CourtListener</h2>
+    <p>To set up a webhook endpoint, begin by logging into CourtListener and going to the <a class="text-primary-600" href="{% url 'view_webhooks' %}">Webhooks panel in your profile</a>:</p>
+    <p>
+      <img src="{% static "png/webhooks-panel-v2.png" %}"
+         alt="screenshot of the webhook panel"
+         class="screenshot"
+         loading="lazy"
+         height="261"
+         width="652">
+    </p>
+
+    <p>Click the “Add webhook” button and the “Add webhook endpoint” modal pops up:</p>
+    <p>
+      <img src="{% static "png/add-webhook-endpoint-v2.png" %}"
+         alt="screenshot of how to add a webhook endpoint"
+         class="screenshot"
+         loading="lazy"
+         height="261"
+         width="652">
+    </p>
+    <p>Complete the box with the following information:</p>
+    <ol>
+      <li><p>The Endpoint URL should be a URL on your server that is long and random. It must be securely served with valid HTTPS.</p></li>
+      <li><p>Select the Event Type for which you wish to receive events.</p></li>
+      <li>
+        <p>Choose the webhook version you wish to set up.</p>
+        <p>We recommend selecting the highest available version for your webhook. Refer to the <a class="text-primary-600" href="{% url "webhooks_docs" version="v2" %}#change-log">Change Log</a> for more details on webhook versions.</p>
+        <p>You can only create one Webhook endpoint for each type of event and version. Please get in touch if this limitation causes difficulty for your application.</p>
+      </li>
+      <li>
+        <p>If you are ready to start receiving events at that URL, check the box to enable the webhook.</p>
+        <p>In order to avoid unnecessary errors and retries, we recommend keeping your endpoint disabled until it is live in your application.</p>
+      </li>
+    </ol>
+    <p>Click “Create webhook”</p>
+    <p>Your Webhook endpoint is now created:</p>
+    <p>
+      <img src="{% static "png/webhook-disabled-v2.png" %}"
+         alt="screenshot of the developer tools panel listing a disabled webhook endpoint"
+         class="screenshot"
+         loading="lazy"
+         height="261"
+         width="652">
+    </p>
+  </c-layout-with-navigation.section>
+
+  <c-layout-with-navigation.section id="testing">
+    <h2>Testing a Webhook Endpoint</h2>
+    <p>Getting a webhook working properly can be difficult, so we have a testing tool that will send you a sample webhook event on demand. </p>
+    <p>To use the tool, go to webhooks panel and click the “Test” button for the endpoint you wish to test:</p>
+    <p>
+      <img src="{% static "png/webhook-disabled-v2.png" %}"
+         alt="screenshot of the developer tools panel listing a disabled webhook endpoint"
+         class="screenshot"
+         loading="lazy"
+         height="261"
+         width="652">
+    </p>
+    <p>In the modal that pops up, there are two methods to test your webhook endpoint.</p>
+    <ol>
+      <li>
+        <p><strong>In the “As JSON” tab</strong>, you can ask our server to send a test event to your endpoint. When you click “Send Webhook Test Event” a new event is created with the information shown and is sent to your endpoint. Test events are not retried, but can be seen in the “Test Logs” tab.</p>
+        <p>
+          <img src="{% static "png/test-json-webhook-event-v2.png" %}"
+             alt="screenshot of the webhook json test modal"
+             class="screenshot"
+             loading="lazy"
+             height="261"
+             width="652">
+        </p>
+      </li>
+      <li>
+        <p><strong>In the “As cURL”</strong> tab, you can copy/paste a curl command that can be used to send a test event to your local dev environment.</p>
+        <p>
+          <img src="{% static "png/test-curl-webhook-event-v2.png" %}"
+             alt="screenshot of the webhook curl test modal"
+             class="screenshot"
+             loading="lazy"
+             height="261"
+             width="652">
+        </p>
+      </li>
+    </ol>
+
+  <h3>Use <code class="text-display-xxs font-semibold">ngrok</code> to test your local endpoint.</h3>
+  <p>During the development process, you may want to test your endpoint in your local environment, before moving it to production. To allow CourtListener to reach your development machine over the Internet, you can use a tool like <code>ngrok</code> which will give you a temporary public HTTPS URL that you can use for testing. To learn more about this, read <a class="underline" href="https://ngrok.com/docs/getting-started" target="_blank" rel="noopener noreferrer">ngrok's documentation</a>.</p>
+  </c-layout-with-navigation.section>
+
+</c-layout-with-navigation>
+{% endblock %}
--- a/cl/api/templates/webhooks-getting-started.html
+++ b/cl/api/templates/webhooks-getting-started.html
@@ -1,6 +1,18 @@
 {% extends "base.html" %}
-{% load static %}
-{% load extras %}
+{% load static extras %}
+
+{% comment %}
+╔═════════════════════════════════════════════════════════════════════════╗
+║                               ATTENTION!                                ║
+║ This template has a new version behind the use_new_design waffle flag.  ║
+║                                                                         ║
+║ When modifying this template, please also update the new version at:    ║
+║ cl/api/templates/v2_webhooks-getting-started.html                       ║
+║                                                                         ║
+║ Once the new design is fully implemented, all legacy templates          ║
+║ (including this one) and the waffle flag will be removed.               ║
+╚═════════════════════════════════════════════════════════════════════════╝
+{% endcomment %}
 
 {% block title %}Getting Started With Webhooks &ndash; CourtListener.com{% endblock %}
 {% block description %}
--- a/cl/assets/tailwind/input.css
+++ b/cl/assets/tailwind/input.css
@@ -87,12 +87,10 @@
   .alert {
     @apply px-4 py-3 text-sm rounded-[10px] border;
   }
-
   .alert-warning {
     @apply alert text-greyscale-950 bg-yellow-50 border-amber-450 border-2 rounded-2xl gap-2.5 p-4 self-stretch;
   }
-
-   .alert-danger {
+  .alert-danger {
     @apply alert bg-red-100 text-red-700 border-red-500;
   }
   .btn {
@@ -124,6 +122,10 @@
   }
   .input-text {
     @apply border border-greyscale-200 rounded-lg h-10 py-2.5 px-3;
+  }
+  /* Screenshots used in Help pages */
+  .screenshot {
+    @apply rounded-2xl block max-w-full;
   }
 
   /* ----------- THIRD PARTY OVERRIDES ----------- */
--- a/cl/assets/tailwind/tailwind.config.js
+++ b/cl/assets/tailwind/tailwind.config.js
@@ -107,6 +107,7 @@ module.exports = {
         lg: ['18px', '28px'],
         'lg-cooper': ['18px', '26px'],
         xl: ['20px', '28px'],
+        'display-xxs': ['18px', '28px'],
         'display-xs': ['24px', '32px'],
         'display-sm': ['30px', '38px'],
         'display-sm-cooper': ['28px', '44px'],


cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3350,140 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+    :return: List of tuples with the following entries:
+    - A flag indicating whether the document needed to be created or updated,
+    - A flag indicating which is set to True when the document was successfully
+    created or updated or when an update was unnecessary,
+    - The primary key of the updated TexasDocument object."""
+    texas_document_merger = functools.partial(
+        merge_texas_document, docket_entry
+    )
+    texas_document_mergers = map(texas_document_merger, documents)
+    return await asyncio.gather(*texas_document_mergers)
```
albertisfu: Here we can use a Celery tasks instead, if you agree with moving these methods to be synchronous.

Also can you also consider to extract the PDF files and store the content in `plain_text`?
The current task we use for it is: `extract_recap_pdf`

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3350,140 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
             "SCOTUS PDF download: Ran into a RequestException. Retrying."
         )
         raise self.retry(exc=exc)
+
+
+async def merge_texas_document(
```
albertisfu: I’d advise against using async methods for code paths that involve database operations, like here and the upstream methods.

We’ve experienced issues related to database inconsistencies when listening to `post_save` signals and triggering Celery tasks with `on_commit`.

See: https://github.com/freelawproject/courtlistener/issues/4623#issuecomment-2814068312

I’m still not entirely sure which exact combination triggers this kind of race condition, but it definitely seems related to `async` methods.

My hypothesis is that it’s a combination of transaction blocks and async connections. Each async operation can run on its own thread, which means a different database connection. That can lead to `on_commit` listening to a different transaction than the one we expect. We'll rely a lot on these signals once we start integrating other services like search and alerts.

What’s your opinion on this? Do you see a strong advantage that would still make it worth using async methods here?

For heavy work like downloading files, we can delegate that to Celery tasks instead.

cl/corpus_importer/tasks.py
albertisfu: Here we can use a Celery tasks instead, if you agree with moving these methods to be synchronous.

Also can you consider to extract the PDF files and store the content in `plain_text`?

The current task we use for it is: `extract_recap_pdf`

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3350,140 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+    """Calculates the sequence numbers for a scraped list of TexasDocketEntry
+    objects to allow consistent matching and merging.
+
+    :param input: The scraped Texas docket data.
+    :return: A list of sequence numbers corresponding to the list of
+    TexasDocketEntry objects in input["case_events"].
+    """
+    return list(
+        [
+            f"{e['date'].isoformat()}-{i}"
```
albertisfu: We should use a zero padded index to properly sort entries:

e.g if we have:

"2024-01-15.1",
    "2024-01-15.2",
    "2024-01-15.10",
    "2024-01-15.11",
    "2024-01-15.3",

sorted:


'2024-01-15.1',
'2024-01-15.10', 
'2024-01-15.11',
'2024-01-15.2',
'2024-01-15.3'

which is wrong.

Using a zero padded index we'll have:


'2024-01-15.001',
'2024-01-15.002',
'2024-01-15.003',
'2024-01-15.010',
'2024-01-15.011'

Also I suggest using the zero padded index per day instead of global, similar as we do in: `calculate_recap_sequence_numbers`

Because if an event is removed or added later on a given date, using a global index makes it more likely to disrupt the sequence of many entries. With an index per day, only the events from that specific day would be affected.

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3350,140 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+    :param sequence_number: The sequence number of the docket entry.
+    :param appellate_brief: Whether the docket entry is an appellate brief.
+    :param input_docket_entry: The docket entry being merged.
+    :return: Tuple with the following entries
+    - A flag indicating whether the docket entry or an attached document needed
+    to be created or updated,
+    - A flag which is set to true when the create/update operations are all
+    either successful or unnecessary,
+    - The primary key of the updated TexasDocketEntry object."""
+    (docket_entry, created) = await TexasDocketEntry.objects.aget_or_create(
```
albertisfu: Here, I suggest using a DB lock to prevent race conditions that can lead to duplicate entries. Since we’ll have an ongoing scraper and also the case mail listener, if both run at the same time for the same case, there is a possibility of a race condition that results in duplicates.

We had this issue with RECAP entries. The solution  there was to add a SELECT … FOR UPDATE on the docket row and use a transaction.atomic() block so that DocketEntry creation happens within the same transaction. That way, if another process tries to add entries to the same docket, it will wait for the first one to complete.

See: `add_create_docket_entry_transaction`. Note that this is an `async` method, this is involved in the errors we have been experiencing when merging RECAPDocuments.

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3350,140 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+        docket=docket,
+        date_filed=input_docket_entry["date"],
+        sequence_number=sequence_number,
```
albertisfu: I think we need to be more careful here. If we rely only on sequence_number, as I mentioned in another comment, there is a possibility that an event is added or removed, which would cause us to match the wrong entry.

What if we first rely on the entry description and/or the main document short description, along with the date, and if that query fails, we fall back to sequence_number?

That way, we can also detect potential duplicate entries and merge them if possible.

Basically, a similar logic to what we have in `get_or_make_docket_entry`.

cl/corpus_importer/tests.py
```diff
@@ -2087,6 +2099,438 @@ def test_merger(self):
...
+    @pytest.mark.asyncio
+    async def test_merge_texas_document_new_document(self):
+        """Can we correctly add a new attachment to an existing docket entry?"""
+        docket_entry = self.docket_coa1_entry
+
+        # Mock PDF download context manager
+        file_mock = MagicMock()
+        self.download_pdf_mock.return_value.__enter__.return_value = file_mock
+
+        input_document = TexasCaseDocument(
```
albertisfu: You can use Dict Factories for this purpose, see:
https://github.com/freelawproject/courtlistener/blob/1ebf468c230d32fe4450965c07e1641fcdbbe827/cl/search/factories.py#L408

cl/corpus_importer/tasks.py
MorganBennetDev: The main motivation for making this async was so that all the PDF files could be downloaded in parallel. Perhaps just have the downloading be asynchronous and then have the DB operations run synchronously? Or do you think it's not worth it (I've asked Brennan if he has any stats on number of attachments per docket)?

Will get started on the `plain_text` extraction. Will this be affected by the fact that `TexasDocument` inherits from `AbstractPDF`?

cl/corpus_importer/tasks.py
MorganBennetDev: That's correct. This was a bit lazy on my part. Will update.

cl/corpus_importer/tasks.py
MorganBennetDev: The choice to use async was mostly just based on trying to make anything that does IO async, but if we know that sync code works and async code sometimes fails (for whatever reason), it's probably best to stick to sync code. Will update.

cl/corpus_importer/tasks.py
albertisfu: > Perhaps just have the downloading be asynchronous and then have the DB operations run synchronously? Or do you think it's not worth it (I've asked Brennan if he has any stats on number of attachments per docket)?

Yeah, we could do that.

However, a few considerations to choose between Celery and asyncio for downloading the PDFs:

Celery has the advantage of being more reliable: if workers die, scheduled or failed tasks can be resumed and the work can be completed.

We also get retries and throttling mechanisms, and it’s easy to scale up.

With asyncio, if the pod dies, that scheduled work won’t be recovered, right?

But also consider the whole workflow since the docket ingestion, entries, documents, PDF extractions. You need to do that work in a celery chain (look at my scotus merger work).


> Will this be affected by the fact that TexasDocument inherits from AbstractPDF?

What impact are you thinking about?

MorganBennetDev: I forgot about that the `plain_text` field is on `AbstractPDF` and not `RECAPDocument`. There shouldn't be any issues.

cl/search/state/texas/factories.py
```diff
@@ -0,0 +1,54 @@
...
+    case_name_full = Faker("case_name", full=True)
+    date_filed = Faker("date_object")
+    case_type = Faker("pystr")
+    parties = List([SubFactory(TexasCasePartyFactory)])
+    trial_court = SubFactory(TexasTrialCourtFactory)
+    case_events = List([SubFactory(TexasDocketEntryFactory)])
+    appellate_briefs = LazyAttribute(
+        lambda d: filter(
+            lambda e: True if random.random() < 0.1 else False, d.case_events
+        )
```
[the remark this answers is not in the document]
MorganBennetDev: /fp There is no return statement here.

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3351,297 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+    output = [
+        merge_texas_document(docket_entry, document) for document in documents
+    ]
+
+    # Perform plaintext extraction on documents that were successfully created
+    # or updated.
+    documents_to_extract = [o[2] for o in output if o[0] and o[1]]
+    # Only spawn the task if we need to extract a document.
+    if len(documents_to_extract) > 0:
+        extract_pdf_document.delay(
```
albertisfu: Here, we’ll be triggering a single `extract_pdf_document` task for all the documents that need extraction. The problem is that, in the upstream method `merge_texas_document`, we’re scheduling another task to download the files: `download_texas_document_pdf`.

This can cause a race condition, where some files may be sent for extraction before the corresponding download task has completed.

To solve this, we first need to ensure that the file has been downloaded.

We can do either of the following:


- Use a Celery `chain` to download all the files in a single task and then chain it to the `extract_pdf_document` task.
- Use a Celery `chain` to download a single file and then chain it with `extract_pdf_document` to extract that single file.

In SCOTUS, I used the single-file approach, which has the advantage of shorter-running tasks instead of long ones, and also allows those tasks to be retried independently more easily.

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3351,297 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+            documents_to_extract, False, False, TexasDocument
+        )
+
+    return output
+
+
+def texas_docket_entry_sequence_numbers(input: TexasCommonData) -> list[str]:
```
albertisfu: Can we move this method to `corpus_importer/utils.py` and rename it to a more generic name? I’m basically using the same method for SCOTUS, so I think it’s a good idea to reuse it for other states in the future as well.

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3351,297 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+
+    :param input: The scraped Texas docket data.
+    :return: A list of sequence numbers corresponding to the list of
+    TexasDocketEntry objects in input["case_events"].
+    """
+    dates = [e["date"].isoformat() for e in input["case_events"]]
+    date_counts: dict[str, int] = {}
+    sequence_numbers = []
+    for entry_date in dates:
+        i = date_counts.get(entry_date, 0)
```
albertisfu: The index should start in 1

```suggestion
        i = date_counts.get(entry_date, 1)
```

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3351,297 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+    - A flag indicating whether the docket entry or an attached document needed
+    to be created or updated,
+    - A flag which is set to true when the create/update operations are all
+    either successful or unnecessary,
+    - The primary key of the updated TexasDocketEntry object."""
+    logger.info(
+        "Merging TexasDocketEntry with sequence number %s into Docket %s",
+        sequence_number,
+        docket.pk,
+    )
```
albertisfu, line 3529: We still need the DB lock using select_for_update here:

https://github.com/freelawproject/courtlistener/blob/15dcc8db786f33673d3e63f6a354f849c548d49a/cl/corpus_importer/tasks.py#L3404

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3351,297 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+    count_matching_entries = docket_entries.count()
+
+    if count_matching_entries == 0:
+        logger.info(
+            "No existing TexasDocketEntry found for sequence number %s on Docket %s. Creating new entry.",
+            sequence_number,
+            docket.pk,
+        )
+        docket_entry = TexasDocketEntry.objects.create(
```
albertisfu: To save a database operation, I suggest only instantiating the TexasDocketEntry and saving it at the end.

```suggestion
        docket_entry = TexasDocketEntry(
```

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3351,297 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+        if matching_sequence_number:
+            logger.info(
+                "Found existing TexasDocketEntry for sequence number %s on Docket %s. Updating entry.",
+                sequence_number,
+                docket.pk,
+            )
+            docket_entry = matching_sequence_number
+            created = False
+        else:
+            logger.info(
```
albertisfu: This could mean that there was a conflict, right? For example, a duplicate entry, or a new entry being introduced or deleted from the docket.

In that case, I think it’s worth reviewing what happened to see if we can adjust something. So logging an error seems best.

```suggestion
            logger.error(
```

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3351,297 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+                "No existing TexasDocketEntry found for sequence number %s on Docket %s. Creating new entry.",
+                sequence_number,
+                docket.pk,
+            )
+            docket_entry = TexasDocketEntry.objects.create(
```
albertisfu:
```suggestion
            docket_entry = TexasDocketEntry(
```

cl/corpus_importer/tests.py
```diff
@@ -2087,6 +2111,628 @@ def test_merger(self):
             )
 
 
+class TexasMergerTest(TestCase):
+    def setUp(self):
```
albertisfu, line 2136: If we want all of these factors to be available in all the tests, we can use the `setUpTestData` class method instead. It is executed once, which will be faster than `setUp`, which runs once per test method.

cl/corpus_importer/tests.py
```diff
@@ -2087,6 +2111,628 @@ def test_merger(self):
...
+            self.extract_pdf_document_patch.start()
+        )
+        self.texas_sc = CourtFactory.create(id="texas_sc")
+        self.texas_cca = CourtFactory.create(id="texas_cca")
+        self.texas_coa1 = CourtFactory.create(id="texas_coa1")
+        self.docket_number_coa1 = "01-25-00011-CV"
+        self.docket_coa1 = DocketFactory.create(
+            court=self.texas_coa1, docket_number=self.docket_number_coa1
+        )
+        self.docket_coa1_entry = TexasDocketEntry.objects.create(
```
albertisfu: Can you please create a `TexasDocketEntryFactory` and use it instead of the model? That way, we can easily get default values in the tests without having to populate all of them.

Then we can add it to `make_dev_data.py`

cl/corpus_importer/tests.py
```diff
@@ -2087,6 +2111,628 @@ def test_merger(self):
...
+        self.download_pdf_patch = patch(
+            "cl.corpus_importer.tasks.download_pdf_in_stream"
+        )
+        self.download_pdf_mock = self.download_pdf_patch.start()
+
+    def tearDown(self):
+        """Tear down patches and remove added objects"""
+        Docket.objects.all().delete()
+        TexasDocketEntry.objects.all().delete()
+        TexasDocument.objects.all().delete()
```
albertisfu: Since these tests are based on TestCase, each test is wrapped in a transaction that is rolled back once the test finishes. Therefore, I think there is no need to delete objects added during the test.

This should speed up tests a bit.

cl/corpus_importer/tests.py
```diff
@@ -2087,6 +2111,628 @@ def test_merger(self):
...
+            str(created_document.document_url)
+            == input_document["document_url"]
+        )
+
+        self.download_task_mock.assert_called_once_with(created_document.pk)
+
+    def test_merge_texas_document_existing_document_no_update(self):
+        """Can we correctly handle an existing document with no update needed?"""
+        docket_entry = self.docket_coa1_entry
+        input_document = TexasCaseDocumentFactory()
```
albertisfu: This is a `DictFactory` so we should rename it to `TexasDocumentDictFactory` this also applies for other `DictFactory` in texas/factories.

cl/corpus_importer/tests.py
```diff
@@ -2087,6 +2111,628 @@ def test_merger(self):
...
+
+        # Create an existing and up to date attachment
+        current_document = TexasDocument.objects.create(
```
albertisfu: Same suggestion as `TexasDocketEntryFactory` let's create a `TexasDocumentFactory` and use it in all the tests that currently are using the model.

cl/corpus_importer/tests.py
```diff
@@ -2087,6 +2111,628 @@ def test_merger(self):
...
+
+        result = download_texas_document_pdf(texas_document.pk)
+
+        assert result is None
+        self.download_pdf_mock.assert_called_once_with(
+            "https://example.com/sample.pdf", texas_document.pk, "texas_"
+        )
+        texas_document.refresh_from_db()
+        assert not texas_document.filepath_local
+
```
albertisfu, line 2775: Can you please add a new test, or add assertions to an existing one, to test the integration of `merge_texas_documents`, specifically to cover `extract_pdf_document`?

Here is an example of what I did for SCOTUS:

https://github.com/freelawproject/courtlistener/blob/15dcc8db786f33673d3e63f6a354f849c548d49a/cl/search/tests/test_scotus_content.py#L151

https://github.com/freelawproject/courtlistener/blob/15dcc8db786f33673d3e63f6a354f849c548d49a/cl/search/tests/test_scotus_content.py#L195

I think you’ll find that it’s necessary to tweak the `microservice` method.

cl/scrapers/tasks.py
```diff
@@ -435,13 +436,14 @@ def find_and_merge_versions(self, pk: int) -> None:
     max_retries=3,
     retry_backoff=10,
 )
-def extract_recap_pdf(
+def extract_pdf_document(
```
albertisfu, line 466: I agree with renaming this method, but we need to consider the following:

- Keep the original `extract_recap_pdf` method as well. If we remove it outright, tasks that are already in the queue will fail, since they were scheduled for `extract_recap_pdf` and the task will no longer exist in the Celery deployment.
- Create an issue to remember to remove the `extract_recap_pdf` task once we know that all queued tasks referring to the old task have been processed.

cl/scrapers/tasks.py
```diff
@@ -435,13 +436,14 @@ def find_and_merge_versions(self, pk: int) -> None:
...
     self,
     pks: int | list[int],
     ocr_available: bool = True,
     check_if_needed: bool = True,
+    model=RECAPDocument,
```
albertisfu: To prevent serialization errors during task initialization, I’d recommend the following pattern:

```suggestion
  model: str = "search.RECAPDocument",
```

cl/scrapers/tasks.py
```diff
@@ -462,26 +464,31 @@ def extract_recap_pdf(
...
+        pks, ocr_available, check_if_needed, model
     )
 
 
-async def extract_recap_pdf_base(
+async def extract_pdf_document_base(
     pks: int | list[int],
     ocr_available: bool = True,
     check_if_needed: bool = True,
+    model=RECAPDocument,
```
albertisfu: Same recommendation as in the previous comment, and here we can do:

```
from django.apps import apps

model_class = apps.get_model(model)
```

albertisfu (changes requested):
Thanks @MorganBennetDev this look good. Just some more changes please.

cl/scrapers/tasks.py
MorganBennetDev: Created #6811 and assigned it to myself.

cl/search/state/texas/factories.py
```diff
@@ -0,0 +1,87 @@
...
+class TexasDocketEntryFactory(DjangoModelFactory):
+    docket = SubFactory(DocketFactory)
+    appellate_brief = Faker("pybool", truth_probability=10)
+    description = Faker("text", max_nb_chars=25)
+    remarks = Faker("text", max_nb_chars=25)
+    disposition = Faker("text", max_nb_chars=25)
+    date_filed = Faker("date_object")
+    entry_type = Faker("pystr", min_chars=3, max_chars=3)
+    sequence_number = LazyAttribute(
+        lambda self: f"{self.date_filed.isoformat()}.{random.randint(1, 3)}"
```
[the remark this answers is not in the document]
MorganBennetDev: /fp Once again there is no return statement here

cl/corpus_importer/tasks.py
```diff
@@ -3179,6 +3194,58 @@ def classify_case_name_by_llm(self, cluster_pk: int, recap_document_id: int):
...
+        # Clear any partial content from previous attempt
+        tmp_file.truncate()
+        with requests.get(
+            url,
+            stream=True,
+            timeout=60,
+            headers={"User-Agent": "Free Law Project"},
+        ) as response:
+            response.raise_for_status()
+            if not is_pdf(response):
```
albertisfu, line 3249: This check is failing for some real Texas documents. The Content-Type header in the documents I found is:

`Application/pdf`

However, the `is_pdf` method checks for an exact match of `application/pdf`.

I think we should update this method in Juriscraper.

Here some examples that failed:
`https://search.txcourts.gov/SearchMedia.aspx?MediaVersionID=ec2ed687-fdea-4e87-b865-3688199757f7&coa=coa01&DT=Other&MediaID=20eca1f0-ee61-4f85-b162-848a7ab205d3`

`https://search.txcourts.gov/SearchMedia.aspx?MediaVersionID=af381195-10c3-4c59-8fb0-67460d389672&coa=coa01&DT=NOA RECD&MediaID=52b72c80-11e6-4f5b-84b2-6020a871f87d`

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3350,284 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+    :return: The primary key of the downloaded TexasDocument instance, or None
+    if the process failed."""
+    try:
+        texas_document = TexasDocument.objects.get(pk=texas_document_pk)
+    except TexasDocument.DoesNotExist:
+        logger.warning(
+            "Texas document PDF download: TexasDocument %s does not exist; skipping.",
+            texas_document_pk,
+        )
+        return None
```
albertisfu, line 3399: In all early returns that return None in a task that has a chained task, it is required to set `self.request.chain = None`; otherwise, the following task will be executed with a potentially incorrect value.

In this case, `extract_pdf_document` would be executed with a `pk` of None, causing the document query to fail in that task.

```suggestion
        self.request.chain = None
        return None
```

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3350,284 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+    - A flag indicating which is set to True when the document was successfully
+    created or updated or when an update was unnecessary,
+    - The primary key of the updated TexasDocument object."""
+    output = [
+        merge_texas_document(docket_entry, document) for document in documents
+    ]
+
+    # Perform plaintext extraction on documents that were successfully created
+    # or updated.
+    documents_to_extract = [o[2] for o in output if o[0] and o[1]]
```
albertisfu: This looks unused now.

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3350,284 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+@transaction.atomic
+def merge_texas_docket_entry(
+    docket: Docket,
+    sequence_number: str,
+    appellate_brief: bool,
+    input_docket_entry: TexasCaseEvent
+    | TexasAppellateBrief
+    | TexasSupremeCourtCaseEvent
+    | TexasSupremeCourtAppellateBrief,
+):
```
albertisfu: The return type hint is missing.

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3364,282 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+                "Failed to download attachment PDF for TexasDocument %s from URL %s.",
+                texas_document.pk,
+                url,
+            )
+            self.request.chain = None
+            return None
+        filename = (
+            f"{texas_document.media_id}-{texas_document.media_version_id}.pdf"
+        )
+        texas_document.filepath_local = File(tmp, name=filename)
```
albertisfu: Here we’re missing setting the `file_size`, `sha1`, and `page_count`. See the SCOTUS documents code for reference:

https://github.com/freelawproject/courtlistener/blob/f7077b0f1c4542c3596bad718fb457194a3a386f/cl/corpus_importer/tasks.py#L3708

For Texas, we already have the `file_size` from the parser. Maybe we can use that, or we can derive it directly from the file, I’m not sure which would be more reliable.

albertisfu: For the sha1, if you look at `download_scotus_document_pdf` and `download_pdf_in_stream` in the SCOTUS PR, you can see that the hash is computed while downloading the document. 

This was suggested by Brennan to avoid loading the file into memory again. You can update your methods here to follow the same approach.

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3364,282 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+    - A flag which is set to true when the create/update operations are all
+    either successful or unnecessary,
+    - The primary key of the updated TexasDocketEntry object."""
+    logger.info(
+        "Merging TexasDocketEntry with sequence number %s into Docket %s",
+        sequence_number,
+        docket.pk,
+    )
+    Docket.objects.select_for_update().get(pk=docket.pk)
+    docket_entries = TexasDocketEntry.objects.filter(
```
albertisfu, line 3531: Would it be possible to update this method to use `.get()` instead of `filter()`, and rely on exceptions rather than doing an extra `count()` query and `first()`?

This would be similar to the optimization we did in the SCOTUS docket entries merger.

cl/corpus_importer/tasks.py
MorganBennetDev: For file size, I think we should rely on the local size after we download. The size we extract from TAMES is rounded to the nearest tenth of a unit so it will be unreliable for larger file sizes.

cl/corpus_importer/tasks.py
```diff
@@ -3179,6 +3197,75 @@ def classify_case_name_by_llm(self, cluster_pk: int, recap_document_id: int):
...
+        (ConnectionError, Timeout),
+        tries=3,
+        delay=0.25,
+        backoff=1,
+    )
+    def download_to_file(tmp_file):
+        tmp_file.seek(0)
+        # Clear any partial content from previous attempt
+        tmp_file.truncate()
+        hasher = hashlib.sha1()
```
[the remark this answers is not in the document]
MorganBennetDev: /ar This is not being used for cryptography

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3370,291 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+            "description": input_document["description"],
+            "media_version_id": input_document["media_version_id"],
+            "document_url": input_document["document_url"],
+        },
+    )
+
+    if (
+        created
+        or str(texas_document.media_version_id)
+        != input_document["media_version_id"]
```
albertisfu, line 3465: After doing some testing, I noticed that we should also check the `filepath_local`. In cases where a previous attempt to download the document failed and the field is still `NULL`, we can try the download and merge the file in future attempts.

```suggestion
        != input_document["media_version_id"] or not texas_document.filepath_local
```

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3370,291 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+    ):
+        texas_document.description = input_document["description"]
+        texas_document.media_version_id = input_document["media_version_id"]
+        texas_document.document_url = input_document["document_url"]
```
albertisfu: I found an issue with some Texas document URLs, for example:

https://search.txcourts.gov/SearchMedia.aspx?MediaVersionID=af381195-10c3-4c59-8fb0-67460d389672&coa=coa01&DT=NOA RECD&MediaID=52b72c80-11e6-4f5b-84b2-6020a871f87d

When trying to re-save the document via the admin, the URLValidator on the URLField raises an error due to the blank space in `NOA RECD`. I think we can solve this by using `urlparse` (and proper URL encoding) to encode the space before saving the URL.

To do this, we’ll need to pull the `document_url` from the default `TexasDocument.objects.get_or_create` logic.

albertisfu (changes requested):
Thanks @MorganBennetDev. I’ve tested this again and found a couple of corner cases that I think are worth handling.

cl/corpus_importer/tasks.py
MorganBennetDev: Do you know what docket this URL comes from? I'm curious what's happening on TAMES's end to cause this.

albertisfu: Sure, here is the case:
https://search.txcourts.gov/Case.aspx?cn=01-25-00994-CR&coa=coa01


The related document is the one highlighted:
Screenshot 2026-02-05 at 11 47 44 p m

However, after checking deeper, it seems that the issue might be in Juriscraper?

Here is the example:

tests/examples/state/texas/common/01-25-00994-CR.json

MorganBennetDev: Interesting. The Texas parsers assume that `href` attributes will be valid URLs that we can use to make requests and just pulls them out verbatim, however that seems to not necessarily be the case. Clicking that link, my browser correctly converts the space to `%20`, which gets the correct document. Should I submit a PR to fix this in Juriscraper or should we try to handle this in CL?

albertisfu: Yeah, for making requests it’s a valid URL. However, it’s better to properly handle spaces to prevent validators errors.

I think it’s better to handle this in Juriscraper so we can reuse the same method to return URLs for other parsers in the future.

MorganBennetDev: Got it. Will start working on this in Juriscraper.

MorganBennetDev: Submitted PR to Juriscraper and assigned Brennan to review [#1795](https://github.com/freelawproject/juriscraper/pull/1795).

cl/corpus_importer/tasks.py
```diff
@@ -3283,3 +3370,292 @@ def download_qp_scotus_pdf(self, docket_id: int) -> None:
...
+        texas_document = TexasDocument.objects.get(pk=texas_document_pk)
+    except TexasDocument.DoesNotExist:
+        logger.warning(
+            "Texas document PDF download: TexasDocument %s does not exist; skipping.",
+            texas_document_pk,
+        )
+        self.request.chain = None
+        return None
+
+    url = texas_document.document_url
```
albertisfu: The `document_url` field was re-named in https://github.com/freelawproject/courtlistener/pull/6807

Could you please update here and in all other outdated places (factories too).

MorganBennetDev: Just pushed

PR #6762, opened 2026-01-10
fix(alerts): Enable waffle switch for stats in SearchAlertsOAESTests

mlissner:
## Fixes
This fixes the flaky test `test_group_alerts_and_hits` that was failing intermittently in CI:
https://github.com/freelawproject/courtlistener/actions/runs/20868964667/job/59966559924?pr=6681

## Summary
The `test_group_alerts_and_hits` test (and other tests in `SearchAlertsOAESTests` that check stats) was flaky because it checked for stats recorded in Redis, but the `tally_stat` function requires the "increment-stats" waffle switch to be active to record stats. 

This also adds a cache prefix to various tests that would otherwise be flakey if run in parallel. This is needed because Django Waffle has a cache of its own.

Without these changes, in `test_group_alerts_and_hits`, the `tally_stat` function returns early at line 90-91 of `cl/stats/utils.py`:
```python
if not switch_is_active("increment-stats"):
    return
```

This resulted in the stat count being 0 instead of the expected 2.

### Changes:
1. Added `@override_switch("increment-stats", active=True)` to the `SearchAlertsOAESTests` class
2. Added `@override_settings(WAFFLE_CACHE_PREFIX="test_<name>")` to ALL test classes that use `@override_switch` to prevent cache conflicts between parallel test runs:
   - `StatTests` in `cl/stats/tests.py`
   - `PrometheusIntegrationTestBase` in `cl/stats/tests.py`
   - `SearchAlertsOAESTests` in `cl/alerts/tests/tests.py`
   - `RECAPAlertsSweepIndexTest` in `cl/alerts/tests/tests_recap_alerts.py`

## Deployment

**This PR should:**

🤖 Generated with [Claude Code](https://claude.ai/code)


mlissner:
Marking as P1 since the flakey tests are driving us all crazy.

albertisfu (approved):
Looks good to me, auto-merge enabled.

Thank you!

PR #6766, opened 2026-01-11
feat(recap): Add X-Ray redaction checking to PDF ingestion pipeline

### LINKED ISSUE #6691: Add X-Ray to CourtListener ingestion pipeline
MVP is to:
 - Add X-Ray as a dependency
 - Check PACER docs for bad redactions during ingestion
 - Send an email to admins about the problem, including the text and bbox.

Eventually:
 - Add a boolean about the problem to the DB?
 - Automatically notify the filer? 
 - Other stuff too.

mlissner:
## Fixes

Fixes #6691

## Summary

This PR integrates Doctor's X-Ray endpoint into the RECAP PDF ingestion pipeline to detect bad redactions (text visible under redaction boxes). When bad redactions are found, an email alert is sent to `redactions@free.law` with details about the exposed text and its location.

**Changes:**
- Add `BAD_REDACTION_EMAIL` Django setting (configurable, default: `redactions@free.law`)
- Add `check-redactions` endpoint config for Doctor microservice (`/utils/check-redactions/pdf/`)
- Add `check_redactions_service()` wrapper with retry logic in `microservice_utils.py`
- Add `send_bad_redaction_email()` helper function in `recap/tasks.py`
- Integrate redaction check into `process_recap_pdf()` after page count
- Add email template at `cl/alerts/templates/emails/bad_redaction_alert.txt`
- Add 5 tests covering the redaction checking feature

**Key behavior:**
- Redaction check runs after page count but before document save
- If bad redactions are found, email is sent with document details, exposed text, and bounding boxes
- Redaction check failures are logged but don't block PDF ingestion

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

1. No special deployment steps required.

### DIFF
--- a/cl/recap/tasks.py
+++ b/cl/recap/tasks.py
@@ -11,6 +11,7 @@ from multiprocessing import process
 from typing import Any
 from zipfile import ZipFile
 
+import httpx
 import requests
 from asgiref.sync import async_to_sync, sync_to_async
 from botocore import exceptions as botocore_exception


cl/settings/project/email.py
```diff
@@ -20,6 +20,7 @@
 
 BCC_EMAIL_ADDRESS = "bcc@free.law"
 PARTNERSHIP_EMAIL_ADDRESS = "partnerships@free.law"
+BAD_REDACTION_EMAIL = env("BAD_REDACTION_EMAIL", default="redactions@free.law")
```
mlissner: This doesn't need to be an env.

cl/settings/project/microservices.py
```diff
@@ -62,6 +62,10 @@
         "url": f"{DOCTOR_HOST}/utils/document-number/pdf/",
         "timeout": 60,
     },
+    "check-redactions": {
+        "url": f"{DOCTOR_HOST}/utils/check-redactions/pdf/",
+        "timeout": 60 * 5,
```
mlissner: Ten seconds is enough.

cl/recap/tasks.py
```diff
@@ -272,6 +275,46 @@ async def mark_pq_status(
     return pq.status, getattr(pq, message_property_name)
 
 
+async def send_bad_redaction_email(
```
mlissner: Move this to the cl.recap.utils file.

cl/recap/tasks.py
```diff
@@ -272,6 +275,46 @@ async def mark_pq_status(
...
+    rd: RECAPDocument,
+    redactions: dict,
+) -> None:
+    """Send email alert when bad redactions are detected.
+
+    :param rd: The RECAPDocument with bad redactions
+    :param redactions: Dict of page_num -> list of {text, bbox}
+    """
+    from django.core.mail import send_mail
+    from django.template import loader
```
mlissner: Never do inline imports. Move these to the top.

cl/recap/tests/tests.py
mlissner, line 1: There's too much repetition in these tests. Pull out common patterns so the testing logic is more clear.

mlissner, line 1: Don't do inline comments.

mlissner:
Throwing this into Whitney because it's small, old, and would have found Taylor's phone number. :)

cl/recap/tasks.py
```diff
@@ -469,6 +473,24 @@ async def process_recap_pdf(pk, subdocket_replication: bool = False):
                 )
             rd.file_size = rd.filepath_local.size
 
+            # Check for bad redactions using X-Ray
```
albertisfu: Two thoughts about this:

- Is there a reason to execute this before saving the document? One downside of doing so is that if the RD is created within the same `process_recap_pdf` workflow, you won’t get the RECAPDocument ID in the email because it will still be `None`.

- I assume the `check_redactions_service` endpoint can take several seconds to complete, especially for large documents. I don’t see that we need anything from the response before saving the document. Would it be better to move this process to a Celery task, similar to the `extract_recap_pdf` task? That way, the entire process, including sending the email,  would run in the background without blocking the `process_recap_pdf` thread.

cl/recap/tasks.py
```diff
@@ -469,6 +473,24 @@ async def process_recap_pdf(pk, subdocket_replication: bool = False):
...
+                    # Response: {"error": false, "results": {"1": [{"text": "...", "bbox": ...}]}}
+                    if not redaction_data.get("error") and redaction_data.get(
+                        "results"
+                    ):
+                        await send_bad_redaction_email(
+                            rd, redaction_data["results"]
+                        )
+            except Exception as e:
+                # Log but don't fail ingestion if redaction check fails
+                logger.warning(
```
albertisfu: I believe it's better to use `logger.error` instead? I think it’s worth letting it fail loudly, at least in the beginning, so we can identify potential issues and fix them if possible.

mlissner:
Good call about moving it to Celery. Done. I also took the opportunity to strip out some of the async stuff, which was no longer needed. I have to admit the code sure is nicer without it. 

Back to you, @albertisfu, for further review.

cl/recap/tasks.py
```diff
@@ -3637,3 +3641,31 @@ def do_recap_document_fetch(epq: EmailProcessingQueue, user: User) -> None:
         process_recap_email.si(epq.pk, user.pk),
         extract_pdf_document.s(),
     ).apply_async()
+
+
+@app.task(
+    autoretry_for=(Exception,),
```
albertisfu: We can auto-retry only for potential known errors in a microservice call.

```suggestion
    autoretry_for=(
        httpx.ConnectError,
        httpx.ConnectTimeout,
        httpx.ReadTimeout,),
```

cl/recap/tasks.py
```diff
@@ -3637,3 +3641,31 @@ def do_recap_document_fetch(epq: EmailProcessingQueue, user: User) -> None:
...
+    Runs as a background Celery task so it doesn't block PDF ingestion.
+
+    :param rd_pk: The PK of the RECAPDocument to check
+    """
+    rd = RECAPDocument.objects.get(pk=rd_pk)
+    try:
+        redaction_response = async_to_sync(check_redactions_service)(rd)
+    except Exception:
+        logger.error("Redaction check failed for RD %s", rd_pk, exc_info=True)
+        raise
```
albertisfu, line 3669: And then we can just call the microservice without catching any errors.

```suggestion
    redaction_response = async_to_sync(check_redactions_service)(rd)
```

Transient httpx errors will be retried, while other unknown errors will be raised. No logging is required, and there is no need to catch the error since a `microservice` failure won't mess up the document ingestion.

mlissner: I like it. Clever, thank you!

mlissner:
I accepted your excellent suggestions and I'm setting to merge. Seems like this must be ready! Thank you!

albertisfu (approved):
Looks great. Merging...

Issue #6678, opened 2025-12-19
Block invalid filters on the API


mlissner:
### Describe the issue.

The Django REST Framework ignores invalid filters. When you apply a filter to a query that isn't configured,  an unfiltered query set is requested and a huge hit to our DB ensues. 

This used to be an uncommon problem, but vibe coders have officially made it worse over the past few weeks as their coding agents do invalid queries.

I asked in the django-filters repo if they'd consider changing this: 

https://github.com/carltongibson/django-filter/issues/1740#issuecomment-3674576312

Alas, they won't, and that's fine, but they pointed me to this post that seems to offer a very simple way to do this: 

https://github.com/carltongibson/django-filter/discussions/1719

We should see how hard that'd be and give it a go. 

## Roll Out

The easiest way to roll this out is to log problematic queries at first, then enable it fully after we've contacted and resolved issues. That will help us identify bugs in our approach as well. Logging to redis feels like a simple way to go, maybe we can access our logs in loki though and that will work too.

### Related Issue

https://github.com/carltongibson/django-filter/issues/1740#issuecomment-3674576312

### What type of issue is this?

To Do

mlissner:
This is important b/c it should help prevent downtimes. I expect when we do this it will break some implementations, and we'll have to help them fix things.

mlissner:
This is deployed. I'm re-opening as a reminder. Here's the schedule:

1. Today: Launched.
2. In two weeks, on Feb. 3rd: 
   - Move to blocking mode by setting `BLOCK_UNKNOWN_FILTERS` to `on` in our secrets, and deploying. 
   - Remove the cronjob so it doesn't email anymore. Things are now throwing errors.
4. A week or so later, remove the logging code from the PR so that it's not around forever, and remove the `BLOCK_UNKNOWN_FILTERS` var, since it's enabled by default.

mlissner:
OK, we're now in blocking mode, and the cronjob is deleted. Whew!

There were still about 30-40 people per day that were getting the emails. Hopefully this won't affect too many people.

mlissner:
https://github.com/freelawproject/courtlistener/issues/6771 is the issue to remove the unused blocking code. It's a clean up task. I've scheduled it for next week.

PR #6769, opened 2026-01-11
feat(api): Block invalid API filter parameters

mlissner:
## Fixes
Fixes: #6678

## Summary
This PR adds detection and blocking of unknown/invalid API filter parameters. This is needed because vibe coders are bringing down our system with invalid filters. This PR makes it so we check all filters if they're invalid. If so, we log the errors to Redis so we can send the user a nightly email, and we block the request if an env is set to true. 

**Key changes:**
- Add `UnknownFilterParamValidationBackend` filter backend that validates query parameters against the view's FilterSet
- Support nested RelatedFilter validation (e.g., `cluster__docket__court` is valid for OpinionFilter)
- Log unknown parameters to Redis with user, endpoint, param name, count, and timestamps
- Add `BLOCK_UNKNOWN_FILTERS` setting to control behavior (log-only vs block with 400)
- Add `cl_notify_bad_api_filter_users` management command to send email notifications
- Fix SQLCommenter recursion when accessing unevaluated `SimpleLazyObject` user

**Behavior:**
- When `BLOCK_UNKNOWN_FILTERS=False` (default): Unknown params are logged to Redis, request proceeds normally
- When `BLOCK_UNKNOWN_FILTERS=True`: Returns 400 error with list of unknown parameters

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

1. Set `BLOCK_UNKNOWN_FILTERS=on` in environment when ready to enforce blocking (can start with `off` for logging-only)


cl/api/utils.py
```diff
@@ -91,17 +91,269 @@ class HyperlinkedModelSerializerWithId(serializers.HyperlinkedModelSerializer):
...
 
 
+class UnknownFilterParameterError(Exception):
+    """Raised when unknown filter parameters are detected in API requests."""
+
+    def __init__(self, unknown_params: set[str]) -> None:
+        self.unknown_params = unknown_params
+        super().__init__(
+            f"Unknown filter parameters: {sorted(unknown_params)}"
+        )
```
mlissner: This appears to be unused.

cl/api/utils.py
```diff
@@ -91,17 +91,269 @@ class HyperlinkedModelSerializerWithId(serializers.HyperlinkedModelSerializer):
...
+        # Dynamic fields
+        "fields",
+        "omit",
+        # Counting (v4)
+        "count",
+        # Search API specific
+        "type",
+        "q",
+        "court_id",
+        "highlight",
```
mlissner: Do not apply any of this to the search API.

cl/api/utils.py
```diff
@@ -91,17 +91,269 @@ class HyperlinkedModelSerializerWithId(serializers.HyperlinkedModelSerializer):
...
+    :param filterset_class: The FilterSet class to extract parameters from.
+    :return: Set of all valid parameter names.
+    """
+    if filterset_class is None:
+        return set()
+
+    valid_params: set[str] = set()
+
+    for filter_name in filterset_class.base_filters:
+        valid_params.add(filter_name)
```
mlissner: Does this work properly? If this works, why not just do:

```
return set(filterset_class.base_filters)
```

cl/api/utils.py
```diff
@@ -91,17 +91,269 @@ class HyperlinkedModelSerializerWithId(serializers.HyperlinkedModelSerializer):
...
+    :param endpoint: The API endpoint/view name that was accessed.
+    :param bad_params: Set of invalid parameter names that were used.
+    """
+    if user_id is None:
+        # Can't notify anonymous users, skip logging
+        return
+
+    r = get_redis_interface("STATS")
+    pipe = r.pipeline()
+    current_time = datetime.now(UTC).isoformat()
```
mlissner: Use Django's built in `now()` API.

cl/api/utils.py
```diff
@@ -91,17 +91,269 @@ class HyperlinkedModelSerializerWithId(serializers.HyperlinkedModelSerializer):
...
+                    "unknown_params": sorted(unknown_params),
+                }
+            )
+        else:
+            # Store to Redis for later notification
+            log_bad_filter_params_to_redis(
+                user_id=user_id,
+                endpoint=endpoint,
+                bad_params=unknown_params,
+            )
```
mlissner: When we block, let's also log. So change this to log and then *also* block if in blocking mode.

cl/api/utils.py
```diff
@@ -91,17 +91,269 @@ class HyperlinkedModelSerializerWithId(serializers.HyperlinkedModelSerializer):
...
+            endpoint = parts[4]
+            param = parts[5]
+            data = r.hgetall(key)
+            results.append(
+                {
+                    "endpoint": endpoint,
+                    "param": param,
+                    "count": int(data.get("count", 0)),
+                    "first_seen": data.get("first_seen", ""),
+                    "last_seen": data.get("last_seen", ""),
```
mlissner: Return a datetime object here, not a string.

cl/api/utils.py
```diff
@@ -91,17 +91,269 @@ class HyperlinkedModelSerializerWithId(serializers.HyperlinkedModelSerializer):
...
+    r = get_redis_interface("STATS")
+    pattern = f"{BAD_FILTER_PARAMS_PREFIX}:user:{user_id}:*"
+    keys_to_delete = list(r.scan_iter(match=pattern))
+
+    if keys_to_delete:
+        return r.delete(*keys_to_delete)
+    return 0
+
+
 class DisabledHTMLFilterBackend(RestFrameworkFilterBackend):
```
mlissner, line 310: This work should be implemented as a separate filter class, rather than cramming it into `DisabledHTMLFilterBackend`.

cl/api/templates/emails/bad_api_filter_params_email.txt
```diff
@@ -0,0 +1,54 @@
+Hello{% if name %} {{ name }}{% endif %},
+
+We're writing to let you know that we've detected API requests from your
+account that include filter parameters our system doesn't recognize. We wanted
+to give you advance notice because we'll soon be updating our API to reject
+requests that contain unknown filter parameters.
```
mlissner: Let's do this, but clean up the spacing, please:

```suggestion
We're writing to let you know that API requests from your
account include invalid GET parameters our system doesn't recognize. In the coming days, we will begin returning HTTP 400 errors for such requests. 

This email is your advance notice and your opportunity to address this issue.
```

cl/api/templates/emails/bad_api_filter_params_email.txt
```diff
@@ -0,0 +1,54 @@
...
+
+We apologize for any inconvenience this may cause. This change is necessary to
+protect the integrity of our system and ensure that all API users receive
+accurate, filtered results. When invalid filter parameters are used, they are
+silently ignored, which can lead to unexpected results—often returning far more
+data than intended.
```
mlissner: Let's try this:

```suggestion
data than intended. Previously, this was not a common problem, but
a plethora of vibe-coded solutions now regularly trigger very 
heavy usage of our system that we need to urgently block.
```

cl/api/templates/emails/bad_api_filter_params_email.txt
```diff
@@ -0,0 +1,54 @@
...
+
+    https://github.com/freelawproject/courtlistener/discussions/
+
+Our community and team members are happy to help you resolve any issues.
+
+Thank you for using CourtListener's API, and we appreciate your understanding
+as we work to improve the reliability of our services.
+
+
+Best regards,
```
mlissner, line 63:
```suggestion

This email will be re-sent with daily updates for two weeks until
we begin blocking these requests instead of simply alerting about 
them.

We hope you understand and we appreciate your assistance making our APIs work well for all of our users.

Best regards,
```

cl/api/templates/emails/bad_api_filter_params_email.txt
```diff
@@ -0,0 +1,54 @@
+Hello{% if name %} {{ name }}{% endif %},
+
+We're writing to let you know that we've detected API requests from your
```
mlissner: Try this with the text wrapping properly: 
```suggestion
We're writing to let you know that in the last 24 hours we detected API requests from your
```

cl/api/management/commands/notify_bad_api_filter_users.py
```diff
@@ -0,0 +1,166 @@
...
+            action="store_true",
+            default=False,
+            help="Don't send any emails, just show what would be sent.",
+        )
+        parser.add_argument(
+            "--clear-after-send",
+            action="store_true",
+            default=False,
+            help="Clear Redis records after sending emails (use with caution).",
+        )
```
mlissner, line 41:
```suggestion
        parser.add_argument(
            "--clear-after-send",
            action="store_true",
            default=False,
            help="Clear Redis records after sending emails (use with caution to prevent repeat emails).",
        )
```

mlissner (commented):
This is OK, but needs a lot of changes. Please review and re-submit.

I have not yet reviewed the tests.

cl/api/tests.py
```diff
@@ -4406,3 +4413,226 @@ async def test_docket_list_with_bankruptcy_information_in_fields(
...
+        # Clear any existing bad filter params for the test user
+        self.r = get_redis_interface("STATS")
+        self._clear_test_redis_keys()
+
+    def tearDown(self) -> None:
+        self._clear_test_redis_keys()
+
+    def _clear_test_redis_keys(self) -> None:
+        """Clear Redis keys for the test user."""
+        pattern = f"api:bad_filter_params:user:{self.user.pk}:*"
```
mlissner, line 4457: Throughout this file, this should import the key instead of repeating it.

cl/api/tests.py
```diff
@@ -4406,3 +4413,226 @@ async def test_docket_list_with_bankruptcy_information_in_fields(
...
+            reverse("court-list", kwargs={"version": "v4"}),
+            {"bad_param": "value"},
+        )
+
+        self.assertEqual(response.status_code, HTTPStatus.OK)
+        # No Redis keys should be created for anonymous users
+        # Check for any keys (anonymous would have None as user_id)
+        pattern = "api:bad_filter_params:user:None:*"
+        keys = list(self.r.scan_iter(match=pattern))
+        self.assertEqual(len(keys), 0)
```
mlissner, line 4599: Factor this out of the tests so it's not repeated so much.

cl/api/tests.py
```diff
@@ -4406,3 +4413,226 @@ async def test_docket_list_with_bankruptcy_information_in_fields(
...
+        self.assertIn("first_seen", data)
+        self.assertIn("last_seen", data)
+
+    def test_anonymous_users_not_logged(self) -> None:
+        """Verify that anonymous user requests don't create Redis entries."""
+        # Don't login - make anonymous request
+        response = self.client.get(
+            reverse("court-list", kwargs={"version": "v4"}),
+            {"bad_param": "value"},
+        )
```
mlissner, line 4592: Factor this out of the tests so it's not repeated so much.

mlissner (commented):
One more round of comments.

cl/api/utils.py
```diff
@@ -91,6 +91,214 @@ class HyperlinkedModelSerializerWithId(serializers.HyperlinkedModelSerializer):
...
+    filter_instance = base_filters.get(prefix)
+    if filter_instance is None:
+        return False
+
+    if not isinstance(filter_instance, RelatedFilter):
+        return False
+
+    # Recursively validate rest against the related filterset
+    related_filterset = filter_instance.filterset
+    return is_valid_filter_param(rest, related_filterset)
```
albertisfu: I found that this recursion can introduce a vulnerability that could lead to a DOS attack.

Since this logic recursively checks related filters and has no stop condition other than finishing all the provided filters, a request with repeatedly nested but valid filters could cause it to run indefinitely until a maximum recursion error is raised.

For instance, consider the following API request:

`/api/rest/v4/dockets/?clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket__clusters__docket=69638964`


It’s a valid request, but this is checking filters recursively and leads to a maximum recursion error.

I think we can address this issue in one of the following ways:

- Track nested filters during recursion and disallow circular references, for example:
clusters → dockets → clusters
If a filter has already been seen before, we simply treat it as invalid and stop processing it.


- Add a max_recursion parameter, for example 4, and pass a recursion counter so we only allow up to four levels of recursion. I don’t think we have nested filters with more than three levels in practice.

cl/api/utils.py
```diff
@@ -91,6 +91,214 @@ class HyperlinkedModelSerializerWithId(serializers.HyperlinkedModelSerializer):
...
+    # Split on first __ to get prefix and rest
+    prefix, rest = param.split(LOOKUP_SEP, 1)
+
+    # Check if prefix is a RelatedFilter
+    filter_instance = base_filters.get(prefix)
+    if filter_instance is None:
+        return False
+
+    if not isinstance(filter_instance, RelatedFilter):
+        return False
```
albertisfu: We can remove the `filter_instance is None` if since it's already cover by the `isinstance` check.
```suggestion
    if not isinstance(filter_instance, RelatedFilter):
        return False
```

cl/lib/sqlcommenter.py
```diff
@@ -77,12 +78,23 @@ def get_context(self) -> dict[str, Any]:
...
+            if isinstance(user_obj, SimpleLazyObject):
+                # Access _wrapped to check if lazy object has been evaluated.
+                # Once evaluated, it proxies to the User model.
+                if (
+                    user_obj._wrapped is not empty  # type: ignore[attr-defined]
+                    and user_obj.is_authenticated  # type: ignore[attr-defined]
+                ):
+                    user = user_obj.pk  # type: ignore[attr-defined]
+            elif user_obj.is_authenticated:
+                user = user_obj.pk
```
albertisfu, line 97: This fix is already in prod. We can remove the code here.

https://github.com/freelawproject/courtlistener/blob/main/cl/lib/sqlcommenter.py

cl/lib/tests.py
```diff
@@ -1715,6 +1716,44 @@ def test_get_context_truncates_path(self):
 
         self.assertEqual(result["url"], "/very/long…")
 
+    def test_get_context_with_unevaluated_lazy_user(self) -> None:
```
albertisfu, line 1720: Also we can remove all the changes in this file since they're in main now.

cl/api/utils.py
mlissner: The second solution seems simple enough, let's do that.

cl/lib/sqlcommenter.py
mlissner: Darn, I thought I played the git game properly to make this work properly. Not this time. Agreed, let's remove this.

mlissner: Ah, I did. Just had to rebase! I needed this change to test locally so that's why it was here briefly...

mlissner:
OK, I added a depth limit to the recursive loop (nice catch), rebased to nuke the SQL commenter stuff, and fixed your other little things. Let me know what you think!

albertisfu (approved):
Thanks! The recursion issue is solved for `is_valid_filter_param`.

However, I found a similar issue is also present in all the filters that use
`rest_framework_filters.RelatedFilter`.

That [library](https://github.com/philipn/django-rest-framework-filters) hasn’t been updated in years, so I’m not sure whether it’s worth suggesting a fix there.

Once this is merged, we need to set a cron job, right?

mlissner:
Yep, I'll create a cronjob and get it landed so that it can start today. 

I agree that it's not likely that the upstream package is going to do anything about a DOS problem, but we should file it. I think it's probably unethical not to even if we know it'll be ignored. 

This is set to auto-merge. Thank you!

PR #6639, opened 2025-12-05
feat(stats): introduce prometheus client

elisa-a-v:
#6636 

## Summary

This PR adds `django-prometheus` to CourtListener, which exports metrics by exposing an endpoint (`/prometheus/metrics`) for Prometheus to scrape.

The package includes a bunch of built-in metrics and allows the use of the python client (`prometheus_client`) to extend and introduce custom metrics, so as a POC we also introduce a custom metric for search queries with two label keys:
- `query_type` (semantic vs keyword)
- `method` (web vs api) this one wasn't mentioned in the issue but it seems useful and since it has low cardinality it [doesn't add too much storage size](https://prometheus.io/docs/practices/naming/#labels).

Built-in metrics have a namespace of `courtlistener` as defined by `PROMETHEUS_METRIC_NAMESPACE`. It seems like the custom metrics don't use this namespace; instead of using the same one, I used `cl_` to differentiate between built-in and custom metrics. What do we think about this?

<details>
<summary>Example of the endpoint response including the custom metric for search queries</summary>

```
# HELP python_gc_objects_collected_total Objects collected during gc
# TYPE python_gc_objects_collected_total counter
python_gc_objects_collected_total{generation="0"} 20118.0
python_gc_objects_collected_total{generation="1"} 12526.0
python_gc_objects_collected_total{generation="2"} 987.0
# HELP python_gc_objects_uncollectable_total Uncollectable objects found during GC
# TYPE python_gc_objects_uncollectable_total counter
python_gc_objects_uncollectable_total{generation="0"} 0.0
python_gc_objects_uncollectable_total{generation="1"} 0.0
python_gc_objects_uncollectable_total{generation="2"} 0.0
# HELP python_gc_collections_total Number of times this generation was collected
# TYPE python_gc_collections_total counter
python_gc_collections_total{generation="0"} 2595.0
python_gc_collections_total{generation="1"} 235.0
python_gc_collections_total{generation="2"} 6.0
# HELP python_info Python platform information
# TYPE python_info gauge
python_info{implementation="CPython",major="3",minor="13",patchlevel="10",version="3.13.10"} 1.0
# HELP process_virtual_memory_bytes Virtual memory size in bytes.
# TYPE process_virtual_memory_bytes gauge
process_virtual_memory_bytes 2.339135488e+09
# HELP process_resident_memory_bytes Resident memory size in bytes.
# TYPE process_resident_memory_bytes gauge
process_resident_memory_bytes 4.8422912e+08
[...]

PR #6776, opened 2026-01-12
feat(stats): Add Prometheus metrics for monitoring

### LINKED ISSUE #6637: Add more custom metrics in prometheus
### Describe the issue.

Refer to [this comment](https://github.com/freelawproject/courtlistener/issues/6260#issuecomment-3514181033)

The metrics for search queries will be included in #6636

### Related Issue

_No response_

### Screenshots

_No response_

To Do

### LINKED ISSUE #6638: Introduce celery queue monitoring with prometheus
### Describe the issue.

Refer to [this comment](https://github.com/freelawproject/courtlistener/issues/6260#issuecomment-3508666708)

### Related Issue

_No response_

### Screenshots

_No response_

To Do

elisa-a-v:
## Summary

- Adds Prometheus metrics for comprehensive platform monitoring, building on the existing `cl_search_queries_total` counter
- Implements custom `CeleryQueueCollector` for queue length monitoring that works correctly with multiple gunicorn workers (computes values at scrape time rather than storing in process-local memory)
- Instruments search duration, alerts, webhooks, and account lifecycle events
- Skip debug webhook events in Prometheus metrics to match logging behavior
- Make Celery queue collector registration idempotent via a small helper + unit test

**Metrics implemented:**
- `cl_search_queries_total` - Counter for search queries (labels: `query_type`, `method`)
- `cl_search_duration_seconds` - Histogram for search latency (labels: `query_type`, `method`)
- `cl_alerts_sent_total` - Counter for alerts (labels: `alert_type`)
- `cl_webhooks_sent_total` - Counter for webhooks (labels: `event_type`)
- `cl_accounts_created_total` - Counter for account creation
- `cl_accounts_deleted_total` - Counter for account deletion
- `cl_celery_queue_length` - Gauge for celery queue lengths (via custom collector)

Closes #6637
Closes #6638

## Test plan

- [ ] Run `make test p=cl/stats/tests.py` to verify all new metric tests pass
- [ ] Start local dev server and verify `/monitoring/prometheus/metrics` endpoint returns all new metrics
- [ ] Verify celery queue metrics appear with correct queue labels
- [ ] Test search queries (web and API) and confirm duration histogram is populated
- [ ] Trigger an alert and verify `cl_alerts_sent_total` increments
- [ ] Trigger a webhook and verify `cl_webhooks_sent_total` increments


cl/alerts/management/commands/cl_send_recap_alerts.py
mlissner, line 1: As far as I can tell, doing a prometheus metric in a django command won't accomplish anything since it isn't scrapable. Do you know of a solution to this? I don't.

cl/alerts/tasks.py
mlissner, line 1: Code looks good, but I don't know how we get metrics from celery workers. Hm. This code will be running in a worker that won't get scraped. Always something to make things trickier!

cl/api/utils.py
```diff
@@ -1123,6 +1125,18 @@ def update_webhook_event_after_request(
...
             handle_webhook_events(results, webhook_event.webhook.user)
+
+            # Track webhook metrics in Prometheus
+            event_type_map = {
+                WebhookEventType.DOCKET_ALERT: "webhooks.sent.docket_alert",
+                WebhookEventType.SEARCH_ALERT: "webhooks.sent.search_alert",
+                WebhookEventType.RECAP_FETCH: "webhooks.sent.recap_fetch",
+                WebhookEventType.OLD_DOCKET_ALERTS_REPORT: "webhooks.sent.old_docket_alerts_report",
+                WebhookEventType.PRAY_AND_PAY: "webhooks.sent.pray_and_pay",
+            }
```
mlissner: Are these better-handled as labels for a `webhooks_sent` metric?

cl/lib/elasticsearch_utils.py
```diff
@@ -2387,7 +2387,15 @@ def fetch_es_results(
         multi_search = multi_search.add(main_query).add(main_doc_count_query)
         if child_total_query:
             multi_search = multi_search.add(child_total_query)
+        start_time = time.perf_counter()
         responses = multi_search.execute()
+        duration = time.perf_counter() - start_time
+
+        # Record search duration metric
+        from cl.stats.metrics import record_search_duration
```
mlissner: Does this need to be inline here?

cl/stats/metrics.py
```diff
@@ -1,23 +1,118 @@
...
+        for queue in settings.CELERY_QUEUES:
+            try:
+                length = get_queue_length(queue)
+                gauge.add_metric([queue], length)
+            except Exception:
+                logger.exception(
+                    "Failed to get queue length for %s",
+                    queue,
+                )
+        yield gauge
```
mlissner, line 75: It might be better just to let Sentry catch and deal with this, since it gives us lots of nice functionality. You can send the exception to sentry inside this exception by importing the sentry lib.

cl/alerts/management/commands/cl_send_recap_alerts.py
```diff
@@ -717,7 +717,11 @@ def query_and_send_alerts(
         )
 
     # Log and tally the total alerts sent
-    tally_stat("alerts.sent", inc=total_alerts_sent_count)
+    tally_stat(
+        "alerts.sent",
+        inc=total_alerts_sent_count,
+        prometheus_handler_key="alerts.sent.search",
```
mlissner: Could we do a label here for search's `type` parameter, so we can see if we're sending more opinion alerts, recap alerts, etc?

cl/users/signals.py
```diff
@@ -205,3 +207,13 @@ def webhook_created_or_updated(
             ):
                 return
         notify_new_or_updated_webhook.delay(instance.pk, created=False)
+
+
+@receiver(
```
mlissner: Can you double check? I think we'll want to handle this manually because we don't actually delete the model, we mark it as deleted. If you check the view that does the delete, it should be pretty clear.

mlissner (changes requested):
Looks good. I've a couple larger questions, and a few tweaks and thoughts. Thank you!

cl/alerts/management/commands/cl_send_recap_alerts.py
elisa-a-v: Prometheus actually supports monitoring 3 types of services: online-serving systems, offline processing, and batch jobs. We could [push to a PushGateway](https://prometheus.io/docs/instrumenting/pushing/), but that would require infra changes, and since we're already logging things in redis we can do a collector, which reads redis at scrape time.

I tried this in latest changes, but haven't thoroughly tested that it works correctly.

cl/alerts/tasks.py
elisa-a-v: same as with the ones from commands, these are now scraped from redis using a collector. I only did minor testing though!

cl/users/signals.py
elisa-a-v: yeah nice catch 😅 

Moved the metric increment to the `delete_account` view

cl/stats/utils.py
```diff
@@ -52,6 +56,46 @@ def get_milestone_range(start, end):
...
+        raise ValueError(
+            f"Labels for '{name}' must be {expected}, got {list(labels.keys())}"
+        )
+
+    # Check each value is valid
+    for label_name, value in labels.items():
+        allowed_enum = STAT_LABEL_VALUES.get(label_name)
+        if allowed_enum:
+            allowed_values = [e.value for e in allowed_enum]
+            if value not in allowed_values:
```
mlissner: I think: 

```suggestion
        allowed_enum = STAT_LABEL_VALUES.get(label_name, [])
        allowed_values = [e.value for e in allowed_enum]
        if value not in allowed_values:
```

cl/lib/search_utils.py
```diff
@@ -326,10 +326,8 @@ def store_search_api_query(
         if is_semantic
         else SearchQuery.KEYWORD,
     )
-    prometheus_key = (
-        f"search.queries.{'semantic' if is_semantic else 'keyword'}.api"
-    )
-    record_prometheus_metric(prometheus_key, 1)
+    query_type = "semantic" if is_semantic else "keyword"
+    search_queries_total.labels(query_type=query_type, method="api").inc()
```
ERosendo: I like this approach.

However, this method short-circuits when the `store-search-api-queries` flag is disabled. We added this flag to help during periods of high database usage by stopping write operations from this method. With the added line incrementing `search_queries_total`, disabling the flag would also prevent the Prometheus counter from being incremented.

Is that the intended behavior, or should the metrics increment regardless of the flag? Should we move the `search_queries_total` increment outside the flag check so we maintain metrics visibility even when database writes are disabled?

cl/search/views.py
```diff
@@ -248,9 +250,15 @@ def show_results(request: HttpRequest) -> HttpResponse:
...
-                prometheus_handler_key="search.queries.keyword.web",
+                StatMetric.SEARCH_RESULTS,
+                labels={
+                    "query_type": StatQueryType.KEYWORD,
+                    "method": StatMethod.WEB,
+                },
             )
+            search_queries_total.labels(
+                query_type="keyword", method="web"
+            ).inc()
```
ERosendo: Regarding the redundancy you mentioned: I believe `search_queries_total` is one of those cases. It was originally introduced in #6639 to make this value scrapable, but since this PR updates the `tally_stat` method to retrieve all those metrics directly from Redis using a collector, we should only keep one of them.

Additionally, for consistency, I think we should consider moving the logic to increment the stats/Prometheus counters to the `store_search_query` method, similar to what we're doing for the API in this PR

cl/stats/utils.py
```diff
@@ -79,27 +123,34 @@ def tally_stat(
...
+    :param name: The stat name (e.g., "search.results" or StatMetric.SEARCH_RESULTS)
+    :param inc: The amount to increment (default: 1)
+    :param date_logged: The date to log the stat for (default: today)
+    :param labels: Optional dict of label names to values for Prometheus metrics
     """
     if not switch_is_active("increment-stats"):
         return
 
+    if labels:
+        _validate_labels(name, labels)
```
ERosendo, line 117: I noticed that `labels` are only validated when they’re provided, but some metrics (like `search.results`) require labels. If no labels are passed, a key that doesn’t follow the expected pattern could be written to Redis. Is that okay?

cl/stats/utils.py
```diff
@@ -79,27 +123,34 @@ def tally_stat(
...
         date_logged = current_dt.date()
 
     key = f"{name}.{date_logged.isoformat()}"
 
-    if prometheus_handler_key:
-        record_prometheus_metric(prometheus_handler_key, inc)
-
-    return _update_cached_stat(key, inc, date_logged)
+    result = _update_cached_stat(key, inc, date_logged)
+    _update_prometheus_stat(name, inc, labels)
```
ERosendo: I noticed this results in two separate network round-trips to Redis, which adds unnecessary latency( especially on search endpoints where this code path is hit frequently).

Each operation requires a full network round-trip:

- `_update_cached_stat`: Pipeline with `INCRBY` + `EXPIRE`.
- `_update_prometheus_stat`:  Single `INCRBY`.

Also, If the second Redis call fails, this can lead to inconsistent data between the date-based stats and the Prometheus metrics.

I think we should combine both updates into a single pipeline to improve performance and ensure atomicity.

cl/search/views.py
elisa-a-v: I like that idea but the `tally_stat` call in `show_results` is in a different part of the code than `store_search_query`. I'm not sure if that separation is deliberate or not, but if it is and we move `tally_stats` into `store_search_query`, we'll add new calls to `tally_stat` where it was skipped before, specifically when a user is editing an alert.

If this is better, happy to do the change, but it might be best to do that in a separate PR, what do you think?

elisa-a-v:
Thanks for the feedback! I've addressed your comments in the code, and about performance:

**Write path** (`tally_stat`):
Prometheus `incrby` piggybacks on a rount-trip that was already happening now that they're executed in a pipeline, so the overhead is minimal.

**Read path** (`StatMetricsCollector.collect()`):
- Replaced individual GETs with a single `mget`
- Key count bounded by label combinations, and enum validation prevents cardinality explosion

So it should be fine, although there's one thing I do want to flag: Prometheus Redis keys have no TTL, but they only increase and the number of keys is capped by the limited cardinality. We currently have 12 keys total, and adding a new label value requires a code change to the enums in constants.py, so this won't grow silently.

search.results (query_type × method):
- `search.results:keyword:web`
- `search.results:keyword:api`
- `search.results:semantic:web`
- `search.results:semantic:api`

alerts.sent (alert_type):
- `alerts.sent:docket`
- `alerts.sent:recap`
- `alerts.sent:scheduled`

webhooks.sent (event_type):
- `webhooks.sent:docket_alert`
- `webhooks.sent:search_alert`
- `webhooks.sent:recap_fetch`
- `webhooks.sent:old_docket_alerts_report`
- `webhooks.sent:pray_and_pay`

cl/search/views.py
ERosendo: > we'll add new calls to tally_stat where it was skipped before, specifically when a user is editing an alert.


You’re absolutely right. right now the tally_stat call is skipped when a user is editing an alert. I think we can leave it as it is for now, and the redesign will likely give us a good opportunity to revisit and improve this view in the future.

elisa-a-v:
@mlissner your request for changes is blocking merge, wanna take another look and approve yourself or shall I go ahead and dismiss that review?

mlissner:
If you and Eduardo are happy, that's good enough for me!

PR #6778, opened 2026-01-13
6537 Implement SCOTUS parties merger

albertisfu:
## Fixes

Fixes: #6537

## Summary

This PR introduces the parties merger for SCOTUS dockets.

Initially, I thought about creating new merge methods to ingest SCOTUS parties, since they are in a different format than PACER parties. However, I realized that the best approach is to normalize the SCOTUS parties into the PACER format and then use the existing party-ingestion methods without any changes.

The normalization consists of introducing attorney metadata to match the PACER format, where the contact is the attorney organization data.

Another tweak is to set the role to `LEAD_ATTORNEY` when `is_counsel_of_record` is `True`. In all other cases, the role is set to `UNKNOWN`.

I also extended the tests to confirm that the data is properly merged into the party models.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/recap/mergers.py
+++ b/cl/recap/mergers.py
@@ -1275,7 +1275,9 @@ def disassociate_extraneous_entities(
     ).delete()
 
 
-def normalize_scotus_parties(parties: list) -> list:
+def normalize_scotus_parties(
+    parties: list[dict[str, Any]],
+) -> list[dict[str, Any]]:
     """Transform SCOTUS court party data to PACER-like format.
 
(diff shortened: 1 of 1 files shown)


mlissner:
Works for me if it works for you guys!

Issue #6790, opened 2026-01-14
RECAPDocument admin search times out on any query


elisa-a-v:
## Problem

Searching in `/admin/search/recapdocument/?q=<any_value>` times out in production.

### What works
- **Admin list view (no search)**: Loads quickly in production
- **Direct Django shell queries**: Fast
  ```python
  RECAPDocument.objects.filter(pk=464673643).count()  # ~0.3s
  list(RECAPDocument.objects.filter(pk=464673643))    # ~0.6s
  ```

### What doesn't work
- **Admin search**: Hangs indefinitely on any search term, regardless of whether the pk exists

### Ruled out
- **DB query performance**: Shell queries against the same data return quickly
- **Specific search terms**: All searches hang, not just certain pks

### Current configuration

`RECAPDocumentAdmin` (cl/search/admin.py):
- Inherits from `CursorPaginatorAdmin` (django-admin-cursor-paginator)
- `search_fields = ("pk__exact",)`

### Tested locally (did not help)
- Setting `show_query_result_count = False` - hang persists

### Still unknown
- Exact location in Django admin/CursorPaginatorChangeList where the hang occurs
- Whether this is related to recent Django 6.0.1 upgrade

## Next steps to try
- Check postgres for long-running queries during the hang
- Git bisect to identify when the issue started

## Related
- RECAPDocumentAdmin also has an N+1 query issue (204 queries on list view due to `__str__` traversing relations) - separate from the search hang but worth fixing with `list_select_related`

mlissner:
I'm also attempting to fix this upstream here: https://github.com/django/django/pull/20538

With the ticket here: https://code.djangoproject.com/ticket/36865#ticket

PR #6792, opened 2026-01-15
fix(admin): Prevent varchar casting in RECAPDocument admin search

mlissner:
## Fixes

Fixes: #6790

## Summary

This PR fixes the RECAPDocument admin search timeout issue by overriding `get_search_results()` to use direct integer comparison instead of varchar casting.

**Root Cause:** Django 6.0.1 introduced a change ([PR #17885](https://github.com/django/django/pull/17885), commit `f223729`) that casts non-text fields to CharField for `__exact` lookups in admin search. This generates SQL like:

```sql
WHERE ("search_recapdocument"."id")::varchar = '123'
```

This cannot use the primary key index and causes full table scans on the RECAPDocument table (hundreds of millions of rows), resulting in timeouts.

**Solution:** Override `get_search_results()` to perform direct integer comparison:

```sql
WHERE "search_recapdocument"."id" = 123
```

Also adds `list_select_related` to fix the N+1 query issue from `__str__` accessing `docket_entry.docket.docket_number`.

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

This only affects the Django admin interface.


mlissner:
I went ahead and merged this because I need it to seal docs and living without has gotten painful. Should be fine (famous last words).

PR #6793, opened 2026-01-15
fix(new site menu): Fix broken nav menu

EstiShay:
## Fixes

This fixes the new site nav menu, which broke in the recent bout of dependency updates.

## Summary

Django Cotton 2.6.0 is more strict about attributes preceded by colons, so Alpine's bindings now must be escaped by using a double-colon. 

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

<details closed>
<details open>
Before: 
image
After:
image

</details>

### DIFF
--- a/cl/lib/s3_cache.py
+++ b/cl/lib/s3_cache.py
@@ -1,0 +1,45 @@
+from math import ceil
+
+from django.conf import settings
+from django.core.cache import caches
+from django.core.cache.backends.base import BaseCache
+from waffle import switch_is_active
+
+
+def get_s3_cache(fallback_cache: str = "db_cache") -> BaseCache:
+    """Get S3 cache in production, fallback cache in dev/test.
+
+    In production, returns the S3 Express cache backend. In development,
+    testing, or when the enable-s3-cache waffle switch is disabled,
+    returns the specified fallback cache to avoid requiring S3 access
+    for local development or to allow quick rollback if issues arise.
+
+    :param fallback_cache: Cache alias to use in dev/test (default: "db_cache")
+    :return: Django cache backend instance
+    """
+    is_dev_or_test = settings.DEVELOPMENT or settings.TESTING
+    if is_dev_or_test or not switch_is_active("enable-s3-cache"):
+        return caches[fallback_cache]
+    return caches["s3"]
+
+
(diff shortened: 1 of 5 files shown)


cl/assets/templates/cotton/layout_with_navigation/index.html
```diff
@@ -8,7 +8,7 @@
 >
   <a href="#article" class="skip-link focus-visible:-translate-y-13 -translate-y-10 -translate-x-1">Skip navigation</a>
   <c-navigation-menu
-    :nav_items="{{ nav_items }}"
+    ::nav_items="{{ nav_items }}"
```
elisa-a-v:
```suggestion
    :nav_items="nav_items"
```

elisa-a-v (changes requested):
Thanks for tracking this down! The nav menu is working again.

I was curious about the fix because the attribute is not in fact an Alpine binding! It's a Cotton dynamic attribute (that's why we needed to parse the JSON in JavaScript), but the syntax was actually never correct.

I dug into the Cotton source a bit to understand why it was working before in the first place. I think it's likely because the DTL gets evaluated during parent template rendering, or via Cotton's fallback in `DynamicAttr.resolve()`), but it's not quite the intended pattern, so how about we do `:nav_items="nav_items"` instead?

I tested this locally and it works 👍🏽 

We should try and avoid mixing syntax moving forward 😅

elisa-a-v (approved):
Thank you! LGTM :shipit:

cl/assets/templates/cotton/layout_with_navigation/index.html
```diff
@@ -8,7 +8,7 @@
 >
   <a href="#article" class="skip-link focus-visible:-translate-y-13 -translate-y-10 -translate-x-1">Skip navigation</a>
   <c-navigation-menu
-    :nav_items="{{ nav_items }}"
+    :nav_items="nav_items"
```
wrabit, line 11: Just a note, since cotton 2.6.0 you can also achieve the same with:

`nav-items=nav_items` 

(hyphen choice in attribute name is down to preference, but hyphen `-` is more typical html style)

PR #6803, opened 2026-01-17
legal: add TM symbol for major "Citegeist" mentions

anseljh:
The rule of thumb for using trademark symbols is to use the appropriate symbol on the first instance on a page.


mlissner:
Thanks. Set for auto-merge.

Issue #6795, opened 2026-01-15
Create independent entry and document models for SCOTUS dockets.


albertisfu:
### Describe the issue.

We’ve agreed that it’s better to have independent models for SCOTUS docket entries and documents, similar to what we have for Texas dockets.

So we need to create these models for SCOTUS.

### What type of issue is this?

Research

PR #6807, opened 2026-01-21
6795 create SCOTUS docket entry and documents models

albertisfu:
## Fixes

Fixes: #6795 

## Summary

This PR introduces the new independent models for SCOTUS docket entries and SCOTUS documents.

**SCOTUSDocketEntry**
```
entry_number
description
date_filed
sequence_number
```

**SCOTUSDocument**
```
docket_entry
description
document_number
attachment_number
document_url
```

These new models are used in the refactor of: https://github.com/freelawproject/courtlistener/pull/6709

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


cl/search/models.py
```diff
@@ -4026,3 +4026,80 @@ class CaseTransfer(AbstractDateTimeModel):
...
+    description = models.TextField(blank=True)
+    document_number = models.IntegerField(
+        blank=True,
+        null=True,
+    )
+    attachment_number = models.SmallIntegerField(
+        blank=True,
+        null=True,
+    )
+    document_url = models.URLField()
```
MorganBennetDev: Should we store the file name part of the URL as a separate field similar to how we're storing `media_id` and `media_version_id` in `TexasDocument` now that we're not trying to extend `RECAPDocument`?  It might make merging easier if we stored all the URL components you mentioned in [this comment](https://github.com/freelawproject/courtlistener/pull/6709#discussion_r2687025173) separately.

mlissner:
I'm going to trust you guys on this one. The good news is the new models are new models, so if they're wrong, they're easy to fix!

cl/search/models.py
albertisfu: Sure. So you’re suggesting that we store the file name as something like: `20221123131814988_42997%20pdf%20Ebner.pdf`?

So we could then use this value to determine whether we need to update the PDF or simply ignore it during a re-scrape of the docket if it’s the same?

Should we preserve the `document_url` field, or should we use the different parameters we have to reconstruct it?

MorganBennetDev: **TL;DR**: I don't have a strong preference on having a `document_url` field or not, but I think we should set a standard.

1. Yes

2. Yes

3. Currently `RECAPDocument` seems to reconstruct the document URL from its stored parts and the new Texas documents use a stored `document_url` field. If we're going to continue adding new `<origin>Document` models, we probably want to decide on one approach or the other to use for everything. I'm partial to using the extracted parts for matching and using the `document_url` field for downloading since courts can change URL formats, breaking reconstruction logic. On the other hand, if a court changes its URL format, they probably also changed something that will break the parsers so we wouldn't be ingesting documents anyway.

albertisfu: I agree with keeping `document_url` as a field. That could become our standard for future document models as well. I think it’s useful to have the full document URL stored in a field instead of reconstructing it specifically for our replication users.

About the file name, I don’t think it’s helpful to have a separate field for it. We can use a model property to extract it at ingestion time from `document_url`. That would be a fast operation.

For Texas, the `media_id` is useful to keep in a separate field since we’ll use it as part of the document slug, so it needs to be queryable. In SCOTUS, we’ll use `document_number` for the same purpose.

What do you think?

MorganBennetDev: Having a standard `document_url` field seems good to me.

The distinction between `media_id` and SCOTUS file names makes sense to me. I'll try to keep it in mind in the future.

mlissner: Just a drive-by thought, but is it weird to have `XYZDocument.document_url` instead of `XYZDocument.url`?

albertisfu: Yeah. I just did the change `document_url` -> `url`

We have the same field in `TexasDocument`. The migration has already been applied, but I think we’re still in time to modify the field name to have a standardized URL field name in Texas too??

mlissner: Not sure if that's a question directed to me, but if we're migrating SCOTUS, I imagine it's doable to migrate Texas, and probably worth getting consistent.

MorganBennetDev: Will create a tiny issue for the Texas migration and assign it to myself so I don't forget.

MorganBennetDev: Ah I just saw you committed the change here.

MorganBennetDev (approved):
Looks good. Oddly, regenerating the migrations seems to have moved creation of foreign key columns from the `CREATE TABLE` statements to `ALTER TABLE` statements later in the migration.

albertisfu:
> Looks good. Oddly, regenerating the migrations seems to have moved creation of foreign key columns from the CREATE TABLE statements to ALTER TABLE statements later in the migration.

Yeah, it seems that adding and renaming the Texas `document_url` field triggered this change. So Django detected that it was better to create the new tables without the foreign keys and add them later.

Issue #6458, opened 2025-10-28
Make API throttles configurable via admin or a new interface


mlissner:
### Describe the issue.

We have a few API variables that we tweak [in the code](https://github.com/freelawproject/courtlistener/blob/main/cl/settings/third_party/rest_framework.py) to set API limits: 

 - `DEFAULT_THROTTLE_RATES` — this sets the default rates
 - `OVERRIDE_THROTTLE_RATES` — this lets us override the rates on a per-user basis
 - `CITATION_LOOKUP_OVERRIDE_THROTTLE_RATES` — to override the citation lookup API rates on a per-user basis

All of these work fine, but they require a pull request to tweak, which isn't great, particularly as Eri does more of this work. It'd be better if these were configured another way. 

In the past I've avoided tweaking this because you just can't beat the performance of an environment variable. Doing a DB lookup for every request isn't fun, and even doing a redis lookup ain't great. 

But we can probably fix this by caching a function that looks up the limits for five minutes at a time. There's a library that does this, but ChatGPT gave a quick solution that looks (at a glance) fine: 

https://chatgpt.com/share/69013069-ad18-8013-8482-c935fc8e5a55

You end up with this: 

```
@timed_cache(300)
def get_api_limits():
    # do some stuff...
    return api_limits
```

That'd run once per pod, but it seems simple enough. 

From there, we need an interface to configure the values. We could do that via admin, probably, with a field on the UserProfile, or perhaps we'd do better with cl.com/set-api-limit/ or whatever (but that seems harder?).

And maybe we have a per-user field for standard API limits and another for their citation lookup limit, then we delete `OVERRIDE_THROTTLE_RATES` and `CITATION_LOOKUP_OVERRIDE_THROTTLE_RATES`?

### What type of issue is this?

To Do

mlissner:
@eriodiah, just FYI, we won't be able to get to this for awhile, so you'll have to continue editing the settings file and doing PRs to remove people's access. Let me know if you need support with that.

mlissner:
Below is the plan that Claude and I came up with. It replaces hardcoded throttle settings with a database-backed `APIThrottle` model that supports throttling and blocking users. 

This is split into two PRs for safe rollout. The first deploys the models and loads the existing throttles into the database. The second applies the throttles to the API. 

[...]

PR #6824, opened 2026-01-25
feat(api): Add APIThrottle model for configurable rate limits

mlissner:
## Fixes
Partial fix of #6458 (PR 1 of 2)

## Summary
This PR adds the `APIThrottle` model and admin interface to allow managing API rate limits via Django admin instead of hardcoding them in settings. This is PR 1 of 2 for issue #6458.

**Changes included:**
- `APIThrottle` is a new model that tracks API throttles
- `APIThrottleAdmin` lets us see and adjust the new models
- `import_throttle_overrides` is a management command to migrate existing settings

**PR 2 (#6825) will:**
- Add a high-performance distributed cache decorator
- Update throttle classes to use the model
- Remove hardcoded settings and migration script from this PR

## Deployment

**This PR should:**

- [ ] `skip-deploy` (skips everything below)
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

1. Run migrations: `docker exec cl-django python manage.py migrate api`
2. Run import script: `docker exec cl-django python manage.py import_throttle_overrides` (use --dry-run first)
3. Verify data in admin at `https://www.courtlistener.com/admin/api/apithrottle/`
4. Then merge and deploy PR 2 (#6825)

PR #6825, opened 2026-01-25
feat(api): Switch throttle overrides to database model

mlissner:
## Fixes
Fixes: #6458

**Note:** This PR depends on #6824 (PR 1) being merged and deployed first.

## Summary
This PR switches throttle rate overrides from hardcoded settings to database lookups using the `APIThrottle` model. This is PR 2 of 2 for issue #6458.

**Changes included:**
 - `tiered_cache` decorator in `cl/lib/decorators.py` that provides three-tier caching: memory (per-process) -> Redis (shared) -> function
 - `get_all_throttle_overrides()` in `cl/api/utils.py` to loads throttle overrides from the DB or cache (it uses the new decorator)
 - Throttle classes now check the database for overrides

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

**Prerequisites:**
1. PR 1 (#6824) must be merged and deployed first

**Then deploy this PR** - no additional steps required.

### DIFF
--- a/cl/stats/management/commands/ip_lookup.py
+++ b/cl/stats/management/commands/ip_lookup.py
@@ -1,0 +1,98 @@
+import sys
+from datetime import timedelta
+
+from django.contrib.auth.models import User
+from django.core.management.base import BaseCommand
+from django.urls import reverse
+from django.utils.timezone import now
+
+from cl.lib.redis_utils import get_redis_interface
+
+
+class Command(BaseCommand):
+    help = "Look up users associated with IP addresses from Redis API logs"
+
+    def add_arguments(self, parser) -> None:
+        parser.add_argument(
+            "ips",
+            nargs="*",
+            help="IP addresses to look up",
+        )
+        parser.add_argument(
+            "--stdin",
+            action="store_true",
+            help="Read IP addresses from stdin (one per line)",
+        )
+        parser.add_argument(
+            "--days",
+            type=int,
+            default=14,
+            help="Number of days to search (default: 14)",
+        )
+
+    def handle(self, *args, **options) -> None:
+        ips = list(options["ips"])
+
+        if options["stdin"]:
+            for line in sys.stdin:
+                ip = line.strip()
+                if ip:
+                    ips.append(ip)
+
+        if not ips:
+            self.stderr.write("No IP addresses provided")
+            return
+
+        days = options["days"]
+        found_ips: set[str] = set()
+        r = get_redis_interface("STATS")
+
+        self.stdout.write(
+            f"Searching {len(ips)} IPs across last {days} days...\n"
+        )
+        self.stdout.write("-" * 80)
+
+        # Search most recent days first
+        for i in range(days):
+            date_str = (now() - timedelta(days=i)).strftime("%Y-%m-%d")
+            key = f"api:v4.d:{date_str}.ip_map"
+
+            for ip in ips:
+                if ip in found_ips:
+                    continue
+
+                user_pk = r.hget(key, ip)
+                if user_pk:
+                    found_ips.add(ip)
+                    self._print_result(ip, user_pk, date_str)
+
+        # Report IPs not found
+        not_found = set(ips) - found_ips
+        if not_found:
+            self.stdout.write(f"\nNot found in last {days} days:")
+            for ip in not_found:
+                self.stdout.write(f"  {ip}")
+
+        self.stdout.write("-" * 80)
+        self.stdout.write(
+            f"Found: {len(found_ips)}, Not found: {len(not_found)}"
+        )
+
+    def _print_result(self, ip: str, user_pk: str, date_str: str) -> None:
+        try:
+            user = User.objects.get(pk=int(user_pk))
+            admin_url = reverse("admin:auth_user_change", args=[user.pk])
+            self.stdout.write(
+                f"\n{ip}\n"
+                f"  User: {user.username} ({user.email})\n"
+                f"  Date: {date_str}\n"
+                f"  Admin: https://www.courtlistener.com{admin_url}\n"
+                f"  Zoho: https://crm.zoho.com/crm/freelawproject/search"
+                f"?searchword={user.pk}&isRelevance=false"
+            )
+        except User.DoesNotExist:
+            self.stdout.write(
+                f"\n{ip}\n"
+                f"  User PK: {user_pk} (NOT FOUND IN DATABASE)\n"
+                f"  Date: {date_str}"
+            )


cl/api/management/commands/import_throttle_overrides.py
mlissner, line 1: This file can be removed in this PR since it will have been used by now.

cl/api/tests.py
```diff
@@ -4675,3 +4684,184 @@ def test_is_valid_filter_param_max_depth_prevents_dos(self) -> None:
...
+        self.assertIn("rate", ctx.exception.message_dict)
+
+
+class ThrottleOverrideIntegrationTest(TestCase):
+    """Integration tests for throttle overrides using the APIThrottle model."""
+
+    def setUp(self) -> None:
+        from django.core.cache import cache
+
+        from cl.lib.decorators import clear_tiered_cache
```
mlissner: All dependencies must be at the top.

cl/lib/decorators.py
```diff
@@ -15,6 +16,85 @@
...
 logger = logging.getLogger(__name__)
 
+# In-memory cache for tiered_cache decorator (per-process, fastest tier)
+_memory_cache: dict[str, tuple[float, Any]] = {}
+
+T = TypeVar("T")
+
+
+def tiered_cache(
+    timeout: int = 300,
```
mlissner: No default timeout here.

cl/lib/decorators.py
```diff
@@ -15,6 +16,85 @@
...
+    """Tiered caching decorator: memory -> Django cache (Redis) -> function.
+
+    Implements a three-tier caching strategy for optimal performance:
+    - Tier 1: In-memory dict (fastest, per-process)
+    - Tier 2: Django cache backend (Redis, shared across processes)
+    - Tier 3: Execute the wrapped function (slowest, e.g., DB query)
+
+    The memory cache is checked first for speed, then Redis for cross-process
+    sharing, and finally the function is called if neither has the value.
+
```
mlissner, line 43: Explain the data model of the _memory_cache object here.

cl/lib/decorators.py
```diff
@@ -15,6 +16,85 @@
...
+            if value is not None:
+                # Store in memory cache for faster subsequent access
+                _memory_cache[cache_key] = (current_time + timeout, value)
+                return value
+
+            # Tier 3: Call the function
+            value = func(*args, **kwargs)
+
+            # Store in both caches
+            cache.set(cache_key, value, timeout)
```
mlissner: Set this value for one second less than the memory cache, so that the memory cache always lasts slightly longer than the redis cache. This will ensure that the memory cache does not expire before the redis cache. 

If we don't do this, the memory cache could get refreshed from the redis cache when the redis cache only has 1 second remaining on it's TTL. That'd give the memory cache the full timeout again, which would be bad. This would be more likely under high load, but it would double the cache length.

cl/lib/decorators.py
```diff
@@ -15,6 +16,85 @@
...
 
+# In-memory cache for tiered_cache decorator (per-process, fastest tier)
+_memory_cache: dict[str, tuple[float, Any]] = {}
+
+T = TypeVar("T")
+
+
+def tiered_cache(
+    timeout: int = 300,
+    cache_alias: str = "default",
```
mlissner: Assume default redis cache here.

cl/lib/tests.py
```diff
@@ -2337,3 +2337,131 @@ def test_make_s3_cache_key_persistent_no_prefix_in_dev(self) -> None:
         base_key = "clusters-mlt-es:123"
         result = make_s3_cache_key(base_key, None)
         self.assertEqual(result, base_key)
+
+
+class TieredCacheTest(SimpleTestCase):
+    """Tests for the tiered_cache decorator."""
+
+    def setUp(self) -> None:
+        from cl.lib.decorators import clear_tiered_cache
```
mlissner: All imports at the top.

cl/lib/decorators.py
```diff
@@ -15,6 +16,85 @@
...
+    return decorator
+
+
+def clear_tiered_cache() -> None:
+    """Clear the in-memory tier of the tiered cache.
+
+    Useful for testing or when you need to force a refresh.
+    Note: This only clears the memory tier, not the Django cache tier.
+    """
+    _memory_cache.clear()
```
mlissner, line 104: Clear the redis cache here too, and then make the test setup/teardown methods more granular (so they are less likely to interfere with other tests).

cl/lib/tests.py
```diff
@@ -2337,3 +2337,131 @@ def test_make_s3_cache_key_persistent_no_prefix_in_dev(self) -> None:
...
+            self.call_count += 1
+            return {"key": "value"}
+
+        # First call
+        result1 = get_data()
+        self.assertEqual(result1, {"key": "value"})
+        self.assertEqual(self.call_count, 1)
+
+        # Clear memory cache but leave Redis cache
+        clear_tiered_cache()
```
mlissner: This will need to be tweaked with clear_tiered_cache also clearing the redis cache.

cl/settings/third_party/rest_framework.py
```diff
@@ -26,159 +26,9 @@
...
-        "WarrenLex": "20000/hour",  # For big litigation days (wow)
-        "quevon24": "500000/hour",  # Perform tests, clone cases in local env
-    },
-    "CITATION_LOOKUP_OVERRIDE_THROTTLE_RATES": {
-        "jafrank": "180/minute",
-        "LexText": "360/minute",
-    },
+    # Note: Throttle overrides are now stored in the database (APIThrottle model)
+    # and managed via Django admin. Use the import_throttle_overrides command
+    # to migrate existing overrides from this file to the database.
```
mlissner: We don't need this comment here. Remove it.

cl/lib/decorators.py
```diff
@@ -13,8 +14,100 @@
...
+# - cached_value: the return value of the decorated function
+_memory_cache: dict[str, tuple[float, Any]] = {}
+
+T = TypeVar("T")
+
+
+def tiered_cache(
+    timeout: int,
+) -> Callable[[Callable[..., T]], Callable[..., T]]:
+    """Tiered caching decorator: memory -> Django cache (Redis) -> function.
```
ERosendo, line 34: This method looks great, but it would only work for sync methods.

mlissner: Funny. Claude asked if I wanted to make it support async and I decided nah, out of scope until needed. :)

PR #6828, opened 2026-01-26
fix(bug): Fix path bug

### LINKED ISSUE #6692: [bug] API help pages with version in path show legacy template
### What happened?

The first API help page landed, and it looks great 🎉 but we're only getting the new template when browsing this path: https://www.courtlistener.com/help/api/rest/pacer/

However, going to the versioned path still shows the legacy template.

For added context: at first, the API help pages were not working at all with `IncrementalNewTemplateMiddleware`, and were later enabled by [this PR](https://github.com/freelawproject/courtlistener/pull/6283).

### Expected behavior

When the `use_new_design` flag is active, he PACER API help page should get the new template no matter what path we're using to browse it.

### Actual behavior

https://www.courtlistener.com/help/api/rest/v4/pacer/ doesn't get the new template, while https://www.courtlistener.com/help/api/rest/pacer/ does. They should both get the new one when the flag is active.

### Steps to reproduce

In production, browse both paths while logged into an account that has the `use_new_design` flag enabled, and you should see the legacy template in the versioned path, and the new template otherwise.

In localhost, make sure the flag is active, and browse both paths:
- http://localhost:8000/help/api/rest/v4/pacer/ will show the legacy template
- http://localhost:8000/help/api/rest/pacer/ will show the new template

### Screenshots

_No response_

Medium

EstiShay:
## Fixes
This fixes #6692 

## Summary

This PR adds a check for whether a file name exists prior to passing it along. Also fixes situation where `citation-lookup-api-help-None.html` was being passed.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

<details closed>
I'm skipping screenshots because if this works, you should see the new style instead of the old one, which is pretty self-evident.

### DIFF
--- a/cl/lib/middleware.py
+++ b/cl/lib/middleware.py
@@ -100,7 +100,11 @@ class IncrementalNewTemplateMiddleware:
                 get_template(old_template[0])
                 old_template = old_template[0]
             except TemplateDoesNotExist:
-                old_template = old_template[1] if len(old_template) > 1 else old_template[0]
+                old_template = (
+                    old_template[1]
+                    if len(old_template) > 1
+                    else old_template[0]
+                )
 
         if not isinstance(old_template, str):
             return response


cl/lib/middleware.py
```diff
@@ -94,8 +94,20 @@ def process_template_response(self, request, response):
...
+                path_exists = True
+            except TemplateDoesNotExist:
+                path_exists = False
+
+            if path_exists:
+                old_template = old_template[0]
+            elif len(old_template) > 1:
+                old_template = old_template[1]
+            else:
+                old_template = old_template[0]
```
albertisfu, line 101: I think we can achieve the same behavior with this more compact approach.

```suggestion
            try:
                get_template(old_template[0])
                old_template = old_template[0]
            except TemplateDoesNotExist:
                old_template = old_template[1] if len(old_template) > 1 else old_template[0]
```

EstiShay: Excellent suggestion

Issue #5942, opened 2025-07-10
Stop using docket.count


mlissner:
Now that our generic counter is working nicely, we need to update our code to use the new table, rather than using `docket.count`. For example, I think we use it in a management command that does the iquery stuff, and I think it's used in docket sitemaps, but it is probably used in other places too.

mlissner:
In #6845, we purged another of the docket.count uses. What's left, @ERosendo?

PR #6845, opened 2026-01-29
feat(sitemap): Replaces `Docket.view_count` with GenericCount subquery

ERosendo:
## Fixes

This PR relates to #5942 but doesn't fully resolve it.

## Summary

This PR replaces the use of Docket.view_count with a GenericCount subquery in the docket sitemap generation. Previously, the sitemap was directly accessing the view_count field on the Docket model. This change migrates to using the GenericCount table, which is the new centralized approach for tracking view counts across the application.

The implementation uses Django ORM annotations with a subquery to fetch view counts from the GenericCount table and join them with the docket queryset. 

I chose the ORM-based solution even though the query plan is not optimal. A raw SQL query would likely be more performant, but it is not compatible with the underlying logic used to compute the sitemap. The sitemap generation relies on Django's ORM queryset methods and conventions, which would require significant refactoring to accommodate raw SQL queries. 

For reference, here's the more performant raw SQL query that couldn't be used:

```python
raw_sql = """
    SELECT
        sd.id,
        sd.date_modified,
        sd.slug,
        COALESCE(gc.value, 0) AS view_count_gc
    FROM search_docket sd
    LEFT JOIN LATERAL (
        SELECT value
        FROM favorites_genericcount
        WHERE label = 'd.' || sd.id::text || ':view'
    ) gc ON true
    WHERE
        NOT sd.blocked
        AND sd.source = ANY(%s)
        AND (gc.value > 10 OR sd.date_filed > %s)
    ORDER BY sd.id
"""
```

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


mlissner (approved):
I had to do some research, but I learned a bit. LGTM!

PR #6853, opened 2026-02-02
feat(prayers): unlimited api prayers

### LINKED ISSUE #6570: Allow unlimited Pray & Pay requests by API
### Describe the issue.

We released webhooks and API documentation for P&P, and I just talked to our first customer about how they might make use of it. The big catch is that you can only make five prayers per day (or 15 if you're a member). Weak. They want to follow 10k cases, and Pray for documents from all of them. 

So, we need to:
 - Allow unlimited prayers by API
 - Don't put those prayers in the leaderboard
 - Maybe don't send emails for those prayers

I assume this means:
 - Adding a column to the prayer table for whether the prayer is from the API or not.
 - Using that column to not show things in the leaderboard
 - Using that column to not send emails when the prayers are granted

I suppose there's an easier way to do this which is to simply allow unlimited prayers via the API, and let them email and show up in the leaderboard. That'd be very easy, but right now the leaderboard says: 

Image

Those stats would go wild if we allowed API users, so probably better not to do that.

----

I don't think this is a priority yet, but I wanted to make an issue for it for discussion and in case somebody wants to pick it up and work on it.

### Related Issue

A client wants this: https://crm.zoho.com/crm/freelawproject/tab/Accounts/7012837000001544001

### Screenshots

_No response_

To Do

bhsauro:
## Fixes

This fixes #6570 

## Summary
This PR implements unlimited prayers via the REST API while maintaining rate limits for web UI prayers. 

The following changes were implemented:
- Added `source` SmallIntegerField to `Prayer` model to distinguish API-created prayers from web UI prayers (`WEBSITE = 1`, `API = 2`)
- Created migration for the new field with default `WEBSITE`
- API prayers (created via `/api/rest/v4/prayers/`) are now unlimited - no rate limiting
- `PrayerViewSet.perform_create()` sets `source=API` for all API-created prayers
- Removed `prayer_eligible()` check from API prayer creation
- Web UI prayers (created via `create_prayer()` function) remain rate-limited at 2/day (or 6/day for members)
- All existing web UI functionality unchanged
- API prayers are now excluded from the leaderboard (`get_top_prayers()` filters to `source=WEBSITE`), summary statistics (`get_lifetime_prayer_stats()` filters to `source=WEBSITE`), prayer count displays (`get_prayer_counts_in_bulk()` filters to `source=WEBSITE`), and email notifications (only web UI prayers trigger emails)
- API prayers do still trigger webhooks (both API and web UI)

The following tests were added:
- Verify API prayers have `source=Prayer.API`
- Verify API prayers are unlimited (no rate limiting)
- Verify API prayers excluded from leaderboard
- Verify API prayers excluded from stats
- Verify API prayers don't send emails
- Verify web UI prayers still respect rate limiting

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


mlissner:
Thanks @bhsauro. We're at an off-site this week, so very little coding is getting done, but we'll get to this as soon as possible!

bhsauro:
@mlissner sounds good, thanks for letting me know and enjoy the off-site!

bhsauro:
hi @albertisfu just wanted to check in to see if there's anything else I can do to support on this request!

albertisfu:
Hi @bhsauro, sorry for the delay on this PR. I'll be reviewing it today.

cl/favorites/models.py
```diff
@@ -171,6 +171,11 @@ class Prayer(models.Model):
         choices=STATUSES,
         default=WAITING,
     )
+    via_api = models.BooleanField(
+        help_text="Whether the prayer was created via the API (unlimited)",
+        default=False,
+        db_index=True,
+    )
```
albertisfu: I'd prefer to see a `SmallIntegerField` instead of a `BooleanField`:

```
    WEBSITE = 1
    API = 2
    SOURCES = (
        (WEBSITE, "Website"),
        (API, "API request"),
    )
    source = models.SmallIntegerField(
        help_text="The interface used to create the prayer.", choices=SOURCES
    )

```

This way, we can easily add a new source in the future if necessary.

Also, I don’t think the index will be helpful here, since it won’t be used by the planner in most queries.

You can see in the model that we have three different composite indexes:

```
indexes = [
            # When adding a new document to RECAP, we'll ask: What outstanding
            # prayers do we have for this document?
            # When loading the prayer leader board, we'll ask: Which documents
            # have the most outstanding prayers?
            models.Index(
                fields=["recap_document", "status"],
                name="favorites_prayer_recap_document_id_status_82e2dbbb_idx",
            ),
            # When loading docket pages, we'll ask (hundreds of times): Did
            # user ABC pray for document XYZ?
            models.Index(
                fields=["recap_document", "user"],
                name="favorites_prayer_recap_document_id_user_id_c5d30108_idx",
            ),
            # When a user votes, we'll ask: How many outstanding prayers did
            # user ABC make today?
            models.Index(
                fields=["date_created", "user", "status"],
                name="favorites_prayer_date_created_user_id_status_880d7280_idx",
            ),
        ]
```

Right now, these indexes are the ones we rely on to make prayer queries more performant.

So, the planner will prefer these indexes instead of using the source index filter alone. However, this is not a problem, since most of the work will be done using the current composite indexes. The additional filtering by source, which currently has only two possible values, will be fast because the heavy lifting is already handled by the composite indexes.

An alternative would be to recreate the composite indexes and add the `source` field, but that would be a pain, and the performance gains would be minimal, so it’s not worthwhile.

cl/favorites/utils.py
```diff
@@ -274,50 +277,53 @@ def send_prayer_emails(instance: RECAPDocument) -> None:
...
 
     docket = instance.docket_entry.docket
     docket_entry = instance.docket_entry
     document_url = instance.get_absolute_url()
-    num_waiting = granted_prayers.count()
     doc_price = price(instance)
 
+    # Send webhooks for all prayers, emails only for web UI prayers
     messages = []
+    web_ui_prayer_count = sum(1 for p in granted_prayers if not p.via_api)
```
albertisfu: This will iterate over the entire queryset, and then below we iterate again over `granted_prayers`, which will duplicate the number of queries. Instead, would it be possible to include the  `source` (`via_api`) as a filter in the database query?

cl/favorites/utils.py
albertisfu: I believe we should add the source filter to this query so that other users who receive this email don’t get inflated numbers caused by API requests.

albertisfu (changes requested):
Thank you @bhsauro this looks good!

I’ve left a couple of suggestions, and here are a few more:

- **get_prayer_counts_in_bulk**
 It looks like we need to add the source filter to this query as well, so that API prayers are not taken into account when displaying the prayer counts of a document in the frontend.

- Can you please update the documentation related to this?
 https://www.courtlistener.com/help/api/rest/recap/#prayer-limitations

  So now we can indicate that prayers via the API are unlimited.

- Could you please create the .sql files for the migration? Here is the documented process:
https://wiki.free.law/c/courtlistener/dev-guide/database-migrations#steps-to-making-new-migrations

  We basically need a .sql and _customers.sql file.

cl/favorites/models.py
bhsauro: thanks for explaining this!

bhsauro:
hey @albertisfu just another ping to let you know that this is ready for review

mlissner:
Looking at this now, several months later, I see a couple things we should do differently: 

 - Prayers should allow rate limiting via a scope, like we do for other APIs, rather than be unlimited, which is risky. I don't know what the right rate is. We've got 5/day by default. These are lightweight, so we can go fairly big, I think. Let's set a default of **250/day and 1/s**. I think that should be enough for most use cases including backfilling.
 - The HTML docs have all moved to the wiki, so they don't need to be here anymore (though we'll want a tweak for the wiki).

We have one potential client that's interested in this right now, but so we'll pick this up again if they come through.

PR #6862, opened 2026-02-03
Texas docket and lower court merger

MorganBennetDev:
## Summary

Implements mergers for Texas lower court data and dockets to complete the full merging functionality.

Additionally, converts mergers to return a `MergeResult` struct, which provides more details about the results of a merge attempt.

## Deployment
    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -112,7 +112,7 @@ dependencies = [
   "django-cotton>=2.6.0",
   "django-cursor-pagination>=0.3.0",
   "django-elasticsearch-dsl>=8.0",
-  "juriscraper>=2.6.68",
+  "juriscraper>=3.0.0",
   "instructor>=1.14.1",
   "django-s3-express-cache>=0.1.0",
   "zohocrmsdk8-0==4.0.0",
--- a/cl/scrapers/management/commands/cl_back_scrape_citations.py
+++ b/cl/scrapers/management/commands/cl_back_scrape_citations.py
@@ -9,6 +9,7 @@ downloaded. If we find an Opinion we don't have in the database,
 we ingest it as in a regular scrape
 """
 
+from asgiref.sync import sync_to_async
 from django.db import IntegrityError
 from django.utils.encoding import force_bytes
 from juriscraper.lib.exceptions import BadContentError
@@ -27,7 +28,7 @@ class Command(cl_back_scrape_opinions.Command):
     scrape_target_descr = "citations"
     juriscraper_module_type = "opinions"
 
-    def scrape_court(
+    async def scrape_court(
         self,
         site,
         full_crawl: bool = False,
@@ -48,7 +49,7 @@ class Command(cl_back_scrape_opinions.Command):
             it's case data
         """
         court_str = site.court_id.split(".")[-1].split("_")[0]
-        court = Court.objects.get(id=court_str)
+        court = await Court.objects.aget(id=court_str)
         dup_checker = DupChecker(court, full_crawl=True)
 
         for case in site:
@@ -62,7 +63,7 @@ class Command(cl_back_scrape_opinions.Command):
                 continue
 
             try:
-                content = site.download_content(
+                content = await site.download_content(
                     case["download_urls"], media_root=settings.MEDIA_ROOT
                 )
             except BadContentError:
@@ -71,7 +72,9 @@ class Command(cl_back_scrape_opinions.Command):
             sha1_hash = sha1(force_bytes(content))
 
             try:
-                cluster = Opinion.objects.get(sha1=sha1_hash).cluster
+                cluster = await sync_to_async(
+                    lambda: Opinion.objects.get(sha1=sha1_hash).cluster
+                )()
             except Opinion.DoesNotExist:
                 # populate special key to avoid downloading the file again
                 case["content"] = content
@@ -84,7 +87,9 @@ class Command(cl_back_scrape_opinions.Command):
                     citation or parallel_citation,
                 )
 
-                self.ingest_a_case(case, None, True, site, dup_checker, court)
+                await self.ingest_a_case(
+                    case, None, True, site, dup_checker, court
+                )
                 continue
 
             for cite in [citation, parallel_citation]:
@@ -95,11 +100,13 @@ class Command(cl_back_scrape_opinions.Command):
                 if not citation_candidate:
                     continue
 
-                if citation_is_duplicated(citation_candidate, cite):
+                if await sync_to_async(citation_is_duplicated)(
+                    citation_candidate, cite
+                ):
                     continue
 
                 try:
-                    citation_candidate.save()
+                    await citation_candidate.asave()
                     logger.info(
                         "Saved citation %s for cluster %s", cite, cluster
                     )
--- a/cl/scrapers/management/commands/cl_back_scrape_opinions.py
+++ b/cl/scrapers/management/commands/cl_back_scrape_opinions.py
@@ -1,4 +1,4 @@
-import time
+import asyncio
 
 from juriscraper import AbstractSite
 from juriscraper.AbstractSite import logger
@@ -45,7 +45,7 @@ class Command(cl_scrape_opinions.Command):
         super().add_arguments(parser)
         add_backscraper_arguments(parser)
 
-    def parse_and_scrape_site(
+    async def parse_and_scrape_site(
         self,
         mod: AbstractSite,
         options: dict,
@@ -68,7 +68,7 @@ class Command(cl_scrape_opinions.Command):
         """
         court_str = mod.__name__.split(".")[-1].split("_")[0]
         logger.info(f'Using court_str: "{court_str}"')
-        for site in site_yielder(
+        async for site in site_yielder(
             mod.Site(
                 backscrape_start=options.get("backscrape_start"),
                 backscrape_end=options.get("backscrape_end"),
@@ -77,15 +77,15 @@ class Command(cl_scrape_opinions.Command):
             mod,
             save_response_fn=save_response,
         ):
-            site.parse()
-            self.scrape_court(site, full_crawl=True)
+            await site.parse()
+            await self.scrape_court(site, full_crawl=True)
 
             if wait := options["backscrape_wait"]:
                 logger.info(
                     "Sleeping for %s seconds before continuing backscrape",
                     wait,
                 )
-                time.sleep(wait)
+                await asyncio.sleep(wait)
 
     def save_everything(self, items, backscrape=True):
         super().save_everything(items, backscrape)
--- a/cl/scrapers/management/commands/cl_back_scrape_oral_arguments.py
+++ b/cl/scrapers/management/commands/cl_back_scrape_oral_arguments.py
@@ -1,4 +1,4 @@
-import time
+import asyncio
 
 from juriscraper.AbstractSite import logger
 from juriscraper.lib.importer import site_yielder
@@ -15,11 +15,11 @@ class Command(cl_scrape_oral_arguments.Command):
         super().add_arguments(parser)
         add_backscraper_arguments(parser)
 
-    def parse_and_scrape_site(self, mod, options: dict):
+    async def parse_and_scrape_site(self, mod, options: dict):
         court_str = mod.__name__.split(".")[-1].split("_")[0]
         logger.info(f'Using court_str: "{court_str}"')
 
-        for site in site_yielder(
+        async for site in site_yielder(
             mod.Site(
                 backscrape_start=options.get("backscrape_start"),
                 backscrape_end=options.get("backscrape_end"),
@@ -28,12 +28,12 @@ class Command(cl_scrape_oral_arguments.Command):
             mod,
             save_response_fn=save_response,
         ):
-            site.parse()
-            self.scrape_court(site, full_crawl=True, backscrape=True)
+            await site.parse()
+            await self.scrape_court(site, full_crawl=True, backscrape=True)
 
             if wait := options["backscrape_wait"]:
                 logger.info(
                     "Sleeping for %s seconds before continuing backscrape",
                     wait,
                 )
-                time.sleep(wait)
+                await asyncio.sleep(wait)
--- a/cl/scrapers/management/commands/cl_scrape_opinions.py
+++ b/cl/scrapers/management/commands/cl_scrape_opinions.py
@@ -5,7 +5,7 @@ import traceback
 from datetime import date
 from typing import Any
 
-from asgiref.sync import async_to_sync
+from asgiref.sync import async_to_sync, sync_to_async
 from django.core.files.base import ContentFile
 from django.core.management.base import CommandError
 from django.db import transaction
@@ -283,7 +283,7 @@ class Command(ScraperCommand):
             help="Disable duplicate aborting.",
         )
 
-    def scrape_court(
+    async def scrape_court(
         self,
         site,
         full_crawl: bool = False,
@@ -293,10 +293,12 @@ class Command(ScraperCommand):
         # Get the court object early for logging
         # opinions.united_states.federal.ca9_u --> ca9
         court_str = site.court_id.split(".")[-1].split("_")[0]
-        court = Court.objects.get(pk=court_str)
+        court = await Court.objects.aget(pk=court_str)
 
         dup_checker = DupChecker(court, full_crawl=full_crawl)
-        if dup_checker.abort_by_url_hash(site.url, site.hash):
+        if await sync_to_async(dup_checker.abort_by_url_hash)(
+            site.url, site.hash
+        ):
             logger.debug("Aborting by url hash.")
             return
 
@@ -316,7 +318,7 @@ class Command(ScraperCommand):
                 next_date = None
 
             try:
-                self.ingest_a_case(
+                await self.ingest_a_case(
                     item, next_date, ocr_available, site, dup_checker, court
                 )
                 added += 1
@@ -338,9 +340,9 @@ class Command(ScraperCommand):
         )
 
         if update_site_hash:
-            dup_checker.update_site_hash(site.hash)
+            await sync_to_async(dup_checker.update_site_hash)(site.hash)
 
-    def get_opinions_content(
+    async def get_opinions_content(
         self, case_dict: dict, site, court, dup_checker, next_case_date
     ) -> list[tuple[dict, bytes, str]]:
         """Downloads opinions and checks if the content is duplicated
@@ -363,7 +365,7 @@ class Command(ScraperCommand):
 
         # download content
         for sub_opinion in opinions_to_download:
-            content = site.download_content(
+            content = await site.download_content(
                 sub_opinion["download_urls"], media_root=settings.MEDIA_ROOT
             )
             opinions_content.append(
@@ -393,7 +395,7 @@ class Command(ScraperCommand):
 
             # Duplicates will raise errors
             try:
-                dup_checker.press_on(
+                await sync_to_async(dup_checker.press_on)(
                     Opinion,
                     case_dict["case_dates"],
                     next_case_date,
@@ -420,7 +422,7 @@ class Command(ScraperCommand):
 
         return opinions_content
 
-    def ingest_a_case(
+    async def ingest_a_case(
         self,
         item,
         next_case_date: date | None,
@@ -429,17 +431,25 @@ class Command(ScraperCommand):
         dup_checker: DupChecker,
         court: Court,
     ):
-        opinions_content = self.get_opinions_content(
+        opinions_content = await self.get_opinions_content(
             item, site, court, dup_checker, next_case_date
         )
 
-        child_court = get_child_court(item.get("child_courts", ""), court.id)
-
-        docket, opinions, cluster, citations, originating_court_info = (
-            make_objects(item, child_court or court, opinions_content)
+        child_court = await sync_to_async(get_child_court)(
+            item.get("child_courts", ""), court.id
         )
 
-        save_everything(
+        (
+            docket,
+            opinions,
+            cluster,
+            citations,
+            originating_court_info,
+        ) = await sync_to_async(make_objects)(
+            item, child_court or court, opinions_content
+        )
+
+        await sync_to_async(save_everything)(
             items={
                 "docket": docket,
                 "opinions": opinions,
@@ -450,7 +460,7 @@ class Command(ScraperCommand):
         )
 
         for opinion in opinions:
-            extract_opinion_content.delay(
+            await sync_to_async(extract_opinion_content.delay)(
                 opinion.pk,
                 ocr_available=ocr_available,
                 juriscraper_module=site.court_id,
@@ -463,9 +473,9 @@ class Command(ScraperCommand):
                 item["case_names"].encode(),
             )
 
-    def parse_and_scrape_site(self, mod, options: dict):
-        site = mod.Site(save_response_fn=save_response).parse()
-        self.scrape_court(site, options["full_crawl"])
+    async def parse_and_scrape_site(self, mod, options: dict):
+        site = await mod.Site(save_response_fn=save_response).parse()
+        await self.scrape_court(site, options["full_crawl"])
 
     def handle(self, *args, **options):
         super().handle(*args, **options)
@@ -501,7 +511,7 @@ class Command(ScraperCommand):
                 i += 1
                 continue
             try:
-                self.parse_and_scrape_site(mod, options)
+                async_to_sync(self.parse_and_scrape_site)(mod, options)
             except Exception as e:
                 capture_exception(
                     e, fingerprint=[module_string, "{{ default }}"]
--- a/cl/scrapers/management/commands/cl_scrape_oral_arguments.py
+++ b/cl/scrapers/management/commands/cl_scrape_oral_arguments.py
@@ -1,7 +1,7 @@
 from datetime import date
 from typing import Any
 
-from asgiref.sync import async_to_sync
+from asgiref.sync import async_to_sync, sync_to_async
 from celery.canvas import chain
 from django.core.files.base import ContentFile
 from django.db import transaction
@@ -107,7 +107,7 @@ class Command(cl_scrape_opinions.Command):
     scrape_target_descr = "oral arguments"
     juriscraper_module_type = "oral_args"
 
-    def ingest_a_case(
+    async def ingest_a_case(
         self,
         item,
         next_case_date: date | None,
@@ -117,14 +117,14 @@ class Command(cl_scrape_opinions.Command):
         court: Court,
         backscrape: bool = False,
     ):
-        content = site.download_content(
+        content = await site.download_content(
             item["download_urls"], media_root=settings.MEDIA_ROOT
         )
         # request.content is sometimes a str, sometimes unicode, so
         # force it all to be bytes, pleasing hashlib.
         sha1_hash = sha1(force_bytes(content))
 
-        dup_checker.press_on(
+        await sync_to_async(dup_checker.press_on)(
             Audio,
             item["case_dates"],
             next_case_date,
@@ -139,16 +139,20 @@ class Command(cl_scrape_opinions.Command):
         )
         dup_checker.reset()
 
-        docket, audio_file = make_objects(item, court, sha1_hash, content)
+        docket, audio_file = await sync_to_async(make_objects)(
+            item, court, sha1_hash, content
+        )
 
-        save_everything(
+        await sync_to_async(save_everything)(
             items={"docket": docket, "audio_file": audio_file},
             backscrape=backscrape,
         )
-        chain(
-            process_audio_file.si(audio_file.pk),
-            transcribe_from_open_ai_api.si(audio_file.pk),
-        ).apply_async()
+        await sync_to_async(
+            chain(
+                process_audio_file.si(audio_file.pk),
+                transcribe_from_open_ai_api.si(audio_file.pk),
+            ).apply_async
+        )()
 
         logger.info(
             "Successfully added audio file %s: %s",
--- a/cl/scrapers/test_assets/test_opinion_scraper.py
+++ b/cl/scrapers/test_assets/test_opinion_scraper.py
@@ -10,7 +10,8 @@ class Site(OpinionSite):
     def __init__(self):
         super().__init__()
         self.court_id = self.__module__
-        self.url = join(
+        self.url = "http://test"
+        self.mock_url = join(
             settings.INSTALL_ROOT,
             "cl/scrapers/test_assets/test_opinion_site.xml",
         )
--- a/cl/scrapers/test_assets/test_oral_arg_scraper.py
+++ b/cl/scrapers/test_assets/test_oral_arg_scraper.py
@@ -10,7 +10,8 @@ class Site(OralArgumentSite):
     def __init__(self):
         super().__init__()
         self.court_id = self.__module__
-        self.url = join(
+        self.url = "http://test"
+        self.mock_url = join(
             settings.INSTALL_ROOT,
             "cl/scrapers/test_assets/test_oral_arg_site.xml",
         )
--- a/cl/scrapers/tests.py
+++ b/cl/scrapers/tests.py
@@ -7,6 +7,7 @@ from pathlib import Path
 from unittest import mock
 from unittest.mock import MagicMock, patch
 
+import httpx
 import responses
 from asgiref.sync import async_to_sync
 from django.conf import settings
@@ -159,8 +160,8 @@ class ScraperIngestionTest(ESIndexTestCase, TestCase):
 
         site = test_opinion_scraper.Site()
         site.method = "LOCAL"
-        parsed_site = site.parse()
-        cl_scrape_opinions.Command().scrape_court(
+        parsed_site = async_to_sync(site.parse)()
+        async_to_sync(cl_scrape_opinions.Command().scrape_court)(
             parsed_site, full_crawl=True, ocr_available=False
         )
 
@@ -305,14 +306,14 @@ class ScraperIngestionTest(ESIndexTestCase, TestCase):
 
         site = test_oral_arg_scraper.Site()
         site.method = "LOCAL"
-        parsed_site = site.parse()
+        parsed_site = async_to_sync(site.parse)()
 
         with self.captureOnCommitCallbacks(execute=True):
             with mock.patch(
                 "cl.lib.celery_utils.get_task_wait"
             ) as patched_wait:
                 patched_wait.return_value = 0
-                cl_scrape_oral_arguments.Command().scrape_court(
+                async_to_sync(cl_scrape_oral_arguments.Command().scrape_court)(
                     parsed_site, full_crawl=True
                 )
 
@@ -356,18 +357,19 @@ class ScraperIngestionTest(ESIndexTestCase, TestCase):
                 content["payload"]["results"][0]["local_path"]
             )
 
-    def test_parsing_xml_opinion_site_to_site_object(self) -> None:
+    async def test_parsing_xml_opinion_site_to_site_object(self) -> None:
         """Does a basic parse of a site reveal the right number of items?"""
-        site = test_opinion_scraper.Site().parse()
+        site = await test_opinion_scraper.Site().parse()
         self.assertEqual(len(site.case_names), 6)
 
-    def test_parsing_xml_oral_arg_site_to_site_object(self) -> None:
+    async def test_parsing_xml_oral_arg_site_to_site_object(self) -> None:
         """Does a basic parse of an oral arg site work?"""
-        site = test_oral_arg_scraper.Site().parse()
+        site = await test_oral_arg_scraper.Site().parse()
         self.assertEqual(len(site.case_names), 2)
 
     @patch(
-        "cl.scrapers.management.commands.cl_scrape_opinions.Command.get_opinions_content"
+        "cl.scrapers.management.commands.cl_scrape_opinions.Command.get_opinions_content",
+        new_callable=mock.AsyncMock,
     )
     def test_scrape_multiple_opinions_per_cluster(
         self, patched_get_opinions_content
@@ -402,7 +404,7 @@ class ScraperIngestionTest(ESIndexTestCase, TestCase):
         mock_site.url = "111"
         mock_site.hash = "234"
 
-        cl_scrape_opinions.Command().scrape_court(mock_site)
+        async_to_sync(cl_scrape_opinions.Command().scrape_court)(mock_site)
 
         # a single cluster with 2 sub opinions was created
         clusters = OpinionCluster.objects.filter(
@@ -805,48 +807,45 @@ class AudioFileTaskTest(TestCase):
 class ScraperContentTypeTest(TestCase):
     def setUp(self):
         # Common mock setup for all tests
-        self.mock_response = mock.MagicMock()
+        self.mock_response = mock.AsyncMock(httpx.Response)
         self.mock_response.content = b"not empty"
         self.mock_response.headers = {"Content-Type": "application/pdf"}
         self.site = test_opinion_scraper.Site()
         self.site.method = "GET"
         self.logger = logger
 
-    @mock.patch("requests.Session.get")
-    def test_unexpected_content_type(self, mock_get):
+    @mock.patch("httpx.AsyncClient.get")
+    async def test_unexpected_content_type(self, mock_get):
         """Test when content type doesn't match scraper expectation."""
         mock_get.return_value = self.mock_response
         self.site.expected_content_types = ["text/html"]
-        self.assertRaises(
-            UnexpectedContentTypeError,
-            self.site.download_content,
-            "/dummy/url/",
-        )
+        with self.assertRaises(UnexpectedContentTypeError):
+            await self.site.download_content("/dummy/url/")
 
-    @mock.patch("requests.Session.get")
-    def test_correct_content_type(self, mock_get):
+    @mock.patch("httpx.AsyncClient.get")
+    async def test_correct_content_type(self, mock_get):
         """Test when content type matches scraper expectation."""
         mock_get.return_value = self.mock_response
         self.site.expected_content_types = ["application/pdf"]
 
         with mock.patch.object(self.logger, "error") as error_mock:
-            _ = self.site.download_content("/dummy/url/")
+            _ = await self.site.download_content("/dummy/url/")
 
             self.mock_response.headers = {
                 "Content-Type": "application/pdf;charset=utf-8"
             }
             mock_get.return_value = self.mock_response
-            _ = self.site.download_content("/dummy/url/")
+            _ = await self.site.download_content("/dummy/url/")
             error_mock.assert_not_called()
 
-    @mock.patch("requests.Session.get")
-    def test_no_content_type(self, mock_get):
+    @mock.patch("httpx.AsyncClient.get")
+    async def test_no_content_type(self, mock_get):
         """Test for no content type expected (ie. Montana)"""
         mock_get.return_value = self.mock_response
         self.site.expected_content_types = None
 
         with mock.patch.object(self.logger, "error") as error_mock:
-            _ = self.site.download_content("/dummy/url/")
+            _ = await self.site.download_content("/dummy/url/")
             error_mock.assert_not_called()
 
 
@@ -895,10 +894,14 @@ class ScrapeCitationsTest(TestCase):
         with (
             mock.patch(f"{cmd}.sha1", side_effect=self.hashes),
             mock.patch.object(
-                self.mock_site, "download_content", return_value="placeholder"
+                self.mock_site,
+                "download_content",
+                new=mock.AsyncMock(return_value="placeholder"),
             ),
         ):
-            cl_back_scrape_citations.Command().scrape_court(self.mock_site)
+            async_to_sync(cl_back_scrape_citations.Command().scrape_court)(
+                self.mock_site
+            )
 
         citations = Citation.objects.filter(cluster=self.cluster).count()
         self.assertEqual(citations, 1, "Exactly 1 citation was expected")
--- a/cl/scrapers/utils.py
+++ b/cl/scrapers/utils.py
@@ -14,7 +14,6 @@ from juriscraper import AbstractSite
 from juriscraper.AbstractSite import logger
 from lxml import html
 from reporters_db import REPORTERS
-from requests import Response, Session
 
 from cl.citations.utils import map_reporter_db_cite_type
 from cl.corpus_importer.utils import winnow_case_name
@@ -157,27 +156,20 @@ def get_child_court(child_court_name: str, court_id: str) -> Court | None:
     return child_court
 
 
-@retry(
-    (
-        httpx.NetworkError,
-        httpx.TimeoutException,
-    ),
-    tries=3,
-    delay=5,
-    backoff=2,
-    logger=logger,
-)
-def test_for_meta_redirections(r: Response) -> tuple[bool, str | None]:
+async def test_for_meta_redirections(
+    r: httpx.Response,
+) -> tuple[bool, str | None]:
     """Test for meta data redirections
 
     :param r: A response object
     :return:  A boolean and value
     """
-    extension = async_to_sync(microservice)(
+    response = await microservice(
         service="buffer-extension",
         file=r.content,
         params={"mime": True},
-    ).text
+    )
+    extension = response.text
 
     if extension == ".html":
         html_tree = html.fromstring(r.text)
@@ -199,15 +191,17 @@ def test_for_meta_redirections(r: Response) -> tuple[bool, str | None]:
     return False, None
 
 
-def follow_redirections(r: Response, s: Session) -> Response:
+async def follow_redirections(
+    r: httpx.Response, s: httpx.AsyncClient
+) -> httpx.Response:
     """
     Parse and recursively follow meta refresh redirections if they exist until
     there are no more.
     """
-    redirected, url = test_for_meta_redirections(r)
+    redirected, url = await test_for_meta_redirections(r)
     if redirected:
         logger.info(f"Following a meta redirection to: {url.encode()}")
-        r = follow_redirections(s.get(url), s)
+        r = await follow_redirections(await s.get(url), s)
     return r
 
 
--- a/uv.lock
+++ b/uv.lock
@@ -504,7 +504,7 @@ requires-dist = [
     { name = "ipython", specifier = ">=9.9.0" },
     { name = "itypes", specifier = ">=1.1.0" },
     { name = "judge-pics", specifier = ">=2.0.5" },
-    { name = "juriscraper", specifier = ">=2.6.68" },
+    { name = "juriscraper", specifier = ">=3.0.0" },
     { name = "kombu", specifier = ">=5.5.1" },
     { name = "lxml", specifier = ">=6.0.2" },
     { name = "markdown2", specifier = ">=2.5.4" },
@@ -1855,7 +1855,7 @@ wheels = [
 
 [[package]]
 name = "juriscraper"
-version = "2.8.4"
+version = "3.0.0"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "certifi" },
@@ -1866,6 +1866,7 @@ dependencies = [
     { name = "feedparser" },
     { name = "geonamescache" },
     { name = "html5lib" },
+    { name = "httpx", extra = ["http2"] },
     { name = "lxml" },
     { name = "nh3" },
     { name = "python-dateutil" },
@@ -1873,9 +1874,9 @@ dependencies = [
     { name = "selenium" },
     { name = "tldextract" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/d1/40/0d63fde43d58e4968bf180651b72a872e1d1577b06c7d5d846bb38b963f5/juriscraper-2.8.4.tar.gz", hash = "sha256:7d752312fbb115c02a58dd543e31275d24aa5acf01a41a7118e77878cf76fac0", size = 385461, upload-time = "2026-03-17T21:45:30.341Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/9f/ff/7f34ccf378c996cee9e679a563fc3d6991043f2ff9f0d275b4ae8bbc3e52/juriscraper-3.0.0.tar.gz", hash = "sha256:02c2c67ecfd3a43ef444bdde7776b69f9997855529644f86a49d71e02643f075", size = 385434, upload-time = "2026-03-18T15:10:23.134Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/35/f0/1fb2f9f71326c13ec81df9c915d0bd155fc6a172c8766e9c66a354751a55/juriscraper-2.8.4-py3-none-any.whl", hash = "sha256:02a2689949697b87516135d33f3a342151d228baeb6c800dc7c272696aeaf0bf", size = 611354, upload-time = "2026-03-17T21:45:28.398Z" },
+    { url = "https://files.pythonhosted.org/packages/9b/4e/151154b483ab6e391a301f192c5ddab12f773f13c6f9269297ad997e56ce/juriscraper-3.0.0-py3-none-any.whl", hash = "sha256:1c59c60a99f644de13689f777ec237c8e08bd3f08ce6d7214f19c6ce02cce326", size = 611776, upload-time = "2026-03-18T15:10:21.408Z" },
 ]
 
 [[package]]


cl/search/state/texas/factories.py
```diff
@@ -51,11 +95,14 @@ class TexasCommonDataDictFactory(DictFactory):
...
     case_events = List([SubFactory(TexasDocketEntryDictFactory)])
     appellate_briefs = LazyAttribute(
-        lambda d: filter(
-            lambda e: True if random.random() < 0.1 else False, d.case_events
+        lambda d: list(
+            filter(
+                lambda e: True if random.random() < 0.1 else False,
+                d.case_events,
+            )
         )
```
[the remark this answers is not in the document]
MorganBennetDev: /fp

MorganBennetDev:
@albertisfu Failing lint is because I'm using a type from my Juriscraper PR that hasn't yet been pushed to PyPI, but otherwise this should be ready to go.

cl/search/models.py
```diff
@@ -4007,21 +4007,13 @@ class CaseTransfer(AbstractDateTimeModel):
...
         on_delete=models.CASCADE,
         related_name="case_transfer_destination_court",
     )
-    destination_docket = models.ForeignKey(
-        "search.Docket",
-        on_delete=models.CASCADE,
-        related_name="case_transfer_destination_docket",
-    )
+    destination_docket_number = models.TextField()
     transfer_date = models.DateField()
```
albertisfu: @MorganBennetDev Could you please move these model changes and their migrations to a separate PR? That way, we can also include the migrations needed to increase the SCOTUS and Texas document URL length in that PR.

MorganBennetDev: @albertisfu Just submitted a PR with this migration and the others (#6913).

cl/corpus_importer/tasks.py
```diff
@@ -3431,9 +3442,52 @@ def download_texas_document_pdf(
         return texas_document_pk
 
 
+class MergeResult[T = int](NamedTuple):
```
albertisfu, line 3932: I’m receiving some type hint complaints from my IDE in this class and its methods.

I tested this refactored approach, and it looks good,  there are no type complaints:


```
class MergeResult(NamedTuple):
    """Stores data about the result of an attempted merge operation."""

    create: bool
    """Whether a document needed to be created."""
    update: bool
    """Whether a document needed to be updated."""
    success: bool
    """Whether the operation was successful."""
    pk: int | None
    """The primary key of the created or updated object."""

    @classmethod
    def created(cls, pk: int) -> MergeResult:
        """Shorthand for the result of a successful creation operation."""
        return cls(create=True, update=False, success=True, pk=pk)

    @classmethod
    def updated(cls, pk: int) -> MergeResult:
        """Shorthand for the result of a successful update operation."""
        return cls(create=False, update=True, success=True, pk=pk)

    @classmethod
    def failed(cls) -> MergeResult:
        """Shorthand for the result of a failed merge operation."""
        return cls(create=False, update=False, success=False, pk=None)

    @classmethod
    def unnecessary(cls, pk: int) -> MergeResult:
        """Shorthand for the result of an unnecessary merge operation."""
        return cls(create=False, update=False, success=True, pk=pk)
```

cl/corpus_importer/tasks.py
albertisfu, line 3876: Currently, we’re experiencing rate limit issues when downloading the final cases from Texas. Brennan suspects that the court is actively decreasing the rate limit for us.

I noticed that the document download process is not rate limited right now. Since these tasks are scheduled as part of the docket merge process and executed in the default Celery queue, which has many workers, we can easily exceed the rate limit.

One option is to use `throttle_task` to ensure that only the allowed number of tasks are executed within a defined time window.

@Brennan-Chesley-FLP, what rate limit would you suggest for downloading PDFs?

```suggestion
)
@throttle_task("2/s")
```

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3707,413 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
     :param parties: The parties involved in the Texas case.
+    :return: A MergeResult indicating the operation succeeded. Note that
+        create and update flags are always False and pk is always None since
+        add_parties_and_attorneys does not return this information.
     """
     add_parties_and_attorneys(docket, normalize_texas_parties(parties))
+    return MergeResult(create=False, update=False, success=True, pk=None)
+
+
+def texas_js_court_id_to_court_id(js_court_id: str) -> str:
```
albertisfu: Would it be better to move this to `corpus_importer.utils`?

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3707,413 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+
+def texas_originating_court_to_court_id(
+    court_data: TexasOriginatingAppellateCourt | TexasOriginatingDistrictCourt,
+) -> str | None:
+    """Attempts to translate Juriscraper Texas originating court data to a
+    CourtListener Court ID.
+
+    :param court_data: The originating court data from Juriscraper.
+    :return: The matching Court ID or None if no court could be found."""
+    # TODO Replace with JS CourtID enum values when dependency is updated
```
albertisfu: Now that we have released the latest version Juriscraper, can we apply this change?

cl/corpus_importer/utils.py
```diff
@@ -1293,6 +1293,30 @@ def create_docket_entry_sequence_numbers(
     return sequence_numbers
 
 
+def juriscraper_to_cl_court_id(js_court_id: str) -> str | None:
```
albertisfu: Is this method still used?

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3707,413 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+    docket_data: TexasCourtOfAppealsDocket
+    | TexasCourtOfCriminalAppealsDocket
+    | TexasSupremeCourtDocket,
+) -> MergeResult:
+    """Merges scraped data from a Texas docket into the `Docket` table.
+
+    :param docket_data: The scraped Texas docket data.
+    :return: The result of the merge operation."""
+    court = Court.objects.get(
+        pk=texas_js_court_id_to_court_id(docket_data["court_id"])
```
albertisfu, line 4721: Currently, we have dockets generated from case law opinions that point to `texapp` (Court of Appeals of Texas) instead of using the individual Texas appellate courts like `txctapp1`, `txctapp2`, etc (which are disabled right now).

I remember we agreed that we should try to match those dockets and update their `court_id` to follow the new disaggregated approach.

To do that, I think we should run a query that uses the incoming appellate court ID along with `texapp`, and always update the `court_id` to the new value. This way, we can match an existing docket in `texapp` and update it to `txctapp1`, for example.

@grossir, do you see any issues for case law opinions if we do this?

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3707,413 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+    )
+    with transaction.atomic():
+        Docket.objects.select_for_update().get(
+            docket_number=docket_data["docket_number"], court_id=court.pk
+        )
```
albertisfu: We don’t need this DB lock here, since `SELECT FOR UPDATE` in PostgreSQL only works on existing rows. So it won’t prevent any race conditions in this case. Also, we don’t expect high concurrency when merging docket objects.

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3707,413 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+    originating_court_information.docket_number = originating_court_data[
+        "case"
+    ]
+    originating_court_information.court_reporter = originating_court_data[
+        "reporter"
+    ]
+    originating_court_information.assigned_to_str = originating_court_data[
+        "judge"
+    ]
+    # TODO Get judge from PeopleDB to add
```
albertisfu: Can we use `lookup_judge_by_full_name_and_set_attr` for this?

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3707,413 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+
+
+def merge_texas_case_transfers(
+    docket: Docket,
+    docket_data: TexasCourtOfAppealsDocket
+    | TexasCourtOfCriminalAppealsDocket
+    | TexasSupremeCourtDocket,
+) -> MergeResult:
+    """Merge appeal and work sharing information into the given Texas docket.
+
```
albertisfu, line 4534: Could you please explain and document the assumptions we’re making here to create case transfers?

Also, while reviewing the `CaseTransfer` model changes, I noticed that we won’t have an easy way to link `CaseTransfer` to a docket in order to display this data. I think we may need to add a foreign key to the docket we’re merging. What do you think?

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3707,413 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+    )
+    with transaction.atomic():
+        Docket.objects.select_for_update().get(
+            docket_number=docket_data["docket_number"], court_id=court.pk
+        )
+        docket_number = docket_data["docket_number"]
+        try:
+            docket = Docket.objects.get(
+                court_id=court.pk, docket_number=docket_number
+            )
```
albertisfu: Instead of using the docket_number and replicating the matching docket logic, can we use `find_docket_object` instead?

For two reasons:

- It uses `docket_number_core`, which has a database index, so the lookup is faster.
- The docket-matching logic is centralized.

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3707,413 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+            court.pk,
+        )
+
+    entry_merge_results = [
+        merge_texas_docket_entry(
+            docket, sequence_number, appellate_brief, entry
+        )
+        for sequence_number, appellate_brief, entry in zip(
+            create_docket_entry_sequence_numbers(docket_data["case_events"]),
+            generate_texas_appellate_brief_flags(
```
albertisfu: Can you explain what `generate_texas_appellate_brief_flags` does?

Also, are we merging appellate briefs and case events as docket entries? We want to merge case events and appellate briefs to create `TexasDocketEntry` instances, using `date_filed` to sort them properly.

albertisfu (changes requested):
Thanks @MorganBennetDev I have a couple of questions and some suggestions.

cl/corpus_importer/tasks.py
MorganBennetDev: Will do on Monday. Forgot to go through TODOs when updating.

cl/corpus_importer/tasks.py
MorganBennetDev: The appellate briefs table on TAMES appears to always be a subset of the case events table, so we extract both of them but only use the case events table to populate docket entries. In order to avoid losing data, we use `generate_texas_appellate_brief_flags` to set the `TexasDocketEntry.appellate_brief` field based on whether or not an entry from the case events table also appears in the appellate briefs table.

cl/corpus_importer/tasks.py
MorganBennetDev: We'll only have `CaseTransfer` objects if at least one end of the transfer is in our DB so we could add back one foreign key. The issue I see there is that we might run into cases where courts are only tracking one end of a transfer (e.g. only appealed from or appealed to data).

Maybe we could store both the docket number and a nullable docket foreign key for both ends of transfers and add a constraint that at least one of the two foreign keys has to be non null. Then when merging transfers we'd

1. Check if there is an existing transfer from/to the court and docket number we're merging
2. If there is, update the foreign key
3. If not, create a new `CaseTransfer` object with a null foreign key on the appropriate end

cl/corpus_importer/utils.py
MorganBennetDev: It is not. Will remove Monday

cl/corpus_importer/tasks.py
MorganBennetDev: My IDE typechecker was broken for a bit and I hadn't bothered to fix it yet when I wrote this. I did push a fix in the importer command PR but forgot to update it here.

cl/corpus_importer/tasks.py
MorganBennetDev: @albertisfu Implemented the approach I described in #6913. Probably best to discuss it there. Will add more explanation to this docstring as well.

pyproject.toml
```diff
@@ -112,7 +112,7 @@ dependencies = [
   "django-cotton>=2.6.0",
   "django-cursor-pagination>=0.3.0",
   "django-elasticsearch-dsl>=8.0",
-  "juriscraper>=2.6.68",
+  "juriscraper>=2.7.7",
```
albertisfu: no need to update Juriscraper since right now it's up to do date in main v2.7.8

you can revert this change and the one in uv.lock

cl/corpus_importer/management/utils.py
```diff
@@ -0,0 +1,243 @@
...
+            for row_idx, download_args in islice(
+                enumerate(download_inputs), start_row, None
+            ):
+                throttle.maybe_wait()
+                chain(
+                    self.download_task()
+                    .si(*download_args)
+                    .set(queue=retrieval_queue),
+                    self.parse_task().s().set(queue=parse_queue),
+                    self.merge_task().s().set(queue=ingesting_queue),
```
albertisfu: Parsing is a relatively fast operation, so we can include it inside the same `merge_task`.

That way we don't enqueue too many tasks.

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3799,436 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+                docket_data["case_events"], docket_data["appellate_briefs"]
+            ),
+            docket_data["case_events"],
+        )
+    ]
+
+    merge_case_transfer_result = merge_texas_case_transfers(
+        docket, docket_data
+    )
```
albertisfu: It seems we’re merging CaseTransfer for every single Texas case, right?

When looking at the Juriscraper examples, only COA has `transfer_from` and `transfer_to` fields.

Is there a reason to generate `CaseTransfer` for cases that belong to COCA and SC as well?

Is the data merged into `OriginatingCourtInformation` not sufficient?

I think this behavior will mirror the data available in TAMES, right?

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3799,466 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+                docket_source=Docket.SCRAPER,
+                allow_create=False,
+            )
+            if docket is not None:
+                logger.info(
+                    "Disaggregating Texas appellate docket %s", docket_number
+                )
+                docket.court = court
+        else:
+            docket = async_to_sync(find_docket_object)(
```
albertisfu: Is it correct for this to happen within this `else` condition?

If docket is `None` from the previous path, this lookup will never be reached, and the docket object will remain `None`. As a result, statements like:

`docket.date_filed = docket_data["date_filed"]`

Will fail.

Instead, shouldn’t it be something like:

```
if docket is None:
     docket = async_to_sync(find_docket_object)
```

cl/corpus_importer/utils.py
```diff
@@ -1299,3 +1306,47 @@ class DownloadPDFResult:
...
+    """Translates a Juriscraper Texas court ID to a CourtListener Court ID.
+
+    :param js_court_id: The court ID extracted from Juriscraper.
+    :return: The corresponding Court ID."""
+    if js_court_id == CourtID.SUPREME_COURT.value:
+        return "tex"
+    if js_court_id == CourtID.COURT_OF_CRIMINAL_APPEALS.value:
+        return "texcrimapp"
+    # Court of appeals
+    appellate_number = str(int(js_court_id[len("texas_coa") :]))
```
albertisfu: This version is more readable:

```suggestion
    appellate_number = str(int(js_court_id.removeprefix("texas_coa")))
```

cl/corpus_importer/utils.py
```diff
@@ -1299,3 +1306,47 @@ class DownloadPDFResult:
...
+    if appellate_number == "13":
+        appellate_number = "13A"
```
albertisfu: I asked Gianfranco, and he agreed that we should use `txctapp13` instead of txctapp13A. This will not be used. `txctapp13` is already created.

cl/corpus_importer/utils.py
```diff
@@ -1299,3 +1306,47 @@ class DownloadPDFResult:
...
+    :param court_data: The originating court data from Juriscraper.
+    :return: The matching Court ID or None if no court could be found."""
+    court_type = court_data["court_type"]
+    if court_type == CourtType.APPELLATE.value:
+        return texas_js_court_id_to_court_id(court_data["court_id"])
+    if court_type == CourtType.DISTRICT.value:
+        district_number = court_data["district"]
+        if district_number:
+            if district_number > 1:
+                district_number = district_number + 1
```
albertisfu: Why is the addition of 1 required here?

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3799,466 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+                    )
+                transfer.origin_court = Court.objects.get(pk=appeals_court_id)
+                transfer.origin_docket_number = appeals_court["case_number"]
+        elif docket_data["court_id"] == CourtID.SUPREME_COURT.value:
+            if appeals_court["court_id"] == CourtID.UNKNOWN:
+                logger.warning(
+                    "Found appellate court with unknown ID (docket %s)",
+                    docket.docket_number,
+                )
+                appeals_court_id = "texapp"
```
albertisfu: same here.

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3799,466 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+                        "origin_docket"
+                    ],
+                    destination_court=docket.court,
+                    destination_docket_number=docket.docket_number,
+                    destination_docket=docket,
+                    # If the transfer date is absent or empty, assume it matches the filing date
+                    transfer_date=transfer_from_date
+                    if transfer_from_date
+                    else docket_data["date_filed"],
+                    transfer_type=CaseTransfer.WORKLOAD,
```
albertisfu: is `transfer_from` data always related to `CaseTransfer.WORKLOAD?`

cl/corpus_importer/management/utils.py
```diff
@@ -0,0 +1,243 @@
...
+    appellate_court: str
+    court_code: str
+
+    @field_validator("date_filed", mode="before")
+    def date_filed_validator(cls, v):
+        return date(*(time.strptime(v, "%m/%d/%Y")[0:3]))
+
+
+@app.task(
+    bind=True,
```
albertisfu: I don’t think this is required since we don’t access the task attributes. Otherwise, it will throw an error.
` _corpus_download_task() takes 2 positional arguments but 3 were given`

Also, we don’t usually put tasks outside tasks.py due to discoverability issues. Did you [test](https://github.com/freelawproject/courtlistener/wiki/Working-with-Celery-on-CourtListener#launching-the-built-in-dashboard) this locally? If possible, we should keep tasks in the same place, or create a separate tasks file specifically for this command/texas, while ensuring there are no discoverability issues.

cl/corpus_importer/tasks.py
albertisfu, line 4752: Is it correct for this to happen within this `else` condition?

If docket is `None` from the previous path, this lookup will never be reached, and the docket object will remain `None`. As a result, statements like:

`docket.date_filed = docket_data["date_filed"]`

Will fail.

Instead, shouldn’t it be something like:

```
if docket is None:
     docket = async_to_sync(find_docket_object)
```

cl/corpus_importer/utils.py
albertisfu: This version is more readable:

```suggestion
    appellate_number = str(int(js_court_id.removeprefix("texas_coa")))
```

cl/corpus_importer/utils.py
albertisfu: I asked Gianfranco, and he agreed that we should use `txctapp13` instead of txctapp13A. This will not be used. `txctapp13` is already created.

cl/corpus_importer/utils.py
albertisfu: Why is the addition of 1 required here?

cl/corpus_importer/utils.py
```diff
@@ -1299,3 +1306,47 @@ class DownloadPDFResult:
...
+            return f"texdistct{district_number}"
+        return "texdistct"
+    if court_type == CourtType.BUSINESS.value:
+        return "texbizct"
+    if court_type == CourtType.MUNICIPAL.value:
+        return "texctyct"
```
albertisfu: In the case of `texctyct`, why do we use this court ID instead of `texctyct1`, `texctyct2`, etc.? Is it not possible to disaggregate the court in this case using the data we have?

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3799,466 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+                    ]["case"]
+                else:
+                    logger.error(
+                        "Unable to determine trial court ID for Texas docket %s to create CaseTransfer",
+                        docket.docket_number,
+                    )
+                    return MergeResult.failed()
+            else:
+                if appeals_court["court_id"] == CourtID.UNKNOWN:
+                    logger.warning(
```
albertisfu, line 4626: Let's log an error here instead of a warning. So we can review the issue.

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3799,466 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+                        "Found appellate court with unknown ID (docket %s)",
+                        docket.docket_number,
+                    )
+                    appeals_court_id = "texapp"
```
albertisfu: Since we’ll use the disaggregated courts for `texapp`, I don’t think it’s correct to default `to texapp`, since we want to stop using it.

Why don’t we just return `MergeResult.failed()?`

cl/corpus_importer/tasks.py
albertisfu: same here.

cl/corpus_importer/tasks.py
albertisfu: is `transfer_from` data always related to `CaseTransfer.WORKLOAD?`

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3799,466 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+        if i == len(appellate_briefs):
+            flags.append(False)
+            continue
+        if case_event == appellate_briefs[i]:
+            flags.append(True)
+            i += 1
+        else:
+            flags.append(False)
+
+    return flags
```
albertisfu: What about this more readable version?

```suggestion
    brief_iter = iter(appellate_briefs)
    next_brief = next(brief_iter, None)
    flags = []
    for case_event in case_events:
        if next_brief is not None and case_event == next_brief:
            flags.append(True)
            next_brief = next(brief_iter, None)
        else:
            flags.append(False)
    return flags
```

cl/corpus_importer/tasks.py
MorganBennetDev: That was the intended functionality. Thanks for the catch.

cl/corpus_importer/utils.py
MorganBennetDev: In CourtListener, the first district court is `texdistct1` then "Texas Judicial District Court 1A" is `texdistct2` and then the IDs continue normally (2nd to `texdistct3`, 3rd -> `texdistct4`, etc.). This workaround seemed easier than redoing the IDs for 509 entries in the `Court` table.

cl/corpus_importer/utils.py
MorganBennetDev: I could not find a good way to disaggregate this, no. City court names are some of the least consistent court names in TAMES, and they typically include no indication of _what_ city they're from. So to match them we'd have to create a bunch of manual cases based on the county and name of the court, which seemed out of scope.

cl/corpus_importer/tasks.py
MorganBennetDev: @albertisfu After some research, I've found this info on transfers.

1. Transfers to and from the 15th COA are _never_ for workload reasons ([Texas Goverment Code 73.001](https://statutes.capitol.texas.gov/?tab=1&code=GV&chapter=GV.73&artSec=)).
2. The Supreme Court may transfer dockets for any "good cause." It seems that "good cause" means workload balancing once per quarter ([source](https://www.txcourts.gov/2ndcoa/practice-before-the-court/docket-equalization-transfers/)).

I will update this logic accordingly.

MorganBennetDev:
@albertisfu Addressed feedback and replied to your questions.

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3803,577 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+                pacer_case_id=None,
+                docket_number=docket_number,
+                federal_defendant_number=None,
+                federal_dn_judge_initials_assigned=None,
+                federal_dn_judge_initials_referred=None,
+                docket_source=Docket.SCRAPER,
+                allow_create=True,
+            )
+        docket.docket_number = docket_number
+        docket.docket_number_core = make_docket_number_core(docket_number)
```
albertisfu: I found that we’re not properly handling the creation of docket_number_core for Texas dockets.
The goal of this method is to produce a simplified version of the docket number that is clean, unique, and works correctly with the docket_number_core database index.

We should take into account all the possible docket number formats for Texas cases.

For example, I tested the following docket numbers:

```
04-97-00972-CV
01-18-00277-CR
AP-77,129
WR-70,849-04
A-4369-A
C-2302
```

None of these are returning the correct docket_number_core.

e.g:
"04-97-00972-CV" -> 04000097 which is incomplete.
"AP-77,129" returns an empty string.

The only Texas format currently working properly is: 
25-1066

So we should ensure that `clean_docket_number` correctly handles Texas docket numbers so that they are accepted by this method.

You can review and add tests for Texas docket number variations in:
`test_clean_docket_number` and `test_making_scotus_docket_number_core`

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3803,577 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+                    lower_court_data
+                )
+            else:
+                lower_court_data = docket_data["appeals_court"]
+                lower_court_id = texas_js_court_id_to_court_id(
+                    lower_court_data["court_id"]
+                )
+
+        if lower_court_id is not None:
+            court = Court.objects.get(pk=lower_court_id)
```
albertisfu: In case the `lower_court_id` doesn’t exist in the database, could you catch it, log it as an error, and allow the docket to be created without an `appeal_from` value?

That way, we can identify the missing court and create it if necessary. Then we can revisit those dockets and populate the field accordingly.

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3803,577 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+        elif meta.court_code.startswith("coa"):
+            parser = TexasCourtOfAppealsScraper(meta.court_code)
+        else:
+            logger.error(
+                "Unrecognized Texas court type %s. Cannot parse docket %s.",
+                meta.court_code,
+                meta.case_number,
+            )
+            task.request.chain = None
+            return MergeResult.failed()
```
albertisfu: It seems this is a good opportunity to use match-case instead of if-elif-else.

cl/corpus_importer/utils.py
```diff
@@ -1299,3 +1306,48 @@ class DownloadPDFResult:
...
+
+    :param js_court_id: The court ID extracted from Juriscraper.
+    :return: The corresponding Court ID or None if invalid."""
+    if js_court_id == CourtID.SUPREME_COURT.value:
+        return "tex"
+    if js_court_id == CourtID.COURT_OF_CRIMINAL_APPEALS.value:
+        return "texcrimapp"
+    if js_court_id == CourtID.UNKNOWN.value:
+        logger.error("Unknown court ID: %s", js_court_id)
+        return None
```
albertisfu, line 1334: Also here we can use a match-case statement.

cl/corpus_importer/utils.py
```diff
@@ -1299,3 +1306,48 @@ class DownloadPDFResult:
...
+    :return: The matching Court ID or None if no court could be found."""
+    court_type = court_data["court_type"]
+    if court_type == CourtType.APPELLATE.value:
+        return texas_js_court_id_to_court_id(court_data["court_id"])
+    if court_type == CourtType.DISTRICT.value:
+        district_number = court_data["district"]
+        if district_number:
+            if district_number > 1:
+                district_number = district_number + 1
+            return f"texdistct{district_number}"
```
albertisfu:
```suggestion
        if district_number and district_number > 1:
            return f"texdistct{district_number + 1}"
```

MorganBennetDev: This will not produce the same output.

cl/corpus_importer/management/commands/import_texas_dockets.py
```diff
@@ -0,0 +1,54 @@
...
+            lambda r: Path(r[1]).suffix == ".html" and "searches" not in r[1],
+            map(lambda r: (r[0].strip(), r[1].strip()), csv_reader),
+        )
+
+        previous_key_stem = None
+        for html_row in html_rows:
+            html_bucket, html_key = html_row
+            html_path = Path(html_key)
+            docket_name = html_path.stem
+            if previous_key_stem and docket_name.startswith(previous_key_stem):
```
albertisfu: It seems that we need a more reliable check here.

Using docket_name.startswith(previous_key_stem) can cause non-duplicate cases to be skipped.

For example:

```
12-34.html
12-345.html
```

These are different cases, but the second one would be skipped.

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3805,590 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+    :param docket_data: The docket data from Juriscraper.
+    :return: The result of the merge operation."""
+    created = False
+    if not docket.originating_court_information:
+        created = True
+        docket.originating_court_information = OriginatingCourtInformation()
+
+    originating_court_information = docket.originating_court_information
+    originating_court_data = docket_data["originating_court"]
+
```
albertisfu, line 4496: Can you also set `docket_number_raw`? It should be the same value as docket_number.

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3805,590 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+                court_name = court.full_name
+        if not court_name:
+            logger.warning(
+                "Failed to find court ID %s while populating appeal_from field for Texas docket %s in court %s",
+                lower_court_id,
+                docket.pk,
+                court.pk,
+            )
+            # Assumes that we will only fail to generate a court ID for trial courts and never appellate courts
+            court_name = lower_court_data["name"]
```
albertisfu: Let's be safe:

```suggestion
            court_name = lower_court_data.get("name", "")
```

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3805,590 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+        logger.info(
+            "Updating lower court info with court %s (ID %s).",
+            court_name,
+            lower_court_id,
+        )
```
albertisfu: I don’t think this log will be useful.

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3805,590 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+            else:
+                lower_court_data = docket_data["appeals_court"]
+                lower_court_id = texas_js_court_id_to_court_id(
+                    lower_court_data["court_id"]
+                )
+
+        court_name = None
+        if lower_court_id is not None:
+            try:
+                court = Court.objects.get(pk=lower_court_id)
```
albertisfu: Use a different variable name, maybe `lower_court`,  so it doesn’t override the `court` from the docket being merged, which could cause a mismatch of the `court_id` in the following loggers.

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3805,590 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+            ):
+                lower_court_data = docket_data["originating_court"]
+                lower_court_id = texas_originating_court_to_court_id(
+                    lower_court_data
+                )
+            else:
+                lower_court_data = docket_data["appeals_court"]
+                lower_court_id = texas_js_court_id_to_court_id(
+                    lower_court_data["court_id"]
+                )
```
albertisfu: It seems that we can simplify this if-else statement like this:

```suggestion
            if (
            docket_data["court_type"] == CourtType.APPELLATE.value
            or docket_data["appeals_court"][
            "court_id"] == CourtID.UNKNOWN.value
        ):
            lower_court_data = docket_data["originating_court"]
            lower_court_id = texas_originating_court_to_court_id(
                lower_court_data)
        else:
            lower_court_data = docket_data["appeals_court"]
            lower_court_id = texas_js_court_id_to_court_id(
                lower_court_data["court_id"])
```

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3805,590 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+                if meta.court_code.startswith("coa"):
+                    parser = TexasCourtOfAppealsScraper(meta.court_code)
+                else:
+                    logger.error(
+                        "Unrecognized Texas court type %s. Cannot parse docket %s.",
+                        meta.court_code,
+                        meta.case_number,
+                    )
+                    task.request.chain = None
+                    return MergeResult.failed()
```
albertisfu: We can move the remaining if-else to a case statement.

```suggestion
        match meta.court_code:
            case "cossup":
                parser = TexasSupremeCourtScraper()
            case "coscca":
                parser = TexasCourtOfCriminalAppealsScraper()
            case court_code if court_code.startswith("coa"):
                parser = TexasCourtOfAppealsScraper(court_code)
            case _:
                logger.error(
                    "Unrecognized Texas court type %s. Cannot parse docket %s.",
                    meta.court_code,
                    meta.case_number,
                )
                task.request.chain = None
                return MergeResult.failed()
```

MorganBennetDev:
@albertisfu Addressed your feedback. Regarding the incorrect docket file filtering, I've updated the method according to @Brennan-Chesley-FLP's comments in Slack. Quoting it here for organization

> For each of the docket pages, there's a .html, a _meta.json, and a _header.json file that will all be duplicated.

> If there's a _meta_X.json file, then the _X.html file can also be ignored.

> But the dockets use comma->underscore conventions for numbering, and don't always end in a -XX suffix, so it's not entirely reliable to just look for those underscore+number suffixes on the .html files.

cl/corpus_importer/tasks.py
```diff
@@ -4154,56 +4154,46 @@ def merge_texas_docket(
...
-        if not court_name:
+                docket.appeal_from = lower_court
+                lower_court_name = lower_court.full_name
+        if not lower_court_name:
             logger.warning(
                 "Failed to find court ID %s while populating appeal_from field for Texas docket %s in court %s",
                 lower_court_id,
                 docket.pk,
-                court.pk,
+                lower_court.pk,
```
albertisfu: It seems that in this specific log, this should be court instead of `lower_court`, right?

Since `lower_court` will be undefined if `lower_court_name` is `None`, because the lower court doesn’t exist?

cl/corpus_importer/tasks.py
```diff
@@ -4212,7 +4202,7 @@ def merge_texas_docket(
         logger.error(
             "Failed to merge party data for Texas docket %s in court %s",
             docket.docket_number,
-            court.pk,
+            lower_court.pk,
```
albertisfu: Also, should it be `court` instead of `lower_court`, since we’re referring to the `court` the docket belongs to?

cl/corpus_importer/tasks.py
```diff
@@ -4235,7 +4225,7 @@ def merge_texas_docket(
         logger.error(
             "Failed to merge CaseTransfer data for Texas docket %s in court %s",
             docket.docket_number,
-            court.pk,
+            lower_court.pk,
```
albertisfu: same here?

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3805,579 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+
+    :param case_events: A list of TexasCaseEvent objects.
+    :param appellate_briefs: A list of TexasAppellateBrief objects.
+    :return: A list of booleans indicating whether the corresponding entry is
+    an appellate brief."""
+    brief_iter = iter(appellate_briefs)
+    next_brief = next(brief_iter, None)
+    flags = []
+    for case_event in case_events:
+        if next_brief is not None and case_event == next_brief:
```
albertisfu: I found that the assignment of appellate brief flags is not working properly.

It’s failing because of this exact check.

Here is an example case where I found the issue:
https://search.txcourts.gov/Case.aspx?cn=25-0719&coa=cossup

Here is the brief dict:
`{'remarks': 'Petition for Review filed on behalf of Ramon Cervantes, Jr.', 'date': datetime.date(2025, 8, 21), 'type': 'Petition for Review', 'attachments': [{'document_url': 'https://search.txcourts.gov/SearchMedia.aspx?MediaVersionID=e771932c-f14d-4d26-972b-4621e0535224&coa=cossup&DT=BRIEFS&MediaID=b717e466-3977-413f-872e-e8d1b5ed1649', 'description': '', 'media_id': 'b717e466-3977-413f-872e-e8d1b5ed1649', 'media_version_id': 'e771932c-f14d-4d26-972b-4621e0535224', 'file_size_str': '1.77 MB', 'file_size_bytes': 1770000}, {'document_url': 'https://search.txcourts.gov/SearchMedia.aspx?MediaVersionID=e8c8fc2a-115c-4737-8f95-c0ed72d45a0b&coa=cossup&DT=PET+FOR+REVIEW+FILED&MediaID=5390c6b6-6c12-4bd9-88e5-163e99677b8f', 'description': 'Notice', 'media_id': '5390c6b6-6c12-4bd9-88e5-163e99677b8f', 'media_version_id': 'e8c8fc2a-115c-4737-8f95-c0ed72d45a0b', 'file_size_str': '36 KB', 'file_size_bytes': 36000}], 'description': 'Petitioner'}`

Here is the same entry but from events:

`{'remarks': 'Petition for Review filed on behalf of Ramon Cervantes, Jr.', 'date': datetime.date(2025, 8, 21), 'type': 'Petition for Review', 'attachments': [{'document_url': 'https://search.txcourts.gov/SearchMedia.aspx?MediaVersionID=e771932c-f14d-4d26-972b-4621e0535224&coa=cossup&DT=BRIEFS&MediaID=b717e466-3977-413f-872e-e8d1b5ed1649', 'description': '', 'media_id': 'b717e466-3977-413f-872e-e8d1b5ed1649', 'media_version_id': 'e771932c-f14d-4d26-972b-4621e0535224', 'file_size_str': '1.77 MB', 'file_size_bytes': 1770000}, {'document_url': 'https://search.txcourts.gov/SearchMedia.aspx?MediaVersionID=e8c8fc2a-115c-4737-8f95-c0ed72d45a0b&coa=cossup&DT=PET+FOR+REVIEW+FILED&MediaID=5390c6b6-6c12-4bd9-88e5-163e99677b8f', 'description': 'Notice', 'media_id': '5390c6b6-6c12-4bd9-88e5-163e99677b8f', 'media_version_id': 'e8c8fc2a-115c-4737-8f95-c0ed72d45a0b', 'file_size_str': '36 KB', 'file_size_bytes': 36000}], 'disposition': ''}`

You can see they are different: the brief has a Description, while the case event has a Disposition.

So they don’t match, and the flag remains `False`.

Maybe we can filter out fields that are always present in both objects before comparing them.

Also, for appellate briefs, we have the `Description`, but we’re not merging it. We should merge it into the consolidated Docket Entry.

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3805,579 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+    """Merge originating court information into the given Texas docket.
+
+    :param docket: The docket to add the originating court to.
+    :param docket_data: The docket data from Juriscraper.
+    :return: The result of the merge operation."""
+    created = False
+    if not docket.originating_court_information:
+        created = True
+        docket.originating_court_information = OriginatingCourtInformation()
+
```
albertisfu, line 4494: I noticed that we’re missing the court data when we create the OCI from the Trial Court Information when there is also Appeals Information, for example:
https://search.txcourts.gov/Case.aspx?cn=25-0719&coa=cossup

In other sources like RECAP, the “Appeal From” is connected to the OCI data.

But here, in Texas, the appeal_from value comes from the Appeals Information, while the OCI is generated from the Trial Court Information.

So it seems we’re missing the court from the trial court data.

What do you think we should do?

I have a couple of ideas:

- Add a court field to the OCI data, although that would be redundant for RECAP.
- Create an additional CaseTransfer in those cases to hold the full Trial Court Information data.
Also, in CaseTransfer, I think we’re missing relevant data such as the Judge.
Should we also store the Disposition field?

cl/corpus_importer/tasks.py
MorganBennetDev: Looking at the code, this log is redundant with the one on line 4180. Removing it.

cl/corpus_importer/tasks.py
MorganBennetDev: It seems when I renamed `court` to `lower_court`, this was changed, making the incorrect logging more apparent.

cl/corpus_importer/tasks.py
MorganBennetDev: This is an oversight on my part. Appellate briefs and case events will always have that difference regardless of the court, and I forgot I set it up that way. Will fix also add a test for `generate_appellate_brief_flags`.

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3805,585 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+    if not party_merge_result.success:
+        logger.error(
+            "Failed to merge party data for Texas docket %s in court %s",
+            docket.docket_number,
+            court.pk,
+        )
+
+    entry_merge_results = [
+        merge_texas_docket_entry(
+            docket, sequence_number, appellate_brief, entry
```
albertisfu: It seems that we still need to merge the entry description from the appellate brief so we can include all the entry data from both the case event and the appellate brief. Right now, `TexasDocketEntry.description` remains empty.

cl/corpus_importer/tasks.py
albertisfu: I'm not sure whether this is better handled here or in Juriscraper, but we're creating empty attorney instances when there are no representatives for the party.

See example:

https://github.com/freelawproject/juriscraper/blob/d70a2c6887fe45bf134d6a3c24eb32a74211a85a/tests/examples/state/texas/sc/A-4369-A.json

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3805,585 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+    | TexasCourtOfCriminalAppealsDocket
+    | TexasSupremeCourtDocket,
+) -> MergeResult:
+    """Merge originating court information into the given Texas docket.
+
+    :param docket: The docket to add the originating court to.
+    :param docket_data: The docket data from Juriscraper.
+    :return: The result of the merge operation."""
+    created = False
+    if not docket.originating_court_information:
```
albertisfu, line 4491: I also noticed that we're creating empty OCI instances when there is no OCI data. See example:
https://github.com/freelawproject/juriscraper/blob/d70a2c6887fe45bf134d6a3c24eb32a74211a85a/tests/examples/state/texas/sc/A-4369-A.json

Would checking `court_type == texas_unknown` be enough to decide not to create the OCI instance?

cl/corpus_importer/tasks.py
```diff
@@ -3657,5 +3805,585 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+                federal_dn_judge_initials_referred=None,
+                docket_source=Docket.SCRAPER,
+                allow_create=True,
+            )
+        docket.add_scraper_source()
+        docket.docket_number = docket_number
+        docket.docket_number_core = make_texas_docket_number_core(
+            docket_number
+        )
+        docket.date_filed = docket_data["date_filed"]
```
albertisfu, line 4767: It seems we're missing the merge of `case_name` and `case_name_full` 🙃.

We should update the tests to confirm that both fields are being merged.

cl/corpus_importer/tasks.py
```diff
@@ -3830,27 +3830,34 @@ def merge_texas_docket_originating_court(
...
-    ]
-    originating_court_information.assigned_to_str = originating_court_data[
-        "judge"
-    ]
-    originating_court_id = texas_originating_court_to_court_id(
-        originating_court_data
-    )
+    if (
+        docket_data["court_type"] == CourtType.APPELLATE.value
+        or docket_data["appeals_court"]["court_id"] == CourtID.UNKNOWN.value
```
albertisfu: We also have this condition within `merge_texas_docket`. Can we create a helper method to centralize this condition and prevent duplicated code?

cl/corpus_importer/tasks.py
MorganBennetDev: Probably best to handle in JS, but putting in handling on both sides seems like a good idea.

cl/corpus_importer/tasks.py
MorganBennetDev: Yes, I think so.

MorganBennetDev:
@albertisfu Addressed your comments. Regarding the `TrialCourtData` model: would it be reasonable to include the mergers for that model in the PR creating it?

albertisfu:
> Regarding the TrialCourtData model: would it be reasonable to include the mergers for that model in the PR creating it?

I think it’s better to constrain that PR only to the model.

Then, before we merge it into main, you can merge that PR into this PR and work on the merger. Once we confirm everything works as expected and ensure the model doesn’t require any tweaks, we can ask the infra team to merge the PR and apply the migrations. After that, we can merge this PR into main with no conflicts.

cl/corpus_importer/management/commands/download_texas_documents.py
```diff
@@ -0,0 +1,104 @@
...
+            default="celery",
+            help="The celery queue for PDF extraction tasks.",
+        )
+        parser.add_argument(
+            "--delay",
+            type=float,
+            default=1.0,
+            help="Seconds to sleep between scheduling tasks.",
+        )
+
```
albertisfu, line 153: Can you also add an `only_extraction` argument like in SCOTUS?

We’ll definitely have extraction issues, so we want to run a second round of extraction only for the files that failed the first time.

cl/corpus_importer/tasks.py
```diff
@@ -3915,61 +4155,61 @@ def merge_texas_document(
...
-        created
-        or str(texas_document.media_version_id)
-        != input_document["media_version_id"]
-        or not texas_document.filepath_local
-    ):
+    if needs_update:
         texas_document.description = input_document["description"]
         texas_document.media_version_id = input_document["media_version_id"]
         texas_document.url = input_document["document_url"]
+        texas_document.filepath_local = ""
```
albertisfu, line 4189: Would we need to delete the file from S3 as well to prevent orphaned files?

```suggestion
        if texas_document.filepath_local:
            texas_document.filepath_local.delete(save=False)
        texas_document.filepath_local = ""
```

cl/corpus_importer/tasks.py
```diff
@@ -3915,61 +4155,61 @@ def merge_texas_document(
...
+        # Using this as a quick and dirty proxy for stale attachments, because
+        # I don't want to do a migration.
```
albertisfu: I understand you're assigning these values so that, in case the file changes and we need to download and extract these files outside of this process, we can identify which documents need to be downloaded and extracted. Correct?

How would a migration help here?

cl/corpus_importer/tasks.py
```diff
@@ -4129,5 +4381,615 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+@app.task(
+    bind=True,
+    max_retries=5,
+    ignore_result=True,
+)
+@time_call(logger)
+def texas_ingest_docket_task(
+    task: Task,
+    i: tuple[bytes, TexasDocketMeta],
+    download_attachments: bool = False,
```
albertisfu: I suggest that `download_attachments` be True by default here and downstream in all the methods. This way, you won’t need to make additional tweaks for the tests to pass. Also, considering that for the ongoing scraper and email notifications the default behavior will be to download attachments.

cl/corpus_importer/tasks.py
```diff
@@ -3903,9 +3929,223 @@ def download_texas_document_pdf(
...
+                court_id=court_id,
+                event_date=None,
+                require_living_judge=False,
+            )
+        else:
+            judge = None
+    else:
+        court = None
+        court_name = originating_court["name"]
+        judge = None
```
albertisfu: Can we remove the else statements by doing this?

```suggestion
    court = None
    court_name = originating_court["name"]
    judge = None
    if court_id:
        court = Court.objects.get(pk=court_id)
        court_name = court.full_name
        if judge_name:
            judge = async_to_sync(lookup_judge_by_full_name)(
                name=judge_name,
                court_id=court_id,
                event_date=None,
                require_living_judge=False,
            )
```

cl/search/factories.py
```diff
@@ -463,5 +465,53 @@ class ScotusDocketDataFactory(DictFactory):
...
+    court_name = Faker("court_name")
+    court = SubFactory(CourtFactory)
+    punishment = Faker("pystr")
+    county = Faker("pystr")
+
+    class Meta:
+        model = TrialCourtData
+
     docket_entries = List([SubFactory(SCOTUSDocketEntryDataFactory)])
     parties = List([SubFactory(SCOTUSPartyDataFactory)])
```
albertisfu: It seems `docket_entries` and parties were removed from `ScotusDocketDataFactory` and added here by mistake.

cl/corpus_importer/tasks.py
```diff
@@ -3903,9 +3929,223 @@ def download_texas_document_pdf(
...
+        )
+        return MergeResult.unnecessary(None)
+    dn_trial = originating_court["case"]
+    judge_name = originating_court["judge"]
+    reporter = originating_court["reporter"]
+    punishment = originating_court["punishment"]
+    county = originating_court["county"]
+    court_id = texas_originating_court_to_court_id(originating_court)
+    if court_id:
+        court = Court.objects.get(pk=court_id)
```
albertisfu: This can raise an error if the court doesn't exist in the DB. Can you catch the error and just log it instead of raising it? That way the instance can still be ingested, and we can later review which courts are missing.

albertisfu: I'm seeing this could also be an issue for `merge_texas_case_transfers`. Would it be possible that when doing `Court.objects.get` within `merge_texas_case_transfers`, if the court is not found, we catch the error and skip merging the case transfer? We should also log the error, including the `docket_number`, so the docket merge can succeed. Then, in a second pass, we can review which courts are missing, add them, and reprocess the dockets that failed so the missing `CaseTransfer` entries can be merged.

cl/corpus_importer/tasks.py
albertisfu: Is this suggestion is still pending?

cl/corpus_importer/tasks.py
albertisfu: I think we're missing this tweak too.

cl/search/admin.py
```diff
@@ -373,6 +374,37 @@ class BankruptcyInformationAdmin(admin.ModelAdmin):
...
+    search_fields = (
+        "origin_docket_number",
+        "destination_docket_number",
+    )
+    readonly_fields = (
+        "date_created",
+        "date_modified",
+    )
+
+
```
albertisfu, line 408: Can you add an admin for `TrialCourtData` please?

cl/corpus_importer/tasks.py
```diff
@@ -3903,9 +3929,223 @@ def download_texas_document_pdf(
...
+        court_name = court.full_name
+
+        if judge_name:
+            judge = async_to_sync(lookup_judge_by_full_name)(
```
albertisfu, line 4015: Can you confirm whether here and in other Texas judge lookups the method is able to match a judge name in the Texas format?

I'm seeing judges like:

`Honorable William Stacy Trotter`

 Do words like `Honorable` before the name could prevent a match? If so, would the judge names need to be cleaned before performing the lookup?

cl/corpus_importer/management/commands/download_texas_documents.py
```diff
@@ -0,0 +1,104 @@
...
+    :param delay: Seconds to sleep between scheduling tasks.
+    :return: None
+    """
+    docs = (
+        TexasDocument.objects.filter(Q(filepath_local=""))
+        .exclude(
+            ocr_status__in=(
+                TexasDocument.OCR_UNNECESSARY,
+                TexasDocument.OCR_COMPLETE,
+            )
```
albertisfu: If `filepath_local=""` is empty, I think we don’t need these additional filters, since we must download the file and extract it.

cl/corpus_importer/management/commands/download_texas_documents.py
```diff
@@ -0,0 +1,104 @@
...
+    `ocr_status` not unnecessary or complete, then schedules a download ->
+    extraction chain for each.
+
+    :param download_queue: The celery queue for download tasks.
+    :param extraction_queue: The celery queue for extraction tasks.
+    :param delay: Seconds to sleep between scheduling tasks.
+    :return: None
+    """
+    docs = (
+        TexasDocument.objects.filter(Q(filepath_local=""))
```
albertisfu:
```suggestion
        TexasDocument.objects.filter(filepath_local="")
```

cl/search/factories.py
MorganBennetDev: Messed up the merge. Fixing.

cl/corpus_importer/tasks.py
MorganBennetDev: At the time, I didn't like that there wasn't a specific value on the model indicating that a file had been changed, but looking at it again, this seems fine now.

cl/corpus_importer/tasks.py
MorganBennetDev: Sorry, forgot to implement this.

cl/corpus_importer/tasks.py
MorganBennetDev: This should be implemented. See `test_merge_texas_docket_entry_add_document` for assertions on functionality.

cl/corpus_importer/tasks.py
MorganBennetDev: Testing locally, `nameparser.HumanName` (which `lookup_judge_by_full_name` calls when it receives a string type input) seems to handle titles correctly so this shouldn't be an issue.

MorganBennetDev:
@albertisfu Addressed your comments and did some cleanup in `merge_texas_case_transfers` so it should be much more readable now (also uncovered a bug while I did that).

cl/corpus_importer/tasks.py
```diff
@@ -4129,5 +4392,583 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+
+    originating_court = docket_data["originating_court"]
+    oc_type = originating_court["court_type"]
+    oc_dn = originating_court["case"]
+    appeals_court = docket_data.get("appeals_court", {})
+    ac_id = appeals_court.get("court_id", "")
+    ac_dn = appeals_court.get("case_number", "")
+    trial_court_id = texas_originating_court_to_court_id(originating_court)
+    appeal_transfer_origin_court_id: str | None = ""
+    appeal_transfer_origin_dn = ""
```
albertisfu, line 4552: These variables seems to be unused.

cl/corpus_importer/tasks.py
```diff
@@ -4039,52 +4275,51 @@ def merge_texas_docket_entry(
...
         logger.info(
             "Found existing TexasDocketEntry for sequence number %s on Docket %s. Updating entry.",
             sequence_number,
             docket.pk,
         )
-        created = False
+
+    created = False
+    if not docket_entry:
+        logger.error(
```
albertisfu: When this was within the context of except `TexasDocketEntry.MultipleObjectsReturned:`, it made sense to log it as `logger.error` so we could review the case and determine why there were duplicated docket entries.

However, now this error is logged even for new entries, which will be quite noisy.

My suggestion is that if the docket entry is new, we don't log anything. If it is created after matching multiple potential duplicate entries, then we log an error.

cl/corpus_importer/tasks.py
```diff
@@ -4129,5 +4392,583 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+                lower_court_data["court_id"]
+            )
+        else:
+            lower_court_data = docket_data["originating_court"]
+            lower_court_id = texas_originating_court_to_court_id(
+                lower_court_data
+            )
+
+        # Assumes that we will only fail to generate a court ID for trial courts and never appellate courts
+        lower_court_name = lower_court_data.get("name", "")
```
albertisfu: If the `lower_court_name` is coming from appeals_court, this is not the correct key. Instead, it should be `district`; otherwise, the `appeal_from_str` field will remain empty.

cl/corpus_importer/tasks.py
```diff
@@ -4129,5 +4392,583 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+
+    if docket_data["court_type"] == CourtType.SUPREME.value:
+        trial_court_result = merge_texas_trial_court_data(docket, docket_data)
+    else:
+        trial_court_result = MergeResult.unnecessary(None)
+
+    party_merge_result = merge_texas_parties(docket, docket_data["parties"])
+
+    entry_merge_results = [
+        merge_texas_docket_entry(
```
albertisfu: I'm still seeing the missing docket entry `description` mentioned in a previous comment.

Here is an example of the problem:

The entry description is a field available in appellate briefs. However, right now we are just adding an `appellate_brief` flag to case events when we determine that it is an appellate brief and merging the case event. As a result, we are missing the description field available in appellate briefs.

We should incorporate the description into the docket entry data for appellate briefs.

Image

Image

cl/corpus_importer/tasks.py
MorganBennetDev: They are on lines 4625 (`appeal_transfer_origin_court_id`) and 4639 (`appeal_transfer_origin_dn`).

cl/corpus_importer/tasks.py
MorganBennetDev: In the case of appellate courts, `lower_court_id` will be set and `lower_court = Court.objects.get(pk=lower_court_id)` will execute without an exception so the `else` block will be called, setting `lower_court_name` to `lower_court.full_name`. Should I add a comment explaining this or is there a better approach?

cl/corpus_importer/tasks.py
MorganBennetDev: I see now. Apologies for not noticing this. Writing a fix and updating tests so they'll catch this.

MorganBennetDev:
@albertisfu Addressed feedback and fixed the description issue. Also left some replies I need clarification on.

cl/corpus_importer/tasks.py
albertisfu: Thanks for removing the `logger.error` when creating a new docket entry every time, that’s correct.

However, I still think we need to add `logger.error` back in the highlighted scenario.

Screenshot 2026-03-18 at 9 22 03 a m

Since this occurs after refining the queryset with `sequence_number`, and we previously had a `MultipleObjectsReturned` error, it means we could end up creating a potential duplicate entry that should be reviewed.

So replacing the `pass` with a `logger.error` would be helpful.

cl/corpus_importer/tasks.py
albertisfu: Sorry, I didn’t explain this correctly.

I know we’re assuming this will never fail for cases that contain `appeals_court` data, since we’re supposed to have all the appellate court IDs in our system.

However, I think it’s worthwhile not to make that assumption. We should be defensive here and consider that it’s possible we don’t have a court ID present in the `appeals_court` data (due to some weirdness we haven't seen yet).

If that happens, `lower_court_name` will never be assigned from `lower_court.full_name`, and it will remain empty.

I’d suggest always populating `lower_court_name` from `appeals_court["district"]` when `appeals_court` data is available. That way, we always preserve the original value from the case, and it will still be assigned even if the court ID is not found in our system.

cl/corpus_importer/tasks.py
albertisfu: Oh, I see now. My IDE flagged these as unused, but I understand now that these declarations help mypy infer the types.

cl/corpus_importer/tasks.py
MorganBennetDev: Got it. Will also add an error log in that case, since I think it indicates something has gone wrong that needs to be reviewed.

Edit: There is already an error logged in that case.

cl/corpus_importer/tasks.py
MorganBennetDev: I have the Pyright feature set up for PyCharm locally which matches mypy pretty closely and works a lot better than the third-party mypy extension.

cl/corpus_importer/tasks.py
```diff
@@ -4129,5 +4445,540 @@ def merge_texas_parties(docket: Docket, parties: list[TexasCaseParty]) -> None:
...
+        lower_court_name = lower_court_data.get("name", "")
+        if lower_court_id:
+            try:
+                lower_court = Court.objects.get(pk=lower_court_id)
+            except Court.DoesNotExist:
+                logger.error(
+                    "Could not find lower court with ID %s to set appeal_from for Texas docket.",
+                    lower_court_id,
+                )
+                if lower_court_data["court_type"] == CourtType.APPELLATE.value:
```
albertisfu: Here there is a potential error: 

```
  File "/opt/courtlistener/cl/corpus_importer/tasks.py", line 4797, in merge_texas_docket
    if lower_court_data["court_type"] == CourtType.APPELLATE.value:
       ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^
KeyError: 'court_type'
```

The problem I see is that `lower_court_data` is assigned previously either with `appeals_court` or `originating_court` data.

However when this is assigned with `appeals_court` data. It doesn't have a `court_type`.

I was thinking in something like:

```
        if texas_docket_has_appellate_info(docket_data):
            lower_court_data = docket_data["appeals_court"]
            lower_court_id = texas_js_court_id_to_court_id(
                lower_court_data["court_id"]
            )
            lower_court_name = lower_court_data.get("district", "")
        else:
            lower_court_data = docket_data["originating_court"]
            lower_court_id = texas_originating_court_to_court_id(
                lower_court_data
            )
            lower_court_name = lower_court_data.get("name", "")

        if lower_court_id:
            try:
                lower_court = Court.objects.get(pk=lower_court_id)
                docket.appeal_from = lower_court
            except Court.DoesNotExist:
                logger.error(
                    "Could not find lower court with ID %s to set appeal_from for Texas docket.",
                    lower_court_id,
                )
        docket.appeal_from_str = lower_court_name
        docket.save()
```

Can you cover this escenario in a test please? that `appeal_from_str` is populated even if the `court_id` is not found in the DB.

albertisfu (approved):
Thanks for all the hard work here @MorganBennetDev this is now ready to be merged!

PR #6867, opened 2026-02-04
feat(help): Add new Field Help template

### LINKED ISSUE #6866: Apply new design to Field help page
### Describe the issue.

Refs #5353 
Replaces #5838 

### Related Issue

_No response_

### Screenshots

_No response_

To Do

EstiShay:
## Fixes

Resolves #6866 
Replaces #5838 
 
## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

<details closed>
<details open>
image
</details>
<details open>
image
</details>


cl/api/templates/v2_field-help.html
```diff
@@ -0,0 +1,91 @@
...
+
+{% block title %}Advanced Help for API Fields – CourtListener.com{% endblock %}
+{% block og_title %}Advanced Help for API <Fields></Fields> – CourtListener.com{% endblock %}
+
+{% block description %}Learn more about specific fields in the CourtListener API.{% endblock %}
+{% block og_description %}Learn more about specific fields in the CourtListener API.{% endblock %}
+
+{% block footer-scripts %}
+  {% include "includes/anchors.html" %}
+{% endblock %}
```
elisa-a-v: 👀

cl/api/templates/v2_field-help.html
```diff
@@ -0,0 +1,91 @@
...
+  </c-code>
+  <p>There are three sections:</p>
+  <ol>
+    <li><strong>$type</strong>: This is the type of object that has been returned, for example, "docket" indicates that you have gotten a docket as your result.</li>
+    <li><strong>$id</strong>: This is a numeric ID for the document. This value increments as we add content to the system. Note that due to deletions and modifications the numeric IDs are not always sequential, but they will never be duplicated within a document type.</li>
+    <li>
+      <p><strong>$name-of-the-case</strong>: This is the "<a class="underline" href="https://en.wikipedia.org/wiki/Slug_%28publishing%29">slug</a>" of the document, and generally mirrors its case name. This value can change if we clean up a case name, but provided it is not omitted completely, this part of the URL can be any value without affecting the page that is loaded.</p>
+      <p>Put another way, we load pages based on the <code>$id</code> and <code>$type</code>, not by the name of the case.</p>
+    </li>
+  </ol>
```
elisa-a-v: nitpick alert: could we fix the indentation here please? 🙏🏽 

```suggestion
    <c-code>
/$type/$id/$name-of-the-case/
    </c-code>
    <p>There are three sections:</p>
    <ol>
      <li><strong>$type</strong>: This is the type of object that has been returned, for example, "docket" indicates that you have gotten a docket as your result.</li>
      <li><strong>$id</strong>: This is a numeric ID for the document. This value increments as we add content to the system. Note that due to deletions and modifications the numeric IDs are not always sequential, but they will never be duplicated within a document type.</li>
      <li>
        <p><strong>$name-of-the-case</strong>: This is the "<a class="underline" href="https://en.wikipedia.org/wiki/Slug_%28publishing%29">slug</a>" of the document, and generally mirrors its case name. This value can change if we clean up a case name, but provided it is not omitted completely, this part of the URL can be any value without affecting the page that is loaded.</p>
        <p>Put another way, we load pages based on the <code>$id</code> and <code>$type</code>, not by the name of the case.</p>
      </li>
    </ol>
```

elisa-a-v (changes requested):
Looks great, but I think maybe both lists in this page could be turned into [description lists](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl), which would be better semantics and therefore better a11y. What do you think? We have an example of a description list in the new homepage.

Also just noting that this one has the same HTML structure issue as others flagged in #6987 so if we don't fix that here, we should add a comment or tweak the description of #6987 to include this template too.

EstiShay:
> Also just noting that this one has the same HTML structure issue as others flagged in #6987 so if we don't fix that here, we should add a comment or tweak the description of #6987 to include this template too.

I added this file to the list in that issue.

PR #6892, opened 2026-02-09
6115 SCOTUS case importer command

### LINKED ISSUE #6115: Make script to import SCOTUS html, htm, and JSON files
### Describe the issue.

For the data that we already scraped https://github.com/freelawproject/ai-research/issues/25, and for the data that we will scrape on an ongoing basis https://github.com/freelawproject/courtlistener/issues/6109, we need to make a script to import them to be used by the parsers https://github.com/freelawproject/courtlistener/issues/6104

### Related Issue

_No response_

### Screenshots

_No response_

To Do

albertisfu:
## Fixes

Fixes #6115

## Summary

This PR introduces the `import_scotus_dockets` command, which triggers a Celery task that processes SCOTUS docket files stored in S3, parses them, and then triggers a secondary task to ingest the dockets, docket entries, and documents into the database.

It works by reading an S3 inventory file, iterating through the rows, and triggering the download, parsing, and ingestion for each entry. The command stores a checkpoint in Redis every 100 cases, so if the command crashes, it can be easily resumed.

The command can be run with:

`./manage.py import_scotus_dockets --inventory-file inventory/scotus_test.csv --retrieval-queue retrieval --ingesting-queue ingest --inventory-rows 3`

We’ll need to wait for the SCOTUS inventory to be available before running this command:
https://github.com/freelawproject/infrastructure/issues/605

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`


cl/corpus_importer/tasks.py
Brennan-Chesley-FLP, line 3452: There's a potential race condition here if we process in parallel the same doc twice. Just flagging this in case (like in the Texas/TAMES case) that's a possibility.

cl/corpus_importer/tasks.py
```diff
@@ -3717,6 +3724,48 @@ def download_scotus_document_pdf(self: Task, doc_pk: int) -> int | None:
...
+        "html": SCOTUSDocketReportHTML,
+        "htm": SCOTUSDocketReportHTM,
+    }
+    if not (parser_cls := parser_map.get(ext)):
+        logger.error("Unsupported file extension: %s for %s", ext, s3_key)
+        self.request.chain = None
+        return None
+
+    parser = parser_cls()
+    parser._parse_text(content)
```
Brennan-Chesley-FLP, line 3812: Do we want to wrap this in a try, or fail fast on potential errors?

cl/corpus_importer/management/commands/import_scotus_dockets.py
```diff
@@ -0,0 +1,134 @@
...
+                        queue=retrieval_queue
+                    ),
+                    process_scotus_docket.s().set(queue=ingesting_queue),
+                ).apply_async()
+                time.sleep(delay)
+
+                if row_idx % 100 == 0:
+                    processed = row_idx - start_row
+                    progress = (
+                        f" ({processed / total_rows:.1%})"
```
Brennan-Chesley-FLP: I have a slight preference for either report progress across all resumes or remaining work here. (either not subtracting the start_row from processed, or subtracting it from total_rows when calculating progress). (row_idx-start_row)/row_idx/total_rows would also be fine, I think.

Brennan-Chesley-FLP:
Looks good. Three little operational questions/suggestions, but each are ignorable if you'd like!

cl/corpus_importer/tasks.py
albertisfu: The DB lock here protects the entire transaction, so no other worker can merge docket entries at the same time. However, it was still possible to merge SCOTUS documents in parallel. I’ve moved the `merge_scotus_document` process inside the same transaction so it’s also protected by the same DB lock.

Let me know what you think.

cl/corpus_importer/tasks.py
albertisfu: Yeah, I think it’s better to let potential errors fail loudly so we can debug them and fix any issues if necessary. We’ll see if there are errors we can safely ignore, but for now, it seems better to fail loudly.

cl/corpus_importer/management/commands/import_scotus_dockets.py
albertisfu: Sure. I've updated the log to consider the overall progress.

albertisfu:
Thanks @Brennan-Chesley-FLP I've applied your suggestions here. Let me know what do you think.

cl/corpus_importer/tasks.py
Brennan-Chesley-FLP: You're right! I had in my mind a different primary key scheme for this where this could potentially have a weird race. I think we still may get entirely duplicate dockets if we inadvertently downloaded two copies of the same docket, but there's no race.

zvisrosen:
Sorry for wandering in without much knowledge, but was this to pull SCOTUS docs from their docket or to get the older dockets from the National Archives available here? https://catalog.archives.gov/search?recordGroupNumber=267&objectType=%22html%22

mlissner:
Just from the SCOTUS website for now!

PR #6896, opened 2026-02-10
feat(help): Update new Docs template

### LINKED ISSUE #6837: Apply updated design to Docs landing page
### Describe the issue.

Refs #5353 
See also #5854 

To Do

EstiShay:
## Fixes

This resolves #6837 

## Summary

This PR includes another layout component that is built off the `layout-with-navigation` component but eliminates the navigation, adds a more flexible table that allows for adjusting `colspan`, and adds the new table to the components library.

I'm not thrilled with how the content table turned out; I'd rather make it more adaptable, with more options for merging columns, but I'm not comfortable taking any more time tweaking it, with so many other pages waiting.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

1. To deploy...

<details closed>
<details open>
image
<details open>
image
</details>

### DIFF
--- a/cl/api/templates/v2_docs.html
+++ b/cl/api/templates/v2_docs.html
@@ -39,7 +39,7 @@
     >
       <tr class="border-b-greyscale-200 border-b">
           <c-flex-cell header aria-label="Replication" class="text-center">
-            <a href="{% url "replication_docs" %}" class="btn-primary justify-center items-center gap-4 inline-flex min-w-[124px] w-fit py-2 px-1">Replication</a>
+            <c-button variant="primary" href="{% url "replication_docs" %}" class="justify-center min-w-[124px]">Replication</c-button>
           </c-flex-cell>
           <c-flex-cell>Instant</c-flex-cell>
           <c-flex-cell>Complete</c-flex-cell>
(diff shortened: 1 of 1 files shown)


cl/assets/templates/cotton/options_table.html
elisa-a-v, line 1: This looks almost identical to the data table component, but less flexible. I may be missing something though but isn't it possible to tweak the data table without breaking previous uses by extending it to support whatever use case we want to cover with this new component?

I love refactors and would be more than happy to hop on a huddle for some pair programming if that helps!

cl/assets/tailwind/input.css
```diff
@@ -127,6 +127,9 @@
   .screenshot {
     @apply rounded-2xl block max-w-full;
   }
+  .subtext {
+    @apply text-xs text-greyscale-700;
+  }
```
elisa-a-v: Are we sure we need this? so far we're only using it once, which makes me feel it's a bit [premature](https://v3.tailwindcss.com/docs/reusing-styles#avoiding-premature-abstraction) to turn into its own class. We want to avoid adding new classes to keep with tailwind's best practices.

elisa-a-v (changes requested):
Thanks! Looks good for the most part but doesn't look exactly like the new design just yet:
1. We're missing the eyebrow
2. Spacing is a bit too tight between elements and sections
3. Heading width should be capped so it wraps
4. Buttons should be the same width
5. Vertical padding should be slightly smaller for each row in the table

Figma elements have all the details for specific widths, paddings, etc.

<details>
<summary>Figma</summary>

Image

</details>

<details>
<summary>Current implementation</summary>

Image

</details>

Other than that, I just have a couple of questions about the new component and utility class introduced here, both explained inline.

EstiShay:
Fixes for `v2_components.html` front end errors are in PR #6973 , so I hid that comment from the bot.

elisa-a-v:
I was tempted to merge just to keep the flex table since it's a new component that seems like it could be handy, but nothing else uses them. Given that we're moving the help pages from CL to the new wiki, I think it's best if we just close this, then we can always come back to this if we need the component in the future.

Thank you for the updates anyway, it does look much better now! 👍🏽

elisa-a-v:
Blocked by #7169, once that is merged we need to update the buttons in this one.

EstiShay:
I've looked over the recent changes and this is good to go as soon as the new `c-button` component is approved and added.

Issue #6885, opened 2026-02-07
Fix Similar Cases feature that is currently blockd


mlissner:
### What happened?

We discovered that our outage last week was caused by the "Similar Cases" feature, which makes pretty nasty queries on Elastic. Somebody was crawling it, and that was enough to bring us down. 

To bring the site up, we blocked access to any URL containing "related:". That's how things stand, and we've received no complaints yet.

There are a few things we can do: 

1. We can make the Similar Cases feature faster. There are a number of settings we can play with: 

    ```
    RELATED_COUNT = 20
    RELATED_USE_CACHE = True
    RELATED_CACHE_TIMEOUT = 60 * 60 * 24 * 7
    RELATED_MLT_MAXQT = 10
    RELATED_MLT_MINTF = 5
    RELATED_MLT_MAXDF = 1000
    RELATED_MLT_MINWL = 3
    RELATED_MLT_MAXWL = 0
    ```
    
    Some ideas:
     - Reduce `RELATED_COUNT` to 10 instead of 20
     - Play with the QT, TF, DF, and WL settings. [See here](https://www.elastic.co/docs/reference/query-languages/query-dsl/query-dsl-mlt-query#mlt-query-term-selection). I just skimmed those docs and didn't see much that looked super helpful. We did optimize these when we first made this system.

1. We can block going beyond page 5 of related results.

1. We can make the cache a month instead of a week.

1. Throttle the feature at the IP level? This doesn't help much for API users and seems hard though.

1. Switch Similar Cases to use semantic search.

    The way this would work is to pull the embedding(s) for the case from S3, then make a semantic search query with it, excluding itself from the results. Should be reliable, fast, and should provide better results than the MLT algorithm. 

    Questions:
     - What do we do when there are multiple embeddings for a single opinion? Just use the first? Try to get them all? Adds complexity.
     - If we do this, can we remove any index parameters from our case law index? Term vectors, for example? That might be nice to be able to reclaim that memory.

----

I think the best approach might be to reduce the result count to ten, block going past page 5, and lengthen the cache. That should help and be enough to allow the feature back. 

With that done, I think the long-term strategy should be to switch to semantic search. The speed will be consistent and the quality will be better.

@rachlllg, can you please chime in with any thoughts you have about the SS approach. 

@ERosendo, can you give me some time estimates, please, of how hard each part of this might be? 

@flooie, copying you FYSA!

### Expected behavior

The feature should work and perform well.

### Actual behavior

[...]

PR #6904, opened 2026-02-10
fix(search): increases `RELATED_CACHE_TIMEOUT` to 30 days

ERosendo:
## Fixes

This fix related to #6885  but doesn't fully solve it.

## Summary

This PR increases `RELATED_CACHE_TIMEOUT` from 7 days to 30 days.

After experiencing high CPU usage caused by crawlers hitting the Similar Cases feature, we identified the expensive MLT queries against ES as the root cause. Extending the TTL to 30 days reduces how often these costly queries need to be recomputed, lowering the overall load on ES.

This is part of a broader effort to prevent future outages related to this feature. Additional changes (tuning MLT parameters, reducing result count, capping pagination depth) will follow in separate PRs.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


mlissner (approved):
Yep!

PR #6907, opened 2026-02-10
feat(ai): add GenAI batch wrapper and management commands

quevon24:
## Summary

- Add `GoogleGenAIBatchWrapper`, a wrapper around the Google GenAI API supporting system prompt caching, batch request preparation, execution, result downloading, and response parsing
- Add `_ResponseValidator` helper to normalize and extract text/errors from both batch and single API response formats
- Add `send_gemini_file_batches` management command to fetch PDFs from S3, create LLMRequest/LLMTask records, and submit batch jobs to the Gemini API. Includes `--batch-size` control (default 1000) to automatically split large S3 paths into multiple LLMRequests with all-or-nothing transaction semantics
- Add `check_gemini_batch_status` management command to poll in-progress batch jobs, download results, update task statuses, and save response files. Uses `JobState` enums from the Google GenAI library instead of hardcoded strings, and handles `JOB_STATE_PARTIALLY_SUCCEEDED` by processing available results and marking orphaned tasks as failed
- Add `google-genai` dependency
- Add tests for the wrapper, response validator, and both management commands covering success paths, validation, API failures, batch splitting, and edge cases
- Depends on https://github.com/freelawproject/courtlistener/pull/6654 which provides the Django models, migrations, and admin

Paired with @quevon24 and [Claude Code](https://claude.com/claude-code)

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

1. To deploy...


cl/ai/management/commands/check_gemini_batch_status.py
```diff
@@ -0,0 +1,243 @@
...
+                "different API keys for different tasks."
+            )
+
+        pending_requests = LLMRequest.objects.filter(
+            is_batch=True,
+            status=LLMTaskStatusChoices.IN_PROGRESS,
+            provider=LLMProvider.GEMINI,
+        )
+        logger.info(
+            f"Found {len(pending_requests)} pending batch requests to check."
```
grossir: Use pending_requests.count() to prevent evaluating the queryset at once

cl/ai/management/commands/check_gemini_batch_status.py
```diff
@@ -0,0 +1,243 @@
...
+                        ContentFile(
+                            json.dumps(
+                                res["raw_result"],
+                                indent=2,
+                                ensure_ascii=False,
+                            ).encode("utf-8")
+                        ),
+                    )
+                except (TypeError, ValueError) as e:
+                    logger.warning(
```
grossir: this should be a logger.error, at least at the beggining, so we can now why it failed

cl/ai/management/commands/send_gemini_batches.py
```diff
@@ -0,0 +1,557 @@
...
+                task.input_file.save(filename, ContentFile(file_content))
+                logger.info(f"  - Stored file: {filename}")
+            else:
+                task.input_file.name = s3_key
+                task.save()
+                logger.info(f"  - Referenced S3 file: {s3_key}")
+
+            temp_file = tempfile.NamedTemporaryFile(
+                delete=False, suffix=os.path.splitext(filename)[1]
+            )
```
grossir: do we need some logic to ensure all temp_files are deleted once the command exits?

I remember we had a problem with doctor. I think by default the tmp directory is not cleaned

cl/ai/tests.py
```diff
@@ -34,19 +40,1617 @@ def test_create_llm_task(self):
...
+    )
+    def test_send_gemini_batches_with_store_input_files(
+        self,
+        mock_boto3_client,
+        mock_wrapper_class,
+        mock_tempfile,
+        mock_remove,
+    ):
+        """Test --store-input-files flag stores files in CourtListener storage."""
+        from django.core.management import call_command
```
grossir: this import is repeated a bunch of times, why not just a single time at the top of the file?

cl/ai/management/commands/send_gemini_batches.py
quevon24: There is a function that helps do this: `cleanup_temp_files`, but your comment helped me identify a bug

cl/ai/management/commands/check_gemini_batch_status.py
```diff
@@ -0,0 +1,243 @@
...
+                    logger.error(
+                        f"Could not serialize JSON for task {task.llm_key}: {e}"
+                    )
+                    task.response_file.save(
+                        f"{task.llm_key}_result.txt",
+                        ContentFile(str(res["raw_result"]).encode("utf-8")),
+                    )
+            task.save()
+
+        request.status = LLMTaskStatusChoices.FINISHED
```
grossir: should use LLMRequestStatusChoices? Similarly in other requests

cl/ai/management/commands/check_gemini_batch_status.py
```diff
@@ -0,0 +1,243 @@
...
+        )
+    except Exception as e:
+        logger.exception(
+            f"Unexpected error in check_gemini_batch_status for request {request.pk}"
+        )
+        sentry_sdk.capture_exception(e)
+
+        request.status = LLMTaskStatusChoices.FAILED
+        request.date_completed = now()
+        request.save()
```
grossir: what happens if this was a network error, or an internal server error, that is actually retryable?

We would be marking the request as failed and we shouldn't

Maybe you could anticipate some network error classes and mark them as still in progress?

Also, should use 
LLMRequestStatusChoices

cl/ai/management/commands/send_gemini_batches.py
```diff
@@ -0,0 +1,558 @@
...
+            api_model_name=model,
+            status=LLMTaskStatusChoices.IN_PROGRESS,
+            date_started=now(),
+        )
+        llm_request.prompts.set([system_prompt, user_prompt])
+        logger.info(f"Created LLMRequest with ID: {llm_request.pk}")
+
+        tasks_data, temp_files = fetch_files_and_create_tasks(
+            llm_request, s3_path, bucket, store_files
+        )
```
grossir: A LLMRequest is saved to DB, then we check for files `fetch_files_and_create_tasks`. If something goes wrong, we delete the LLMRequest and cascade delete the created tasks. 

Wouldn't it be better to
- create the object, but not save it to the DB
- create the tasks objects and other related objects
- only save all the objects once we are done with `fetch_files_and_create_tasks`

We do it this way when ingesting opinions, seem like a cleaner way in case something goes wrong

cl/ai/management/commands/check_gemini_batch_status.py
```diff
@@ -0,0 +1,243 @@
...
+            status=LLMTaskStatusChoices.SUCCEEDED
+        ).count()
+        request.failed_tasks = request.tasks.filter(
+            status=LLMTaskStatusChoices.FAILED
+        ).count()
+        request.date_completed = now()
+        request.save()
+
+
+def process_failed_request(request: LLMRequest, job_state_name: str) -> None:
```
grossir, line 172: should update the failed tasks counter?

cl/ai/management/commands/check_gemini_batch_status.py
```diff
@@ -0,0 +1,243 @@
...
+from django.db import transaction
+from django.utils.timezone import now
+
+from cl.ai.llm_providers.google import GoogleGenAIBatchWrapper
+from cl.ai.models import LLMProvider, LLMRequest, LLMTaskStatusChoices
+from cl.lib.command_utils import VerboseCommand
+
+logger = logging.getLogger(__name__)
+
+COMPLETED_STATES = {
```
grossir, line 85: should we include here JOB_STATE_PARTIALLY_SUCCEEDED

https://github.com/googleapis/python-genai/blob/1fdb4b87aaec6e58b415168ea5893c0e901819a9/google/genai/types.py#L611

grossir, line 85: about this, given that the APIs tend to change, shouldnt we use the package `from google.genai import types`. We are already using it cl.ai.llm_providers.google


```
        if job.state != types.JobState.JOB_STATE_SUCCEEDED:
```

This is better than a hardcoded string agains future changes

cl/ai/tests.py
```diff
@@ -50,3 +60,1562 @@ def test_create_llm_task(self):
...
+                # Execute command
+                call_command("check_gemini_batch_status")
+
+                # Assertions - all three states follow the same code path
+                mock_wrapper.download_results.assert_not_called()
+
+                # Check LLMRequest
+                llm_request.refresh_from_db()
+                self.assertEqual(
+                    llm_request.status, LLMTaskStatusChoices.FAILED
```
grossir: this is a request, so the used enum should LLMRequestStatusChoices

cl/ai/tests.py
```diff
@@ -50,3 +60,1562 @@ def test_create_llm_task(self):
...
+        mock_wrapper.download_results.return_value = "{}"
+        mock_wrapper.process_results.return_value = []
+
+        # Execute command
+        call_command("check_gemini_batch_status")
+
+        # Assertions
+        running_request.refresh_from_db()
+        self.assertEqual(
+            running_request.status, LLMTaskStatusChoices.IN_PROGRESS
```
grossir: should use LLMRequestStatusChoices

cl/ai/tests.py
```diff
@@ -50,3 +60,1562 @@ def test_create_llm_task(self):
...
+        )
+        self.assertIsNone(running_request.date_completed)
+
+        succeeded_request.refresh_from_db()
+        self.assertEqual(
+            succeeded_request.status, LLMTaskStatusChoices.FINISHED
```
grossir: same as previous comments

cl/ai/management/commands/check_gemini_batch_status.py
```diff
@@ -0,0 +1,286 @@
...
+    (reported to Sentry) to ensure one failing request doesn't stop
+    processing of subsequent requests.
+
+    :param wrapper: Google GenAI batch wrapper used to query job status.
+    :param request: The in-progress LLMRequest to check.
+    """
+    try:
+        job = wrapper.get_job(request.batch_id)
+
+        if job.state.name not in COMPLETED_STATES:
```
grossir: `job.state.name` is a string
COMPLETED_STATES contains enum types

This works, but it wouldn't pass a type checker?

```python
In [1]: from google.genai.types import JobState

In [2]: JobState.JOB_STATE_SUCCEEDED.name
Out[2]: 'JOB_STATE_SUCCEEDED'

In [3]: JobState.JOB_STATE_SUCCEEDED.name in {JobState.JOB_STATE_SUCCEEDED,}
Out[3]: True

In [4]: JobState.JOB_STATE_FAILED.name in {JobState.JOB_STATE_SUCCEEDED,}
Out[4]: False

In [5]: str(JobState.JOB_STATE_SUCCEEDED)
Out[5]: 'JobState.JOB_STATE_SUCCEEDED'

In [6]: type(JobState.JOB_STATE_SUCCEEDED)
Out[6]: <enum 'JobState'>

In [7]: type(JobState.JOB_STATE_SUCCEEDED.name)
Out[7]: str
```

cl/ai/management/commands/check_gemini_batch_status.py
```diff
@@ -0,0 +1,286 @@
...
+            logger.info(
+                f"  - Job state is '{job.state.name}'. Skipping for now."
+            )
+            return
+
+        if job.state.name == "JOB_STATE_SUCCEEDED":
```
grossir: should use the library, not a hardcoded string 

if job.state.name == JobState.JOB_STATE_SUCCEEDED.name:

cl/ai/management/commands/check_gemini_batch_status.py
```diff
@@ -0,0 +1,286 @@
...
+        pending_requests = LLMRequest.objects.filter(
+            is_batch=True,
+            status=LLMRequestStatusChoices.IN_PROGRESS,
+            provider=LLMProvider.GEMINI,
+        )
+        logger.info(
+            f"Found {pending_requests.count()} pending batch requests to check."
+        )
+
+        if not pending_requests:
```
grossir: should be `if not pending_requests.exists()`, because `not` will evaluate the queryset 

But even better, just save the `pending_requests.count()` into a varialbe

Then, you can do `if count == 0` instead of `if not pending_requests:` or `if not pending_requests.exists()`

cl/ai/management/commands/check_gemini_batch_status.py
grossir: also, what about handling JOB_STATE_PARTIALLY_SUCCEEDED ?

Some tasks have succeded, some have failed. It doesn't fit any of the current conditional clauses

cl/ai/management/commands/send_gemini_batches.py
```diff
@@ -0,0 +1,545 @@
...
+    :returns: A list of dicts with ``llm_key`` and ``input_file_path`` keys.
+    """
+    logger.info(f"Fetching PDF files from S3: s3://{bucket}/{s3_path}")
+    tasks_data: list[dict] = []
+
+    for filename, file_content, s3_key in get_s3_file_list(
+        s3_path=s3_path,
+        bucket_name=bucket,
+        file_extension=".pdf",
+    ):
```
grossir: this will iterate over all the files matching the S3 path. Is there a batch size limit in the Google batch API? There should be a batch size control somewhere

quevon24: There is no API limit but the documentation mention this:

> Break up very large batches: While the target turnaround time is 24 hours, actual processing time can vary based on system load and job size. For large jobs, consider breaking them into smaller batches if intermediate results are needed sooner.

https://ai.google.dev/gemini-api/docs/batch-api?batch=file#best-practices

So probably it is a good idea to set a limit and create ore LLMRequest objects if we reach the limit or what do you think?

quevon24:
I renamed `send_gemini_batches` to `send_gemini_file_batches` since this command is specific to S3 file (PDF) processing. This leaves room for a future `send_gemini_text_batches` command for plain text tasks like case name extraction.

cl/ai/llm_providers/google.py
```diff
@@ -0,0 +1,306 @@
...
+
+    def download_results(self, job: types.BatchJob) -> str:
+        """
+        Downloads the result file of a completed job and returns its raw content.
+
+        :param job: A completed Google GenAI BatchJob object.
+        :return: The raw JSONL content of the result file as a string.
+        :raises ValueError: If the job has not succeeded.
+        """
+        if job.state != types.JobState.JOB_STATE_SUCCEEDED:
```
grossir: we should also keep jobs with state  JOB_STATE_PARTIALLY_SUCCEEDED ?

cl/ai/management/commands/check_gemini_batch_status.py
```diff
@@ -0,0 +1,302 @@
...
+                "Request %s has been in progress since %s "
+                "(over %s). Marking as failed.",
+                request.pk,
+                request.date_started,
+                MAX_REQUEST_AGE,
+            )
+            sentry_sdk.capture_exception(e)
+            request.status = LLMRequestStatusChoices.FAILED
+            request.date_completed = now()
+            request.save()
```
grossir: should the tasks be marked as failed too?

cl/ai/management/commands/send_gemini_file_batches.py
```diff
@@ -0,0 +1,592 @@
...
+        s3_files = get_s3_file_list(
+            s3_path=s3_path,
+            bucket_name=bucket,
+            file_extension=".pdf",
+        )
+
+        all_temp_files: list[str] = []
+        batch_number = 0
+        try:
+            with transaction.atomic():
```
grossir: I think the transaction.atomic block should be inside the loop

say that you get 5 chunks. If this succeeds for the first chunk, and fails at the second; even the first LLMRequest will be deleted, but it will have been sent to Google. We will lose track of that batch, and will send it again, making us incur in a duplicated cost.

cl/ai/management/commands/check_gemini_batch_status.py
```diff
@@ -0,0 +1,302 @@
...
+    with transaction.atomic():
+        request.status = LLMRequestStatusChoices.FAILED
+        request.date_completed = now()
+        request.failed_tasks = request.tasks.all().count()
+        for task in request.tasks.all():
+            task.status = LLMTaskStatusChoices.FAILED
+            task.error_message = (
+                f"Batch job failed with state: {job_state_name}"
+            )
+            task.save()
```
grossir: wouldn't an update be better

request.tasks.all().update(status=, error_message=)

cl/ai/management/commands/send_gemini_file_batches.py
```diff
@@ -0,0 +1,612 @@
...
+    PromptTypes,
+)
+from cl.lib.command_utils import VerboseCommand
+
+logger = logging.getLogger(__name__)
+
+# Supported Gemini models for batch operations
+SUPPORTED_GEMINI_MODELS = {
+    "gemini-3-flash-preview",
+    "gemini-3-pro-preview",
```
grossir: this one has already been deprecated
https://ai.google.dev/gemini-api/docs/deprecations?_gl=1*12zemyl*_up*MQ..&gclid=CjwKCAjw687NBhB4EiwAQ645djA8nqZ7RQuvhM1x1DKhHNCQ1J4Y91jfyhUBy3r0QJvJzXg2n_-xLxoCJn4QAvD_BwE&gbraid=0AAAAACn9t66ov9QSAsrRN-H-dfsnmRGAJ#gemini-3-models

cl/ai/management/commands/send_gemini_file_batches.py
```diff
@@ -0,0 +1,679 @@
...
+    :param system_prompt_id: PK of the system Prompt.
+    :param user_prompt_id: PK of the user Prompt.
+    :returns: Set of S3 keys that have already been processed.
+    """
+    return set(
+        LLMTask.objects.filter(
+            status__in=[
+                LLMTaskStatusChoices.SUCCEEDED,
+                LLMTaskStatusChoices.FINISHED,
+            ],
```
grossir, line 467: we should also avoid stuff in progress?

quevon24: No, I think we should only skip files that we know completed successfully (SUCCEEDED or FINISHED). If a task is still IN_PROGRESS, we don't know the outcome yet. If we skip it and the batch ends up failing, that file would never get reprocessed. Better to have an occasional duplicate than a silently skipped file. If someone re-runs the command after the in-progress batch finishes, check_gemini_batch_status will have resolved those tasks to SUCCEEDED or FAILED, and dedup will handle them correctly at that point.

cl/ai/management/commands/send_gemini_file_batches.py
```diff
@@ -0,0 +1,679 @@
...
+    --cache-name        Stable name for system prompt cache (for cost savings)
+                        Example: "scan-extraction-cache-v1"
+
+    --bucket            S3 bucket name (default: AWS_STORAGE_BUCKET_NAME)
+                        Must be in allowed list for security
+
+    --store-input-files Store files in CourtListener storage (creates duplicates)
+                        Default: Only S3 references are stored (no duplication)
+
+    --batch-size        Maximum files per batch request (default: 5000)
```
grossir, line 39: default is 1000?

PR #6913, opened 2026-02-12
Update `SCOTUSDocument`, `TexasDocument`, and `CaseTransfer` models

MorganBennetDev:
## Summary

Makes the following changes to models

`SCOTUSDocument`: Increase `max_length` of `url` field to 1,000 characters to avoid validation errors when ingesting SCOTUS docket attachments.

`TexasDocument`: Increase `max_length` of `url` field to 250 characters to avoid validation errors when ingesting Texas docket attachments.

`CaseTransfer`: Change `origin_docket` and `destination_docket` foreign key fields to `origin_docket_number` and `destination_docket_number` text fields. This is because we have no guarantee that dockets on both ends of this relationship will exist while populating this table. In fact, they may never exist if the transfer is to or from a court we do not scrape.

## Deployment

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`


cl/search/models.py
```diff
@@ -4007,26 +4028,39 @@ class CaseTransfer(AbstractDateTimeModel):
         on_delete=models.CASCADE,
         related_name="case_transfer_origin_court",
     )
+    origin_docket_number = models.TextField()
     origin_docket = models.ForeignKey(
         "search.Docket",
         on_delete=models.CASCADE,
         related_name="case_transfer_origin_docket",
+        blank=True,
```
albertisfu, line 4036: If we want this foreign key to be optional, we also need to add null=True, since blank=True only works at the validation level.

```suggestion
        blank=True,
        null=True,
```

cl/search/models.py
```diff
@@ -4007,26 +4028,39 @@ class CaseTransfer(AbstractDateTimeModel):
...
         "search.Court",
         on_delete=models.CASCADE,
         related_name="case_transfer_destination_court",
     )
+    destination_docket_number = models.TextField()
     destination_docket = models.ForeignKey(
         "search.Docket",
         on_delete=models.CASCADE,
         related_name="case_transfer_destination_docket",
+        blank=True,
```
albertisfu, line 4049: same here
```suggestion
        blank=True,
        null=True,
```

cl/search/models.py
```diff
@@ -4007,26 +4028,39 @@ class CaseTransfer(AbstractDateTimeModel):
         on_delete=models.CASCADE,
         related_name="case_transfer_origin_court",
```
albertisfu, line 4029: Should `origin_court` and `destination_court` also be optional, since we don’t know what data will be available at creation time?

albertisfu (changes requested):
Thanks @MorganBennetDev I just left a couple of comments.

cl/search/models.py
MorganBennetDev: @albertisfu Both courts seem to always be available at creation time for Texas. I don't think we should make either optional since then the only thing we have to match against when populating missing transfer data is docket numbers, which are not guaranteed to be unique between courts.

albertisfu (approved):
Thanks @MorganBennetDev this is ready to be merged.

PR #6940, opened 2026-02-13
fix(tames) Only keep latest copy of docket html,meta,json

Brennan-Chesley-FLP:
## Summary

This allows alternate naming strategies to be passed to the S3GlacierInstantRetrievalStorage constructor, with a default of the incrementing renaming strategy that is currently in use. It also sets the docket scraper for Texas cases to use the clobbering naming strategy.

## Deployment

**This PR should:**

    - [X] `skip-web-deploy`
    
    
    - [X] `skip-celery-deploy`
    
    
    - [X] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

This change should only affect one management command.

### DIFF
--- a/cl/lib/storage.py
+++ b/cl/lib/storage.py
@@ -141,6 +141,7 @@ class S3GlacierInstantRetrievalStorage(S3Storage):
     ) -> str:
         return self.naming_strategy(self, name, max_length)
 
+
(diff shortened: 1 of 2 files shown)


Brennan-Chesley-FLP:
Code was merged as part of #6959.

PR #6949, opened 2026-02-14
fix(alerts): Refines `scheduled_alert_hits_limit_reached`

### LINKED ISSUE #6947: Fix scheduled_alert_hits_limit_reached, which does not appear to be limiting the number of cases that can be included in a search alert.
### Describe the issue.

We found an alert in https://github.com/freelawproject/courtlistener/issues/6918#issuecomment-3900155039 

that contains a huge number of cases per alert. So it does seem that `scheduled_alert_hits_limit_reached` is not working properly to limit the number of cases an alert can contain.

We should review this method so it properly limit the number of cases to `SCHEDULED_ALERT_HITS_LIMIT` setting.

### Related Issue

_No response_

### Screenshots

_No response_

Research

ERosendo:
## Fixes

This fixes #6947

## Summary

This PR fixes the `scheduled_alert_hits_limit_reached` method not properly enforcing the `SCHEDULED_ALERT_HITS_LIMIT` for opinion alerts.

The root cause was that opinion alerts incorrectly set `child_document=True` in `percolator_response_processing`. When `child_document=True`, `scheduled_alert_hits_limit_reached` counts hits per parent object (cluster), instead of counting the total number of distinct clusters scheduled for a given alert. This meant the method was checking "how many opinions from a given cluster had been scheduled?" which was not the intended constraint, so the overall limit was never enforced.

The fix sets `child_document=False` for opinions, which directs `scheduled_alert_hits_limit_reached` to count distinct `object_id` values (clusters) across the alert, properly enforcing `SCHEDULED_ALERT_HITS_LIMIT`.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/audio/feeds.py
+++ b/cl/audio/feeds.py
@@ -19,7 +19,6 @@ class JurisdictionPodcast(JurisdictionFeed):
         "the CourtListener.com initiative. Not an official podcast."
     )
     subtitle = description
-    summary = description
     iTunes_name = "Free Law Project"
     iTunes_email = "feeds@courtlistener.com"
     iTunes_image_url = f"{static('png/producer-2000x2000.png')}"
@@ -115,7 +114,51 @@ class AllJurisdictionsPodcast(JurisdictionPodcast):
 
 
 class SearchPodcast(JurisdictionPodcast):
-    title = "CourtListener.com Custom Oral Argument Podcast"
+    def _get_search_summary(self, request):
+        """Build a human-readable search summary from GET params.
+
+        Caches on the request to avoid re-parsing the form in
+        title/description/subtitle.
+        """
+        if hasattr(request, "_summary_cache"):
+            return request._summary_cache
+
+        summary = ""
+        search_form = SearchForm(request.GET)
+        if search_form.is_valid():
+            cd = search_form.cleaned_data
+            courts: list[bool] = [
+                v for k, v in cd.items() if k.startswith("court_")
+            ]
+            selected = courts.count(True)
+            court_count_human = (
+                "All" if selected == len(courts) else str(selected)
+            )
+            summary = search_form.as_text(court_count_human)
+
+        request._summary_cache = summary
+        return summary
+
+    def title(self, obj):
+        if summary := self._get_search_summary(obj):
+            return f"{summary} — CourtListener Oral Arguments"
+        return "CourtListener.com Custom Oral Argument Podcast"
+
+    def description(self, obj):
+        if summary := self._get_search_summary(obj):
+            return (
+                f"Oral arguments matching: {summary}. "
+                "A custom podcast by Free Law Project via "
+                "CourtListener.com."
+            )
+        return (
+            "A chronological podcast of oral arguments with improved "
+            "files and meta data. Hosted by Free Law Project through "
+            "the CourtListener.com initiative."
+        )
+
+    def subtitle(self, obj):
+        return self.description(obj)
 
     def get_object(self, request, get_string):
         return request
--- a/cl/audio/tests.py
+++ b/cl/audio/tests.py
@@ -210,6 +210,38 @@ class PodcastTest(ESIndexTestCase, TestCase):
         )
         self.assertTrue(pubdate_not_present)
 
+    def test_search_podcast_has_individualized_metadata(self) -> None:
+        """Does the search podcast include query-specific title and
+        description?"""
+        params = {
+            "q": "magnitsky",
+            "type": SEARCH_TYPES.ORAL_ARGUMENT,
+        }
+        response = self.client.get(
+            reverse("search_podcast", args=["search"]),
+            params,
+        )
+        self.assertEqual(200, response.status_code)
+        xml_tree = etree.fromstring(response.content)
+
+        # Title should contain the query and CourtListener branding
+        title = xml_tree.xpath("//channel/title")[0].text  # type: ignore
+        self.assertIn("magnitsky", title)
+        self.assertIn("CourtListener", title)
+
+        # Description should contain the query
+        description = xml_tree.xpath("//channel/description")[0].text  # type: ignore
+        self.assertIn("magnitsky", description)
+
+        # iTunes subtitle should also contain the query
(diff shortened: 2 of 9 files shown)


cl/alerts/utils.py
```diff
@@ -435,12 +434,8 @@ def scheduled_alert_hits_limit_reached(
             content_type=content_type,
             object_id=object_id,
         ).count()
-        hits_per_result = (
-            settings.OPINION_HITS_PER_RESULT
-            if is_opinion_hit
-            else settings.RECAP_CHILD_HITS_PER_RESULT
-        )
-        hits_limit = hits_per_result + 1
+        # Only RECAP documents use child_document=True
```
albertisfu: It seems that the same issue can impact RECAP search alerts triggered by documents.

Since `child_document` will be True for them, if we ingest 1,000 RDs from different cases, they will fall into this condition and be counted as child hits instead of as a global count of cases per alert.

As a result, all 1,000 alerts will be included in the email.

I'm thinking that a possible solution would be to introduce an additional `scheduled_alert_hits_limit_reached` condition when scheduling alerts: one to count scheduled child documents, and another to count cases in an alert using only the `content_type`.

I think this would also be valid for Opinion alerts. Considering that all opinion alerts are triggered by opinions and not clusters, `child_document` would still be True for all of them. Therefore, the limit of opinions within a cluster in an alert should also be enforced to `OPINION_HITS_PER_RESULT`.

cl/alerts/utils.py
```diff
@@ -458,8 +453,7 @@ def scheduled_alert_hits_limit_reached(
 
     if hits_count >= hits_limit:
         if child_document:
-            log_function = logger.error if is_opinion_hit else logger.info
```
albertisfu, line 461: The reason for using `logger.error` for opinion alerts is that all alerts triggered by opinions are based on opinion instances. We want to be notified if a case has more than `OPINION_HITS_PER_RESULT`, so we can update this value if necessary, since we do not expect an opinion cluster to have more than 20 opinions and we wanted to display all of them.

cl/alerts/utils.py
ERosendo: So the idea is to check both counts for opinions and RECAP documents. For these two record types, we would check:

- Whether the number of child hits for a given parent (cluster/docket) has reached `OPINION_HITS_PER_RESULT` / `RECAP_CHILD_HITS_PER_RESULT`
- Whether the global number of distinct cases for the same alert has reached `SCHEDULED_ALERT_HITS_LIMIT`

If either condition is true, we return `True`. Is that correct?

cl/alerts/utils.py
```diff
@@ -416,6 +416,11 @@ def scheduled_alert_hits_limit_reached(
     """Check if the alert hits limit has been reached for a specific alert-user
      combination.
 
+    For child documents (RECAPDocument, Opinion), two limits are checked:
+        1. Per-parent limit: caps child hits per parent object.
+        2. Global alert limit: caps distinct parent objects per alert.
+    If either limit is reached, returns True.
+
```
albertisfu, line 423: I believe we still need an additional condition to correctly count the number of hits allowed per user/alert.

For instance, with this approach, let’s imagine that we have:

```
SCHEDULED_ALERT_HITS_LIMIT = 5
RECAP_CHILD_HITS_PER_RESULT = 3
```


- Case 1
   - child 1
   - child 2
   - child 3

- Case 2
   - child 1
   - child 2
   - child 3

- Case 3

- Case 4
   - child 1
 
- Case 5
   - child 1


In this example `global_hits_count` will prevent scheduling more child documents into the cases that have not reached the limit of child hits, like Case 3, Case 4 and Case 5.

The limit in this example is 5 cases per user-alert and 3 documents per case.

I think we should check an additional condition to determine what to count.

My idea is the following:

- Check whether there is at least one scheduled hit for the case (parent or child) (grouped by docket_id or cluster_id). If so, that means the case has already been accounted to be part of the email. In that scenario, we should do an additional count to count the child hits within the case to determine whether the hit should be scheduled.

- If there is no scheduled hit for the case, that means it is a new case for the alert–user relation. In that scenario, count the number of cases (`global_hits_count`) already scheduled for the given alert–user pair and determine whether the case should be scheduled.

In that way we can properly limit the number of cases per alert and child hits per case.


cl/alerts/utils.py
```diff
@@ -424,56 +429,51 @@ def scheduled_alert_hits_limit_reached(
...
+                    alert_pk,
+                    object_id,
+                    child_hits_count,
+                )
+                return True
+            # Child limit not reached, allow this hit.
+            return False
+
+    # Global alert limit
+    global_hits_count = (
```
albertisfu, line 464: Found an additional issue with this check.

It seems that the global case count per alert can be bypassed by child hits.

Imagine the following scenario:

```
SCHEDULED_ALERT_HITS_LIMIT = 5
RECAP_CHILD_HITS_PER_RESULT = 3
```

Scheduled case hits:

Case ID 1: content_type → Search | docket and object_id = 1
Case ID 2: content_type → Search | docket and object_id = 2
Case ID 3: content_type → Search | docket and object_id = 3
Case ID 4: content_type → Search | docket and object_id = 4
Case ID 5: content_type → Search | docket and object_id = 5

`SCHEDULED_ALERT_HITS_LIMIT` reached.

New incoming RD hit related to Case ID 6:
content_type → Search | recap document and object_id = 6

In this scenario, base_filter is using content_type = Search | recap document.

So when doing:

`base_filter.only("object_id").distinct("object_id").count()`

It returns 1, which is lower than `SCHEDULED_ALERT_HITS_LIMIT`.

As a result, a new child hit is scheduled for a new case. Now we have 6 cases, which exceeds `SCHEDULED_ALERT_HITS_LIMIT`.

Therefore, we should ensure we do not exceed the global case limit when scheduling a child hit for a new case.

Issue #6901, opened 2026-02-10
Custom podcasts should have individualized metadata


anseljh:
### Describe the issue.

Currently, all custom podcasts have the same RSS metadata, causing them to have the same title and description in podcatcher apps: "CourtListener.com Custom Oral Argument Podcast".

Example: 
https://www.courtlistener.com/podcast/search/?q=magnitsky

```xml
<rss version="2.0">
<channel>
<title>CourtListener.com Custom Oral Argument Podcast</title>
<link>https://www.courtlistener.com/</link>
<description>
A chronological podcast of oral arguments with improved files and meta data. Hosted by Free Law Project through the CourtListener.com initiative. Not an official podcast.
</description>
<atom:link href="https://www.courtlistener.com/podcast/search/" rel="self"/>
<language>en-us</language>
<copyright>Created for the public domain by Free Law Project</copyright>
<lastBuildDate>Mon, 26 Jan 2026 00:00:00 -0500</lastBuildDate>
<itunes:subtitle>
A chronological podcast of oral arguments with improved files and meta data. Hosted by Free Law Project through the CourtListener.com initiative. Not an official podcast.
</itunes:subtitle>
<itunes:author>Free Law Project</itunes:author>
<itunes:summary>
A chronological podcast of oral arguments with improved files and meta data. Hosted by Free Law Project through the CourtListener.com initiative. Not an official podcast.
</itunes:summary>
<itunes:explicit>no</itunes:explicit>
<itunes:owner>
<itunes:name>Free Law Project</itunes:name>
<itunes:email>feeds@courtlistener.com</itunes:email>
</itunes:owner>
<itunes:image href="https://storage.courtlistener.comhttps://storage.courtlistener.com/static/png/producer-2000x2000.645f9b18960d.png"/>
<itunes:category text="Government & Organizations"/>
<item>
<title>Rinat Akhmetshin v. William Browder</title>
<link>
https://www.courtlistener.com/audio/102248/rinat-akhmetshin-v-william-browder/
</link>
<description>Rinat Akhmetshin v. William Browder</description>
<dc:creator>Court of Appeals for the D.C. Circuit</dc:creator>
<pubDate>Mon, 26 Jan 2026 00:00:00 -0500</pubDate>
<guid>
https://www.courtlistener.com/audio/102248/rinat-akhmetshin-v-william-browder/
</guid>
[...]

PR #6960, opened 2026-02-16
Add individualized metadata to custom search podcasts

mlissner:
## Fixes

Fixes: #6901

## Summary

Custom search podcasts at `/podcast/search/?q=...` all shared the same static title ("CourtListener.com Custom Oral Argument Podcast") and description, making them indistinguishable in podcatcher apps.

This PR makes the `SearchPodcast` title, description, and subtitle dynamic based on the search query parameters by reusing `SearchForm.as_text()`. The dynamic part is placed first in the title so it's visible in truncated podcast app UIs:

```
Courts: All › Query: magnitsky — CourtListener Oral Arguments
```

Court count is computed correctly by counting selected `court_*` fields vs total, mirroring the logic in `merge_form_with_courts()`.

Also:
 - Removes dead code: `_court_count` and `_stat_count` were set in `SearchForm.clean()` but never read anywhere in the codebase.
 - Cleans up a flakey test. It just needed an order_by, but I cleaned it up since I was looking at it.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [X] `skip-celery-deploy`
    - [X] `skip-cronjob-deploy`
    - [X] `skip-daemon-deploy`

### DIFF
--- a/cl/audio/feeds.py
+++ b/cl/audio/feeds.py
@@ -157,6 +157,9 @@ class SearchPodcast(JurisdictionPodcast):
             "the CourtListener.com initiative."
         )
 
+    def subtitle(self, obj):
+        return self.description(obj)
+
     def get_object(self, request, get_string):
         return request
 
--- a/cl/audio/tests.py
+++ b/cl/audio/tests.py
@@ -223,9 +223,6 @@ class PodcastTest(ESIndexTestCase, TestCase):
         )
         self.assertEqual(200, response.status_code)
         xml_tree = etree.fromstring(response.content)
-        namespaces = {
-            "itunes": "https://www.itunes.com/dtds/podcast-1.0.dtd",
-        }
 
         # Title should contain the query and CourtListener branding
         title = xml_tree.xpath("//channel/title")[0].text  # type: ignore
@@ -235,6 +232,15 @@ class PodcastTest(ESIndexTestCase, TestCase):
         # Description should contain the query
         description = xml_tree.xpath("//channel/description")[0].text  # type: ignore
         self.assertIn("magnitsky", description)
+
+        # iTunes subtitle should also contain the query
+        subtitle = xml_tree.xpath(
+            "//channel/itunes:subtitle",
+            namespaces={
+                "itunes": "https://www.itunes.com/dtds/podcast-1.0.dtd",
+            },
+        )[0].text  # type: ignore
+        self.assertIn("magnitsky", subtitle)
 
     def test_catch_es_errors(self) -> None:
         """Can we catch es errors and just render an empy podcast?"""


cl/audio/feeds.py
```diff
@@ -115,7 +115,46 @@ def items(self, obj):
...
+        """
+        if hasattr(request, "_summary_cache"):
+            return request._summary_cache
+
+        summary = ""
+        search_form = SearchForm(request.GET)
+        if search_form.is_valid():
+            cd = search_form.cleaned_data
+            courts = [v for k, v in cd.items() if k.startswith("court_")]
+            selected = sum(1 for v in courts if v)
```
ERosendo: The `court_` fields are defined as `BooleanField` in [cl/search/forms.py:517](https://github.com/freelawproject/courtlistener/blob/e304f4ccdc2d86fee46b5c08e20cf8bd70d7d39d/cl/search/forms.py#L517-L522), so courts will always be a list of booleans. Given that, I think we can simplify this section slightly:

```diff
- selected = sum(1 for v in courts if v)
+ selected = courts.count(True)
```

To make this even clearer for future developers, we could add a type hint:

```suggestion
            courts : list[bool] = [v for k, v in cd.items() if k.startswith("court_")]
            selected = courts.count(True)
```

This makes the intent explicit and makes the `count(True)` usage self-explanatory. In contrast, the current implementation requires the reader to mentally parse the generator expression to realize that it’s simply counting truthy values.

cl/audio/tests.py
```diff
@@ -210,6 +210,32 @@ def test_do_search_podcasts_have_content(self) -> None:
...
+        }
+        response = self.client.get(
+            reverse("search_podcast", args=["search"]),
+            params,
+        )
+        self.assertEqual(200, response.status_code)
+        xml_tree = etree.fromstring(response.content)
+        namespaces = {
+            "itunes": "https://www.itunes.com/dtds/podcast-1.0.dtd",
+        }
```
ERosendo: The `namespaces` dict is defined but never used in this test. Should this be removed, or was it intended for an `xpath` query that wasn't added?

cl/audio/feeds.py
```diff
@@ -115,7 +115,46 @@ def items(self, obj):
 
 
 class SearchPodcast(JurisdictionPodcast):
-    title = "CourtListener.com Custom Oral Argument Podcast"
+    def _get_search_summary(self, request):
+        """Build a human-readable search summary from GET params.
+
+        Caches on the request to avoid re-parsing the form in
+        title/description/subtitle.
```
ERosendo, line 121: This docstring mentions that caching avoids re-parsing the form in "title/description/subtitle".

However, there's no `subtitle()` method override in `SearchPodcast`. The parent class `JurisdictionPodcast` defines `subtitle = description` as a class attribute (line 21), which means it will still use the static fallback description rather than the dynamic one.                                                     

Should we add a `subtitle()` method to match the dynamic behavior, or update the docstring to remove the reference to subtitle?

cl/audio/tests.py
mlissner: Fixed by adding another test for the subtitle field that uses the namespace var.

cl/audio/feeds.py
mlissner: I don't think I ever knew the `.count` method on lists. Handy, thanks. Adapted accordingly.

cl/audio/feeds.py
mlissner: Good catch. While I was working, I considered having the subtitle node, but removed it at the end since it seemed redundant, but let's add it back so it doesn't get the value from the parent class.

I also noticed that `JurisdictionPodcast.summary` is dead code because it gets the value from `cl.lib.postcast.iTunesPodcastsFeedGenerator`, so I removed it too.

mlissner:
Add comments should be addressed. Back to you, @ERosendo. Thanks for the fast review!

PR #6966, opened 2026-02-16
feat: add automated frontend checks for PR reviews

elisa-a-v:
## Summary

Adds a GitHub Actions workflow that automatically checks HTML and CSS files in PRs for common frontend issues. Posts inline annotations and a summary comment when findings exist.

**Checks include:**
- Accessibility: tabnabbing prevention, tabindex validation, unstyled links (v2 templates only)
- Redesign conventions: Alpine.js usage (no jQuery, no `@` shortcuts, no inline `x-data`), Cotton components (no `{% include %}`), no Font Awesome in new templates
- Template structure: base template inheritance, legacy/v2 sync notices, new-stack leakage detection
- CSS: raw CSS detection in `input.css`
- Component library: warns when new Cotton components are added without updating `v2_components.html`

**PR comment behavior:**
- Posts a markdown summary comment when warnings or errors are found
- On subsequent pushes, previous summary comments are minimized as "outdated" (preserving history) and a fresh one is posted if findings remain
- Silent when clean — no comment posted, no noise
- Also writes a job summary visible in the Actions tab

**Known limitation:** The workflow only triggers on PRs that touch `.html` or `input.css` files. If frontend files are removed from a PR after a previous run posted a summary comment, that comment won't be automatically minimized (since the workflow doesn't run at all). This needs to be handled manually for now.

**Other details:**
- Uses `--name-status` to detect file additions vs modifications (e.g. new component check only fires on added/copied files)
- Sync notice check reads legacy counterparts from disk, not just from the PR's changed files
- Concurrency group cancels in-progress runs on same PR
- Check names in output match function names for grepability

## Deployment

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/.github/scripts/frontend_checks.py
+++ b/.github/scripts/frontend_checks.py
@@ -145,7 +145,8 @@ def check_tabnabbing(lines: list[str]) -> list[tuple[int, str]]:
                     (
                         i,
                         'target="_blank" without rel containing "noopener" or '
-                        '"noreferrer" — tabnabbing risk',
+                        '"noreferrer" — tabnabbing risk '
+                        "(https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/noopener)",
                     )
                 )
     return results
@@ -162,7 +163,8 @@ def check_tabindex(lines: list[str]) -> list[tuple[int, str]]:
                 (
                     i,
                     f'tabindex="{m.group(1)}" — never use tabindex > 0; '
-                    f"use 0 (focusable) or -1 (programmatic only)",
+                    f"use 0 (focusable) or -1 (programmatic only) "
+                    f"(https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex)",
                 )
             )
     return results
@@ -220,7 +222,7 @@ def check_alpine_shortcuts(
             else:
                 msg = (
                     f'"{shortcut}" — use "x-on:{m.group(1)}" instead '
-                    f"(CSP-safe Alpine prefix)"
+                    f"(https://github.com/freelawproject/courtlistener/wiki/New-Frontend-Architecture#alpine-conventions)"
                 )
             results.append((i, msg))
     return results
@@ -236,7 +238,9 @@ def check_font_awesome(lines: list[str]) -> list[tuple[int, str]]:
             results.append(
                 (
                     i,
-                    "Font Awesome class detected — use SVG icons in new templates",
+                    "Font Awesome class detected — use the {% svg %} template tag instead "
+                    "(see cl/custom_filters/templatetags/svg_tags.py, "
+                    "icons live in cl/assets/static-global/svg/)",
                 )
             )
     return results


.github/scripts/frontend_checks.py
```diff
@@ -0,0 +1,771 @@
...
+    return results
+
+
+def check_placeholder_text(lines: list[str]) -> list[tuple[int, str]]:
+    """Flag TODO, TBD, FIXME, Lorem ipsum outside of comments."""
+    results = []
+    pattern = re.compile(
+        r"\bTODO\b|\bTBD\b|\bFIXME\b|Lorem ipsum", re.IGNORECASE
+    )
+    inline_comment_re = re.compile(r"(<!--.*?-->|\{#.*?#\})")
```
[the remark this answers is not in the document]
elisa-a-v: this is intended, multiline comments are handled separately.

.github/scripts/frontend_checks.py
```diff
@@ -0,0 +1,898 @@
...
+    results = []
+    pattern = re.compile(r'tabindex\s*=\s*["\']([1-9]\d*)["\']')
+    for i, line in enumerate(lines, 1):
+        m = pattern.search(line)
+        if m:
+            results.append(
+                (
+                    i,
+                    f'tabindex="{m.group(1)}" — never use tabindex > 0; '
+                    f"use 0 (focusable) or -1 (programmatic only)",
```
mlissner: Are you saying that tabindex is basically dead? I still thought it was useful, but maybe browsers are just this good at it now?

mlissner: Maybe a reference for some of these rules would be useful?

.github/scripts/frontend_checks.py
```diff
@@ -0,0 +1,898 @@
...
+    """No Font Awesome classes in v2_ templates."""
+    results = []
+    # Matches class attributes containing the "fa" base class used by FA
+    pattern = re.compile(r'class\s*=\s*["\'][^"\']*\bfa\b')
+    for i, line in enumerate(lines, 1):
+        if pattern.search(line):
+            results.append(
+                (
+                    i,
+                    "Font Awesome class detected — use SVG icons in new templates",
```
mlissner: Maybe a tip here on how to find these?

.github/scripts/frontend_checks.py
elisa-a-v: Noo tabindex is not dead, but the only recommended values are -1 and 0, you can see this in [mdn](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex) and a lot of a11y guidelines mention this too, e.g. [w3](https://www.w3.org/WAI/WCAG21/Techniques/html/H4).

The strong recommendation is to write the document with the elements in a logical order, since tab order is primarily dictated by the order in which the elements appear in the document. `tabindex` is now mostly for programmatic focusing: using -1 removes it from the order but enables focusing the element using JS, and 0 is the default and recommended for focusable elements.

image

A link to the mdn docs is now included in the log for easy reference.

.github/scripts/frontend_checks.py
elisa-a-v: Great idea, thanks! added 👍🏽

elisa-a-v:
> it'd be nice to have more information about some of the checks

Great idea, thank you! Added some context to 4 checks:
- tabindex check now includes a link to the mdn docs discouraging positive values
- Font Awesome check now points to svg template tag and directory and instructs the use of `{% svg %}`
- tabnabbing check includes link to [`noopener` docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/rel/noopener)
- Alpine prefix check points to the Alpine conventions section [in our wiki](https://github.com/freelawproject/courtlistener/wiki/New-Frontend-Architecture#alpine-conventions)

PR #6978, opened 2026-02-18
feat(tames) add task for case subscription

Brennan-Chesley-FLP:
## Summary

This adds a celery task to subscribe to a batch of cases in TAMES. It requires a username and password to be set via env vars. 

## Deployment

**This PR should:**

    - [X] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [X] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

Before deployment, ensure the TAMES_USERNAME and TAMES_PASSWORD env vars are set for celery workers.

### DIFF
--- a/cl/scrapers/migrations/0006_accountsubscription.py
+++ b/cl/scrapers/migrations/0006_accountsubscription.py
@@ -14,7 +14,7 @@ class Migration(migrations.Migration):
             name='AccountSubscription',
             fields=[
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
-                ('scraper', models.IntegerField(choices=[(1, 'Texas')], help_text='The scraper/site the account is for.')),
+                ('scraper', models.IntegerField(choices=[(1, 'TAMES CaseMail')], help_text='The scraper/site the account is for.')),
                 ('email', models.EmailField(help_text='The email address for the account.', max_length=254)),
                 ('user_name', models.CharField(help_text='The username of the account.', max_length=30)),
                 ('first_subscription', models.DateField(help_text='The date of the first subscription.')),


cl/scrapers/tasks.py
```diff
@@ -893,3 +898,112 @@ def subscribe_to_scotus_updates(self: celery.Task, pk: int) -> None:
...
+                case_number,
+                e,
+            )
+            failed.append(case)
+
+    if failed:
+        raise self.retry(
+            args=(failed,),
+            exc=SubscriptionFailure(
+                f"{len(failed)}/{len(cases)} cases failed"
```
Brennan-Chesley-FLP, line 1099: We can use this batch method with retries, similar to how [find_citations_and_parentheticals_for_opinion_by_pks](https://github.com/freelawproject/courtlistener/blob/a5f96af77a2407bbdf8b008fe8d071da8e9d8141/cl/citations/tasks.py#L205-L218) operates. 

Or we can break it down by task with either @throttle_task or chaining. I had trouble with both of these methods locally, but I suspect this is my local setup and if anyone would prefer, we can use one of those strategies and I'll prioritize getting my local setup in a state where I can test it.

Brennan-Chesley-FLP:
Natural back pressure for this task with serial requests is probably enough, we're looking at probably 1-3 seconds per subscription depending on how many subscriptions for the account, and something on the order of 100 new cases expected in a batch, so this should run for something less than 5 minutes most of the time.

Brennan-Chesley-FLP:
This requires information from the Texas docket merger to be fully implemented (the list of newly created cases), but the code is mostly complete here, and can be reviewed for a good idea of how it will work once that can be wired up.

Brennan-Chesley-FLP:
Todo before this change can be merged:
- [ ] Square away TAMES_USER in environment
- [ ] Set up SES rules for texas.recap.email
- [ ] Merge [CaseMail endpoint](https://github.com/freelawproject/courtlistener/pull/6988) code

cl/scrapers/models.py
```diff
@@ -102,3 +104,47 @@ class PACERMobilePageData(AbstractDateTimeModel):
 
     def __str__(self) -> str:
         return f"<{self.pk}: Docket {self.docket_id} crawled at {self.date_last_mobile_crawl} with {self.count_last_mobile_crawl} results>"
+
+
+class Scraper(models.IntegerChoices):
```
albertisfu, line 109: Can we use a more specific name, please? Maybe

```suggestion
class ScraperSite(models.IntegerChoices):
```

cl/scrapers/models.py
```diff
@@ -102,3 +104,47 @@ class PACERMobilePageData(AbstractDateTimeModel):
...
+    TEXAS = 1, "Texas"
```
albertisfu: Not sure if we can find many different email notification systems in a single state, but if that's possible, could we use the domain to be more descriptive?

```suggestion
    TXCOURTS = 1, "Texas Case Mail"
```

cl/scrapers/tasks.py
```diff
@@ -892,3 +901,186 @@ def subscribe_to_scotus_updates(self: celery.Task, pk: int) -> None:
...
+
+    payload.update(
+        {
+            "ctl00$ctl00$BaseContentPlaceHolder1$ContentPlaceHolder1$txtUserName": tames_username,
+            "ctl00$ctl00$BaseContentPlaceHolder1$ContentPlaceHolder1$txtPassword": tames_password,
+            "ctl00$ctl00$BaseContentPlaceHolder1$ContentPlaceHolder1$cmdLogon": "Logon",
+        }
+    )
+
+    resp = session.post(CASEMAIL_LOGIN_URL, data=payload, timeout=30)
```
albertisfu: also here:
`{"User-Agent": "Free Law Project"}`

cl/scrapers/tasks.py
```diff
@@ -892,3 +901,186 @@ def subscribe_to_scotus_updates(self: celery.Task, pk: int) -> None:
...
+    :param cases: List of dicts, each with "court", "case", and "date_filed" keys.
+    :raises SubscriptionFailure: If login fails.
+    """
+    if not settings.TAMES_USER:  # type: ignore[misc]
+        raise ConfigurationException("TAMES_USER must be set.")
+    tames_user = json.loads(settings.TAMES_USER)
+    subscription_tracker, _created = AccountSubscription.objects.get_or_create(
+        scraper=Scraper.TEXAS,
+        user_name=tames_user["username"],
+        defaults={"email": tames_user["email"]},
```
albertisfu: There is an issue with the dates here.

`first_subscription` and `last_subscription` in AccountSubscription default to date.today.

When the instance is created, the default dates belong to the current day.

If you send cases to subscribe like:

```
{"court": "cossup", "case": "24-0001", "date_filed": "01/15/2025"}
{"court": "cossup", "case": "24-0002", "date_filed": "02/20/2025"}

```
first_subscription is correct, since you’ll always send a case filed on or before today.

However, last_subscription is set to today instead of the latest date_filed.

This is caused by logic in include_subscriptions that does:

```
all_dates = dates | {self.first_subscription, self.last_subscription}
self.last_subscription = max(all_dates)

```

Since `self.last_subscription` is initialized to today when the instance is created, it will always be the greatest date, causing a mismatch with the actual case subscription date range.


It'd be a good idea to cover this escenario in a test to prevent a regression.

cl/scrapers/tasks.py
```diff
@@ -892,3 +901,186 @@ def subscribe_to_scotus_updates(self: celery.Task, pk: int) -> None:
...
+
+TAMES_FAILURES_CACHE_KEY = "scraper:tames:subscription_failures"
+TAMES_FAILURES_TTL = 60 * 60 * 24 * 28  # 4 weeks
+
+
+@app.task(
+    bind=True,
+    max_retries=3,
+    retry_backoff=10,
+    autoretry_for=(SubscriptionFailure,),
```
albertisfu: We can remove these two (retry_backoff and autoretry_for ) if we’re already doing manual retries for this exception.

cl/scrapers/tasks.py
```diff
@@ -892,3 +901,186 @@ def subscribe_to_scotus_updates(self: celery.Task, pk: int) -> None:
...
+
+    try:
+        _login_tames(session, tames_user)
+    except (requests.RequestException, SubscriptionFailure):
+        if self.request.retries == self.max_retries:
+            redis.set(
+                TAMES_FAILURES_CACHE_KEY,
+                json.dumps(cases),
+                ex=TAMES_FAILURES_TTL,
+            )
```
albertisfu: If `max_retries` is reached here, the following `raise self.retry` will fail with a `MaxRetriesExceededError`.

So we should tweak the logic to prevent this error and properly register the cases that failed at this stage.

cl/scrapers/tasks.py
```diff
@@ -892,3 +901,186 @@ def subscribe_to_scotus_updates(self: celery.Task, pk: int) -> None:
...
+    if not failed:
+        redis.delete(TAMES_FAILURES_CACHE_KEY)
+        return
+
+    if self.request.retries == self.max_retries:
+        redis.set(
+            TAMES_FAILURES_CACHE_KEY,
+            json.dumps(failed),
+            ex=TAMES_FAILURES_TTL,
+        )
```
albertisfu: I imagine we’re going to subscribe to cases after the ongoing scraper discovers and ingests new cases, so this task will be triggered for many of those new cases.

This task is going to run in the default Celery queue, which has many workers.
Can we ensure this task runs only one at a time?

Are we going to schedule all the new cases in a single task, or ensure that we don’t schedule more than one task at the same time?

Otherwise, I see a couple of places where running multiple tasks of this type could be a problem, for example, when setting up the list of failed cases in Redis, which could be overridden by another task.

Or when updating the `AccountSubscription` dates.

If we identify a risk of race conditions, we should either use a Redis semaphore (e.g., using `create_redis_semaphore`) so only one task runs at a time, or use a different Redis data structure like a Set with `SADD` to add new values atomically. We could also use a database lock (`select_for_update`) when updating the `AccountSubscription` instance.

cl/scrapers/tasks.py
Brennan-Chesley-FLP: Good catch. I'll drop the defaults and construct new AccounSubscriptions explicitly.

cl/scrapers/tasks.py
Brennan-Chesley-FLP: What I think happens here: 
- We check to see if this is our last attempt to log in.
- If it is not, we just retry.
- If it is, we put our list of cases that we weren't able to subscribe to in REDIS, and we throw an error that gets picked up by Sentry, alerting us. 
- We take a manual step : subscribe to the cases ourselves, create a new user, rerun the subscription process the next day, etc. We can look in Sentry or Redis for the list of cases. 

I'm assuming here that the SubscriptionFailure (possibly chained) makes it to Sentry for alerting purposes if we're past our max_retries. Is that a bad assumption?

Is there some other way of registering the cases you'd like?

albertisfu:
@Brennan-Chesley-FLP could you please review the failed missing migration?

Also, we still need the non-customer file here for consistency with the infra team. So, in total, we’ll have 3 migration files. See Morgan’s PR for reference:

https://github.com/freelawproject/courtlistener/pull/6988/changes

And assign it back to myself when it's ready so I can review it.

cl/scrapers/tasks.py
Brennan-Chesley-FLP: My thought is that we'll run this once with any new cases found after each ongoing scrape finishes. While it should be naturally non-concurrent from this, I've added a `redis.lock()` and swapped the redis case subscription list from a kv - clobbered store to a set with add/remove ops. I think this should help smooth over the race conditions you identified.  The AccountSubscription race that you point out here should be fairly unlikely, and self healing on the next run, as the subscription window expands.

Brennan-Chesley-FLP:
I've updated the migration to bring it up to speed with the enum change. I think I have all 3 migrations files (one .py, one .sql, and one _customers.sql). That looks like what the wiki suggests and what the referenced PR has. I don't think I should be adding anything to the bulk data script here since we're not replicating this to customers. Happy to add the appropriate steps if we need it.

cl/scrapers/tasks.py
```diff
@@ -892,3 +901,202 @@ def subscribe_to_scotus_updates(self: celery.Task, pk: int) -> None:
...
+                resp = session.get(
+                    CASEMAIL_CASE_ADD_URL,
+                    params={
+                        "coa": court,
+                        "FullCaseNumber": case_number,
+                        "cID": court,
+                    },
+                    timeout=30,
+                )
+                resp.raise_for_status()
```
albertisfu, line 1035: If the user session expires in the middle of the case subscriptions, I don’t think this will catch that error. As far as I can saw, it will just return the login page with a 200 status code (I tested this by running the subscription request without a user session).

In those cases, it will mark the cases as successful without actually being subscribed.

Can we look for a success message ("Case added successfully") to confirm that the subscription was completed correctly? However not sure if this is the most reliable method to confirm the succesful subscription. I did a test using a non-existing Case Number and the "Case added successfully" was still rendered even thought no subscription were made.

Perhaps we can also check whether the Case Number is rendered on the page?

I noticed the most recent subscription appears last, and I’m not sure if this page uses pagination. If it does, we might not find the recently subscribed case on the same page. Maybe we’ll need to look for the case number on the last page, if it’s possible to navigate directly there. This would require an additional request, but perhaps it’s worth it?

albertisfu (commented):
Thanks @Brennan-Chesley-FLP, this looks good. I left a comment regarding expired sessions that might become a potential issue.

I also have a suggestion related to the retry architecture and potential concurrency issues.

I’m imagining a scenario where the ongoing scraper schedules a `subscribe_to_tames_cases` task, but the Celery workers are too busy to process it. Then, an hour later, the scraper schedules another `subscribe_to_tames_cases` task with different cases to subscribe to. Now the workers are free, and both tasks run at the same time, creating a potential race condition.

While we do have now a Redis lock (which I haven’t tested), I don’t think it’s reliable in this particular escenario since it has a timeout of 60 seconds, which seems too short considering the subscription request can be slow if we have many cases and are subscribing to multiple cases. Also, in your approach, it seems we’re not retrying the cases contained in the task that raises a `LockError`?

My suggestion is the following:

What if, instead of keeping track of failed cases, we keep track of the cases we need to subscribe to in Redis? That way, we wouldn’t need to pass the cases to `subscribe_to_tames_cases`; instead, we store them in Redis. Then, within the task, we retrieve those cases and try to subscribe to all of them, removing each one from the list once we confirm the subscription was successful.

To prevent race conditions, we should only schedule the subscribe_to_tames_cases task if there isn’t another task already running. For that, you can use `create_redis_semaphore`  (with a longer TTL) and `delete_redis_semaphore` (once retries are exhausted or after the task finishes successfully).

See `enqueue_docket_alert` for an example.

I think if we follow this pattern, we won’t have more than one task running at a time, and we probably won’t need to worry about manually running tasks if we see errors in Sentry, since the next task will process all pending cases to subscribe to.

We just need to be aware of cases that remain in the list after multiple attempts. For example, as I mentioned in the other comment, non-existing cases won’t raise an error when we try to subscribe to them. So maybe we can track in Redis the number of attempts per case (as the value of the key), and if that number exceeds a threshold (e.g., 6), we log an error in Sentry for manual review and remove it from the Redis set.

albertisfu:
@Brennan-Chesley-FLP I just realized that `subscribe_to_tames_cases` doesn’t need to be a Celery task. It can just be a regular method called within the same ongoing scraper daemon, right? That would remove all the concurrency problems, and we could get rid of the locks or semaphores. The other suggestion of keeping track of cases to subscribe instead of failed cases, is aligned with this approach.

Brennan-Chesley-FLP:
Rolled this into https://github.com/freelawproject/courtlistener/pull/6982.

PR #6982, opened 2026-02-18
feat(tames) ongoing case scraper

Brennan-Chesley-FLP:
## Summary

This PR adds a polling command for new cases on TAMES. It looks at the first page of search results (going backwards in time from the current date) and if it doesn't match the last cached results, it backfills a configurable number of days or cases. After it has done this, it (will) merges the scraped dockets and subscribes to new cases. 

## Deployment

**This PR should:**

    - [X] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [X] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`


cl/scrapers/management/commands/tames_poller.py
```diff
@@ -0,0 +1,272 @@
...
+                except requests.RequestException:
+                    logger.exception("Failed to fetch case URL %s", case_url)
+
+                if backfill_count is not None and case_count >= backfill_count:
+                    break
+
+            logger.info("Backfill complete. Processed %d cases.", case_count)
+
+        # Merge scraped cases
+        # Subscribe to new cases
```
Brennan-Chesley-FLP: These two comments will be replaced with the actual code pending the respective PRs.

cl/scrapers/management/commands/tames_poller.py
```diff
@@ -0,0 +1,272 @@
...
+            try:
+                self._poll_cycle(options, redis, courts, TAMESScraper)
+            except Exception:
+                logger.exception("Error during poll cycle")
+
+            logger.info(
+                "Sleeping %d minutes before next poll.",
+                options["polling_delay"],
+            )
+            shutdown_event.wait(timeout=polling_delay_seconds)
```
Brennan-Chesley-FLP: I'm assuming that our initial polling_delay is going to be an hour, so the work happens on a background thread so we can easily terminate it (and wait less than an hour for it to shutdown).

Brennan-Chesley-FLP:
Ran this overnight to verify basic polling behavior (every 15 minutes). I'm not sure how often it happens, but it looks like the website was updated at around midnight. Email notifications didn't get sent out until after 9:30. Might be an interesting thing to track as we add more sites/states.

cl/scrapers/management/commands/tames_poller.py
```diff
@@ -0,0 +1,272 @@
...
+        options: dict[str, Any],
+        redis,
+        courts: list[str] | None,
+        polling_delay_seconds: int,
+    ) -> None:
+        while not shutdown_event.is_set():
+            try:
+                self._poll_cycle(options, redis, courts, TAMESScraper)
+            except Exception:
+                logger.exception("Error during poll cycle")
```
MorganBennetDev: We should log the exception as well especially if we're using an `except` clause this broad.

cl/scrapers/management/commands/tames_poller.py
```diff
@@ -0,0 +1,272 @@
...
+                options["polling_delay"],
+            )
+            shutdown_event.wait(timeout=polling_delay_seconds)
+
+    def _poll_cycle(
+        self,
+        options: dict[str, Any],
+        redis,
+        courts: list[str] | None,
+        scraper_class: type,
```
MorganBennetDev: Are we planning to make this command more extensible or can we just use `TAMESScraper` and remove this parameter?

cl/scrapers/management/commands/tames_poller.py
```diff
@@ -0,0 +1,272 @@
...
+        case_rm_args = {
+            "requests_per_second": options["case_rate"],
+            "max_backoff_seconds": options["max_backoff"],
+        }
+
+        with (
+            RateLimitedRequestManager(
+                **search_rm_args
+            ) as search_request_manager,
+            RateLimitedRequestManager(**case_rm_args) as case_request_manager,
```
MorganBennetDev: I think it's more clear to pass the kwargs directly here since we're not doing anything with the dictionaries.

cl/scrapers/management/commands/tames_poller.py
Brennan-Chesley-FLP: `logger.exception` does this automatically (and should work with Sentry).

cl/scrapers/management/commands/tames_poller.py
Brennan-Chesley-FLP: I think I was probably fantasizing about this being reused, so you're right that this is a bit premature.

cl/scrapers/management/commands/tames_poller.py
Brennan-Chesley-FLP: True!

Brennan-Chesley-FLP:
Changed this to target https://github.com/freelawproject/courtlistener/pull/6978 which it depends upon for case subscription.

Brennan-Chesley-FLP:
I pinged Elisa about how to manage exceptions for the frontend testing. I'll update this PR ~or a separate PR~ with that once we've settled on a method.

Brennan-Chesley-FLP:
Depends on [PR 7187](https://github.com/freelawproject/courtlistener/pull/7187) for front-end lint changes.

cl/scrapers/management/commands/tames_poller.py
```diff
@@ -0,0 +1,314 @@
...
+                        case_response,
+                        SCRAPER_CLASS_NAME,
+                        dict(case),
+                        court_code or "unknown_court",
+                    )
+                except requests.RequestException:
+                    logger.exception("Failed to fetch case URL %s", case_url)
+                    continue
+
+                try:
```
MorganBennetDev: We can just save to S3 and then run the same `texas_corpus_download_task` to `texas_ingest_docket` task that the `import_texas_dockets` command uses. Then we don't have to copy the code above or have a docket page test asset (tests for this command would just need to assert that the test was called). One issue is keeping track of whether or not an import succeeded and adding it to the subscription queue if it did, but that shouldn't be a crazy problem to solve. Maybe a subscription celery task that we chain with the parse and merge tasks here?

Brennan-Chesley-FLP: We aborted the subscription celery task due to concurrency concerns, but the latest version of this can still dedup a bit of the code at the expense of a little coupling, reasoning locality, and a little time delay. We pass the subscription info through the existing tasks, and if that arg is not none, add it to the Redis maintained subscription list. The poller just looks at the Redis set for what to try subscribing to on each loop through.

cl/scrapers/management/commands/tames_poller.py
```diff
@@ -0,0 +1,433 @@
...
+            resp = session.get(
+                CASEMAIL_CASE_ADD_URL,
+                params={
+                    "coa": court,
+                    "FullCaseNumber": case_number,
+                    "cID": court,
+                },
+                timeout=30,
+            )
+            resp.raise_for_status()
```
albertisfu, line 182: Moving this suggestion from the closed [PR](https://github.com/freelawproject/courtlistener/pull/6978):

If the user session expires in the middle of the case subscriptions, I don’t think this will catch that error. As far as I can saw, it will just return the login page with a 200 status code (I tested this by running the subscription request without a user session).

In those cases, it will mark the cases as successful without actually being subscribed.

Can we look for a success message ("Case added successfully") to confirm that the subscription was completed correctly? However not sure if this is the most reliable method to confirm the succesful subscription.

cl/scrapers/models.py
```diff
@@ -102,3 +104,45 @@ class PACERMobilePageData(AbstractDateTimeModel):
 
     def __str__(self) -> str:
         return f"<{self.pk}: Docket {self.docket_id} crawled at {self.date_last_mobile_crawl} with {self.count_last_mobile_crawl} results>"
+
+
+class Scraper(models.IntegerChoices):
+    TAMES = 1, "TAMES CaseMail"
+
+
+class AccountSubscription(models.Model):
```
albertisfu: After testing  subscriptions integrated with the poller command, I found that it would be quite helpful to have `date_created` and `date_modified` fields in this model for debugging purposes.

```suggestion
class AccountSubscription(AbstractDateTimeModel):
```

cl/scrapers/management/commands/tames_poller.py
```diff
@@ -0,0 +1,433 @@
...
+                    break
+
+                case_url = case.get("case_url")
+                if not case_url:
+                    logger.warning("Case without case_url: %s", case)
+                    continue
+
+                court_code = case.get("court_code", "")
+                try:
+                    case_response = case_request_manager.get(case_url)
```
albertisfu, line 409: Considering that newer cases appear first in the search results, would it be possible to abort retrieving cases once we encounter a case already present in `cached_urls`? That way, we can avoid re-scraping or re-ingesting cases that have already been processed. This might be particularly helpful if there are many cases within the backfill date range.

cl/scrapers/management/commands/tames_poller.py
```diff
@@ -0,0 +1,433 @@
...
+                "on 403 errors (default: 300)."
+            ),
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        global shutdown_requested
+
+        if (
+            options["case_backfill_count"] is None
+            and options["case_backfill_days"] is None
```
albertisfu: `--case-backfill-days` has a default value of 3, so it will never be `None`, and the `case_backfill_count` setting will never be used. Is that expected?

Can you explain when `case_backfill_count` would be used instead of `case_backfill_days`?

Also, why is `case_backfill_days` configurable? Considering that we’ll be running this command every hour, why would it be necessary to have a value greater than 1 (the current day)?

albertisfu (changes requested):
Thanks @Brennan-Chesley-FLP this looks great.

I just have a couple of questions/suggestions.

BTW, I changed the target branch of this PR to main since it was pointing to the related closed PR.

cl/scrapers/management/commands/tames_poller.py
Brennan-Chesley-FLP: Removed case_backfill_count and changed the default to 1 day. I thought these might be nice for operational flexibility, but I'll defer to your experience here.

cl/scrapers/management/commands/tames_poller.py
Brennan-Chesley-FLP: Looking for this message, or the Cannot-add-twice message.

cl/scrapers/management/commands/tames_poller.py
Brennan-Chesley-FLP: Sure, we can capture some efficiency here.

cl/scrapers/management/commands/tames_poller.py
```diff
@@ -0,0 +1,465 @@
...
+
+            logger.info(
+                "Change detected (%d new URLs). Starting backfill.",
+                len(fresh_urls - cached_urls),
+            )
+
+            # -- Backfill ------------------------------------------------------
+            backfill_days = options["case_backfill_days"]
+
+            if backfill_days is not None:
```
albertisfu: `backfill_days` will never be `None` because `case_backfill_days` defaults to 1.

Do we still need this if/else branch condition?

cl/scrapers/management/commands/tames_poller.py
```diff
@@ -0,0 +1,465 @@
...
+                if case_count % 100 == 0:
+                    logger.info("Dispatched %d cases", case_count)
+
+            logger.info(
+                "Backfill complete. Dispatched %d cases.",
+                case_count,
+            )
+            if not found_stop:
+                logger.error(
+                    "Did not find known case when scraping. Likely gap."
```
albertisfu: This error will be raised on the first run because there are no `cached_urls` yet. Can we omit logging this error when there are no `cached_urls`?

cl/scrapers/management/commands/tames_poller.py
```diff
@@ -0,0 +1,465 @@
...
+            resp = session.get(
+                CASEMAIL_CASE_ADD_URL,
+                params={
+                    "coa": court,
+                    "FullCaseNumber": case_number,
+                    "cID": court,
+                },
+                timeout=30,
+            )
+            resp.raise_for_status()
```
albertisfu, line 182: Should we add a sleep after each subscription so we don’t exceed the rate limit when subscribing to many cases?

```suggestion
            resp.raise_for_status()
            time.sleep(1)
```

cl/scrapers/management/commands/tames_poller.py
Brennan-Chesley-FLP: I'd still like a bit of logging in case we need to debug odd behavior, is an `info` alright?

albertisfu: Yeah, the logging is fine here. I just wanted to see if we could avoid logging the error during the first daemon iteration. Let’s keep it as it is for now. We’ll just ignore that single error in Sentry.

albertisfu (approved):
Thanks for all the work here, @Brennan-Chesley-FLP. This is ready to go. I’ll notify the infra team to merge the PR and apply the migration when they’re ready.

Issue #1757, opened 2021-08-31
Re-evaluate feature set of the Big Cases Bot


mlissner:
It's sort of a special time right now for the bot. Since it's been broken so long, we can more easily deprecate features than normal, and we can change how it works for the better if we have ideas for doing so. In that vein, some thoughts after reviewing the codebase:

## New feature: Bitcoin! 

Sorry, kidding. Haha, no. 

## Considerations for removal/change/deprecation

1. What links should we share? We had a long conversation about this on Slack today, with some options being:
   - Links to encapsulated PDF pages on CL
       - Pros: Easy to find the docket; easy to create alerts for the docket
       - Cons: Doesn't really work on mobile; many clicks to get the actual PDF
   - Links to the PDF in storage.courtlistener.com
       - Pros: So simple; so performant
       - Cons: No extra features now or ever. No docket context; no alerts; nothing extra
   - Links to the docket
       - Pros: Full context
       - Cons: Slow; complicated; too much context?; doesn't show the docket very well; paginated
   - Some kind of simple minimal wrapper of the PDF page on CL that links to the docket and embeds the PDF
       - Pros: Could be performant; could be simple; could provide needed links and features
       - Cons: Doesn't exist; might not work on mobile (iframed PDFs on mobile no good?); 
   - Links to DocumentCloud
       - Pros: Journalists might like this? 
       - Cons: One more API to support; no docket context;

1. Allow curation from the accounts BCB follows instead of a hardcoded whitelist. Right now there's a setting that notes who gets to add dockets to BCB's follow list. Wouldn't it be cool to allow it from any account followed by the BCB?

1. Some sort of system so people can set up CL alerts based on a message to BCB? Like, I saw this tweet today and wanted to follow the docket: https://twitter.com/jonsteingart/status/1432542969656221696. Wouldn't it be nice to reply to it and get it to reply with a link saying how to follow the case? @johnhawkinson rightfully notes that it's kind of annoying to see those kinds of bot conversations on twitter, but I think they might be useful to those that want to follow cases, and fairly easy to ignore, since they'd be in a thread. Still: Maybe there's a neater way to do so.

## Bugs I've noticed

1. Text-only docket entries seem to be an issue. I think I see a fair number of people replying to these saying things like, "Where's the PDF!?"

mlissner:
I'm reaching out to prominent users and bot maintainers and had a great conversation just now with one of them. A couple thoughts:

[...]

PR #6988, opened 2026-02-18
TAMES CaseMail endpoint

### LINKED ISSUE #6685: Implement an API endpoint to ingest Texas CaseMail notifications
### Describe the issue.

Texas Case Mail allows us to get updates on cases as new docket entries are added. Parse those emails to surface the case numbers that we need to reindex. 

We will update our recap.email Lambda function to listen for Texas notifications, so this endpoint should be able to:

- Parse the email metadata
- Identify the related case
- Re-scrape the related case and ingest the new entries/documents related to the notification.

### Related Issue

_No response_

### Screenshots

_No response_

To Do

MorganBennetDev:
Resolves #6685

## Summary

Adds an endpoint to ingest TAMES CaseMail emails at `/api/v4/scrapers/email/state/tx/tames/alert`.

Awaiting completion of the email parser in [#1757](https://github.com/freelawproject/juriscraper/issues/1757) to proceed.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/corpus_importer/tasks.py
+++ b/cl/corpus_importer/tasks.py
@@ -5045,10 +5045,10 @@ def texas_ingest_docket_task(
         )
         task.request.chain = None
         return MergeResult.failed("Docket")
-    return merge_texas_docket(
+    result = merge_texas_docket(
         docket_data, download_attachments=download_attachments
     )
-    if subscription_data and result.create and result.pk is not None:
+    if subscription_data and result.create:
         redis = get_redis_interface("CACHE")
         redis.sadd(TAMES_PENDING_SUBSCRIPTIONS_KEY, subscription_data)
     return result


MorganBennetDev:
This depends on [JS#1812](https://github.com/freelawproject/juriscraper/pull/1812) and [CL#6862](https://github.com/freelawproject/courtlistener/pull/6862) to be merged, but is otherwise complete

cl/recap/tasks.py
```diff
@@ -127,6 +128,20 @@
...
+    # Try to read the file using utf-8.
+    # If it fails, fallback on iso-8859-1
+    try:
+        with bucket.open(message_id, "rb") as f:
+            return f.read().decode("utf-8")
+    except UnicodeDecodeError:
+        with bucket.open(message_id, "rb") as f:
+            return f.read().decode("iso-8859-1")
+    except FileNotFoundError as exc:
+        raise exc
```
Brennan-Chesley-FLP: I don't think it's necessarily a bad idea to add some sort of customization to the alerting here (otherwise this is a noop).

cl/scrapers/views.py
```diff
@@ -0,0 +1,160 @@
...
+        serializer.is_valid(raise_exception=True)
+        self.perform_create(serializer, state=state, site=site)
+        headers = self.get_success_headers(serializer.data)
+        return Response(
+            serializer.data, status=status.HTTP_201_CREATED, headers=headers
+        )
+
+    def perform_create(self, serializer, **kwargs):
+        state = kwargs["state"]
+        site = kwargs["site"]
```
Brennan-Chesley-FLP, line 138: We're calling this up above (line 129) without any kwargs.

cl/scrapers/views.py
```diff
@@ -0,0 +1,160 @@
+from attr import dataclass
```
Brennan-Chesley-FLP: Should this be dataclasses.dataclass , or are you thinking of expanding EmailProcessingTask with some validation or other attr feature in the future?

cl/recap/tasks.py
```diff
@@ -3631,3 +3638,79 @@ def do_recap_document_fetch(epq: EmailProcessingQueue, user: User) -> None:
...
+        requests.ConnectionError,
+        requests.RequestException,
+        requests.ReadTimeout,
+        RedisConnectionError,
+    ),
+    max_retries=10,
+    retry_backoff=2 * 60,
+    retry_backoff_max=60 * 60,
+)
+def process_texas_email(self: Task, epq: EmailProcessingQueue) -> None:
```
Brennan-Chesley-FLP: I think we're calling this with a pk `processing_task_info.task.delay(epq.id)` in `StateEmailEndpoint.perform_create` .

cl/recap/tasks.py
MorganBennetDev: I don't disagree, but right now I'm trying to preserve the exact behavior from `open_and_validate_email_notification` to try and avoid breaking things in unexpected places.

cl/recap/models.py
```diff
@@ -230,32 +231,65 @@ def print_file_contents(self) -> None:
         print(self.file_contents)
 
 
+class EmailSource:
```
Brennan-Chesley-FLP, line 239: We likely want to make https://github.com/freelawproject/courtlistener/blob/5f3ccb728995ec52735956ebfd27bb3872260023/cl/recap/admin.py#L73 aware of this new dynamic

cl/scrapers/views.py
MorganBennetDev: This was a sloppy auto import.

cl/recap/admin.py
```diff
@@ -73,7 +75,11 @@ class PacerHtmlFilesAdmin(CursorPaginatorAdmin):
 def reprocess_failed_epq(modeladmin, request, queryset):
     recap_email_user = User.objects.get(username="recap-email")
     for epq in queryset:
-        do_recap_document_fetch(epq, recap_email_user)
+        if epq.source == EmailSource.STATE:
+            if is_texas_court(epq.court):
+                process_texas_email.apply_async(epq.pk)
```
albertisfu: `apply_async` requires `args` to be a tuple. If you don’t need to use a chain here, you can do this instead.

```suggestion
                process_texas_email.delay(epq.pk)
```

cl/recap/tasks.py
```diff
@@ -132,6 +141,20 @@
...
+    # Try to read the file using utf-8.
+    # If it fails, fallback on iso-8859-1
+    try:
+        with bucket.open(message_id, "rb") as f:
+            return f.read().decode("utf-8")
+    except UnicodeDecodeError:
+        with bucket.open(message_id, "rb") as f:
+            return f.read().decode("iso-8859-1")
+    except FileNotFoundError as exc:
+        raise exc
```
albertisfu: We don’t need to catch `FileNotFoundError` to raise it again. We can let it propagate naturally.

cl/recap/tasks.py
```diff
@@ -3644,6 +3659,117 @@ def do_recap_document_fetch(epq: EmailProcessingQueue, user: User) -> None:
...
+                "status_message",
+            )
+            self.request.chain = None
+            return None
+        case _:
+            docket_parser = TexasCourtOfAppealsScraper(
+                court_id=email_data.court_id
+            )
+
+    res = httpx.get(email_data.url)
```
albertisfu: We do need to set a timeout to prevent hanging. Maybe 30 secs?

cl/recap/tasks.py
```diff
@@ -3644,6 +3659,117 @@ def do_recap_document_fetch(epq: EmailProcessingQueue, user: User) -> None:
...
+        return None
+    try:
+        d = merge_texas_docket(docket_data, download_attachments=True)
+    except Exception as e:
+        async_to_sync(mark_pq_status)(
+            epq,
+            f"Texas docket update error: {e}",
+            PROCESSING_STATUS.FAILED,
+            "status_message",
+        )
```
albertisfu, line 3798:
```suggestion
        )
        self.request.chain = None
        return None
```

cl/scrapers/views.py
```diff
@@ -0,0 +1,163 @@
...
+        )
+
+    def validate(self, attrs):
+        court_id = attrs["court"].pk
+        mail = attrs["mail"]
+        receipt = attrs["receipt"]
+
+        all_court_ids = Court.state_courts.all()
+        if not all_court_ids.filter(pk=court_id).exists():
+            raise ValidationError(f"{court_id} is not a state court ID.")
```
albertisfu: Considering court field is already defined as:
`queryset=Court.state_courts.all()`

Doesn't DRF rejects on-state courts before validate() is called? so this might validation be redundant?

cl/scrapers/views.py
```diff
@@ -0,0 +1,163 @@
...
+    ordering = "-id"
+    cursor_ordering_fields = ["id", "date_created", "date_modified"]
+
+    def get_message_id_from_request_data(self):
+        return self.request.data.get("mail", {}).get("message_id")
+
+    def get_destination_emails_from_request_data(self):
+        return self.request.data.get("receipt", {}).get("recipients")
+
+    def retrieve(self, request, *args, **kwargs):
```
albertisfu: Do we need this retrieve method? I understand it’s called on `GET` requests, but we should only accept `POST` requests that are routed to the create method.

cl/recap/models.py
```diff
@@ -230,32 +231,65 @@ def print_file_contents(self) -> None:
...
+    Represents email notifications from any state court system.
+    """
+    NAMES = (
+        (PACER, "PACER email notification."),
+        (SCOTUS, "SCOTUS email notification."),
+        (STATE, "State email notification."),
+    )
+
+
+@document_model
```
albertisfu: Since this is an existing model, can you please avoid using `@document_model` on it and instead keep the regular `help_text`?

Also it seems we're missing the customer version of the SQL?

cl/lib/storage.py
```diff
@@ -96,6 +96,10 @@ class RecapEmailSESStorage(S3Storage):
     bucket_name = "recap.email"
 
 
+class TexasEmailSESStorage(S3Storage):
+    bucket_name = "texas.email"
```
albertisfu: The bucket name is: `texas-email`

cl/recap/tasks.py
```diff
@@ -132,6 +141,20 @@
 cnt = CaseNameTweaker()
 
 
+def retrieve_email_from_queue(epq: EmailProcessingQueue, bucket: S3Storage):
```
albertisfu: Can we directly pass the message_id instead of the epq?

```suggestion
def retrieve_email_from_queue(message_id:str, bucket: S3Storage):
```

cl/recap/tasks.py
MorganBennetDev: HTTPX has 5-second timeouts on by default, which is why I used it over requests. Can update the value to 30 seconds if you think that's better.

cl/recap/tasks.py
```diff
@@ -3644,6 +3668,127 @@ def do_recap_document_fetch(epq: EmailProcessingQueue, user: User) -> None:
...
+            docket_parser = TexasCourtOfAppealsScraper(
+                court_id=email_data["court_id"]
+            )
+
+    res = httpx.get(
+        email_data["url"],
+        headers={
+            "User-Agent": "Free Law Project",
+        },
+        timeout=30.0,
```
albertisfu, line 3773: It seems that TAMES is performing a redirection here, as I received:

```
exception:                                                                                                                                                               
       HTTPError("Redirect response '302 Found' for url 'https://search.txcourts.gov/Case.aspx?cn=01-25-00982-CV'\nRedirect location:                                     
       'https://search.txcourts.gov/Case.aspx?cn=01-25-00982-CV&coa=coa01'\nFor more information check: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/302")
```

It worked after applying this tweak:

```suggestion
        timeout=30.0,
        follow_redirects=True,
```

cl/scrapers/views.py
```diff
@@ -0,0 +1,147 @@
...
+        required=True,
+    )
+    mail = serializers.JSONField(write_only=True)
+    receipt = serializers.JSONField(write_only=True)
+
+    class Meta:
+        model = EmailProcessingQueue
+        fields = "__all__"
+        read_only_fields = (
+            "error_message",
```
albertisfu, line 62: We need to add `recap_documents` here; otherwise, an error is raised stating that this field is required.

```suggestion
            "error_message",
            "recap_documents",
```

cl/recap/tasks.py
```diff
@@ -3644,6 +3668,127 @@ def do_recap_document_fetch(epq: EmailProcessingQueue, user: User) -> None:
...
+        )
+        self.request.chain = None
+        return None
+    else:
+        async_to_sync(associate_related_instances)(
+            epq,
+            d_id=d.pk,
+            de_id=None,
+            rd_id=None,
+        )
```
albertisfu, line 3810: Here we have this error:

```
  File "/opt/courtlistener/cl/recap/tasks.py", line 280, in associate_related_instances                                                                                   
    await pq.recap_documents.aadd(*rd_id)                                                                                                                                 
          ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^                                                                                                                                 
TypeError: django.db.models.fields.related_descriptors.create_forward_many_to_many_manager.<locals>.ManyRelatedManager.aadd() argument after * must be an iterable, not   
NoneType 
```

The problem is that `associate_related_instances` assigns only the `rd_id` for `EmailProcessingQueue` instances. In this case, we don’t know that value because we’re merging multiple documents from a different model at the same time. Additionally, `Docket` is not a field on the `EmailProcessingQueue` model.

I know that we have logic to identify what docket entries or documents were updated or added during the merge process. I think that would be useful here. Like if during the merge of a email notification we found 2 new entries were added or updated. We link those to the `EmailProcessingQueue` that'll be really helpful for debugging purposes.

The problem I see is that if we add a `ManyToMany` field pointing to `TexasDocketEntry`, it will help here, but for other states we'll need to add a similar field to the DB, which is impractical.

The alternative is using a [Generic relation](https://docs.djangoproject.com/en/6.0/ref/contrib/contenttypes/#generic-relations), but that only supports linking to a single instance.

What if we add our own Generic relation field that supports multiple instances? That would mean adding a `content_type` field like:
`content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)`
and a field to hold the related PKs,  `object_ids`, which could probably be a `JSONField`:
`object_ids = models.JSONField(default=list)`

This relation won't work with the ORM but for debugging purposes would be enough.

What do you think?

MorganBennetDev: > I know that we have logic to identify what docket entries or documents were updated or added during the merge process. I think that would be useful here. Like if during the merge of a email notification we found 2 new entries were added or updated. We link those to the EmailProcessingQueue that'll be really helpful for debugging purposes.

I think we would need to modify the top-level merger code slightly to return this data, but it is available and the change shouldn't be too difficult.

> What if we add our own Generic relation field that supports multiple instances? That would mean adding a `content_type` field like:
`content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)`
and a field to hold the related PKs, `object_ids`, which could probably be a `JSONField`:
`object_ids = models.JSONField(default=list)`

Would `object_ids` be a field on `EmailProcessingQueue`? I think that would work, but I'm not sure we need `ContentType` in that case since we could just store the model in the JSON field.

albertisfu: > Would object_ids be a field on EmailProcessingQueue? I think that would work, but I'm not sure we need ContentType in that case since we could just store the model in the JSON field.

Yeah, `object_ids` would be a field in `EmailProcessingQueue`.

I was thinking of storing the model since, depending on the State, the related IDs would belong to different document models.

I was also thinking of storing only IDs in the JSON field so we can do something like:

```
class EmailProcessingQueue(models.Model):
    ...
    related_model = models.ForeignKey(ContentType, on_delete=models.CASCADE)
    object_ids = models.JSONField(default=list)

    def get_related_objects(self):
        """For debugging, resolves the actual objects."""
        model_class = self.related_model.model_class()
        return model_class.objects.filter(pk__in=self.object_ids)
```

albertisfu:
I saw that the frontend checks are failing due to the HTML example added.

Instead of using the HTML, could you use a `TexasCommonDataDictFactory` and mock the HTML parsing so that it returns the `TexasCommonDataDictFactory` instance instead? That way, we won’t need to include the HTML directly.

cl/corpus_importer/tasks.py
```diff
@@ -4316,11 +4312,18 @@ def merge_texas_docket_entry(
...
+            create=created or any(r.create for r in document_results),
+            update=not created or any(r.update for r in document_results),
+            success=all(r.success for r in document_results),
+            pk=docket_entry.pk,
+        ),
+        [
+            result.pk
+            for result in document_results
+            if result.update or result.create
+        ],
```
albertisfu: Is it possible to retrieve the updated and created documents directly from MergeResult.create and MergeResult.update?

Aren’t they the same values?

That way, we can avoid passing an extra list of values with data we already have?

cl/recap/tasks.py
```diff
@@ -243,11 +276,32 @@ async def associate_related_instances(
...
+            rd_ids = rd_id
+        else:
+            rd_ids = [rd_id]
+        # Kept for compatibility
+        if model_name == "search.RECAPDocument":
+            await pq.recap_documents.aadd(*rd_ids)
+        # It's okay to overwrite these because we should only be calling this
+        # method once per email.
+        document_model = apps.get_model(model_name)
+        pq.content_type = await sync_to_async(
```
albertisfu: It seems this should be `related_model`, as that's the field name.

```suggestion
        pq.related_model = await sync_to_async(
```

Could you also extend the test to assert that the object IDs and the model are properly stored, please?

cl/corpus_importer/tasks.py
albertisfu: There's a mypy error that seems related to this return.

cl/recap/tasks.py
```diff
@@ -3644,6 +3692,135 @@ def do_recap_document_fetch(epq: EmailProcessingQueue, user: User) -> None:
...
+        return None
+    else:
+        docket_pks = result.creates.get("Docket", set())
+        changed_documents = list(
+            result.creates.get("TexasDocument", set())
+            | result.updates.get("TexasDocument", set())
+        )
+        async_to_sync(associate_related_instances)(
+            epq,
+            d_id=next(iter(docket_pks), None),
```
albertisfu: Is there a reason to pass the `d_id` if the `EmailProcessingQueue` doesn't have a docket field?

albertisfu (changes requested):

Just one small comment, and I think this will be ready.

cl/recap/tasks.py
MorganBennetDev: There is not. Removing.

albertisfu:
@MorganBennetDev could you solve merge conflicts in this PR when you have a moment please?

MorganBennetDev:
@albertisfu Resolved. Should be good to go again.

ERosendo:
Hi @MorganBennetDev, could you please take a look at the merge conflicts in this PR when you have a moment?

MorganBennetDev:
@ERosendo Merge conflicts resolved along with a failing test. Should be good to merge now.

PR #6993, opened 2026-02-19
feat(search): limits pagination for MLT queries

### LINKED ISSUE #6885: Fix Similar Cases feature that is currently blockd

ERosendo:
## Fixes

This PR fixes #6885 

## Summary

This PR introduces a pagination depth limit for `related:` search queries to prevent crawlers from triggering expensive deep-pagination MLT queries, which caused the recent outage.                  
                                                                                                         
 Changes:                                                                                                
 - Adds `MAX_RELATED_SEARCH_PAGINATION_DEPTH` setting (default: 5 pages), configurable via environment variable.
 - Updates `check_pagination_depth()` to accept a configurable max depth parameter                       
 - Detects `related:` queries in `fetch_and_paginate_results()` and applies the stricter limit           
 - Returns 403 Forbidden for related queries requesting page 6+       

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


albertisfu (approved):
Thanks @ERosendo this is working as expected. Merging...

mlissner:
With this fixed, do we need to tweak the WAF to allow these again?

ERosendo:
@mlissner Yes, that’s correct.

I can take care of it now.

Issue #7006, opened 2026-02-20
Don't percolate expensive dockets (with many parties)


albertisfu:
### Describe the issue.

During the ES outage today, Eduardo found that many percolator queries were running for several seconds.

One related case was:
https://www.courtlistener.com/docket/6245245/parties/johnson-johnson-talcum-powder-products-marketing-sales-practices-and/

This is a docket with hundreds of parties.

Every time we ingest this docket and the source contains parties, it gets percolated, generating an expensive percolator query.

One potential alternative we discussed is avoiding percolation for dockets with many parties. To do this, we can set a reasonable threshold for the number of parties, so we only percolate dockets with a small number of parties.

Alerts that cannot be triggered by the percolator would instead be handled nightly by the RECAP sweep index approach. For those documents with many parties it'll be cheaper matching alerts by running the query.

### What type of issue is this?

Research

mlissner:
Oh man, I'm not clicking that link. I've fallen for the Talcum Powder litigation parties problem before, and the database did not appreciate me.

This sounds like a good solution to me. We'll want to document it.

PR #7010, opened 2026-02-20
7006 Don’t percolate dockets that exceed the MAX_ATTORNEYS_TO_PERCOLATE attorney count

albertisfu:
## Fixes

Fixes: #7006

## Summary

This PR prevents expensive percolator queries on dockets with large numbers of attorneys.

When parties data is present, we only percolate if the total attorney count is within MAX_ATTORNEYS_TO_PERCOLATE (default: 10). Dockets exceeding that limit skip percolation entirely.

We determined this threshold by running SQL queries to check the average number of parties and attorneys per docket.

Here are the results of the queries run by Eduardo:

avg_parties_per_docket 
------------------------
`   5.3427374196531648`
(1 row)

avg_attorneys_per_docket 
--------------------------
    3.5309373107074822
(1 row)

Based on this, we decided to use a limit of 10 attorneys per docket and made the value configurable so it can be easily increased or decreased.

This should help reduce percolator load when ingesting dockets with a large number of parties. In these cases, we won’t be able to send real-time alerts. Instead, alerts that match these dockets will be sent at the end of the day using the sweep index approach via the `cl_send_recap_alerts` command, which will handle dockets with many parties more efficiently.

There is still room for improvement. For example, we could be more selective and omit parties data from the docket payload so we can still match alerts that don’t use parties as a filter. Currently, we percolate by referencing the `docket_id` from the RECAP index, which is why we can’t omit parties data.

Another scenario not fully covered is when we ingest a docket change from a source that doesn’t include parties, but the docket already has many parties merged from a previous source upload. In this case, the incoming change and the absence of parties data will trigger percolation of the full docket, including parties. This scenario is expected to be very rare, but it is still possible.

To address these cases, we would need to build the docket data dynamically and pass it directly to the percolator instead of referencing the existing document in the index. This would be a significant change and would require extra care. I'll open an issue to explore this approach in the future, as it is likely worth doing. For now, the changes in this PR should significantly reduce percolator load related to parties in most cases.

### Changes:
  - set_skip_percolation_if_parties_data() now returns a bool indicating whether post-merge percolation should happen
  - New` _exceeds_attorney_limit()` helper counts attorneys across parties with early exit for efficiency
  - Added `MAX_ATTORNEYS_TO_PERCOLATE` setting (default: 10)
  - Updated all four docket processing paths (`process_recap_docket, process_recap_appellate_docket, process_recap_acms_docket, merge_pacer_docket_into_cl_docket`) to use the new return
  value.
  - Fixed `process_recap_acms_docket` which was missing the `index_docket_parties_in_es` call entirely
  - Added test covering both over-limit (skipped) and within-limit (percolated) scenarios

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/settings/third_party/elasticsearch.py
+++ b/cl/settings/third_party/elasticsearch.py
@@ -254,7 +254,7 @@ PERCOLATOR_OPINIONS_SEARCH_ALERTS_ENABLED = env(
 
 # The maximum number of attorneys in parties data allowed for percolating a
 # docket after merging parties.
-MAX_ATTORNEYS_TO_PERCOLATE = env("MAX_ATTORNEYS_TO_PERCOLATE", default=100)
+MAX_ATTORNEYS_TO_PERCOLATE = env("MAX_ATTORNEYS_TO_PERCOLATE", default=50)
 
 ################################
 # ES bulk indexing batch size #


cl/settings/third_party/elasticsearch.py
```diff
@@ -252,6 +252,12 @@
     "PERCOLATOR_OPINIONS_SEARCH_ALERTS_ENABLED", default=False
 )
 
+# The maximum number of attorneys in parties data allowed for percolating a
+# docket after merging parties. If the number of attorneys exceeds this limit,
+# the docket is percolated on save instead to avoid expensive percolator
+# queries with large parties data.
+MAX_ATTORNEYS_TO_PERCOLATE = 10
```
ERosendo: Is this comment accurate? It says, “the docket is percolated on save instead,” but from what I can see, percolation is always skipped when the number of attorneys exceeds this value.

Also, it doesn’t seem like we can tweak this threshold without opening another PR. I was thinking we could make it configurable via an environment variable instead.

ERosendo:
@albertisfu I noticed that in the `process_docket_data` method we’re also doing:

```python
if data["parties"]:
    # Index or re-index parties only if the docket has parties.
    index_docket_parties_in_es.delay(d.pk)
```

Did you intentionally leave that out, or should it be updated as well?

ERosendo (changes requested):

mlissner:
I think I'd rather try to get a sense of performance. If 100 attorneys/parties are still fast, we'd want to do that instead. The Talcum case has thousands, right? It's insane! Any way we can check that?

ERosendo:
> Any way we can check that?

@mlissner If we make `MAX_ATTORNEYS_TO_PERCOLATE` configurable, we can control it via an environment variable and monitor ES performance. If we notice that 100 is still too high for ES to handle, we can lower it without opening another PR.

In every warning we received today, the percolator queries with thousands of attorneys were highlighted by Claude as the issue.

ERosendo:
Claude analyzed the query and counted the number of attorneys and parties. Here’s the summary:

```
Field         Array Items
party         2,609
party_id      2,609
attorney      925
attorney_id   925
```

mlissner:
Nice. Let's do that. We should also be sure to put this in the documentation.

cl/settings/third_party/elasticsearch.py
albertisfu: Updated.

albertisfu:
> I noticed that in the process_docket_data method we’re also doing:
> 
> if data["parties"]:
>     # Index or re-index parties only if the docket has parties.
>     index_docket_parties_in_es.delay(d.pk)
> Did you intentionally leave that out, or should it be updated as well?

Yeah, in this case, this is coming from a command to reprocess dockets, so we shouldn’t trigger alerts based on it. I’ve changed it to never percolate in this scenario.

On another note, I’ve decided to apply the further improvement I described above. After looking at some examples to evaluate parties percolation performance, I found that it’s more common to have case uploads with no parties. This means that, in those cases, if it’s an existing case, it will be percolated with all its parties from the document already indexed in ES.

This optimization is even more important for RECAPDocument percolation. I had forgotten that we also percolate RDs with all their parties in order to match cross-object alerts. So, if we receive a docket with hundreds of entries and the docket has many parties, this results in hundreds of heavy percolator queries in a short period of time. I suspect this was the source of the heavy ES percolation load we observed.

The optimization is basically the following:

- **Dockets:** The document to percolate is built from the DB data instead of referencing the indexed document. This allows us to omit parties from the percolated document if the case has more than `MAX_ATTORNEYS_TO_PERCOLATE`. This way, we can still percolate the case to match alerts that don’t involve parties.

- **RECAPDocuments:** When building the document to percolate, we also check whether the number of attorneys is less than `MAX_ATTORNEYS_TO_PERCOLATE`. If it is, parties are included in the document; otherwise, parties are omitted entirely, so alerts that don’t involve parties can still match.


- I also added documentation to explain these limitations regarding alerts and parties.

- Regarding percolator query performance, I ran the following tests:

[example_percolator_query.json](https://github.com/user-attachments/files/25462965/example_percolator_query.json)


- Document with 100 parties: ~1500 ms
- Document with 50 parties: ~800 ms
- Document with 10 parties: ~200 ms

Based on this, it seems that 100 parties is still a high limit. We can set the environment variable to a lower number before merging this PR.

cl/simple_pages/templates/v2_help/alert_help.html
```diff
@@ -243,6 +243,9 @@ <h3>Limitations of RECAP Search Alerts</h3>
     <p>However, because the field that causes the match <code>caseName</code> is at the case level and not tied directly to the individual filing, CourtListener cannot detect that a relevant filing now qualifies for the alert. As a result, the alert will not be triggered in real time.
       Instead, edge-case alerts like this are processed at the end of the day using a different matching strategy. In such cases, the associated webhooks are also triggered at the end of the day, rather than in real time.</p>
     <p>Apart from the edge case described above, real-time search alert emails are triggered within {{ rt_alerts_sending_rate }} minutes of ingesting a case or filing that matches the alert criteria.</p>
+    <h4 id="recap-search-alerts-large-parties">Limitations with Cases that Have Many Parties</h4>
+    <p>To maintain system performance, real-time alert matching skips party information for cases that have a large number of attorneys. If a docket exceeds {{ MAX_ATTORNEYS_TO_PERCOLATE }} attorneys, party, attorney, and firm fields are excluded from real-time alert matching.</p>
```
ERosendo: What do you think about this wording instead?

```suggestion
    <p>To keep the system running smoothly, real-time alerts skip party information for cases with a large number of attorneys. Specifically, if a docket has more than {{ MAX_ATTORNEYS_TO_PERCOLATE }} attorneys, the party, attorney, and firm fields won't be used when processing real-time alerts.</p>
```

cl/simple_pages/templates/help/alert_help.html
```diff
@@ -286,6 +292,9 @@ <h3 id="recap-search-alerts-limitations">Limitations of Real-Time RECAP Search A
   <p>However, because the field that causes the match <code>caseName</code> is at the case level and not tied directly to the individual filing, CourtListener cannot detect that a relevant filing now qualifies for the alert. As a result, the alert will not be triggered in real time.
     Instead, edge-case alerts like this are processed at the end of the day using a different matching strategy. In such cases, the associated webhooks are also triggered at the end of the day, rather than in real time.</p>
   <p>Apart from the edge case described above, real-time search alert emails are triggered within {{ rt_alerts_sending_rate }} minutes of ingesting a case or filing that matches the alert criteria.</p>
+  <h4 id="recap-search-alerts-large-parties">Limitations with Cases that Have Many Parties</h4>
+  <p>To maintain system performance, real-time alert matching skips party information for cases that have a large number of attorneys. If a docket exceeds {{ MAX_ATTORNEYS_TO_PERCOLATE }} attorneys, party, attorney, and firm fields are excluded from real-time alert matching.</p>
```
ERosendo: Same comment about the wording in this case.

cl/search/documents.py
```diff
@@ -1627,6 +1628,23 @@ def prepare(self, instance):
         return data
 
 
+class DocketDocumentPlain(DocketDocument):
+    def prepare_parties(self, instance):
```
ERosendo: I recall there’s an ongoing effort to improve type hint coverage across the codebase. Since this is a new function, it seems like a good opportunity to avoid introducing additional untyped code.


```suggestion
    def prepare_parties(self, instance: Docket) -> dict[str, set]:
```

cl/alerts/utils.py
```diff
@@ -771,17 +771,51 @@ def set_skip_percolation_if_bankruptcy_data(
...
     parties_data: list[dict[str, Any]], d: Docket
-) -> None:
+) -> bool:
     """Set skip percolation flag if parties data is present.
 
+    Since docket percolation references the existing ES document (not inline
+    content), percolating on save would just re-percolate the same document
+    without new parties data. So we always skip percolation on save when
+    parties data is present and decide whether to percolate after merging
+    based on the attorney count limit.
```
ERosendo: I think this docstring might still be referring to the old behavior:

> Since docket percolation references the existing ES document (not inline content)...

After this PR, Docket percolation actually builds the document from the DB using DocketDocumentPlain, rather than pulling it from the ES index. We might want to update the docstring to reflect that so it doesn’t get confusing later on

cl/simple_pages/templates/v2_help/alert_help.html
albertisfu: Thanks, change done!

cl/simple_pages/templates/help/alert_help.html
albertisfu: done

cl/search/documents.py
albertisfu: Sure. Added types here and also added a docstring.

cl/alerts/utils.py
albertisfu: Yeah, this comment was outdated. I’ve rewritten it completely.

albertisfu:
Thanks @ERosendo I've applied the requested changes. 

Also, what do you think about the initial value we should set for `MAX_ATTORNEYS_TO_PERCOLATE`, based on the performance metrics I shared? Maybe start with 50 parties, considering that we also percolate on RD and DocketEntry ingestions?

ERosendo:
> Maybe start with 50 parties, considering that we also percolate on RD and DocketEntry ingestions?

Sure, let's use 50

PR #7025, opened 2026-02-24
fix(navbar): Remove oversized font on Donate link

mlissner:
## Fixes
Fixes: #7021

## Summary
Sets `font-size: 5%` on the `.donate` CSS class in `override.css` (previously `110%`) so the navbar Donate link renders much smaller than other nav items. The red color and bold weight are preserved.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/assets/static-global/css/override.css
+++ b/cl/assets/static-global/css/override.css
@@ -116,6 +116,7 @@ header {
   color: #B53C2C;
   background-color: transparent;
   font-weight: bold;
+  font-size: 5%;
 }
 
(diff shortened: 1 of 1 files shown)


cl/assets/static-global/css/override.css
mlissner, line 1: I want it really small. Like 5%! Not 100%. That's too big.

PR #7027, opened 2026-02-24
Morgan/7001 casetransfer population

### LINKED ISSUE #7001: `CaseTransfer` Foreign Key Population
### Describe the issue.

While working on #6862, I ran into the issue that it is very hard to consistently populate foreign keys on both ends of a `CaseTransfer` object during the docket merging process. The specific issues encountered with Texas were:

1. Texas does not provide a destination docket number for workload transfers out of an appellate court. This means we do not have the information to create a `CaseTransfer` object for that field. We can assume that every transfer out of a court has a corresponding transfer into the destination court and ignore transfers out. However, this means that workload transfers will _never_  have their `origin_docket` field populated. Transfers out don't give us enough information and there is no guarantee on the order dockets are merged in so we can't rely on the origin docket for a transfer in to already exist.
2. Texas does not provide any indication on COA dockets of whether or not they have been appealed. The only source for this information is the corresponding SC or CCA docket. This creates the same issue we have with workload transfers: we are not guaranteed to ever be in a situation where we can populate the `origin_docket` FK on a `CaseTransfer` object.

We have three main approaches we can take in order to populate the `origin_docket` field for `CaseTransfer` objects.

1. Create a command that runs on a daily schedule to update existing `CaseTransfer` objects with null FK's. `CaseTransfer`s require that origin and destination court and docket number be set and at least one of `origin_docket` or `destination_docket` be non-null. We can try to find matching `Docket`s for `CaseTransfer`s with a missing FK, which will solve our problem. This has the downside of repeatedly running against `CaseTransfer`s we will _never_ have one of the FK's for (e.g. appeals from trial courts) and leaving us with incomplete data for up to 24 hours.
2. When the Texas merger creates a `Docket`, check if there is a matching `CaseTransfer` it could populate a FK for. This adds complexity to the Texas merger code, but ensures our data will be complete at the moment we import a docket.
3. Create a specialized method on `Docket` to perform the check and update process for `CaseTransfer`s just described when creating a new `Docket`. So mergers that create `CaseTransfer`s would use this method instead to automatically handle that logic.

### Related Issue

_No response_

### Screenshots

_No response_

To Do

MorganBennetDev:
Resolves #7001

#  Summary

Add `CaseTransfer.fill_null_dockets()` and a `fill_case_transfers` management command to backfill missing docket foreign keys on `CaseTransfer`s.

# Deployment

This PR should:

  - [ ] `skip-web-deploy`
  - [ ] `skip-celery-deploy`
  - [ ] `skip-cronjob-deploy`
  - [ ] `skip-daemon-deploy`


cl/corpus_importer/management/commands/fill_case_transfers.py
```diff
@@ -0,0 +1,20 @@
...
+            type=str,
+            help="The queue to run the update task in.",
+            default="celery",
+        )
+
+    def handle(self, *args, **options):
+        super().handle(*args, **options)
+
+        queue = options["queue"]
+        fill_case_transfer_missing_dockets.delay(queue)
```
albertisfu: Can’t we just run the whole process within the command?

`CaseTransfer.fill_null_dockets()`

That way, it wouldn’t require using a Celery worker apart from the pod where this command runs.

cl/corpus_importer/management/commands/import_texas_dockets.py
```diff
@@ -47,3 +48,7 @@ def download_task() -> app.Task:
     @staticmethod
     def merge_task() -> app.Task:
         return texas_ingest_docket_task
+
+    def handle(self, *args, **options):
+        super().handle(*args, **options)
+        fill_case_transfer_missing_dockets.delay()
```
albertisfu: Same here can we just run `CaseTransfer.fill_null_dockets()`?

cl/corpus_importer/tasks.py
```diff
@@ -4377,3 +4377,18 @@ def texas_corpus_download_task(
         meta = TexasDocketMeta.model_validate_json(f.read())
 
     return content, meta
+
+
+@app.task(
+    ignore_result=True,
+)
+@time_call(logger)
+def fill_case_transfer_missing_dockets():
```
albertisfu: If we follow previous suggestion. We won't need this task.

cl/search/models.py
```diff
@@ -4075,6 +4054,88 @@ class CaseTransfer(AbstractDateTimeModel):
         choices=transfer_type_choices.items(),
     )
 
+    # We currently only generate transfers for state courts, and we do not
+    # scrape trial courts so skip trying to populate fields we'll never be able
+    # to populate.
+    TRACKED_JURISDICTIONS = (Court.STATE_APPELLATE, Court.STATE_SUPREME)
+
+    @classmethod
+    def fill_null_dockets(cls):
```
albertisfu: A couple of suggestions for this method.

 If `missing_origin_docket` and `missing_destination_docket` match too many instances, they can use a lot of memory and cause an OOM error in the pod. 

So my suggestions are the following: use `.iterator()`, which defaults to 2000 instances loaded into memory at once, and create a helper method to prevent duplicate code so we can have something like this (I'm missing logs and details).

```
def _fill_null_docket_side(cls, side: str):
    updated_transfers = []
    total_updated = 0
    for transfer in missing_origin_docket.iterator():
        docket = async_to_sync(find_docket_object)(...)
        if docket:
            setattr(transfer, f"{side}_docket", docket)
            updated_transfers.append(transfer)
        
        if len(updated_transfers) >= 100:
            total_updated += cls.objects.bulk_update(
                updated_transfers, [f"{side}_docket"]
            )
            updated_transfers = []
    
    # Flush remaining
    if updated_transfers:
        total_updated += cls.objects.bulk_update(
            updated_transfers, [f"{side}_docket"]
        )

   return updated, total
```

And then we can call this method inside `fill_null_dockets` for both sides. 
This should be more memory-efficient.

cl/search/models.py
```diff
@@ -4075,6 +4054,88 @@ class CaseTransfer(AbstractDateTimeModel):
...
+        total_origin = missing_origin_docket.count()
+        for transfer in missing_origin_docket:
+            origin_docket = async_to_sync(find_docket_object)(
+                court_id=transfer.origin_court_id,
+                pacer_case_id=None,
+                docket_number=transfer.origin_docket_number,
+                federal_defendant_number=None,
+                federal_dn_judge_initials_assigned=None,
+                federal_dn_judge_initials_referred=None,
+                docket_source=Docket.SCRAPER,
```
albertisfu: Can we omit the source? Otherwise we'll miss existing cases combined with other sources like RECAP +  SCRAPER, etc.

albertisfu:
bwt can you please add an admin for `CaseTransfer` please?

cl/search/models.py
MorganBennetDev: The `docket_source` argument shouldn't factor into the search process. I added so that the source `find_docket_object` uses when creating a docket can be configured as it was previously just setting the source to `PACER`.

cl/corpus_importer/management/commands/fill_case_transfers.py
MorganBennetDev: We could do this, however I remember you said in our 1-1 this week that it should be run in a Celery worker since it could be handling a very large number of updates. I may be misremembering; please correct me if I am.

cl/search/models.py
MorganBennetDev: Will add this.

MorganBennetDev:
@albertisfu Implemented all the requested changes

cl/corpus_importer/management/commands/fill_case_transfers.py
albertisfu: Yeah, I thought we might need to run this every time the ongoing scraper runs. But it’s not necessary since it’s better to run this process once a day. So we can afford for it to be a long-running process executed via the command instead of using a celery task.

This is good now.

albertisfu:
Thanks @MorganBennetDev, this is ready to go. Just one thing, could you please resolve the merge conflict?

MorganBennetDev:
@albertisfu Conflicts resolved. Will merge into main after tests are done.

PR #7028, opened 2026-02-25
fix(search): Adds support for court picker syntax in ES filtering

### LINKED ISSUE #7008: Add support for the court_xx syntax for filtering courts in ES queries
### Describe the issue.

In https://github.com/freelawproject/courtlistener/issues/6958, we found many alerts using the following syntax to filter courts: `court_xx=on`. This is currently not a valid syntax in search queries, so these queries behave as match-all queries.

We agreed that the fastest solution right now is to add support for this syntax at the ES level, so these queries are no longer treated as match-all queries, which negatively impacts performance.

Later, we can consider blocking support for this syntax or adding official support. However, that would require further discussion, since it could impact the courts picker.

### Related Issue

_No response_

### Screenshots

_No response_

Research

ERosendo:
## Fixes

This PR fixes #7008 

## Summary

This PR adds support for the `court_xx=on` syntax when filtering courts in Elasticsearch queries.

Currently, many saved search alerts store court selections using individual checkbox parameters (e.g., `court_scotus=on&court_ca1=on`). However, the ES query builder only reads court filters from `cd["court"]`, a single space-separated string (e.g., `court=scotus ca1`). In the normal UI flow, a JS script converts the checked checkboxes into this hidden `court` field before form submission. When this JS conversion is bypassed (such as using the API) `cd["court"]` remains empty, resulting in no court filter and effectively a match-all query.

While testing, we also noticed that manually modifying the search URL to add courts using the picker syntax and refreshing the page does not update the query. The courts appear only in the picker component, resulting in an inconsistent interface that does not accurately reflect the query generating the displayed results.

This PR fixes the issue by adding a reverse conversion in `SearchForm.clean()` that rebuilds `cd["court"]` from all `court_xx=True` fields after the existing forward conversion. 

It covers three scenarios:

- Only `court_xx=on`: checkbox values are collected into cd["court"].
- Only `court`: forward conversion sets `court_xx=True`, then the reverse conversion collects them.
- Both formats: courts from both sources are merged.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,14 +13,11 @@ This repository is organized in the following way:
 
 ## Getting Involved
 
-If you want to get involved, email us with your contact info or take a look through the [issues list][issues]. There are innumerable things we need help with, but we especially are looking for help with:
+We love volunteers and work with them regularly. If you want to help further our mission of making a better and more fair legal ecosystem, please see our volunteer information page to get started:
 
- - legal research, to fix data errors or other problems (check out the [data-quality label][dq] for some starting points)
- - fixing bugs and building features (most things are written in Python)
- - machine learning or natural language problems.
- - test writing — we always need more and better tests
+https://free.law/volunteer/
 
-In general, we're looking for all kinds of help. Get in touch if you think you have skills we could use or if you have skills you want to learn by improving CourtListener.
+Also, please note that we assign every issue to *somebody*, so if you find an issue that looks interesting, just chime in and say you want to take it over. Usually we're happy to hand things off!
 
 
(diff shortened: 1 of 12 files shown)


albertisfu (approved):
Thanks @ERosendo this looks great. Set to auto-merge.

PR #7032, opened 2026-02-26
feat(recap): Update `find_docket_object` Texas DN formatting

### LINKED ISSUE #7031: Existing Texas `docket_number_core` Values Generated Incorrectly
While reviewing #6862, Alberto discovered that the `make_docket_number_core` method was not working for Texas docket numbers. It generates empty output (`PD-0836-24` to empty string) or drops information (`01-23-00656-CV` to `0100656`) in certain cases. While fixing this issue, we realized that this would affect our ability to match against existing Texas dockets and we found these issues exist in CL data.

To resolve this, I will be creating a PR properly handling Texas docket numbers in `make_docket_number_core` and `find_docket_object`. We will also need to run a command to repopulate existing `docket_number_core` fields with the correct values.

Tagging @grossir for feedback/suggestions since this touches your team's area.

MorganBennetDev:
Resolves #7031.

Adds `make_texas_docket_number_core` method to normalize Texas docket numbers without losing information.

Updates `find_docket_object` method to use `make_texas_docket_number_core` for lookups in Texas state courts.

    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [ ] `skip-cronjob-deploy`
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/lib/tests.py
+++ b/cl/lib/tests.py
@@ -24,8 +24,8 @@ from cl.lib.filesizes import convert_size_to_bytes
 from cl.lib.mime_types import lookup_mime_type
 from cl.lib.model_helpers import (
     clean_docket_number,
-    clean_texas_docket_number,
     clean_scotus_docket_number,
+    clean_texas_docket_number,
     is_docket_number,
     is_texas_court,
     linkify_orig_docket_number,


cl/recap/mergers.py
```diff
@@ -153,11 +153,22 @@ async def find_docket_object(
...
+    elif (
+        court_id == "tex"
+        or court_id == "texcrimapp"
+        or court_id.startswith("txctapp")
+        or court_id.startswith("texdistct")
+        or court_id.startswith("texcrimdistct")
+        or court_id.startswith("texctyct")
+    ):
+        docket_number_core = make_texas_docket_number_core(docket_number)
+        skip_check = True
```
albertisfu, line 163: Can you add a comment explaining why this is not required for Texas?

Basically, that it’s not possible to have dockets with the same `docket_number_core` in Texas, since the generated DNCs are unique.

cl/recap/mergers.py
```diff
@@ -153,11 +153,22 @@ async def find_docket_object(
...
+    if court_id == "scotus":
+        docket_number_core = make_scotus_docket_number_core(docket_number)
+        skip_check = False
+    elif (
+        court_id == "tex"
+        or court_id == "texcrimapp"
+        or court_id.startswith("txctapp")
+        or court_id.startswith("texdistct")
+        or court_id.startswith("texcrimdistct")
+        or court_id.startswith("texctyct")
```
albertisfu: Can you please move these conditions into a method and also use it in the `Docket.save()` method to avoid duplicating code?

cl/lib/model_helpers.py
```diff
@@ -111,6 +111,30 @@ def make_docket_number_core(docket_number: str | None) -> str:
...
+
+    There is overlap between valid Texas docket numbers and valid Federal\
+    docket numbers, but they need to be normalized differently so we need a\
+    separate method.
+
+    :param docket_number: The docket number to normalize.
+
+    :return: The normalized docket number.
+    """
+
```
albertisfu, line 206: I believe we need a cleaner method, such as `clean_docket_number`, perhaps one specific to Texas due to the federal overlap.

While we can ensure we’re properly extracting the cleaned docket number in our Texas scrapers, we can’t guarantee that other scrapers (like the Opinions scraper) are doing the same.

As a result, we could end up with a DN merged by the state scraper as:

`04-97-00972-CV`

While the same DN is merged from a different source as:

`Case Number: 04-97-00972-CV`

In this scenario, the first DN will produce a DNC of `049700972cv`, while the second will produce `casenumber049700972cv`, and the cases won’t match even though they refer to the same case.

To make this method more reliable, I think we should reuse the same regexes we’re using in our Texas scrapers: `DOCKET_NUMBER_REGEXES`. Perhaps we can import them from `juriscraper` so that when the JS regexes are updated, CL automatically picks up those changes as well.

Another suggestion: while reviewing those regexes, I found an edge case:

`responses/dockets/TAMESScraper/cossup/5753.html`

If you try to parse this docket with the Tex SC parser, you’ll get the error:

`Unrecognized docket number format: {docket_number}`

It appears we don’t currently have a valid regex for this DN format. We should add support for it.

To ensure we’re not missing any other DN formats for Texas, could you please download the latest inventory file and check the DNs (for example, extracting the DN from the file path like `25-0943`_1.html) through the regexes in `DOCKET_NUMBER_REGEXES`? We should verify that each number matches at least one regex and report any additional DNs that need to be added to the supported formats in `DOCKET_NUMBER_REGEXES` in JS.

MorganBennetDev:
@albertisfu Addressed your comments. Currently working on getting all the DN variants.

albertisfu (approved):

So once we’ve completed the DN variants check and added the required ones in JS, we should release a new JS version, update JS in CL, and then run the script to repopulate the `docket_number_core` for Texas cases.

Issue #7034, opened 2026-02-26
Regenerate SCOTUS docket_number_core


albertisfu:
### Describe the issue.

We’ll be running the SCOTUS importer to ingest all the scraped SCOTUS dockets we have.

While doing some validations before running it in production, I found that we might not be able to match existing SCOTUS dockets created by Opinions sources.

Here are the two potential cases I’ve detected so far:

**Cases with double docket numbers, e.g.:**

No. 01-8148 (01A587)
https://www.courtlistener.com/opinion/9271897/owsley-v-luebbers/

No. 01A576 (01-8099)
https://www.courtlistener.com/opinion/9271896/bottoson-v-florida/

No. 01-7987 (01A561)
https://www.courtlistener.com/opinion/9271894/anderson-v-davis/

These cases include both the regular SCOTUS docket number and the “A” docket number.

In all the examples I checked, the “A” docket number cannot be found in the SCOTUS docket search, while the regular 00-000 docket number is found and correctly matches the case.

The approach I’m planning is to prioritize the regular 00-000 docket number to generate the `docket_number_core` based on that value. This way, when we merge the case from SCOTUS, it can match properly.

**Docket number with A format:**

No. 01A530
https://www.courtlistener.com/opinion/9271910/urban-v-hurley/

When these appear alone, they can sometimes be found in the SCOTUS docket search.

Right now, the `docket_number_core` for them is empty, since we previously did not account for this format when generating it.

The idea now is to generate the `docket_number_core` based on this format and update the dockets accordingly. That way, when we ingest the docket coming from SCOTUS, it can be matched correctly.

@grossir @mlissner  do you have any thoughts on this? Do you see any issues with the proposed approaches to generate `docket_number_core`?

If this sounds good. I'll create an script to generate `docket_number_cores` for existing SCOTUS content before running the SCOTUS docket importer.

### What type of issue is this?

Research

mlissner:
Can you give a couple input/output examples mapping between the docket number fields? I want to make sure I understand correctly.

albertisfu:
Sure.

Here are some examples:

```
DN: No. 01-8148 (01A587)
Old DNC approach: 01008148
New DNC approach: 01008148
```

No changes, since the regular SCOTUS DN format (00-0000) was already supported.

```
DN: No. 01A576 (01-8099)
Old DNC approach: 01008099
New DNC approach: 01A00576
```

The old approach generated the DNC based on 01-8099 because the “A” DNs were not supported.
[...]

PR #7037, opened 2026-02-27
7034 Extract clean_scotus_docket_number and prioritize NN-NNNN format

albertisfu:
## Fixes
Fixes: #7034

## Summary
This PR extracts SCOTUS docket number cleaning into a dedicated `clean_scotus_docket_number` function and updates `make_scotus_docket_number_core` to use it.

Key changes:
- **New `clean_scotus_docket_number` function** that prioritizes the `NN-NNNN` format (e.g., `01-8148`) over the `NNA` format (e.g., `01A576`) when both are present in a docket number string.
- **Raises `ValueError`** when multiple docket numbers of the same type are found, instead of silently returning empty, so we can review those cases manually.
- **Removes SCOTUS A docket handling from `clean_docket_number`** so the general function no longer handles SCOTUS-specific logic.
- Adds tests for both the new function and the updated core function.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/lib/model_helpers.py
+++ b/cl/lib/model_helpers.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import re
 from collections.abc import Callable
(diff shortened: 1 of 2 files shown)


MorganBennetDev (approved):
Looks good!

PR #7050, opened 2026-03-03
feat(zoho): Integrate contact form with Zoho Desk tickets and CRM leads

### LINKED ISSUE #6750: Zoho Desk ticket automation via API
### Describe the issue.

For ticket management in Zoho Desk, we’d like to use an API to send the Request Type and Assignee to Zoho, along with other ticket info.

**Ticket information to include in the API:**
- Assignee
- Request type
- All other data users provide in the [contact form](https://www.courtlistener.com/contact/) in CourtListener (name, email, subject, etc.)

**A note on “Request Type”:**
- Request Type is selected by the user on the “Contact Us” form.
- The _exception_ is that “Case Removal Requests” should be recategorized as “Sealing Orders” when the description includes any form of these keywords:
- urgent, seal, redacted, pseudonym, anonymity, press.coverage, time.sensitive

**Auto-assign the ticket:**
- RECAP Extension: Eduardo
- Report Data Quality Problem: Bill
- Memberships or Donations: Jenifer
- Data or API Help: Mike
- Case Removals: [removals@free.law](mailto:removals@free.law)
- Sealing Orders: [legal@free.law](mailto:legal@free.law)
- General Support: Mike
- Partnership: [partnerships@free.law](mailto:partnerships@free.law) 
- Volunteering: Mike
- Legal Help: no one (it’s a dead end)

### Related Issue

_No response_

### Screenshots

_No response_

To Do

ERosendo:
## Fixes

This PR fixes #6750 

## Summary

This PR routes contact form submissions to Zoho Desk and Zoho CRM instead of the current Jira email pipeline, The new code is behind a waffle flag called `zoho-desk-tickets` to enable a safe rollout and production testing before fully removing the existing Jira integration.

Key changes: 
- **Zoho Desk:** 
   All non-partnership contact form submissions create a ticket in Zoho Desk via the REST API. Since Zoho Desk does not provide a Python SDK, we reuse the `OAuthToken` class from the Zoho CRM SDK for authentication. The same OAuth app and refresh token are valid for both services. For consistency, tickets use the same plain-text format currently used for Jira email submissions.

- **Zoho CRM leads:**
   Partnership inquiries create a CRM Lead with custom fields (e.g., Background, Funding, Team Size) mapped directly from the form inputs.

- **Sealing order detection:**
   Submissions from government domains (`@uscourts.gov`, `@usdoj.gov`) or removal requests containing sealing-related keywords are automatically recategorized as “Sealing Order” and left unassigned, per #6750 requirements.

- **Celery tasks:**
  Both Zoho operations run as Celery tasks with automatic retries on timeout,. This makes sure the contact form response is never blocked by Zoho API calls.

- **Backward compatible rollout:**
  The feature is disabled by default behind the zoho-desk-tickets waffle flag. Merging this PR will not affect the current contact form behavior. The existing Jira email pipeline remains active until the flag is enabled, allowing us to test the Zoho integration in production without impacting users.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

1. Generate a new Zoho OAuth refresh token with both CRM and Desk scopes. The existing CRM-only token will not work for Desk API calls.

```
ZohoCRM.org.ALL,ZohoCRM.users.ALL,aaaserver.profile.READ,ZohoCRM.modules.ALL,ZohoCRM.settings.fields.ALL,ZohoCRM.settings.related_lists.ALL,Desk.tickets.ALL
```

3. Add the following environment variables:
   - `ZOHO_DESK_DEPARTMENT_ID`: Zoho Desk department ID
   - `ZOHO_DESK_ORG_ID`: Zoho Desk organization ID
   - `ZOHO_DESK_AGENT_RECAP`: Agent ID for RECAP issues
   - `ZOHO_DESK_AGENT_DATA_QUALITY`: Agent ID for data quality issues
   - `ZOHO_DESK_AGENT_MEMBERSHIPS`: Agent ID for membership issues
   - `ZOHO_DESK_AGENT_API`: Agent ID for API issues
   - `ZOHO_DESK_AGENT_REMOVAL`: Agent ID for removal requests
   - `ZOHO_DESK_AGENT_SUPPORT`: Agent ID for general support
   - `ZOHO_DESK_AGENT_VOLUNTEERING`: Agent ID for volunteering inquiries
   - `ZOHO_DESK_AGENT_PARTNERSHIPS`: Agent ID for volunteering inquiries
4. Invite the relevant agents/staff to the Zoho Desk organization account so their agent IDs are valid for ticket assignment.
5. Create the `zoho-desk-tickets` waffle flag (inactive by default) and enable it once the environment is configured.

### DIFF
--- a/cl/corpus_importer/management/commands/download_scotus_pdfs.py
+++ b/cl/corpus_importer/management/commands/download_scotus_pdfs.py
@@ -1,0 +1,165 @@
+import time
+from itertools import batched
+
+from celery import chain
+from django.db.models import Q
+
+from cl.corpus_importer.tasks import download_scotus_document_pdf
+from cl.lib.celery_utils import CeleryThrottle
+from cl.lib.command_utils import VerboseCommand, logger
+from cl.scrapers.tasks import extract_pdf_document
+from cl.search.models import SCOTUSDocument
+
+
+def download_and_extract_scotus_pdfs(
+    download_queue: str,
+    extraction_queue: str,
+    delay: float,
+) -> None:
+    """Download and extract PDFs for SCOTUSDocuments missing a local file.
+
+    Queries SCOTUSDocument instances that have no filepath_local,
+    then schedules a download -> extraction chain for each.
+
+    :param download_queue: The celery queue for download tasks.
+    :param extraction_queue: The celery queue for extraction tasks.
+    :param delay: Seconds to sleep between scheduling tasks.
+    :return: None
+    """
+    docs = (
+        SCOTUSDocument.objects.filter(filepath_local="")
+        .values_list("pk", flat=True)
+        .order_by()
+    )
+    count = docs.count()
+    logger.info("Found %s SCOTUSDocuments needing download.", count)
+    throttle = CeleryThrottle(queue_name=extraction_queue)
+    processed_count = 0
+    for pk in docs.iterator():
+        throttle.maybe_wait()
+        chain(
+            download_scotus_document_pdf.si(pk).set(queue=download_queue),
+            extract_pdf_document.s(
+                check_if_needed=False,
+                model_name="search.SCOTUSDocument",
+            ).set(queue=extraction_queue),
+        ).apply_async()
+        processed_count += 1
+        if processed_count % 100 == 0:
+            logger.info(
+                "Scheduled %s/%s (%s)",
+                processed_count,
+                count,
+                f"{processed_count / count:.0%}",
+            )
+        time.sleep(delay)
+    logger.info(
+        "Scheduled %s/%s",
+        processed_count,
+        count,
+    )
+
+
+def extract_scotus_pdfs(
+    extraction_queue: str,
+    chunk_size: int,
+    delay: float,
+) -> None:
+    """Extract text from SCOTUSDocuments that have a file but incomplete OCR.
+
+    Queries SCOTUSDocument instances that already have a filepath_local but
+    whose OCR status is not complete or unnecessary.
+
+    :param extraction_queue: The celery queue for extraction tasks.
+    :param chunk_size: The number of items to extract per celery task batch.
+    :param delay: Seconds to sleep between scheduling tasks.
+    :return: None
+    """
+    docs = (
+        SCOTUSDocument.objects.exclude(filepath_local="")
+        .exclude(
+            Q(ocr_status=SCOTUSDocument.OCR_COMPLETE)
+            | Q(ocr_status=SCOTUSDocument.OCR_UNNECESSARY)
+        )
+        .values_list("pk", flat=True)
+        .order_by()
+    )
+    count = docs.count()
+    logger.info("Found %s SCOTUSDocuments needing extraction.", count)
+    throttle = CeleryThrottle(queue_name=extraction_queue)
+    processed_count = 0
+    for chunk in batched(docs.iterator(), chunk_size):
+        throttle.maybe_wait()
+        processed_count += len(chunk)
+        extract_pdf_document.si(
+            list(chunk),
+            check_if_needed=False,
+            model_name="search.SCOTUSDocument",
+        ).set(queue=extraction_queue).apply_async()
+        logger.info(
+            "Scheduled %s/%s (%s)",
+            processed_count,
+            count,
+            f"{processed_count / count:.0%}",
+        )
+        time.sleep(delay)
+
+
+class Command(VerboseCommand):
+    help = (
+        "Download and extract PDFs for SCOTUSDocument instances. "
+        "By default, downloads missing PDFs and extracts them. "
+        "Use --only-extraction to skip downloads and only extract "
+        "already-downloaded files."
+    )
+
+    def add_arguments(self, parser):
+        parser.add_argument(
+            "--download-queue",
+            type=str,
+            default="celery",
+            help="The celery queue for PDF download tasks.",
+        )
+        parser.add_argument(
+            "--extraction-queue",
+            type=str,
+            default="celery",
+            help="The celery queue for PDF extraction tasks.",
+        )
+        parser.add_argument(
+            "--only-extraction",
+            action="store_true",
+            default=False,
+            help="Only extract text from already-downloaded PDFs. "
+            "Skips the download step.",
+        )
+        parser.add_argument(
+            "--chunk-size",
+            type=int,
+            default=10,
+            help="The number of PDFs to extract in a single celery task "
+            "(only used with --only-extraction).",
+        )
+        parser.add_argument(
+            "--delay",
+            type=float,
+            default=1.0,
+            help="Seconds to sleep between scheduling tasks.",
+        )
+
+    def handle(self, *args, **options):
+        super().handle(*args, **options)
+
+        extraction_queue = options["extraction_queue"]
+        delay = options["delay"]
+
+        if options["only_extraction"]:
+            chunk_size = options["chunk_size"]
+            logger.info("Extracting SCOTUSDocument PDFs (extraction only).")
+            extract_scotus_pdfs(extraction_queue, chunk_size, delay)
+        else:
+            download_queue = options["download_queue"]
+            logger.info("Downloading and extracting SCOTUSDocument PDFs.")
+            download_and_extract_scotus_pdfs(
+                download_queue, extraction_queue, delay
+            )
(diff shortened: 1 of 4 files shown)


cl/simple_pages/views.py
```diff
@@ -434,19 +439,76 @@ async def contact(
...
+                            label
+                            for label in (
+                                form.label_for("partner_background") or ""
+                            ).split(", ")
+                            if label
+                        ],
+                        partner_background_other=cd.get(
+                            "partner_background_other", ""
+                        ),
+                    )
```
mlissner: Will Python let you do something silly like:

```python
create_zoho_crm_lead_from_contact.delay(
    partner_team_size=form.label_for("partner_team_size")
    or "",
    partner_funding_total=form.label_for(
        "partner_funding_total"
    )
    or "",
    partner_funding_stage=form.label_for(
        "partner_funding_stage"
    )
    or "",
    partner_background=[
        label
        for label in (
            form.label_for("partner_background") or ""
        ).split(", ")
        if label
    ],
    **cd,
)
```

Or is that hopeless?

mlissner (approved):
LGTM. One thing that I thought might be useful, but I'll let Alberto take a real look.

We will need to find a way to notify Eri when new requests come in this way. We don't currently get emails for new leads.

Thanks!

mlissner:
>  Partnership inquiries create a CRM Lead with custom fields (e.g., Background, Funding, Team Size) mapped directly from the form inputs.

I was thinking about this yesterday and I think it's probably the wrong way to do this. I think the way we have things now, entries into Zoho Desk also get added to the CRM, so if we add directly to the CRM, we'll wind up without the item in Zoho Desk (which is fine), and we won't get a notification about it in the CRM (which is not fine). If we just add it to Zoho Desk and assign to Eri, she'll get the notification, it'll wind up in the CRM, and we'll be able to remove code. 

What do you think, @ERosendo?

cl/lib/zoho.py
```diff
@@ -198,6 +199,78 @@ def _build_body_wrapper(
...
+    module_name = "Desk"
+
+    def _get_access_token(self) -> str:
+        """Return a valid access token from the Zoho CRM SDK.
+
+        ``OAuthToken.get_token()`` checks expiry, refreshes via the
+        Zoho accounts API if needed, persists the new token to the
+        ``FileStore``, and returns the access token string.
+        """
+        return Initializer.get_initializer().token.get_token()
```
albertisfu, line 235: Is this method race-condition safe?

I imagine two concurrent form requests could happen almost at the same time, and both of them refresh the token. The first one might then make the request using the first refreshed token, which may have already been invalidated by Zoho.

Is that possible?

cl/simple_pages/tasks.py
```diff
@@ -0,0 +1,141 @@
...
+from zohocrmsdk.src.com.zoho.crm.api.util import Choice
+
+from cl.celery_init import app
+from cl.lib.command_utils import logger
+from cl.lib.zoho import LeadsModule, ZohoDeskClient
+
+
+@app.task(
+    bind=True,
+    autoretry_for=(Timeout,),
```
albertisfu: What about also retrying on `ConnectionError`?

cl/simple_pages/forms.py
```diff
@@ -304,15 +311,22 @@ def email_subject(self) -> str:
         """Subject line for this submission."""
         return f"[CourtListener] Contact: {self.cleaned_data['phone_number']}"
 
-    def render_email_body(self, user_agent: str = "Unknown") -> str:
+    def render_email_body(
+        self,
+        user_agent: str = "Unknown",
+        target: Literal["jira", "zoho_desk"] = "jira",
+    ) -> str:
         """Plain-text email body built from cleaned_data. Includes all relevant fields for the issue type."""
```
albertisfu: Is this docstring still accurate? I mean that we’re using `<br>` as a separator for Zoho emails so this can also be HTML?

I tested it, and it seems Zoho is already sanitizing the HTML, so XSS injections do not seem to be possible. However, users can still embed images and other HTML tags. I'm not sure if that's an issue.

cl/lib/zoho.py
```diff
@@ -198,6 +199,78 @@ def _build_body_wrapper(
         return wrapper
 
 
+class ZohoDeskClient(ZohoModule):
```
albertisfu: This class inherits `ZohoModule` but doesn't use any of the CRM operations.

This only needs the `__init__`

Would it be better to extract the initialization logic into a shared mixin, rather than inheriting the full CRM module interface?

cl/lib/zoho.py
ERosendo: I don't think a race condition can happend here. According to the documentation, each organization can have a maximum of 20 refresh tokens. In our case, we only use one, which we store in Redis and reuse to obtain access tokens. Each refresh token can have up to 15 active (non-expired) access tokens.

Additionally, when a user creates a 16th access token, the system automatically deletes the oldest one. So in your example, if two concurrent requests generate a new access token, both tokens would still be valid. The system would invalidate the oldest existing token instead, not the newly created ones.

You can read more about this here:

https://desk.zoho.com/DeskAPIDocument#OauthTokens#RefreshingAccessTokens

cl/simple_pages/tasks.py
ERosendo: Sure!

cl/lib/zoho.py
ERosendo: That's a great idea. I just added this refactor in [678a4cb](https://github.com/freelawproject/courtlistener/pull/7050/commits/678a4cb7ec88d8a5bf1584e9dd8908a2ae21a598)

cl/simple_pages/forms.py
ERosendo: > Is this docstring still accurate?

Sorry, I forgot to update it in one of my previous commits. I’ve just refined it in [68d01b6](https://github.com/freelawproject/courtlistener/pull/7050/commits/68d01b6b41ced5b2c90bcfe27ccf0c8854d35cf5)

ERosendo:
> I was thinking about this yesterday and I think it's probably the wrong way to do this. I think the way we have things now, entries into Zoho Desk also get added to the CRM, so if we add directly to the CRM, we'll wind up without the item in Zoho Desk (which is fine), and we won't get a notification about it in the CRM (which is not fine). If we just add it to Zoho Desk and assign to Eri, she'll get the notification, it'll wind up in the CRM, and we'll be able to remove code.

@mlissner I like the idea of having Zoho Desk notify Eri when a new ticket is created and keeping both systems (Desk and the CRM) connected. However, I noticed that the integration creates Contacts by default when there’s no corresponding record in the CRM (it uses the email address to match records between systems)

As I mentioned in the message I sent to Eri and you over Slack, I’m not sure whether it’s okay to let the form from CourtListener flood a category that isn’t really meant for those kinds of records.

I’ve been testing the integration a bit more since I sent that message and noticed that the Zoho Desk integration can also match **existing** Leads. So if letting the integration create new Contacts isn’t the right approach, we could instead create Lead records first and then create the Desk ticket from there.

I’m also adding this comment here for context so @albertisfu is in the loop as well.

mlissner:
A heads up here. Once we get this in, I'm going to try to mirror it over on free.law using Netlify workers: 

https://github.com/freelawproject/free.law/issues/446#issuecomment-4043176529

This came up last week as a way of helping Eri. One problem she has is that the contact form currently provides the info@free.law email address, so her workflow involves emails to that address as well as things from Jira (and soon Zoho). If we can centralize everything to go into Zoho from the start, that should help her. 

I've been sort of lazy about fixing the contact form because netlify workers seemed like such a pain to figure out, but I realized that Claude probably can do this work without much effort by looking at this PR, CL's existing code, and the issue description above.

Once the FLP contact form is updated and working, I think we want to remove the contact form code from here, bots.law, the litigant portal, etc., and all contact can be directed through that free.law.

Yes, I realize this means I'm planning the (probable) deprecation of code even before it has been fully reviewed, but I think there's still a lot of value in this PR because it will suss out discussions and issues, bridge us over until we get free.law working, and it will make the free.law code easy (Claude is good at converting code from Python to JS; this Zoho code means the contact form doesn't do email anymore; etc). 

The end result should be better and "this" code will live on over in free.law with a bunch of benefits:
 - One contact form for all of FLP
 - Users learn about FLP before contacting us
 - Eri's workflow is smoother
 - Less code in CourtListener, where things are complicated, and more in free.law where things are simpler
 - Free.law gains its first back end code unlocking that possibility!

I hope this all makes sense!

albertisfu (approved):
Thanks for the changes @ERosendo this is ready to be merged.

PR #7078, opened 2026-03-05
fix(search): Tweak SCOTUS docket number core matching logic

albertisfu:
## Fixes
Fixes: #7034

## Summary
- When multiple SCOTUS docket numbers of the same type (NN-NNNN or NNA) are found, select the lexicographically smallest one instead of returning an empty string. This ensures deterministic docket matching regardless of input order.
- Rename `skip_check` to `skip_dn_core_confirmation` in `find_docket_object` for clarity, and skip the federal docket number confirmation for SCOTUS (multi-number inputs would fail it).
- Add integration test verifying `find_docket_object` consistently matches the same docket when multiple NN-NNNN numbers appear in different orders.

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/lib/model_helpers.py
+++ b/cl/lib/model_helpers.py
@@ -93,7 +93,6 @@ def clean_scotus_docket_number(docket_number: str | None) -> str:
     docket_number = docket_number.lower()
 
     scotus_m = re.findall(r"(?<![^ ,(])\d\d-\d+", docket_number)
-    scotus_a_m = re.findall(r"\b\d{2}a\d{1,5}\b", docket_number)
 
     if len(scotus_m) == 1:
         return scotus_m[0]
@@ -101,7 +100,9 @@ def clean_scotus_docket_number(docket_number: str | None) -> str:
         logger.warning(
             "Multiple NN-NNNN docket numbers found in: %s", docket_number
         )
-        return sorted(scotus_m)[0]
+        return min(scotus_m)
+
+    scotus_a_m = re.findall(r"\b\d{2}a\d{1,5}\b", docket_number)
 
     if len(scotus_a_m) == 1:
         return scotus_a_m[0]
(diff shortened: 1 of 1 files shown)


cl/lib/model_helpers.py
MorganBennetDev: I didn't think of this when reviewing the last PR, but we can move the `scotus_a_m` declaration below the `len(scotus_m)` checks for a slight optimization.

cl/lib/model_helpers.py
```diff
@@ -98,16 +98,18 @@ def clean_scotus_docket_number(docket_number: str | None) -> str:
     if len(scotus_m) == 1:
         return scotus_m[0]
     if len(scotus_m) > 1:
-        logger.error(
+        logger.warning(
             "Multiple NN-NNNN docket numbers found in: %s", docket_number
         )
-        return ""
+        return sorted(scotus_m)[0]
```
MorganBennetDev: We should use `min` here since it will be slightly faster.

cl/lib/model_helpers.py
```diff
@@ -98,16 +98,18 @@ def clean_scotus_docket_number(docket_number: str | None) -> str:
...
 
     if len(scotus_a_m) == 1:
         return scotus_a_m[0]
     if len(scotus_a_m) > 1:
-        logger.error("Multiple NNA docket numbers found in: %s", docket_number)
-        return ""
+        logger.warning(
+            "Multiple NNA docket numbers found in: %s", docket_number
+        )
+        return sorted(scotus_a_m)[0]
```
MorganBennetDev: Also here

MorganBennetDev (changes requested):
Looks good. Just two minor tweaks and this will be ready to go.

albertisfu:
Thanks! Changes applied.

MorganBennetDev:
Looks good @albertisfu. Auto-merge enabled.

PR #7081, opened 2026-03-06
Add `TrialCourtData` Model

MorganBennetDev:
## Summary

Adds a `TrialCourtData` model to `cl.search` as discussed in #6862.
- Allows linking of trial court information to cases which are more than one step above the trial court in the appellate chain.

Updates `document_model` decorator to clean up messy whitespace before setting model documentation parameters.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [ ] `skip-cronjob-deploy`
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/search/models.py
+++ b/cl/search/models.py
@@ -4115,8 +4115,7 @@ class SCOTUSDocument(AbstractDateTimeModel, AbstractPDF):
             "attachment_number",
         )
         ordering = ("document_number", "attachment_number")
-        
-    
+
     def __str__(self) -> str:
         return f"{self.pk}: Docket_{self.docket_entry.docket.docket_number} , document_number_{self.document_number} , attachment_number_{self.attachment_number}"
 


cl/search/models.py
```diff
@@ -4250,3 +4250,52 @@ class Meta:
...
+    :ivar court_name: The name of the court this case was filed in as it
+        appears in the source. May be blank if this information is not
+        available.
+    :ivar court: A foreign key to the Court object corresponding to the court
+        this case was heard in. May be blank if the court is not in the
+        database or is unavailable in the source.
+    """
+
+    docket = models.ForeignKey(
+        Docket, on_delete=models.SET_NULL, blank=True, null=True
```
albertisfu: Is there a reason to keep the `TrialCourtData` if we delete the Docket instance? Would it be better to just remove it so this can be CASCADE instead?

cl/search/models.py
```diff
@@ -4250,3 +4250,52 @@ class Meta:
...
+    docket_number = models.TextField()
+    docket_number_raw = models.TextField()
+    judge = models.TextField(blank=True)
+    reporter = models.TextField(blank=True)
+    date_filed = models.DateField(blank=True, null=True)
+    court_name = models.TextField(blank=True)
+    court = models.ForeignKey(
+        Court, on_delete=models.SET_NULL, blank=True, null=True
+    )
+
```
albertisfu: are we missing the `punishment` field and `county`?

cl/search/models.py
```diff
@@ -4250,3 +4250,52 @@ class Meta:
...
+    pass
```
albertisfu: Why is this `pass` required?

cl/search/models.py
```diff
@@ -4250,3 +4250,52 @@ class Meta:
...
+    :ivar court: A foreign key to the Court object corresponding to the court
+        this case was heard in. May be blank if the court is not in the
+        database or is unavailable in the source.
+    """
+
+    docket = models.ForeignKey(
+        Docket, on_delete=models.SET_NULL, blank=True, null=True
+    )
+    docket_number = models.TextField()
+    docket_number_raw = models.TextField()
```
albertisfu: This can be a chard field and add blank and default:
```suggestion
    docket_number_raw = models.CharField(
        help_text=(
            "The raw docket number value as found on the source,"
            "with no cleaning or transformations applied"
        ),
        blank=True,
        default="",
    )
```

cl/search/models.py
```diff
@@ -4250,3 +4250,52 @@ class Meta:
...
+        available.
+    :ivar court: A foreign key to the Court object corresponding to the court
+        this case was heard in. May be blank if the court is not in the
+        database or is unavailable in the source.
+    """
+
+    docket = models.ForeignKey(
+        Docket, on_delete=models.SET_NULL, blank=True, null=True
+    )
+    docket_number = models.TextField()
```
albertisfu: This should also have `blank=True` in case in some states we don't have this field.

albertisfu (changes requested):
Thanks @MorganBennetDev this looks good just a couple of comments and questions.

Also we should base this PR on main. So it can be merged first than the Texas merger PR.

cl/search/models.py
MorganBennetDev: I left that in by mistake. Removing.

cl/search/models.py
MorganBennetDev: My rationale here was that if we allow creating a `TrialCourtData` object without an associated `Docket` object, we also shouldn't delete it when an associated `Docket` is deleted. I'm open to changing this.

MorganBennetDev:
@albertisfu Implemented the changes you asked for and asked a clarifying question on one comment.

cl/search/models.py
albertisfu: Oh, I see, yeah, you're completely right.

cl/search/models.py
```diff
@@ -4170,3 +4250,61 @@ class Meta:
...
+        help_text=(
+            "The raw docket number value as found on the source,"
+            " with no cleaning or transformations applied"
+        ),
+        blank=True,
+        default="",
+    )
+    judge = models.TextField(blank=True)
+    reporter = models.TextField(blank=True)
+    date_filed = models.DateField(blank=True, null=True)
```
albertisfu, line 4190: I assume this field  `date_filed`  is being added as a provision in case it becomes useful in the future for other states, even if it's not present in Texas. Is that assumption correct?

cl/search/models.py
```diff
@@ -4170,3 +4250,61 @@ class Meta:
...
+    docket_number = models.CharField(blank=True, default="")
+    docket_number_raw = models.CharField(
+        help_text=(
+            "The raw docket number value as found on the source,"
+            " with no cleaning or transformations applied"
+        ),
+        blank=True,
+        default="",
+    )
+    judge = models.TextField(blank=True)
```
albertisfu: Would it be worth also adding a judge field that is a foreign key to `people_db.Person`, so we can look it up and see if we have the judge in our DB, or in the future, when we add more judges, have the possibility of linking them?

If so this field can change to `judge_str`

MorganBennetDev:
@albertisfu Had to force-push to remove the Texas changes from this while keeping the commit log clean so some things may be wonky.

cl/search/models.py
MorganBennetDev: That is correct, yes.

cl/search/models.py
MorganBennetDev: That would likely be helpful. Will add this.

chaco-fl:
RDS Customer Updated

Screenshot 2026-03-16 at 16 44 22

chaco-fl:
CL Replica Update

Screenshot 2026-03-16 at 16 47 01

PR #7090, opened 2026-03-09
Provide better memberships links and change navbar

mlissner:
Three things:

1. This updates our membership links throughout CourtListener to use https://free.law/memberships/ instead of the link on Neon. The free.law link is prettier and clearer. 

3. @nadahlberg noted that the donate link in the header should say "Pricing" and while we don't want to do that, we can use this opportunity to push memberships a bit more, so I switched it from "Donate" to "Join FLP". We'll see if that helps.

3. This removes a dead eslintrc file and tweaks the functional one so that it doesn't convert `class` to `classname`, which it was sneakily doing.

## Deployment

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

## Screenshot

image

### DIFF
--- a/cl/.eslintrc.js
+++ b/cl/.eslintrc.js
@@ -3,6 +3,7 @@ module.exports = {
     node: true,
   },
   root: true,
+  ignorePatterns: ['**/*.html'],
   parser: '@typescript-eslint/parser',
   plugins: ['@typescript-eslint'],
   extends: [


ERosendo:
@mlissner I reviewed the changes and the updated href values look good to me. Should we be worried about the frontend checks failing, or is it okay to merge the PR as it is?

mlissner:
Those are all unrelated. My ambition isn't to fix all of those for the moment! I'll merge, thanks!

Issue #6954, opened 2026-02-14
Run and monitor SCOTUS case importer


albertisfu:
### Describe the issue.

This task is to run and monitor the SCOTUS case importer command: https://github.com/freelawproject/courtlistener/pull/6892

### What type of issue is this?

Research

albertisfu:
The SCOTUS case importer is cruising now.

<img width="1092" height="120" alt="Image" src="https://github.com/user-attachments/assets/8e80eec5-81cf-4ff8-bd8b-242a7032fd01" />

I expect it to finish in around 72 hours. 

I'll be monitoring it.

mlissner:
YAY!

albertisfu:
I found that the celery-default queue was growing significantly.

The problem is that importing a single case can trigger the download of several PDFs and their subsequent extraction. So even if we throttle the case importer rate, we cannot control the actual document download and extraction rate. This was impacting the Celery workers, which were busy extracting PDFs, a task that can take a long time when the PDFs are large.

To prevent this issue, it seems better to avoid downloading PDFs and extracting them when importing SCOTUS cases in bulk.

I opened https://github.com/freelawproject/courtlistener/pull/7096 to separate the metadata ingestion from the PDF download and extraction. We'll be downloading and extracting PDFs in a secondary command and using different celery queues.

For now, I've scaled up the celery-default queue so it can execute the enqueued tasks faster. I'll scale it back down once the queue is cleared.

albertisfu:
We are at 64% of the 191442 SCOTUS cases we'll ingest.

<img width="888" height="101" alt="Image" src="https://github.com/user-attachments/assets/93f60bb7-4fd0-4d96-9bc9-ca4f987bc1d6" />

Here are some numbers so far:

```
In [1]: from cl.search.models import ScotusDocketMetadata, SCOTUSDocument, SCOTUSDocketEntry
   ...: 
   ...: metas = ScotusDocketMetadata.objects.all()
   ...: print("All SCOTUS: ", metas.count())
   ...: 
   ...: scotus_entries = SCOTUSDocketEntry.objects.all()
   ...: print("All SCOTUS entries: ", scotus_entries.count())
   ...: 
   ...: scotus_docs = SCOTUSDocument.objects.all()
   ...: print("All SCOTUS docs: ", scotus_docs.count())
   ...: 
All SCOTUS cases:  123166
All SCOTUS entries:  493241
All SCOTUS docs:  23566
```

[...]

PR #7096, opened 2026-03-11
feat(corpus_importer): Separate SCOTUS PDF downloads from docket ingestion

albertisfu:
## Summary

In https://github.com/freelawproject/courtlistener/issues/6954 I found that the celery-default queue was growing significantly.

The problem is that importing a single case can trigger the download of several PDFs and their subsequent extraction. So even if we throttle the case importer rate, we cannot control the actual document download and extraction rate. This was impacting the Celery workers, which were busy extracting PDFs,  a task that can take a long time when the PDFs are large.

To prevent this issue, it seems better to avoid downloading PDFs and extracting them when importing SCOTUS cases in bulk.

This way we can accelerate the `import_scotus_dockets` command so that the metadata for dockets, entries, and documents is ingested into the database faster.

In parallel, we can run a second command introduced in this PR (`download_scotus_pdfs`) to download the PDFs and extract them in a separate process.

This will also allow us to better control the PDF download rate. As we have seen some 403 errors during downloads, we will be able to slow down the PDF ingestion if necessary.

The command can both download and extract PDFs, or only extract PDFs. Since we have seen some extraction errors when doctor fails to return a response, we can retry the extraction in a second stage.

We can also have independent queues for downloading and extraction.

- Adds a `download_file` flag to the SCOTUS docket processing pipeline (`process_scotus_docket`, `merge_scotus_docket`, `merge_scotus_document`, etc.) so PDF downloads can be skipped during bulk docket imports
- Updates `import_scotus_dockets` to pass `download_file=False`, preventing Celery workers from being overwhelmed during bulk imports
- Adds a new `download_scotus_pdfs` management command that independently downloads and extracts PDFs for SCOTUSDocuments missing local files, with support for an `--only-extraction` mode
- Adds a test verifying that no download chain is triggered when `download_file=False`

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/corpus_importer/management/commands/download_scotus_pdfs.py
+++ b/cl/corpus_importer/management/commands/download_scotus_pdfs.py
@@ -2,6 +2,7 @@ import time
 from itertools import batched
 
 from celery import chain
+from django.db.models import Q
 
 from cl.corpus_importer.tasks import download_scotus_document_pdf
 from cl.lib.celery_utils import CeleryThrottle
@@ -17,7 +18,7 @@ def download_and_extract_scotus_pdfs(
 ) -> None:
     """Download and extract PDFs for SCOTUSDocuments missing a local file.
 
-    Queries SCOTUSDocument instances that have a URL but no filepath_local,
+    Queries SCOTUSDocument instances that have no filepath_local,
     then schedules a download -> extraction chain for each.
 
     :param download_queue: The celery queue for download tasks.
@@ -26,14 +27,13 @@ def download_and_extract_scotus_pdfs(
     :return: None
     """
     docs = (
-        SCOTUSDocument.objects.exclude(url="")
-        .filter(filepath_local="")
+        SCOTUSDocument.objects.filter(filepath_local="")
         .values_list("pk", flat=True)
         .order_by()
     )
     count = docs.count()
     logger.info("Found %s SCOTUSDocuments needing download.", count)
-    throttle = CeleryThrottle(queue_name=download_queue)
+    throttle = CeleryThrottle(queue_name=extraction_queue)
     processed_count = 0
     for pk in docs.iterator():
         throttle.maybe_wait()
@@ -54,10 +54,9 @@ def download_and_extract_scotus_pdfs(
             )
         time.sleep(delay)
     logger.info(
-        "Scheduled %s/%s (%s)",
+        "Scheduled %s/%s",
         processed_count,
         count,
-        f"{processed_count / count:.0%}",
     )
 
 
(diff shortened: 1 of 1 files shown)


albertisfu:
@MorganBennetDev  Once this is merged, I think we should use the same approach for the Texas merger. Maybe we can reuse `download_scotus_pdfs` by giving it a more generic name in the Texas PR and passing the document model as a parameter.

cl/corpus_importer/management/commands/download_scotus_pdfs.py
```diff
@@ -0,0 +1,163 @@
...
+    have not been extracted yet (empty plain_text).
+
+    :param extraction_queue: The celery queue for extraction tasks.
+    :param chunk_size: The number of items to extract per celery task batch.
+    :param delay: Seconds to sleep between scheduling tasks.
+    :return: None
+    """
+    docs = (
+        SCOTUSDocument.objects.exclude(filepath_local="")
+        .filter(plain_text="")
```
MorganBennetDev: Is there any reason we can't filter on `Q(ocr_status=SCOTUSDocument.OCR_FAILED) | Q(ocr_status=SCOTUSDocument.OCR_NEEDED)`?

cl/corpus_importer/management/commands/download_scotus_pdfs.py
```diff
@@ -0,0 +1,163 @@
...
+            check_if_needed=False,
+            model_name="search.SCOTUSDocument",
+        ).set(queue=extraction_queue).apply_async()
+        logger.info(
+            "Scheduled %s/%s (%s)",
+            processed_count,
+            count,
+            f"{processed_count / count:.0%}",
+        )
+        time.sleep(delay)
```
MorganBennetDev, line 105: Do we need `time.sleep` here in addition to the `throttle.maybe_wait` above?

cl/corpus_importer/management/commands/download_scotus_pdfs.py
albertisfu: Good question. The issue here is that, as I’ve seen in the logs, extraction has been failing without Doctor sending a response (Timeout or ReadError). In those cases, no `OCR_FAILED` status is being set.

Also, `OCR_NEEDED` is not the default document OCR status, the default is None.

So I think we could do the following instead:

```
 SCOTUSDocument.objects.exclude(filepath_local="")
        .exclude(
            Q(ocr_status=SCOTUSDocument.OCR_COMPLETE)
            | Q(ocr_status=SCOTUSDocument.OCR_UNNECESSARY)
        )
        .values_list("pk", flat=True)
        .order_by()

```

To avoid checking the plain text.

What do you think?

MorganBennetDev: That seems reasonable to me.

cl/corpus_importer/management/commands/download_scotus_pdfs.py
albertisfu: The reason is that time.sleep will be helpful to throttle `download_scotus_document_pdf`, and `throttle.maybe_wait()` can be helpful to throttle `extract_pdf_document`. I also changed `CeleryThrottle(queue_name=extraction_queue)` to use the extraction queue.

This is because I’m planning to use a single Celery worker for `download_queue`, but multiple workers for `extraction_queue`.

MorganBennetDev (approved):

Issue #7045, opened 2026-03-02
Update the CL contact us form to include a documentation checkbox


JessicaFrank:
### Describe the issue.

The CourtListener contact us form needs to include an affirmative checkbox where users agree that they have reviewed our documentation before reaching out with technical questions. 

We work hard to make our documentation complete, but oftentimes, it is not reviewed before users reach out. Many people could self-help if they did research. We're making improvements to our documentation (wiki forthcoming) that will make the process easier. We also want to be more explicit/direct about requiring that research before people reach out. 

In the contact form workflow, there should be an affirmative box that needs to be checked before the form can be submitted for "Data or API help", "General support", and the "RECAP extension bug" probably. 

It should point to the current documentation, but change to the searchable wiki when available. 

### What type of issue is this?

To Do

mlissner:
I'll take this one on. Should be easy to vibe code and I'll make a decision about whether to wait for the wiki.

PR #7106, opened 2026-03-13
Force people to agree that they read the docs before submitting questions

mlissner:
## Fixes

Fixes: #7045

## Summary

- Adds a `checked_documentation` checkbox to the contact form, required for "General Support", "Data or API Help", and "RECAP Extension Bug" issue types. We hope this will encourage more folks to read the documentation and spare us some repetitive help questions.
- Checkbox label links to the [wiki](https://wiki.free.law), [documentation](/help/), and [discussion forum](https://github.com/freelawproject/courtlistener/discussions)
- Conditionally shown/hidden via Alpine.js (`issue.requiresDocCheck`)
- Server-side validation in `ContactForm.clean()` rejects submissions without the checkbox for those issue types
- Help URL uses `reverse("help_home")` instead of a hardcoded path
- Tests cover both required (support/api/recap) and not-required (other types) scenarios

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

1. No special deployment steps required.

<details closed>
<details open>
image
</details>
<details open>
image
</details>


cl/simple_pages/forms.py
```diff
@@ -221,6 +227,20 @@ class ContactForm(forms.Form):
         label="Links to screenshots, error messages, or logs (if any)",
     )
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        # reverse() can't be called at class definition time because
+        # Django's URL configuration isn't loaded yet during import.
+        help_url = reverse("help_home")
+        self.fields["checked_documentation"].label = mark_safe(
```
mlissner: Why does this use mark_safe and then we also use the `safe` filter?

cl/assets/static-global/js/alpine/components/contact_form.js
```diff
@@ -12,6 +12,7 @@ document.addEventListener('alpine:init', () => {
         isTechnicalSupport: isTechSupport,
         isNotTechnicalSupport: !isTechSupport,
         hasAdditionalFields: isPartnershipsInquiry || isTechSupport,
+        requiresDocCheck: ['support', 'api', 'recap'].includes(this.issueType),
```
elisa-a-v: These values are hard-coding the same ones used in the backend as `ContactForm.DOCUMENTATION_CHECK_TYPES`. To keep one source of truth, we could pass the values from the server, like we're already doing with `techTypes` (see `init()` at the end of this file)

We're using the [`json_script` built-in tag](https://docs.djangoproject.com/en/6.0/ref/templates/builtins/#std-templatefilter-json_script) in the template to inject the Python data into a JSON in the final HTML.

elisa-a-v (changes requested):
Looks great! Just one minor comment about duplicated values to address before merging 🫡

cl/assets/static-global/js/alpine/components/contact_form.js
mlissner: Nice. I like this technique a lot. Done and pushed!

mlissner:
I added one more tweak in 9a7021d46fa1db4cf54f21accd4cae4cf7605f28!

cl/assets/static-global/js/alpine/components/contact_form.js
```diff
@@ -53,6 +54,7 @@ document.addEventListener('alpine:init', () => {
       // Fetch backend information: issue type options and Terms URL
       const issueTypeInput = this.$el.elements['issue_type'];
       if (issueTypeInput) this.issueType = issueTypeInput.value;
+      this.requiresDocCheck = JSON.parse(document.getElementById('requires-check-types').textContent);
```
elisa-a-v:
```suggestion
      this.requiresCheckTypes = JSON.parse(document.getElementById('requires-check-types').textContent);
```

cl/assets/static-global/js/alpine/components/contact_form.js
```diff
@@ -12,6 +12,7 @@ document.addEventListener('alpine:init', () => {
         isTechnicalSupport: isTechSupport,
         isNotTechnicalSupport: !isTechSupport,
         hasAdditionalFields: isPartnershipsInquiry || isTechSupport,
+        requiresDocCheck: this.requiresCheckTypes.includes(this.issueType),
```
elisa-a-v, line 15: there's a naming mismatch here because `init()` sets `this.requiresDocCheck` instead of `this.requiresCheckTypes`, and calling `.includes()` on `undefined` throws a TypeError that takes down the whole Alpine component.

Simply renaming the init function so it assigns what this getter expects fixes the issue because it'll always run before this is ever executed.

mlissner:
Thanks. Suggestion accepted. Sorry for not catching that. I thought I could sneak by and not make a dumb mistake, but clearly not.

Issue #7109, opened 2026-03-16
Implement Django Database Router


ERosendo:
### Describe the issue.

We need to build a custom Django database router that can direct read queries to read-only replicas using multiple strategies. 

The goal is to segment API traffic from frontend traffic at the database level so that spikes in one do not degrade the performance of the other.

The router should support multiple ways to decide which queries go to a replica. These strategies should be composable, so a request could match multiple criteria (e.g., by user and by view) when determining whether it should be routed to a replica.

### Routing Strategies

1. **Per-View Registration**

Allow specific views to be registered to use a read-only replica. This gives us fine-grained control over which endpoints hit the replica.

This could be implemented via a decorator or configuration-based approach (e.g., @use_replica or a list of view names in settings) to mark views whose read queries should be routed to a replica.
This approach lets us roll out replica usage incrementally, starting with high-traffic or read-heavy API views.

2. **Per-User Routing**

Maintain a configurable list of users (or API keys/tokens) whose requests are routed to a replica.

This is especially useful for large API clients—we can route a specific client’s traffic to a replica before rolling the change out more broadly.
The list should be configurable without requiring a deploy (e.g., via environment variables, settings flags, or a simple database/cache lookup).

3. **Percentage-Based Routing**

Send a configurable percentage of API reads to a replica, allowing gradual traffic shifts rather than hard cutovers.

This is useful for safely ramping up replica usage and validating behavior under increasing load.
The percentage should be adjustable without requiring a deploy.

### What type of issue is this?

Research

ERosendo:
After discussing with @albertisfu, the plan for the first stage is to focus on routing all API-sourced requests to the read replica.

Approach:
- All requests coming through the API will be routed to the read-only replica by default.
- Flag-controlled: This will be gated behind a feature flag so we can enable or disable it without a deploy and target specific users if needed.

This simplifies the initial implementation. Instead of implementing per-user filtering, we’ll control routing at view level, with the flag providing flexibility to adjust as needed.

Frontend requests will continue hitting the primary database as usual.

mlissner:
I agree with this approach, but I'd encourage a waffle or env to slide the percentage around, so it might be 10%, then 20, 40, 80, 100, or whatever.

PR #7110, opened 2026-03-17
feat(api): Route API reads to database replica via Django router and Waffle Flag

ERosendo:
## Fixes

This PR relates to #7109 but it doesnt fully resolve it.

## Summary

This PR introduces automatic database routing to direct API read requests to a read-only replica. By offloading read traffic from the primary database, this lays the groundwork for horizontal scaling and better traffic distribution.

This PR does not fully implement all aspects described in the github issue. Instead, it focuses on the core building blocks (database router, middleware, and feature flag control) so we can begin experimenting with a replica-based topology and validate the approach before expanding into additional strategies.

### How it works: 

1. `ReplicaRoutingMiddleware` (`cl/api/middleware.py`): It checks every incoming request against four conditions: 
      - Uses a safe HTTP method (`GET`, `HEAD`, `OPTIONS`)
      - path starts with `/api/rest/` 
      - a replica database is configured
      - the `replica-reads` Waffle Flag is enabled for the request. 
     
     When all pass, it sets a `ContextVar` for the duration of the request.

2. `ReplicaRouter` (`cl/api/routers.py`): Django database router. On `db_for_read`, it checks the `ContextVar`, if set, returns a replica alias via the existing `get_api_read_db()` cycling logic; otherwise returns `None` so Django uses the default database. Writes always go to default.

3. Waffle Flag (replica-reads): Controls rollout from Django admin without a deploy. Supports percentage-based activation, per-user targeting, and per-group targeting.

### Why `ContextVar` + Django database router

Django's database router doesn't receive the HTTP request, it only gets the model and query hints. So the router has no way to know whether the current request is an API call or a frontend page load. We need a way to pass that information from the middleware (which recieves the request) to the router (which decides the database).

`ContextVar` bridges this gap. it provides a clean solution: the middleware sets a value at the start of the request, the router reads it during query execution, and it’s automatically reset once the request completes. This approach works safely across both sync and async code paths. For more details, see: https://peps.python.org/pep-0567/ and https://docs.python.org/3/library/contextvars.html

### Other changes

- Tweaks settings: This PR updates the default value of `API_READ_DATABASES` from "replica" to an empty string ("") to make sure environments without a replica work out of the box. It also switches from `env()` to `env.list()` to guarantee proper list parsing.
- Updates `_get_replica_list`: This method now filters out any aliases specified in `API_READ_DATABASES` that are not defined in `DATABASES`, and falls back to `["default"]` when no valid replicas are available.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/assets/static-global/js/alpine/components/contact_form.js
+++ b/cl/assets/static-global/js/alpine/components/contact_form.js
@@ -12,6 +12,7 @@ document.addEventListener('alpine:init', () => {
         isTechnicalSupport: isTechSupport,
         isNotTechnicalSupport: !isTechSupport,
         hasAdditionalFields: isPartnershipsInquiry || isTechSupport,
+        requiresDocCheck: this.requiresCheckTypes.includes(this.issueType),
         noType: !this.issueType,
         isValidType: this.issueType && this.issueType !== 'legal',
       };
@@ -23,7 +24,7 @@ document.addEventListener('alpine:init', () => {
       const discussionsLink =
         'For community help and to see what others are discussing, visit the <a href="https://github.com/freelawproject/courtlistener/discussions">CourtListener Discussion</a> forum.';
       const hints = {
-        support: `<strong>Need a hand with CourtListener?</strong> First, check the other options above to make sure your question gets to the right team. Then, tell us what you’re trying to do and include the exact page link(s).<br><br>${discussionsLink}`,
+        support: `<strong>Need a hand with CourtListener?</strong> This option goes to our general inbox. Doublecheck the other options in the dropdown above to make sure your question gets to the right team. Then, tell us what you’re trying to do and include the exact page link(s).<br><br>${discussionsLink}`,
         api: `<a href="https://free.law">Free Law Project</a> makes it possible for you and your team to access our data. ${discussionsLink}`,
         recap:
           'Having trouble with the <a href="https://free.law/recap">RECAP extension</a>? Include your browser and version, the page/court link you were on, what you expected, and what happened instead.',
@@ -53,6 +54,7 @@ document.addEventListener('alpine:init', () => {
       // Fetch backend information: issue type options and Terms URL
       const issueTypeInput = this.$el.elements['issue_type'];
       if (issueTypeInput) this.issueType = issueTypeInput.value;
+      this.requiresCheckTypes = JSON.parse(document.getElementById('requires-check-types').textContent);
       this.techTypes = JSON.parse(document.getElementById('tech-types').textContent);
       this.termsURL = JSON.parse(document.getElementById('terms-url').textContent);
     },
--- a/cl/simple_pages/forms.py
+++ b/cl/simple_pages/forms.py
@@ -3,6 +3,8 @@ from typing import Any, Literal
 
 from django import forms
 from django.conf import settings
+from django.urls import reverse
+from django.utils.html import format_html
 from hcaptcha.fields import hCaptchaField
 
 SEALING_KEYWORDS_REGEX = re.compile(
@@ -25,7 +27,7 @@ class ContactForm(forms.Form):
     ISSUE_TYPE_CHOICES = [
         (SUPPORT_REQUEST, "General Support"),
         (PARTNERSHIPS, "Partnership Inquiry"),
-        (API_HELP, "Data or API Help"),
+        (API_HELP, "Data or API Support"),
         (DATA_QUALITY, "Report Data Quality Problem"),
         (RECAP_BUG, "RECAP Extension Bug"),
         (REMOVAL_REQUEST, "Case Removal Request"),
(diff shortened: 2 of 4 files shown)


cl/api/middleware.py
```diff
@@ -0,0 +1,57 @@
...
+        return await self.get_response(request)
+
+    def _should_route_to_replica(self, request: HttpRequest) -> bool:
+        if request.method not in ("GET", "HEAD", "OPTIONS"):
+            return False
+        if not request.path.startswith("/api/rest/"):
+            return False
+        if "replica" not in settings.DATABASES:
+            return False
+        return bool(flag_is_active(request, "replica-reads"))
```
mlissner, line 61: Is it crazy to cache locally here so that w don't add a DB query to every request?

cl/api/routers.py
```diff
@@ -34,6 +40,58 @@ def _get_replica_list():
...
+    def db_for_write(self, model, **hints) -> str | None:
+        return None
+
+    def allow_relation(self, obj1, obj2, **hints) -> bool | None:
+        return True
+
+    def allow_migrate(
+        self, db, app_label, model_name=None, **hints
+    ) -> bool | None:
+        return None
```
mlissner, line 119: Some comments about these methods would be helpful to me.

mlissner:
Looks great, thanks for working on this. I think I found one little performance opportunity (since this will be used a million times/day!), and that's it from me!

cl/api/middleware.py
ERosendo: That’s a good question. I had the same concern initially

I think we’re already covered, though. The `Flag` model inherits from a custom [`BaseModel`](https://github.com/django-waffle/django-waffle/blob/9b583738c46eace91ddb39e5e37e0d957e16eb34/waffle/models.py#L31), and its [`get()`](https://github.com/django-waffle/django-waffle/blob/9b583738c46eace91ddb39e5e37e0d957e16eb34/waffle/models.py#L49-L65) method uses caching under the hood. It checks the cache first, and only hits the DB on a cache miss. After that, the result is cached. Here’s the relevant code:

```python

class BaseModel(models.Model):
    ...
    def get(cls: type[_BaseModelType], name: str) -> _BaseModelType:
        cache = get_cache()
        cache_key = cls._cache_key(name)
        cached = cache.get(cache_key)
        if cached == CACHE_EMPTY:
            return cls(name=name)
        if cached:
            return cached

        try:
            obj = cls.get_from_db(name)
        except cls.DoesNotExist:
            cache.add(cache_key, CACHE_EMPTY)
            return cls(name=name)

        cache.add(cache_key, obj)
        return obj
```

Since [`flag_is_active()`](https://github.com/django-waffle/django-waffle/blob/9b583738c46eace91ddb39e5e37e0d957e16eb34/waffle/__init__.py#L17-L19) just calls `get(flag_name)` and evaluates it, we shouldn’t be adding a DB query on every request in practice.

```
def flag_is_active(request: HttpRequest, flag_name: str, read_only: bool = False) -> bool | None:
    flag = get_waffle_flag_model().get(flag_name)
    return flag.is_active(request, read_only=read_only)
```

So unless we start seeing frequent cache misses or performance issues, adding extra local caching here might be redundant.

cl/api/middleware.py
```diff
@@ -0,0 +1,57 @@
...
+            finally:
+                reset_replica_routing(token)
+        return await self.get_response(request)
+
+    def _should_route_to_replica(self, request: HttpRequest) -> bool:
+        if request.method not in ("GET", "HEAD", "OPTIONS"):
+            return False
+        if not request.path.startswith("/api/rest/"):
+            return False
+        if "replica" not in settings.DATABASES:
```
albertisfu: `replica` is no longer a default value of `API_READ_DATABASES`, so it’s possible that “replica” is not present in `DATABASES`, right? To be consistent, what about something like:

```suggestion
        if not any(
          db in settings.DATABASES for db in settings.API_READ_DATABASES                                                                                                  
      ):
```

mlissner:
I don't know if RDS supports it, but I think there's a way to do synchronous replication so that the write DB doesn't respond that the write is successful until it's replicated. If we do that, we don't have these kinds of timing issues. Might be a mess though if we have to apply this to all writes across everything. Hm.

cl/api/middleware.py
ERosendo: You're right. I'll update the `if` statement

ERosendo:
Thanks both for the feedback. This PR is ready for another review.

I met with Alberto this morning and mentioned that, before merging, we’ll need a new read-only replica. The one currently configured in our settings is also handling client replication connections, and its purpose is different from serving traffic. We’ll need to update the settings to support this experiment.

Longer term, I’m working on a mechanism to add/remove replica credentials without requiring a PR to introduce new environment variables.

cl/api/middleware.py
```diff
@@ -52,6 +52,8 @@ def _should_route_to_replica(self, request: HttpRequest) -> bool:
             return False
         if not request.path.startswith("/api/rest/"):
             return False
-        if "replica" not in settings.DATABASES:
+        if not any(
+            db in settings.DATABASES for db in settings.API_READ_DATABASES
```
albertisfu, line 58: Should we check whether `settings.API_READ_DATABASES `is a list, or convert it to a list before iterating over it?

Since the default value of `settings.API_READ_DATABASES` is a string.

ERosendo: Oh sorry. In a previous commit I updated the settings to use `env.list`, which should handle the parsing. I’ll update the default value to return `None`.

PR #7129, opened 2026-03-19
fix(header): profile hover text, tabnabbing, and frontend-checks skip directive

elisa-a-v:
## Summary
- **Profile button hover fix**: The `.btn-outline` class sets `color: white` on hover, but the button has a near-white hover background — resulting in invisible text. Adds `hover:text-greyscale-800` to keep text visible.
- **Tabnabbing fix**: Adds `rel="noopener noreferrer"` to the Support FLP menu's external links (donate + membership).
- **Frontend-checks skip directive**: Adds a generic `<!-- frontend-checks-skip: check_name -->` HTML comment that lets files opt out of specific linter checks. Only an allowlisted subset can be skipped (`check_hardcoded_ids`, `check_new_stack_leakage`, `check_raw_css`, `check_include_in_v2`) — security and a11y checks stay enforced. Applies it to header.html to suppress hardcoded ID warnings (it's a singleton component).

## Screenshots

<details>
<summary>Profile button hover — Before / After</summary>

image

image

</details>

## Test plan
- [ ] Visit homepage while logged in, hover the Profile button — text and icons should remain visible
- [ ] Visit any other page (e.g. search results), hover the Profile button — same check


mlissner:
Would it make sense to use `{#` and `#}` for these instead of HTML comments that would ship to end users?

PR #7147, opened 2026-03-25
refactor: separate download/extraction and fix DB timeouts in SCOTUS/Texas commands

albertisfu:
## Fixes

This fixes DB connection timeouts (`psycopg.OperationalError: SSL connection has been closed unexpectedly`) in `download_scotus_pdfs` and `download_texas_documents` management commands.

## Summary

- **Add `paginate_docs_queryset` utility** in `cl/corpus_importer/utils.py`: a reusable pk-based keyset pagination helper that replaces `iterator()`. Each batch is an independent query, so the DB connection is released between batches and won't time out during long celery waits.
- **Separate download from extraction**: both commands' download functions no longer chain an extraction step — they only dispatch download tasks.
- **Add `--page-limit` flag** (default 50) to extraction mode: documents exceeding the page limit are skipped, allowing smaller documents to be processed first. Run again with a higher limit for heavier docs.
- **Fix `--only-extraction` in Texas command**: changed from `type=bool` to `action="store_true"` so the flag actually works.

## Deployment

**This PR should:**

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -648,11 +648,11 @@ wheels = [
 
 [[package]]
 name = "courts-db"
-version = "0.10.26"
+version = "0.10.27"
 source = { registry = "https://pypi.org/simple" }
-sdist = { url = "https://files.pythonhosted.org/packages/de/d5/468434ebbef9e041b8358f1aa702f5538acf258f4ef60c61878edb66d391/courts_db-0.10.26.tar.gz", hash = "sha256:06cbbeabd91786b321d46af0a5c96b27a98a397c0a3f256c5682220a14380ee5", size = 139217, upload-time = "2025-08-27T19:57:25.712Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/f9/b5/67faa4ae8e273708256a08e71e9173f4671dbe274a35256c10ec054e446c/courts_db-0.10.27.tar.gz", hash = "sha256:0562bff60ad2267d5aea7a1acf327af4b00fdd893783612135372101a45e313e", size = 139584, upload-time = "2026-03-25T18:46:00.138Z" }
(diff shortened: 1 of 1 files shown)


cl/corpus_importer/management/commands/download_texas_documents.py
```diff
@@ -1,104 +1,117 @@
...
+            ocr_status__in=(
+                TexasDocument.OCR_UNNECESSARY,
+                TexasDocument.OCR_COMPLETE,
             )
         )
-        .values_list("pk", flat=True)
-        .order_by()
-    )
+    ).values_list("pk", "page_count")
     count = docs.count()
```
MorganBennetDev: We can filter for `page_count__lte=page_count` if `page_count` is not `None` and store that as `filtered_docs` to keep things simpler and move the check onto the DB. This also lets us loop over `batched(paginate_docs_queryset(filtered_docs))`, simplifying the code below. I think we should also be able to do this for SCOTUS.

```suggestion
    total_count = docs.count()
    filtered_docs = docs
    if page_count:
        filtered_docs = filtered_docs.filter(page_count__lte=page_count)
    filtered_count = filtered_docs.count()
    skipped_count = total_count - filtered_count
```

albertisfu: Thanks. I checked this query, and it’s performing well. So I applied your suggestion here and for SCOTUS, with a small tweak to also include docs with no `page_count`; otherwise, they would be completely omitted.

This is ready for another review.

albertisfu:
> Looks good. Just one change: I think we can use itertools.batched again now which should make things clearer and more concise.

Good catch. I've re-introduced `itertools.batched`.

PR #7153, opened 2026-03-26
feat(api): Routes DRF-authenticated users to read replicas

ERosendo:
## Fixes

This PR addresses a limitation in the existing replica routing logic.

## Summary

DRF authenticates users inside the view layer, after Django middleware has already run. The `ReplicaRoutingMiddleware` checks the `replica-reads` waffle flag at middleware time, when `request.user` is still `AnonymousUser` for Token/Basic auth users. This means per-user routing silently fails for API-authenticated requests.

This PR:
- Extracts `SAFE_METHODS` and `is_replica_configured()` into `cl/api/routers.py` so eligibility logic is defined once.
- Adds thin DRF authentication subclasses that re-evaluate the waffle flag after authentication, when the real user is known, and set the `ContextVar` for replica routing.
- Adds middleware cleanup so tokens set by auth classes are properly reset after the response.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/api/middleware.py
+++ b/cl/api/middleware.py
@@ -47,10 +47,9 @@ class ReplicaRoutingMiddleware:
                 reset_replica_routing(token)
 
         try:
-            response = self.get_response(request)
+            return self.get_response(request)
         finally:
             self._reset_auth_replica_token(request)
-        return response
 
     async def __acall__(self, request: HttpRequest) -> HttpResponseBase:
         if self._should_route_to_replica(request):
@@ -61,10 +60,9 @@ class ReplicaRoutingMiddleware:
                 reset_replica_routing(token)
 
         try:
-            response = await self.get_response(request)
+            return await self.get_response(request)
         finally:
             self._reset_auth_replica_token(request)
-        return response
 
     def _should_route_to_replica(self, request: HttpRequest) -> bool:
         if request.method not in SAFE_METHODS:


cl/api/middleware.py
```diff
@@ -25,6 +29,11 @@ def __init__(self, get_response) -> None:
         self.get_response = get_response
         self.async_mode = iscoroutinefunction(self.get_response)
 
+    def _reset_auth_replica_token(self, request: HttpRequest) -> None:
+        """Disable replica routing enabled by DRF auth classes."""
+        if getattr(request, "_replica_routing_token", None) is not None:
+            set_replica_routing(False)
```
albertisfu, line 35: Is there a reason to use `set_replica_routing(False)` instead of `reset_replica_routing(token)` like in `__call__`?

I see that we're storing the previous token value in `_replica_routing_token`.

cl/api/middleware.py
```diff
@@ -45,17 +57,16 @@ async def __acall__(self, request: HttpRequest) -> HttpResponseBase:
                 return await self.get_response(request)
             finally:
                 reset_replica_routing(token)
-        return await self.get_response(request)
+
+        response = await self.get_response(request)
+        self._reset_auth_replica_token(request)
```
albertisfu: In the previous lines, within the `if self._should_route_to_replica(request):` block, we’re doing:

we're doing:
```
            try:
                return await self.get_response(request)
            finally:
                reset_replica_routing(token)
```

So `reset_replica_routing` always runs.

To ensure `_reset_auth_replica_token` always runs, should we do the same?

cl/api/authentication.py
```diff
@@ -0,0 +1,99 @@
...
+        return
+
+    # Temporarily override the request user so waffle flag evaluation
+    # can use the authenticated user for targeting.
+    original_user = getattr(django_request, "user", None)
+    if user is not None and user.is_authenticated:
+        django_request.user = user  # type: ignore[assignment]
+
+    try:
+        if not flag_is_active(django_request, "replica-reads"):
```
albertisfu, line 57: `flag_is_active` is a synchronous method when fetching the status from the DB. Is there a risk of raising an error by calling a sync function from an async context here?

I’m also noticing that we’re doing the same in `_should_route_to_replica`, where that method is called from `async def __acall__`. However, we haven’t seen this issue so far 🤔

cl/api/authentication.py
```diff
@@ -0,0 +1,99 @@
...
+    If enabled, the resulting ContextVar token is stored on the underlying
+    Django request object so that middleware can reset it after the response.
+
+    :param request: The DRF request instance.
+    :param user: The authenticated user resolved by DRF. Must be passed
+        explicitly because ``request.user`` is not yet assigned when
+        ``authenticate()`` returns.
+    """
+    django_request: HttpRequest = request._request
+
```
albertisfu, line 42: If `set_replica_routing` is already set by `ReplicaRoutingMiddleware.__call__` (e.g., based on a percentage of traffic), would it be a good idea to skip this, since `set_replica_routing` is already set to `True`?


```suggestion
if _use_replica.get(False):                                                                                                                                             
      return
```

cl/api/middleware.py
ERosendo: Even though we store the token on the request, there’s a chance it was created in a different async context (for example, DRF auth may run in a thread pool via `sync_to_async`). Since ContextVar.reset(token) must be called in the same context where the token was created, this can lead to issues.

Instead of resetting with the token, calling set_replica_routing(False) clears the _use_replica variable from the current context safely.

cl/api/middleware.py
ERosendo: You're right. If `get_response` raises an exception, `_reset_auth_replica_token` never runs and the ContextVar leaks. It should be wrapped in try/finally block. 

Thanks for catching this.

cl/api/authentication.py
ERosendo: i think that's a good idea!

cl/api/middleware.py
```diff
@@ -45,17 +59,18 @@ async def __acall__(self, request: HttpRequest) -> HttpResponseBase:
                 return await self.get_response(request)
             finally:
                 reset_replica_routing(token)
-        return await self.get_response(request)
+
+        try:
+            response = await self.get_response(request)
+        finally:
+            self._reset_auth_replica_token(request)
+        return response
```
albertisfu: Would it be a good idea to return the response within the same try, just like in the `if self._should_route_to_replica(request):` branch?

This way, we avoid a “referenced before assignment” error and instead allow the original exception to propagate.

```suggestion
        try:
            return await self.get_response(request)
        finally:
            self._reset_auth_replica_token(request)
```

cl/api/middleware.py
```diff
@@ -36,7 +45,12 @@ def __call__(
                 return self.get_response(request)
             finally:
                 reset_replica_routing(token)
-        return self.get_response(request)
+
+        try:
+            response = self.get_response(request)
+        finally:
+            self._reset_auth_replica_token(request)
+        return response
```
albertisfu: Same suggestion as in the other comment.

albertisfu (commented):
Thanks @ERosendo just an additional comment.

cl/api/middleware.py
ERosendo: Do you think we might run into a “reference before assignment” exception? I’m fine with the refactor from a consistency standpoint, but if an exception occurs while obtaining the response, the return response statement would become unreachable.

albertisfu: You're completely right.

PR #7160, opened 2026-03-27
tests(state-race) fix race conditions for tally stats and opinion_ids

Brennan-Chesley-FLP:
## Fixes

This creates a mock interface to prevent intermittent failures for tests using stats tallying, and addresses one other small issue in tests when an opinion cluster is assigned id 0.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/.github/scripts/frontend_checks.py
+++ b/.github/scripts/frontend_checks.py
@@ -22,6 +22,37 @@ FAIL = "error"
 WARN = "warning"
 
 # ---------------------------------------------------------------------------
+# Per-file skip directives
+# ---------------------------------------------------------------------------
+
+# Checks that can be skipped via {# frontend-checks-skip: ... #} comments.
+# Only advisory/context-dependent checks belong here — security, a11y, and
+# architecture checks must stay enforced.
+SKIPPABLE_CHECKS = {
+    "check_hardcoded_ids",
+    "check_new_stack_leakage",
+    "check_raw_css",
+    "check_include_in_v2",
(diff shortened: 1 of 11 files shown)


cl/citations/management/commands/reindex_es_cite_fields.py
```diff
@@ -51,7 +51,7 @@ def handle(self, *args, **options):
         start_id = options["start_id"]
         end_id = options["end_id"]
 
-        if not (start_id and end_id):
+        if start_id is None or end_id is None:
```
Brennan-Chesley-FLP, line 54: This was causing a test to sporadically fail when an OpinionCluster id was 0 (Falsey).

docker/django/Dockerfile
```diff
@@ -31,8 +31,8 @@ ENV \
     # Unbuffer Python output for instant logging
     PYTHONUNBUFFERED=1
 
-ARG BUILD_ENV=prod
 FROM build-base AS python-base
+ARG BUILD_ENV=prod
```
Brennan-Chesley-FLP, line 35: This fixes an issue with podman builds for me, arising from what I'm going to call an underspecification in Dockerfile standards.

albertisfu (approved):
LGTM

PR #7165, opened 2026-03-30
fix(tames) allow non-pdf downloads

Brennan-Chesley-FLP:
## Fixes

TAMES includes non pdf downloads. We've identified at least wpd, html, mp3. This allows those files to be downloaded, and when possible (html, wpd) plain_text extracted using the doctor service. Extensions aren't available through TAMES, so the doctor service is also used to infer them. 

IMPORTANT DISCLAIMER: The TexasDocuments model still inherits from AbstractPDF. It's likely that non-pdfs represent a very small percentage of the files that TAMES hosts, so it makes at least some amount of sense to keep this as the representation. The downside of this is two parts:
- It's a slightly misleading name.
- We'll need to guard the views that we make to prevent inline pdf viewers from being used when they don't apply. [We already do this for Opinions](https://github.com/freelawproject/courtlistener/blob/65f177dc35b3acab373e7da08768fa87599bf725/cl/opinion_page/views.py#L867-L888). 

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`


cl/corpus_importer/tasks.py
```diff
@@ -3869,63 +3883,91 @@ def ingest_scotus_docket(docket_data: dict[str, Any]) -> None:
...
-                "Failed to download attachment PDF for TexasDocument %s from URL %s.",
+                "Failed to download document for TexasDocument %s from URL %s.",
                 texas_document.pk,
                 url,
             )
             self.request.chain = None
             return None
+
         tmp, sha1_hash = result
+        content = tmp.read()
```
albertisfu: When doing this, we’re loading the whole file into memory, which makes streaming the download pointless.

Can we just pass the standard 8 KiB to detect the file type?

This seems compatible with Doctor’s `extract_extension` service:

https://github.com/freelawproject/doctor/blob/f0c9e1c117709ac0ac9433caa53164c468997232/doctor/views.py#L297 

```suggestion
        content = tmp.read(8192)
```

cl/corpus_importer/tasks.py
```diff
@@ -3869,63 +3883,91 @@ def ingest_scotus_docket(docket_data: dict[str, Any]) -> None:
...
-            texas_document.page_count = int(response.text)
+
+        if extension == ".pdf":
+            response = async_to_sync(doc_page_count_service)(texas_document)
+            if response.is_success:
+                texas_document.page_count = int(response.text)
+
         texas_document.save()
+
+        if extension not in EXTRACTABLE_TEXAS_EXTENSIONS:
```
albertisfu, line 4000: The current `extract_pdf_document` stores the extracted content directly in the `plain_text` field, which is plain text for PDFs. For HTML and WPD, it contains HTML tags. To be consistent, should we clean up the output so it’s plain text as well? 

Should we also rename `extract_pdf_document` to something more generic?

If we do this, we’ll need to keep the original task in the codebase (as an orphan) and remove it once all the scheduled `extract_pdf_document` tasks have been processed. To prevent celery errors.

cl/corpus_importer/tasks.py
```diff
@@ -3869,63 +3883,91 @@ def ingest_scotus_docket(docket_data: dict[str, Any]) -> None:
...
+            self.request.chain = None
```
albertisfu:
```suggestion
            self.request.chain = None
            return None
```

cl/corpus_importer/tasks.py
Brennan-Chesley-FLP: Indeed! Made this change.

cl/corpus_importer/tasks.py
Brennan-Chesley-FLP: Going through doctor for an initial html hoisting and then tag stripping for html. Just doctor for wpd.

Brennan-Chesley-FLP:
Sorry about the twiddling on this one, I realize there's another change that needs to be included here.

Brennan-Chesley-FLP:
The last commit references code in https://github.com/freelawproject/juriscraper/pull/1897 for identifying the "File not found" pages that TAMES can produce. I'll need to update this PR once that one is merged to bump the version (why the Lint and tests are failing currently).

cl/corpus_importer/tasks.py
```diff
@@ -3884,72 +3910,120 @@ def _download_texas_document_pdf(
...
         logger.warning(
-            "Texas document PDF download: TexasDocument %s does not exist; skipping.",
+            "Texas document download: TexasDocument %s does not exist; skipping.",
             texas_document_pk,
         )
         task.request.chain = None
         return None
 
     url = texas_document.url
+    if url.endswith(" (i)"):
```
Brennan-Chesley-FLP: I'm extremely amenable to a different method of marking these if anyone has suggestions. Happy to add another field to the TexasDocuments if we're okay paying the migration price, or another method if someone prefers.

cl/corpus_importer/tasks.py
```diff
@@ -3312,6 +3316,16 @@ def download_to_file(tmp_file):
...
+def download_pdf_in_stream(
+    url: str,
+    identifier: str | int,
+    temp_prefix: str = "scotus_",
+):
+    return download_document_in_stream(
+        url, identifier, temp_prefix, require_pdf=True
+    )
+
+
```
albertisfu: Instead of creating this wrapper, can we just replace `download_pdf_in_stream` with `download_document_in_stream` using `require_pdf=True`?

I think that will be understandable enough.

cl/corpus_importer/tasks.py
albertisfu: Yeah, I think we should go with adding the new field, as it’s the best long-term solution. Perhaps the right solution is adding a small integer `status` field with options like Available or Missing. Also, I’m not sure if it’s posible, but if we encounter sealed documents, we could easily add that as another option.

This way, we could set `None` for Unknown, Available once we have the file, and Missing if we determine the file cannot be found.

In RECAP we use `is_available`, but that won’t work here since a file can be missing either because the download failed or because the file is actually missing.

cl/corpus_importer/tasks.py
```diff
@@ -4229,8 +4303,8 @@ def merge_texas_document(
                 # prevent mypy from complaining.
                 (
                     lambda pk: lambda: chain(
-                        download_texas_document_pdf.si(pk),
-                        extract_pdf_document.s(
+                        download_texas_document.si(pk),
+                        extract_formatted_text_document.s(
```
albertisfu, line 4313: We’ll also need to pass the `strip_html_tags` argument so we can properly extract HTML files from the `CaseMail` notifications and the ongoing scraper.

cl/corpus_importer/tasks.py
albertisfu: I’m not sure if it’s just the documents I chose for testing, but the extraction of WPD files is returning content with HTML tags.


Image

Here are the documents for reference:
https://search.txcourts.gov/SearchMedia.aspx?MediaVersionID=1991db56-91f2-4f31-acbf-4994f4dc7474&coa=coa01&DT=Other&MediaID=fbf3c0cb-2dec-46b0-87ba-f0f8b5e71209

https://search.txcourts.gov/SearchMedia.aspx?MediaVersionID=1991db56-91f2-4f31-acbf-4994f4dc7474&coa=coa01&DT=Other&MediaID=fbf3c0cb-2dec-46b0-87ba-f0f8b5e71209


If this is common for WPD files. We should also strip tags for WPD files.

Brennan-Chesley-FLP:
> If helpful. In cases like this you can try installing Juriscraper from your branch as a temp solution for testing.
> 
> `uv add git+https://github.com/freelawproject/juriscraper --branch feat/tames-missing-file-page-id`

I have a commit that I float around to add a few lines to my pyproject.toml that lets me reference whatever branch I have checked out locally. 

```
[tool.uv.sources]
juriscraper = { path = "../juriscraper", editable = true }
...
```

cl/search/migrations/0057_texasdocument_processing_state.py
```diff
@@ -0,0 +1,56 @@
...
+            model_name='texasdocument',
+            name='processing_state',
+            field=models.SmallIntegerField(choices=[(0, 'Pending'), (1, 'Downloaded'), (2, 'Bad URL'), (3, 'Summarized'), (4, 'Summary Failed'), (5, 'Sealed')], db_comment='The processing state of the document.', default=0, help_text='The processing state of the document.'),
+        ),
+        migrations.AddField(
+            model_name='texasdocumentevent',
+            name='processing_state',
+            field=models.SmallIntegerField(choices=[(0, 'Pending'), (1, 'Downloaded'), (2, 'Bad URL'), (3, 'Summarized'), (4, 'Summary Failed'), (5, 'Sealed')], db_comment='The processing state of the document.', default=0, help_text='The processing state of the document.'),
+        ),
+        migrations.RunSQL(
```
Brennan-Chesley-FLP: If we'd like this to happen via management command instead of in the migration, I'm happy to move it (and the pure sql versions below).

cl/corpus_importer/tasks.py
Brennan-Chesley-FLP: Added tag_stripping to wpds as well.

cl/search/state/texas/models.py
```diff
@@ -10,6 +10,23 @@
...
+    SUMMARIZED = 3
+    SUMMARY_FAILED = 4
+    SEALED = 5
+    CHOICES = (
+        (PENDING, "Pending"),
+        (DOWNLOADED, "Downloaded"),
+        (BAD_URL, "Bad URL"),
+        (SUMMARIZED, "Summarized"),
+        (SUMMARY_FAILED, "Summary Failed"),
+        (SEALED, "Sealed"),
```
Brennan-Chesley-FLP, line 20: Happy path is `PENDING -> DOWNLOADED -> SUMMARIZED` mp3 only get to `DOWNLOADED` but `SUMMARIZED` would represent transcription if we decide to go that route. Happy to add or remove states if we want here.

Brennan-Chesley-FLP:
Failing front-end lints currently as it includes a test html for the missing_file detection. I'll pull that fix out of the other PR into its own so we can merge it separately.

Brennan-Chesley-FLP:
Depends on [PR 7187](https://github.com/freelawproject/courtlistener/pull/7187) for front-end lint changes.

cl/corpus_importer/tasks.py
```diff
@@ -115,6 +115,7 @@
 from cl.corpus_importer.llm_models import CaseNameExtractionResponse
 from cl.corpus_importer.management.utils import TexasDocketMeta
 from cl.corpus_importer.prompts.system import CASE_NAME_EXTRACT_SYSTEM
+from cl.corpus_importer.state.texas.missing_file import is_missing_file_page
```
albertisfu: We don't usually use a whole file for a small method like this, unless we expect it to grow much in the future, which doesn't seem to be the case here. Instead, could we move it to a miscellaneous file like `utils.py`? 

That way, we can move other related methods there in the future as well.

```suggestion
from cl.corpus_importer.state.texas.utils import is_missing_file_page
```

cl/search/state/texas/models.py
albertisfu: This has states that can be derived from existing fields, which creates a synchronization risk where the state and actual field values can drift apart:
                                                                                                                                                                                
  - PENDING = no filepath_local and no plain_text                                                                                                                               
  - DOWNLOADED = filepath_local is set but no plain_text                                                                                                                        
  - SUMMARIZED = plain_text is non-empty
                                                                                                                                                                                
I'd recommend keeping only the states that cannot be inferred: `BAD_URL`, `SEALED`, and the extraction failure state. 

Make the field nullable, where `None` means "normal processing" and we could add a property or multiple ones to derive the current state from other fields when needed.
                                                                                                                                                                                
I also suggest using `EXTRACTION_FAILED` instead of `SUMMARY_FAILED` which can work for both text and audio documents.                                                                                                    

If you agree with this, we’ll only have a failed status, so we may need to rename `ProcessingState` to something like `DocumentException`.

cl/search/state/texas/models.py
```diff
@@ -80,6 +97,12 @@ class TexasDocument(AbstractDateTimeModel, AbstractPDF):
     media_id = models.UUIDField()
     media_version_id = models.UUIDField()
     url = models.URLField(max_length=250)
+    processing_state = models.SmallIntegerField(
+        choices=ProcessingState.CHOICES,
+        default=ProcessingState.PENDING,
+        help_text="The processing state of the document.",
+        db_comment="The processing state of the document.",
```
albertisfu: This model is using Morgan's `document_model` decorator. So if you add these to the docstrings the model get's auto-documented.

cl/search/migrations/0057_texasdocument_processing_state.py
albertisfu: Yeah, we prefer to run this type of backfill process in a separate script instead of a migration to prevent locking the table for too long. In this model, it won’t be too serious since we can stop the current downloading process while the migration is being applied, but it’s worth mentioning for future cases involving high-usage models.

Anyway, this backfill process won’t be required if we go with the suggestion of only adding failed statuses to the new field. We’ll populate those during the second pass of the download and extraction command.

cl/corpus_importer/tasks.py
```diff
@@ -3884,72 +3908,120 @@ def _download_texas_document_pdf(
...
     except TexasDocument.DoesNotExist:
         logger.warning(
-            "Texas document PDF download: TexasDocument %s does not exist; skipping.",
+            "Texas document download: TexasDocument %s does not exist; skipping.",
+            texas_document_pk,
+        )
+        task.request.chain = None
+        return None
+
+    if texas_document.processing_error == ProcessingError.BAD_URL:
```
albertisfu, line 3917: Here we're aborting the download of documents marked with `BAD_URL`.

However, if the document is later updated and the URL is fixed, we will never try to re-download it due to this check.

So we should clear the `processing_error` status if the document is updated here:

https://github.com/freelawproject/courtlistener/blob/5c7ccf21fb36ad2bbe1dedeccc9a19d77d5ffec1/cl/corpus_importer/tasks.py#L4288

cl/corpus_importer/tasks.py
```diff
@@ -4229,10 +4301,11 @@ def merge_texas_document(
                 # prevent mypy from complaining.
                 (
                     lambda pk: lambda: chain(
-                        download_texas_document_pdf.si(pk),
-                        extract_pdf_document.s(
+                        download_texas_document.si(pk),
+                        extract_formatted_text_document.s(
                             check_if_needed=False,
                             model_name="search.TexasDocument",
+                            strip_html_tags=True,
```
albertisfu, line 4316: In the `download_texas_documents` command, we differentiate between PDF and non-PDF documents, so HTML tags are only stripped from non-PDF documents. Should we apply the same logic here, so we don’t attempt to strip HTML tags from PDF documents?

cl/scrapers/tasks.py
```diff
@@ -574,13 +614,35 @@ async def extract_pdf_document_base(
                 do_extraction=False,
                 update_fields=["ocr_status", "plain_text"],
             )
```
albertisfu: This isn’t a change in this PR, but I noticed that this comparison is incorrect. It uses `isinstance(model, RECAPDocument)` when it should be `isinstance(rd, RECAPDocument)`.

This issue isn’t currently affecting behavior because it falls back to the else branch, which does:
`await rd.asave(update_fields=["ocr_status", "plain_text"])`

This is effectively the same as the default behavior, since `do_extraction` defaults to `False` in the `RECAPDocument.asave`  method

Could you please fix this or remove the redundant code. Whichever you think is best?

cl/corpus_importer/tasks.py
```diff
@@ -3884,72 +3908,120 @@ def _download_texas_document_pdf(
...
+        if extension not in KNOWN_TEXAS_EXTENSIONS:
+            logger.warning(
+                "Texas document download: Unexpected file extension "
+                "'%s' for TexasDocument %s from %s. Proceeding anyway.",
+                extension,
+                texas_document.pk,
+                url,
+            )
+
+        if extension == ".html" and is_missing_file_page(content):
```
albertisfu: I found an issue here: `is_missing_file_page` is using `content`, which is limited to `8192` bytes, but it needs the entire file to properly read the HTML. Otherwise the error page is stored as the HTML file.

It seems that HTML files aren’t too large, so we could load the whole file into memory and do something simple like this:

```suggestion
        if extension == ".html":
            full_content = tmp.read()
            tmp.seek(0)
            if is_missing_file_page(full_content):
```

Brennan-Chesley-FLP: I added a short-circuit if they're over 25k to skip the check, so we'll only be loading small ones here (the missing file pages I could find were 21k, so there's a little bit of buffer room).

cl/corpus_importer/tasks.py
```diff
@@ -3884,72 +3908,127 @@ def _download_texas_document_pdf(
...
+
+        if extension == ".html":
+            tmp.seek(0, 2)
+            file_size = tmp.tell()
+            tmp.seek(0)
+            if file_size <= 25_000:
+                full_content = tmp.read()
+                tmp.seek(0)
+                if is_missing_file_page(full_content):
+                    logger.error(
```
albertisfu: Can we use `logger.warning` instead?

`logger.error` will be quite noisy for these kinds of files, which we should be able to identify if needed.
```suggestion
                    logger.warning(
```

cl/corpus_importer/tasks.py
```diff
@@ -3884,72 +3908,127 @@ def _download_texas_document_pdf(
...
         logger.warning(
-            "Texas document PDF download: TexasDocument %s does not exist; skipping.",
+            "Texas document download: TexasDocument %s does not exist; skipping.",
+            texas_document_pk,
+        )
+        task.request.chain = None
+        return None
+
+    if texas_document.processing_error == ProcessingError.BAD_URL:
+        logger.error(
```
albertisfu: Also here `logger.warning` would be better.

```suggestion
        logger.warning(
```

albertisfu (changes requested):
looks good @Brennan-Chesley-FLP just one last tweak please

Issue #7068, opened 2026-03-03
Implement: Docket entries tab


elisa-a-v:
## Summary

Implement the redesigned Docket Entries tab. This is a larger effort spanning multiple PRs — the filter bar, pagination, and entry rows will be tracked as separate sub-issues.

### Docket entry rows
- Document number, date filed, description
- Column headings same as current CL
- "Main Document" and "Attachment" links with document type labels
- "Download PDF" dropdown button (uses `c-dropdown-button`, [#7064](https://github.com/freelawproject/courtlistener/issues/7064)) in the rightmost column
- "Buy on PACER" button with price for documents not yet in RECAP
- "Share" option removed (per [#5472](https://github.com/freelawproject/courtlistener/issues/5472))

### Search & filter bar
- Search within docket, date range pickers, document range, sort (Asc/Desc)
- Mobile: slide-in filter panel with search, date range, document range, Apply button

### Pagination
- Previous/Next navigation controls
- Document count/range display

### Mobile layout
- Responsive layout as shown in mobile mockups
- Compact entry display with stacked content
- **Tabs should use a `<select>` element on mobile** instead of horizontal scrolling, so all available tabs are visible without scrolling

## Figma

- [Desktop docket entries](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=7537-51189&m=dev)
- [Mobile docket entries](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=9170-61899&m=dev)

## Design references

- [#5472](https://github.com/freelawproject/courtlistener/issues/5472) — Docket Page - Docket Entries Tab
- [#5902](https://github.com/freelawproject/courtlistener/issues/5902) — Design: Mobile for Docket Tabs
- [#1535](https://github.com/freelawproject/courtlistener/issues/1535) — Selecting docket entries is inconsistent

## Dependencies

- [#7067](https://github.com/freelawproject/courtlistener/issues/7067) — Docket page base
- [#7063](https://github.com/freelawproject/courtlistener/issues/7063) — Component: Button
- [#7064](https://github.com/freelawproject/courtlistener/issues/7064) — Component: Dropdown button

Issue #7067, opened 2026-03-03
Implement: Docket page base, metadata & actions


elisa-a-v:
## Summary

Implement the redesigned docket page shell: the case header, metadata section, action buttons, and tab navigation. This is the first implementation PR for the docket page redesign — it sets up the waffle-flagged base template that all subsequent tab implementations build on.

### Case header
- Filed date displayed above the case title (e.g., "Filed: September 26, 2024")
  - **Note**: The current design shows this as a red badge/pill, but it may look too similar to action buttons. Consider using the `c-eyebrow` component instead — discuss with team.
- Case title with docket number — docket number must never wrap to a separate line
- Court name subtitle where applicable (see [#5906](https://github.com/freelawproject/courtlistener/issues/5906))

### Action buttons
- Add a Note, Tags, Get Alerts, View on PACER
- Uses `c-button` component ([#7063](https://github.com/freelawproject/courtlistener/issues/7063)) — chip variant for the compact style shown in designs
- "Tags" should be a dropdown (see [#5906](https://github.com/freelawproject/courtlistener/issues/5906))

### Metadata grid
- Key-value pairs: Last Updated, Assigned To, Citations, Date of Last Known Filing, Cause, Nature of Suit, Jury Demand, Jurisdiction Type
- Uses `c-definition-list` component ([#7065](https://github.com/freelawproject/courtlistener/issues/7065))
- Some metadata should appear as links (e.g., "Assigned To") — see [#5906](https://github.com/freelawproject/courtlistener/issues/5906)

### Tab navigation
- Tabs: Docket Entries, Parties and Attorneys, FJC Integrated Database, Authorities
- Uses existing `c-tabs` component
- Tabs should only appear when relevant data exists (current behavior)
- **Mobile**: `c-tabs` already supports responsive `<select>` fallback instead of horizontal scrolling — use that behavior here, and extend the component if needed

### Mobile
- Mobile metadata layout pending design ([#7066](https://github.com/freelawproject/courtlistener/issues/7066)) — implement what's possible and iterate once design is available

## Figma

[Figma selection](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=7537-55748&m=dev)

## Design references

- [#5449](https://github.com/freelawproject/courtlistener/issues/5449) — Docket Page - Docket Metadata & Actions
- [#5906](https://github.com/freelawproject/courtlistener/issues/5906) — Design: updates to Docket page
[...]

Issue #7063, opened 2026-03-03
Component: Button


elisa-a-v:
## Summary

Create a reusable `c-button` Cotton component for the component library. Currently, buttons in the redesigned pages use CSS classes (`btn-primary`, `btn-outline`) directly on `<button>` and `<a>` elements, but there is no reusable component that standardizes markup, accessibility attributes, and variants.

This component is a prerequisite for several docket page implementation issues.

## Requirements

- **Default slot** for button content — allows flexible content including text, icons, or both
- **Variants**: All 5 variants as defined in the Figma design page (see link below), each available in two sizes (40px and 48px)
- **Chip variant**: An additional smaller variant (single size) for the compact pill-shaped buttons seen in the docket page action bar
- **Element flexibility**: Should render as `<button>` by default, but support rendering as `<a>` when an `href` is provided (for link-styled buttons like "View on PACER")
- **Accessibility**: Proper `aria-label` support, focus styles, disabled state
- **Refactor existing buttons**: All buttons in templates extending `new_base.html` should be migrated to use `c-button` instead of raw CSS classes. Once migration is complete, remove the standalone `btn-primary`, `btn-outline` etc. CSS classes if possible (or deprecate them).
- **Add to the component library page** (`v2_components.html`) with demo of all variants/sizes, props, and code example

## Figma

- [Buttons design](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=1-1183&m=dev)

## Context

- Part of [#4997](https://github.com/freelawproject/courtlistener/issues/4997) (Redesign the docket pages)
- Used across all docket page tabs and throughout other redesigned pages

PR #7166, opened 2026-03-30
feature(docket): Add starting framework for docket page

EstiShay:
## Fixes
<!-- What bugs does this fix? Use this syntax to auto-close the issue: -->
<!-- https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword -->
<!-- E.g.: "Fixes: #XYZ" -->
Related to #7067

## Summary
 First PR for the docket page v2 redesign. Creates the page shell — title, metadata, and tab navigation — as a Cotton component (`<c-docket-page>`), without any tab content yet. Follow-up PRs will add content for each tab (entries, parties, FJC IDB, authorities).
                                                                                                                                                                                                                 
Approach: Instead of a template inheritance chain (like the legacy `docket.html` → `docket_tabs.html` → `base.html`), the shell is a Cotton component with explicit props. Each future tab template will extend `new_base.html` directly and use <`c-docket-page`>, keeping tab templates flat and self-contained.

 Tab navigation uses styled `<a>` links rather than the existing `c-tabs` component. `c-tabs` is designed for client-side Alpine panel switching (`x-show`/`x-data`) where all content is in the DOM simultaneously. Docket tabs are URL-based — each tab is a separate view with its own URL — so we use server-rendered `<a>` links styled to match `c-tabs` visually, with a mobile `<select>` fallback. This also sets us up cleanly for htmx tab switching in a follow-up PR.
                                                                                                                                                                                                                 
What's included:                                                                                                                                                                                               
  -  `<c-docket-page>` Cotton component with all docket metadata ported from legacy `docket_tabs.html`
  - `v2_docket.html` extending `new_base.html` (middleware swaps automatically when `use_new_design flag` is on)                                                                                                       
  - Tab navigation: desktop `<nav>` links styled to match c-tabs, mobile `<select>` with CSP-safe Alpine handler
  - `aria-current="page"` on active tab, focus indicators, labeled nav region                                                                                                                                      
  - `scrollbar-none` Tailwind utility (not built into v3)                                                                                                                                                          
  - Sync notice on legacy docket.html                                                                                                                                                                            
  - Component library entry                                                                                                                                                                                      
                                                                                                                                                                                                                 
Not in this PR: tab content, action buttons (blocked on #7063), htmx tab switching, view changes. Metadata grid layout awaits #7126 (uses flat description list as stand-in).       

 ---                                                                                                                                                                                                            
  Update (April 9): Addressed review feedback:
                                                                                                                                                                                                                 
  - DRY refactor: Moved all metadata and tab conditional logic out of the template and into Python view helpers (build_docket_metadata(), build_bankruptcy_metadata(), build_originating_court_metadata(),     
  build_docket_tabs()). The template now loops over pre-built lists instead of ~200 lines of individual conditionals.                                                                                            
  - View changes: view_docket now calls the helpers and passes metadata, bankruptcy_metadata, originating_court_metadata, and tabs to the template context. Sync DB access is wrapped in sync_to_async for the 
  async view.                                                                                                                                                                                                    
  - Action buttons: Added placeholder <c-button variant="outline"> shells for Add a Note, Tags, Get Alerts, and View on PACER (blocked on menu button component for full interactivity).                       
  - Description list upgrade: Now using <c-description-item> wrappers with responsive grid layout after #7126 merged.                                                                                            
  - Tests added: Unit tests for all four helper functions + integration tests for v2 template rendering with waffle flag.                                                                                        
  - Type hints: Proper model classes (BankruptcyInformation, OriginatingCourtInformation) instead of object; mypy passes clean
  --- 

Update 4/16: Addressing @elisa-a-v's review on commit ccc6d69:                                                                                                                                                              
                                                                                                                                                                                                               
  **HTML out of Python.** `build_originating_court_metadata` returns
  suffix_text/suffix_url data fields instead of constructing HTML strings.
  The `|safe` filter usage is gone.                                                                                                                                                                              
   
  **Component extraction.** New `c-metadata-section` component DRYs the                                                                                                                                          
  three metadata blocks in `docket_page.html`.                                                                                                                                                                 

  **Docs refreshed.** Docket-page prop list updated to match actual props;                                                                                                                                       
  added `c-metadata-section` section; cleaned up prop label formatting.
                                                                                                                                                                                                                 
  **Additional review conducted: A11y.** External links now get                                                                                                                                                
  `aria-label` with "(opens in new tab)"; replaced `title` attr with
  `aria-label`; composed `rel` in a single attribute to fix a latent                                                                                                                                             
  duplicate-attribute bug.
                                                                                                                                                                                                                 
  **Additional review conducted: Pythonic code.** Added `MetadataItem`                                                                                                                                         
  TypedDict; extracted `_person_item` helper (4 call sites); replaced
  `pytz` with stdlib `zoneinfo` (pytz support removed in Django 5.0);                                                                                                                                            
  moved inline imports to module top.
                                                                                                                                                                                                                 
  **Tests.** Added `BuildOriginatingCourtMetadataTest` (3 tests);                                                                                                                                              
  mutation-tested all three.

  **Menu buttons.** Placeholder buttons now incorporate the new menu button component.

 The two latest semgrep findings are false positives (trusted server-side URL construction, no user input path). 
                                                                                                                                                                                                                                                                                                                                                                                                                     
  Claude (Claude Code) helped with this review-response work and drafted                                                                                                                                       
  this PR description update.
  
---

Update 4/28:
- Renamed `authority_count` to `has_authorities` in `views.py` and `utils.py` to accurately reflect that it's a bool (from `aexists()`), not a count. Corrected the type hint from `int` to `bool` in `build_docket_tabs()`.
- URL-encoded all query string values in `utils.py` metadata builders using `django.utils.http.urlencode` to prevent special characters from breaking URLs, and added a test.
- Consolidated `sync_to_async` lambda wrappers for bankruptcy/originating court metadata into a single `_get_related` helper that isolates DB access, then calls the pure-Python builders directly.
- Extracted citation-building logic from `templatetags/extras.py` into `build_citation_string()` in `utils.py`. The template tag now calls the helper, inverting the dependency.
- Fixed merge regression in `v2_components.html`: restored `'variant'` keys (were incorrectly changed to `'style': 'btn-*'`) and restored the `buttons` prop description.
- Moved inline imports in new tests to the top of the file.
- The `check_v2_register` warning will be handled in the #7068 branch.

---

## Deployment

**This PR should:**
<!-- The following labels control the deployment of this PR if they’re applied. -->
<!-- Please put an "X" in the box on ones that apply. -->
<!-- For more details on what pods are affected by each label, see the wiki -->
<!-- https://github.com/freelawproject/courtlistener/wiki/Pull-requests-%60skip%E2%80%90%7Btype%7D%E2%80%90deploy%60-labels -->

<!-- Check here if the entire deployment can be skipped -->
<!-- This might be the case for a small fix, a tweak to documentation or something like that. -->
    <!-- Check here if the web tier can be skipped -->
    <!-- This is the case if you're working on code that doesn't affect the front end, like management commands, tasks, or documentation. -->
    - [ ] `skip-web-deploy`
    <!-- Check here if the deployment to celery can be skipped -->
    <!--This is the case if you make no changes to tasks.py or the code that tasks rely on. -->
    - [x] `skip-celery-deploy`
    <!-- check this if deployment to cron jobs can be skipped -->
    <!-- This is the case if no changes are made that affect cronjobs. -->
    - [x] `skip-cronjob-deploy`
    <!-- Deployment of daemons can be skipped -->
    <!-- This is the case if you haven't updated daemons or the code they depend on. -->
    - [x] `skip-daemon-deploy`

<!-- DELETE this section if your PR doesn't require screenshots. -->
<!-- If this changes the front end, please include desktop and mobile screenshots or videos showing the new feature. -->
<details closed>
<details open>
<img width="1380" height="853" alt="image" src="https://github.com/user-attachments/assets/fd880acd-6690-49f4-85b2-d64625c04f40" />
</details>
<details open>
<img width="618" height="747" alt="image" src="https://github.com/user-attachments/assets/c34b7f44-48c4-4ac6-b5f9-b7fced1e906e" />
</details>
<!-- END DELETE -->

<!-- Thank you for contributing and filling out this form! -->

_Work created with Claude. PR description written by Claude._


elisa-a-v (changes requested):
Awesome, this a great start! As discussed in our meeting, there's a lot of repetition that I think we can handle in an easier way by building the data in the view and just looping in the template.

**Metadata: build a list in the view**

```python
# In utils.py or the view
def build_docket_metadata(docket, timezone_str):
    """Build metadata items for the docket page description list."""
    items = []
    if docket.source in docket.RECAP_SOURCES:
        items.append({"label": "Last Updated", "value": localtime(docket.date_modified, pytz.timezone(timezone_str))})
    if docket.panel_str:
        items.append({"label": "Panel", "value": docket.panel_str})
    if docket.assigned_to:
        items.append({"label": "Assigned To", "value": docket.assigned_to.name_full, "url": docket.assigned_to.get_absolute_url()})
    elif docket.assigned_to_str:
        items.append({"label": "Assigned To", "value": docket.assigned_to_str, "url": f'/?type=r&assigned_to="{docket.assigned_to_str}"', "nofollow": True})
    if docket.date_filed:
        items.append({"label": "Date Filed", "value": docket.date_filed})
    if docket.cause:
        items.append({"label": "Cause", "value": docket.cause, "url": f'/?type=r&cause="{docket.cause}"', "nofollow": True})
    # ... same pattern for all fields
    return items
```

Same approach for bankruptcy and originating court (so three separate lists), then the entire metadata section in the template collapses to a loop:

```html
<c-description-list class="grid grid-cols-1 md:grid-cols-2 lg:grid-cols-4 gap-4 mt-4">
  {% for item in docket_metadata %}
    <c-description-item>
      <c-description-term>{{ item.label }}</c-description-term>
      <c-description-details>
        {% if item.url %}
          <a class="links" href="{{ item.url }}" {% if item.nofollow %}rel="nofollow"{% endif %}>{{ item.value }}</a>
        {% else %}
          {{ item.value }}
        {% endif %}
      </c-description-details>
    </c-description-item>
  {% endfor %}
</c-description-list>
```

**Tabs: same idea**

Build the tab list in the view so both desktop nav and mobile select loop over the same data:

```python
tabs = [{"label": "Docket Entries", "url": docket.get_absolute_url(), "key": "entries"}]
if parties:
    tabs.append({"label": "Parties and Attorneys", "url": reverse("docket_parties", kwargs={"docket_id": docket.pk, "slug": docket.slug}), "key": "parties"})
if has_idb_data:
    tabs.append({"label": "FJC Integrated Database", "url": reverse("docket_idb_data", kwargs={"docket_id": docket.pk, "slug": docket.slug}), "key": "idb"})
if authority_count:
    tabs.append({"label": "Authorities", "url": reverse("docket_authorities", kwargs={"docket_id": docket.pk, "slug": docket.slug}), "key": "authorities"})
```

```html
{# Desktop #}
<nav class="hidden md:flex gap-1 mt-8 overflow-x-auto scrollbar-none" aria-label="Docket sections">
  {% for tab in tabs %}
    <a href="{{ tab.url }}"
       {% if active_tab == tab.key %}aria-current="page"{% endif %}
       class="px-5 py-3 text-sm font-normal rounded-t-2xl whitespace-nowrap focus-visible:ring-2 focus-visible:ring-primary-400 focus-visible:outline-none {% if active_tab == tab.key %}bg-greyscale-100 text-greyscale-900{% else %}bg-greyscale-50 text-greyscale-600{% endif %}">
      {{ tab.label }}
    </a>
  {% endfor %}
</nav>

{# Mobile #}
<div class="bg-greyscale-100 rounded-t-2xl md:hidden pt-3 px-4 mt-8" x-data="docketPage">
  <select x-on:change="navigateToSelected" aria-label="Select docket section" class="w-full">
    {% for tab in tabs %}
      <option value="{{ tab.url }}" {% if active_tab == tab.key %}selected{% endif %}>{{ tab.label }}</option>
    {% endfor %}
  </select>
</div>
```

This would take the component from ~425 lines down to maybe ~150, moving the conditional logic to Python where it's easier to maintain and test.

elisa-a-v (commented):
Really nice, the DRY refactor is a big improvement! Moving the metadata logic into Python helpers makes the template much more readable. I think we can push it just a little bit further though:

### 1. HTML construction in Python (`utils.py`)

`build_originating_court_metadata` constructs raw HTML strings for the "Appealed From" field — `<a>` tags with classes, `rel` attributes, etc. This mixes concerns: the helper should prepare **data**, and the template should handle **markup**.

Suggestion: return the pieces as separate keys in the dict and let the template compose them:

```python
item = {"label": "Appealed From", "value": appeal_court_name}
if og_info.docket_number:
    item["suffix_text"] = og_info.docket_number
    if docket.appeal_from:
        item["suffix_url"] = f"/?type=r&docket_number={og_info.docket_number}&court={docket.appeal_from.pk}"
        item["nofollow"] = True
    elif og_info.administrative_link:
        item["suffix_url"] = og_info.administrative_link
        item["is_external"] = True
items.append(item)
```

Then the template renders the parenthetical suffix as a link when `suffix_url` is present. This keeps all HTML in templates where it belongs, and makes it easier to update styling later.

### 2. Extract a `c-metadata-section` component

The main metadata, bankruptcy, and originating court sections all follow the same pattern: optional heading + description list + loop over items with url/safe/plain rendering. Consider extracting a `c-metadata-section` component that takes `:items` and an optional `title`:

```html
<c-metadata-section :items="metadata" />
<c-metadata-section :items="bankruptcy_metadata" title="Bankruptcy Information" />
<c-metadata-section :items="originating_court_metadata" title="Originating Court Information" />
```

The component handles the `<h2>` (when title is provided), the `<c-description-list>`, the item loop, and all the rendering variants (url/nofollow/safe/plain/suffix/external). Bankruptcy items just happen not to use any link keys — no special-casing needed. This keeps the docket page component focused on page structure (eyebrow, title, actions, tabs) rather than metadata rendering details.

---

Also flagging that the component library needs to be updated now that props have changed, it's using stale attributes in the examples. You may be aware but just in case 👌🏽

cl/opinion_page/views.py
```diff
@@ -397,15 +401,35 @@ def paginate_docket_entries(docket_entries, docket_page):
         rd.prayer_count = prayer_counts.get(rd.id, 0)
         rd.prayer_exists = existing_prayers.get(rd.id, False)
 
+    parties = await docket.parties.aexists()
+    has_idb_data = bool(docket.idb_data_id)
+    authority_count = await docket.ahas_authorities()
```
elisa-a-v: `authority_count` is a bool (from `aexists()`), not a count. We should either rename the variable, or change its value.

To prevent messing with the current view, if the context is considerably different than it currently is, we could consider doing a router view like we're doing with the new homepage.

cl/opinion_page/utils.py
```diff
@@ -11,11 +13,14 @@
...
 from django.http import HttpRequest
 from django.shortcuts import aget_object_or_404  # type: ignore[attr-defined]
+from django.urls import reverse
+from django.utils.timezone import localtime
 from django_elasticsearch_dsl.search import Search
 from elasticsearch.exceptions import ApiError, ConnectionTimeout, RequestError
 from elasticsearch_dsl import Q
 
 from cl.alerts.models import DocketAlert
+from cl.custom_filters.templatetags.extras import citation
```
elisa-a-v: `citation()` is a Django template tag (`@register.simple_tag` in `templatetags/extras.py`) being imported and called directly from Python. Template tags live in `templatetags/` because they're meant for template rendering, so pulling one into `utils.py` is a backwards dependency. We should extract the citation-building logic into a plain helper in `utils.py` and have the template tag call the helper.

cl/opinion_page/views.py
```diff
@@ -397,15 +401,35 @@ def paginate_docket_entries(docket_entries, docket_page):
...
+                lambda: build_originating_court_metadata(
+                    docket,
+                    getattr(docket, "originating_court_information", None),
+                )
+            )(),
+            "tabs": build_docket_tabs(
+                docket, parties, has_idb_data, authority_count
+            ),
         }
     )
```
elisa-a-v, line 443: This is mostly a readability suggestion, not a functional issue: the `sync_to_async(lambda: ...)` wrappers exist because `getattr(docket, "bankruptcy_information", ...)` is a reverse OneToOne that triggers a DB query, and async Django forbids sync DB access. But each wrapper covers *both* the DB access *and* the pure-Python `build_*` call — which makes it look like `build_*` itself needs a sync context when it doesn't. Isolating just the lookups into one helper makes the sync/async boundary explicit and does both lookups in a single thread-pool trip:

  ```python
  @sync_to_async
  def _get_related(docket):
      return (
          getattr(docket, "bankruptcy_information", None),
          getattr(docket, "originating_court_information", None),
      )

  bankr_info, og_info = await _get_related(docket)

  context.update({
      "bankruptcy_metadata": build_bankruptcy_metadata(bankr_info),
      "originating_court_metadata": build_originating_court_metadata(docket, og_info),
  })
  ```

cl/opinion_page/tests.py
```diff
@@ -2513,3 +2513,236 @@ def test_cluster_redirection(self):
...
+        docket = DocketFactory(court=self.appellate_court, source=Docket.RECAP)
+        items = build_originating_court_metadata(docket, None)
+        self.assertEqual(items, [])
+
+    def test_appealed_from_links_lower_court_docket_number(self) -> None:
+        """When the lower court is known, the lower-court docket number
+        renders as a nofollow RECAP search link via data fields — never as
+        a raw HTML string in the value."""
+        from cl.opinion_page.utils import build_originating_court_metadata
+        from cl.search.models import OriginatingCourtInformation
```
elisa-a-v: Any reason to inline these and all the other imports in the tests?

cl/opinion_page/utils.py
elisa-a-v, line 1: URL construction in this file doesn't encode special characters. DB field values get dropped straight into the query string via f-strings at lines 88, 186, 196, 206. If a value contains `&`, `#`, `"`, `%`, or spaces, the resulting URL breaks.

Fix with `urlencode`:

```python
from django.utils.http import urlencode

"url": f"/?{urlencode({'type': 'r', 'cause': f'\"{docket.cause}\"'})}",
```

Carry-forward from the legacy page — not introduced here — but worth fixing while these helpers are new code.

cl/opinion_page/templates/cotton/docket_page.html
```diff
@@ -0,0 +1,220 @@
...
+         {% if active_tab == tab.key %}aria-current="page"{% endif %}
+         class="px-5 py-3 text-sm font-normal rounded-t-2xl whitespace-nowrap focus-visible:ring-2 focus-visible:ring-primary-400 focus-visible:outline-none {% if active_tab == tab.key %}bg-greyscale-100 text-greyscale-900{% else %}bg-greyscale-50 text-greyscale-600{% endif %}">
+        {{ tab.label }}
+      </a>
+    {% endfor %}
+  </nav>
+
+  {# Mobile select #}
+  <div class="bg-greyscale-100 rounded-t-2xl md:hidden pt-3 px-4 mt-8" x-data="docketPage">
+    <select x-on:change="navigateToSelected" aria-label="Select docket section" class="w-full">
```
elisa-a-v, line 204: This `change` handler fails WCAG 3.2.2 ("On Input"): for keyboard users, arrow keys cycle through values and fire `change` per press, so the page navigates before they reach their intended choice. However, we have the same pattern in the tabs component, so fixing this may require some design decisions. I've opened an issue #7260 so we don't block this PR

elisa-a-v (changes requested):
Thanks for the updates and the patience with the review 🙏🏽

Refactor looks great, I still have a few comments and questions left inline

cl/opinion_page/tests.py
EstiShay: Not to my knowledge. I decided to keep them inline only to follow already-established conventions in this file.

elisa-a-v (approved):
Nice, thank you for the updates, looks great!

Issue #7145, opened 2026-03-24
Texas Importer Issues


MorganBennetDev:
This is a tracking issue for problems encountered in the Texas merging process and will be updated as we find them.

## Merging Issues

These may or may not prevent dockets from being imported and will require a fix to CourtListenere.

### [One or more merger steps failed](https://freelawproject.sentry.io/issues/7358918533/events/cf4d2279e4714678a4df3dce34e6ca69/?project=5257254)

This issue won't stop dockets from being merged it just may lead to some missing info we need to fill in later.

This does not appear to be associated with any other error logs so it can only be one of two paths. The only way for `success` to be `False` but no errors to be logged is if:
    1. The originating court has unknown type
    2. There was a Supreme Court case with no originating or appellate information.
I suspect these are mostly option 2, but we'll have to take a look at the warnings once the merger run is complete to know for sure.

### django.db.transaction.TransactionManagementError: An error occurred in the current transaction. You can't execute queries until the end of the 'atomic' block.

This will likely prevent dockets from being merged. I suspect it's an uncommon path in the top level atomic block being hit, but I'll have to do more investigation.

### psycopg.errors.UniqueViolation: duplicate key value violates unique constraint "people_db_attorneyorganization_lookup_key_key"

This should not prevent dockets from being merged. Unsure of the cause.

MorganBennetDev:
### Court with ID None not found while populating CaseTransfer.origin_court

and

### Unknown court ID: texas_unknown

These are the same issue. There's no check for unknown court IDs when handling appellate workload transfers since I assumed that was already handled by the check for the transfer existing, but apparently not. We should add a check to `merge_texas_case_transfers` and add one of the dockets as an example to Juriscraper.

MorganBennetDev:
### Unrecognized docket number format: "WR-09,671-03" ([example](https://search.txcourts.gov/Case.aspx?cn=WR-09,671-03&coa=coscca))

This is an issue with the docket number validation in Juriscraper. The validation regex for writs in CCA requires that the first digit of the number be non-zero because this was the case for all examples we found. However, it does not seem this is a correct assumption.

This appears to be a relatively prevalent issue and is preventing ingestion of dockets (since they fail the parsing step).

albertisfu:
I've moved the parsing issues to Juriscraper: https://github.com/freelawproject/juriscraper/issues/1890

And assigned @Brennan-Chesley-FLP  so we can divide the workload.

[...]

PR #7177, opened 2026-03-31
feat(texas): Normalize appellate docket number before attempting lookup

MorganBennetDev:
When populating case transfer docket FKs, normalize the docket numbers for appellate dockets before attempting lookup as they are often oddly formatted.

This should solve the `CaseTransfer` docket number cleaning issue mentioned in #7145 and allow us to run the command without issue.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/corpus_importer/tasks.py
+++ b/cl/corpus_importer/tasks.py
@@ -4518,7 +4518,7 @@ def merge_texas_docket_originating_court(
             "Skipping merge of OCI for Texas docket %s due to unknown originating court type.",
             docket.docket_number,
         )
-        return MergeResult.failed()
+        return MergeResult.unnecessary(None)
 
     created = False
     if not docket.originating_court_information:
@@ -4623,7 +4623,7 @@ def merge_texas_case_transfers(
                     docket.docket_number,
                 )
 
-                return MergeResult.failed()
+                return MergeResult.unnecessary(None)
 
             logger.warning(
                 "Found Texas SC docket with originating information but no appellate information (docket number %s). Falling back to using trial court to create appeal type transfer.",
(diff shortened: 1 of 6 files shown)


cl/lib/model_helpers.py
```diff
@@ -175,6 +175,41 @@ def is_texas_court(court_id: str) -> bool:
...
+        if len(parts) != 3:
+            return docket_number
+        int_parts = []
+        for part in parts:
+            if not re.fullmatch(r"^\d+$", part):
+                return docket_number
+            int_parts.append(int(part))
+        return f"{int_parts[0]:0>2}-{int_parts[1]:0>2}-{int_parts[2]:0>5}-{suffix}"
+    joined = "".join(parts)
+    return f"{joined[0:2]}-{joined[2:4]}-{joined[4:]}-{suffix}"
```
albertisfu: I think this logic is the same, but a bit clearer.

```suggestion
    if len(parts) < 2:
        return docket_number
    suffix = parts.pop()
    if not all(part.isdigit() for part in parts):
        return docket_number
    # Three numeric parts (e.g. "09-01-404"): zero-pad to NN-NN-NNNNN
    if len(parts) == 3:
        return f"{int(parts[0]):0>2}-{int(parts[1]):0>2}-{int(parts[2]):0>5}-{suffix}"
    # Joined digits total 9 (e.g. "0199-01326"): re-slice to NN-NN-NNNNN
    joined = "".join(parts)
    if len(joined) == 9:
        return f"{joined[:2]}-{joined[2:4]}-{joined[4:]}-{suffix}"
    return docket_number
```

PR #7178, opened 2026-04-01
feat(commands): add --auto-resume and Redis logging to download commands

albertisfu:
## Fixes
N/A

## Summary
Adds `--auto-resume` flag and Redis progress logging (every 100 documents) to `download_texas_documents` and `download_scotus_pdfs` management commands. So we can ignore documents that failed to be downloaded (due to not found file error) after the command resumes.

- Each command gets a `compose_redis_key()` function and uses `log_last_document_indexed` / `get_last_parent_document_id_processed` from `cl.lib.indexing_utils`
- On `--auto-resume`, the queryset is filtered by the last pk stored in Redis, respecting `--download-order` direction

## Deployment

**This PR should:**

### DIFF
--- a/cl/corpus_importer/management/commands/download_texas_documents.py
+++ b/cl/corpus_importer/management/commands/download_texas_documents.py
@@ -3,7 +3,10 @@ from itertools import batched
 
 from django.db.models import Q
 
-from cl.corpus_importer.tasks import download_texas_document_pdf, logger
+from cl.corpus_importer.tasks import (
+    download_texas_document_pdf_unthrottled,
+    logger,
+)
 from cl.corpus_importer.utils import paginate_docs_queryset
 from cl.lib.celery_utils import CeleryThrottle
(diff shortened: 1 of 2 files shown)


albertisfu:
@MorganBennetDev I’ve pushed an additional tweak here.

I found that when scheduling `download_texas_document_pdf` tasks via the `download_texas_documents` command, we were throttling the task with `throttle_task`, which made the CeleryThrottle in the command useless. This is because `throttle_task` reschedules the task using countdown, which moves the task outside the main queue to the `unacked` queue, as described in https://github.com/freelawproject/courtlistener/issues/4077#issuecomment-2133791186
. In this case, `throttle.maybe_wait()` was mostly doing nothing. There were no Celery runaway issues because we have an additional sleep.

To make things consistent, I’ve created a `download_texas_document_pdf_unthrottled` task for use within the download command.

PR #7181, opened 2026-04-01
fix(texas): Importer fixes for #7145

MorganBennetDev:
Two fixes for errors encountered during Texas docket merging (see #7145):

1. Add None check for workload transfer origin court ID in merge_texas_case_transfers. When texas_js_court_id_to_court_id returns None for unknown courts, skip the workload transfer with a warning instead of passing None to Court.objects.get().

2. Wrap AttorneyOrganization.objects.create() in a savepoint in add_attorney. The enclosing add_parties_and_attorneys is @transaction.atomic, so an IntegrityError from a race condition on create() would abort the transaction. The recovery get() then failed with TransactionManagementError. The savepoint ensures only it rolls back, not the outer transaction.

Once all fixes have been added, I'll move this PR out of draft.

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/.github/scripts/frontend_checks.py
+++ b/.github/scripts/frontend_checks.py
@@ -22,6 +22,37 @@ FAIL = "error"
 WARN = "warning"
 
 # ---------------------------------------------------------------------------
+# Per-file skip directives
+# ---------------------------------------------------------------------------
+
+# Checks that can be skipped via {# frontend-checks-skip: ... #} comments.
+# Only advisory/context-dependent checks belong here — security, a11y, and
+# architecture checks must stay enforced.
+SKIPPABLE_CHECKS = {
+    "check_hardcoded_ids",
+    "check_new_stack_leakage",
+    "check_raw_css",
+    "check_include_in_v2",
+}
+
+_skip_directive_re = re.compile(r"\{#\s*frontend-checks-skip:\s*(.+?)\s*#\}")
+
+
+def _parse_skip_checks(lines: list[str]) -> set[str]:
+    """Parse ``{# frontend-checks-skip: ... #}`` comments.
+
+    Returns the intersection of requested skips with SKIPPABLE_CHECKS,
+    so non-allowlisted checks cannot be bypassed.
+    """
+    skip: set[str] = set()
+    for line in lines:
+        m = _skip_directive_re.search(line)
+        if m:
+            skip.update(name.strip() for name in m.group(1).split(","))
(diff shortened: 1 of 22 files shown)


MorganBennetDev:
Returned `MergeResult.unnecessary(None)` instead of `MergeResult.failed()` when skipping imports, but I'd like to add a `skipped` field to `MergeResult` after the rewrite in #6988 is merged and log that at the end of the merger method. I'd also like to make the log better than "Something failed, oops." after that's merged (since we'll have the information to do so).

cl/recap/mergers.py
```diff
@@ -332,7 +332,10 @@ def add_attorney(atty, p, d):
                 )
             except AttorneyOrganization.DoesNotExist:
                 try:
-                    org = AttorneyOrganization.objects.create(**atty_org_info)
+                    with transaction.atomic():
+                        org = AttorneyOrganization.objects.create(
+                            **atty_org_info
+                        )
```
Brennan-Chesley-FLP, line 338: This is my understanding of the fix. Looks good to me. 

```mermaid
sequenceDiagram
    participant P1 as Process 1
    participant DB as PostgreSQL
    participant P2 as Process 2

    P1->>DB: BEGIN (add_parties_and_attorneys @transaction.atomic)
    P1->>DB: AttorneyOrganization.objects.get(lookup_key="X")
    DB-->>P1: DoesNotExist

    Note over P2: Enter Villain
    P2->>DB: AttorneyOrganization.objects.create(lookup_key="X")
    DB-->>P2: OK ✓

    P1->>DB: SAVEPOINT (with transaction.atomic) ← fix
    P1->>DB: AttorneyOrganization.objects.create(lookup_key="X")
    DB-->>P1: IntegrityError (duplicate key)
    P1->>DB: ROLLBACK TO SAVEPOINT ← only savepoint is aborted
    Note over P1: Catches IntegrityError, retries get()
    P1->>DB: AttorneyOrganization.objects.get(lookup_key="X")
    DB-->>P1: OK ✓
    P1->>DB: COMMIT
```

Brennan-Chesley-FLP:
Though it conjures hilarious images of gluttony to see `unnecessary(None)` in the code and I hate to see it go for that reason, I agree that your follow up would make it a bit more obvious what's going on. Overall looks good to me.

PR #7187, opened 2026-04-03
feat(CI) add frontend lint exception

Brennan-Chesley-FLP:
## Fixes

This allows exceptions to the front end lint check, so that HTML files used for tests that won't ever be shown to our end users (hopefully we're sparing them) don't get flagged. 

## Summary

This adds a `--skip-files` flag which will accept globs and ignore matching files when running the checks. It also adds an initial glob for html within an apps `test_assets` directory, which may cover all of our bases. 

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1855,7 +1855,7 @@ wheels = [
 
 [[package]]
 name = "juriscraper"
-version = "3.0.6"
+version = "3.0.7"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "certifi" },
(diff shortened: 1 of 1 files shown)


Brennan-Chesley-FLP:
I have multiple open PRs that need this so I'm trying to get ahead of the merge conflicts.

.github/workflows/frontend-lint.yml
```diff
@@ -38,7 +38,8 @@ jobs:
             --repo-root . \
             --format github \
             --changed-files changed_files.txt \
-            --summary-file summary.md
+            --summary-file summary.md \
+            --skip-files 'cl/*/test_assets/**/*.html'
```
albertisfu: Is the `/**/` right pattern here?

I ran:

```
In [2]: fnmatch.translate('cl/*/test_assets/**/*.html')
Out[2]: '(?s:cl/(?>.*?/test_assets/)(?>.*?/).*\\.html)\\Z'
```

Which requires at least one character before the final `/`. I think this could cause files directly in `test_assets` to be ignored, just like the one in your other PR:

[‎cl/corpus_importer/test_assets/missing_file.html](https://github.com/freelawproject/courtlistener/pull/7165/changes#diff-75c73a8a331a86c8489a70e57da73ac226f50230645ebcf64de927742038374e) 


Would this pattern be better?
```
In [3]: fnmatch.translate('cl/*/test_assets/*.html')
Out[3]: '(?s:cl/(?>.*?/test_assets/).*\\.html)\\Z'
```

albertisfu (commented):
This looks good. Just a question/suggestion.

.github/workflows/frontend-lint.yml
Brennan-Chesley-FLP: Indeed! I was bash globing here. Thanks for the catch.

PR #7211, opened 2026-04-10
fix(donate): Set `.edu` membership payment status to succeeded

ERosendo:
## Summary

`.edu` membership webhooks from Neon CRM don't include a `payments` array in their payload. Our code in `_get_membership_data` returns an empty string when no payments are present, and `_map_payment_status_value` maps that to `PENDING` via its default case. Since `.edu` memberships are free, no subsequent payment webhook ever arrives to correct this, leaving these records permanently stuck as "Awaiting payment."

This PR updates `_handle_membership_creation` to check if the membership is EDU-level and the webhook had no payment info. When both are true, `payment_status` is set to `SUCCEEDED` instead of falling through to the default `PENDING`.

Existing affected records will be corrected via a script in production after this is merged.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


cl/donate/api_views.py
```diff
@@ -362,10 +362,17 @@ def _handle_membership_creation(self, webhook_data) -> None:
...
-        payment_status = self._map_payment_status_value(
-            membership_data["paymentStatus"]
-        )
+        # .edu memberships are free. If Neon didn't send payment info,
+        # treat them as succeeded rather than pending.
+        if (
+            membership_level == NeonMembershipLevel.EDU
+            and not membership_data["paymentStatus"]
+        ):
+            payment_status = MembershipPaymentStatus.SUCCEEDED
```
albertisfu: In case Neon’s logic changes in the future and it starts sending payment info with a failed status, It seems like a good idea to log an error so we can review the membership manually and update the logic accordingly.

albertisfu (commented):
Thanks @ERosendo this looks good. Just the suggestion we talked about.

albertisfu (approved):
Looks good. Thank you.

PR #7215, opened 2026-04-13
feat(api): Make API usage programmatically accessible

### LINKED ISSUE #7114: Make API usage API-accessible
### What happened?

For the MCP, it would be useful to be able to programmatically access your API usage, so you can see how much of it you have used and whether you're getting near to your usage cap. 

We should make a basic API for this so folks can pull what we've got. 

The API should also provide their access level (how many requests/s they're allowed). This should be in a list or dict of some kind since they can have multiple rates (10/day & 1/hour, or 1/s for search and 20/hour for citation lookup, etc). 

When we do, we need to update the page here with the information: 

https://wiki.free.law/c/courtlistener/help/api/faqs/how-can-i-check-my-api-usage

@nadahlberg, please let us know if you've already filed this one or if you have more to add.

ERosendo:
## Fixes

This PR fixes #7114

## Summary
This PR adds `GET /api/rest/v4/api-usage/`,  a read-only endpoint that lets authenticated users inspect their own API consumption programmatically. Primarily motivated by MCP integrations that want to monitor usage.

The response includes:
- **`current_usage`**: per-scope throttle status (requests made, remaining, reset time) sorted by utilization descending, so the limit closest to being hit appears first. Scopes are discovered from `DEFAULT_THROTTLE_RATES`, so adding a new rate limit to settings makes it visible here automatically.

- **`historical_usage`**: 14-day daily request counts plus a running total, sourced from existing Redis stats via `invert_user_logs`.

- **`membership`**: `NeonMembership` level and active status, or null.

Here's an example: 

```json
{
    "current_usage": [
        {
            "scope": "user",
            "rate": "5000/hour",
            "requests_made": 17,
            "requests_allowed": 5000,
            "requests_remaining": 4983,
            "window_duration_seconds": 3600,
            "reset_at": "2026-04-13T17:35:10.122831+00:00",
            "blocked": false
        },
        {
            "scope": "citations",
            "rate": "60/min",
            "requests_made": 0,
            "requests_allowed": 60,
            "requests_remaining": 60,
            "window_duration_seconds": 60,
            "reset_at": null,
            "blocked": false
        }
    ],
    "historical_usage": {
        "2026-04-02": 7,
        "total": 7
    },
    "membership": null
}
```

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


cl/api/views.py
```diff
@@ -388,3 +397,145 @@ async def wiki_data(request: HttpRequest) -> JsonResponse:
...
+            override = overrides.get(throttle_type) if throttle_type else None
+            if override is not None:
+                blocked, override_rate = override
+                if override_rate:
+                    rate = override_rate
+
+            num_requests, duration = parse_rate(rate)
+            cache_key = f"throttle_{scope}_{user.pk}"
+            history = cache.get(cache_key, [])
+            valid = [ts for ts in history if ts > current_time - duration]
```
albertisfu: If you make a citations request and there are entries in the citations throttle history, this will raise:

`TypeError: '>' not supported between instances of 'list' and 'float'`

This happens because the throttle history is a two-dimensional array like:

`[[1, 1777779929.627806]]`

So, we should properly extract the entries from the citations count and compute the current and remaining bucket, since the budget for citations is based on matched citations.

We might also need to rename these keys:

```
"requests_made"
"requests_allowed"
"requests_remaining"
```
to something that represents the number of matched citations instead of requests.

we should cover the described issues in a test to prevent regresessions.

albertisfu (changes requested):
Thanks @ERosendo this looks good.

I just have one comment and some additional notes here:

- `ApiUsageViewSet` is itself rate-limited by the same user throttle it reports on. This means that if the user gets throttled, they won’t be able to see their usage. For instance, if a user is using this endpoint to check `reset_at` after being throttled in order to retry requests, they won’t be able to call this endpoint once they are throttled.

  I’m not sure if this is intended behavior, but I think it could be confusing for users.
  
  Maybe we can allow some extra requests to `ApiUsageViewSet` before it gets throttled itself, so the endpoint remains available when users need it most, right after they’ve been throttled on regular endpoints.

- Also, we should update this PR with main and ensure the updated `APIThrottle` constraint takes effect in this PR: `["user", "throttle_type", "rate"]`. Currently, this PR only supports one throttle type per user, as the constraint is `["user", "throttle_type"]`.

   We can also expand the tests to cover multidimensional throttle usage reports.

cl/api/views.py
```diff
@@ -276,3 +286,55 @@ async def wiki_data(request: HttpRequest) -> JsonResponse:
...
+
+    permission_classes = [IsAuthenticated]
+    throttle_classes: list = []
+    pagination_class = None
+
+    def _get_historical_usage(self, user: User) -> dict[str, int]:
+        """14-day daily request counts from Redis."""
+        start = datetime.today() - timedelta(days=14)
+        end = datetime.today()
+        data = invert_user_logs(start, end, add_usernames=False)
```
albertisfu: This might be outside the scope of this PR; however, I wanted to point it out because now that the API usage is no longer throttled, the Redis work performed by `invert_user_logs` might become expensive.

I noticed that `invert_user_logs` pulls the counts for all users using:
```
pipe.zrange(
                f"api:{version}.user.d:{d}.counts",
                0,
                -1,
                withscores=True,
            )
```

But at the end, the invert_user_logs callers only use the data for a single user, as in the line below:

`data.get(user.pk, {"total": 0})`

What if, instead of pulling all counts for every user, we only retrieved the count for the relevant user?

`pipe.zscore(f"api:{version}.user.d:{d}.counts", user_id)`

Here's the refactor Claude suggested after asking for this performance improvement:

```
def get_user_api_usage(
    user_id: int,
    start: str | datetime,
    end: str | datetime,
) -> dict[str, int]:
    """Get one user's daily API usage counts over a date range.

    Combines v3 and v4 API counts using O(1) ZSCORE lookups against the
    per-day sorted sets that LoggingMixin populates, so the cost is
    independent of how many users hit the API.

    :param user_id: The pk of the user whose usage to look up
    :param start: Beginning date (inclusive) for the query range
    :param end: End date (inclusive) for the query range

    :return: Dictionary mapping ISO dates with usage to their counts, in
        chronological order, followed by a 'total' key. Empty if the user
        made no requests in the range.
    """
    pipe = r.pipeline()

    dates = make_date_str_list(start, end)
    versions = ["v3", "v4"]
    for d in dates:
        for version in versions:
            pipe.zscore(f"api:{version}.user.d:{d}.counts", user_id)

    # results contains alternating v3/v4 scores (or None) for each date
    # queried, e.g. [v3_day1, v4_day1, v3_day2, v4_day2, ...].
    results = pipe.execute()

    out: dict[str, int] = {}
    total = 0
    for d, scores in zip(dates, batched(results, len(versions))):
        if count := sum(int(score) for score in scores if score):
            out[d] = count
            total += count
    if out:
        out["total"] = total
    return out
```

cl/api/views.py
```diff
@@ -276,3 +286,55 @@ async def wiki_data(request: HttpRequest) -> JsonResponse:
...
+    Returns current throttle usage, 14-day historical usage, and membership
+    information.
+
+    This endpoint is intentionally not rate-limited. Users must be able to
+    inspect their current usage, including ``reset_at``, even after being
+    throttled on other endpoints, when this information is most useful.
+    """
+
+    permission_classes = [IsAuthenticated]
+    throttle_classes: list = []
```
albertisfu: Would it be crazy if we created an `ApiUsageRateThrottle` scope with a relatively high default rate per day or hour and used it here? I'm a bit concerned that, since this endpoint is unthrottled, it could become an easy target for abuse.

For example, if we defined a default rate of `60/h` for the `api_usage` scope, users could make, on average, one request per minute to monitor their API consumption or plan retries. That way, requests to this endpoint wouldn't count against their regular API rate limit, but the endpoint itself would still be protected by throttling.

We could document the default limit and possibly include it in the api-usage response so users can see their remaining `api_usage` requests and plan accordingly. If they exceed the `api_usage` limit, the endpoint would simply be throttled like any other endpoint.

cl/api/utils.py
```diff
@@ -959,6 +960,133 @@ def get_recent_api_request_count(user: User, window_seconds: int) -> int:
...
+    for rate in rates:
+        limit, duration = parse_rate(rate)
+        cutoff = now - duration
+        in_window = [(w, ts) for w, ts in weighted_history if ts > cutoff]
+        used = sum(w for w, _ in in_window)
+
+        reset_at = None
+        if in_window:
+            # Oldest in-window entry frees a slot when it exits the window.
+            oldest_ts = min(ts for _, ts in in_window)
```
albertisfu: Found an edge case that might create a conflict here. Steps to reproduce:

* User starts with a defined limit, e.g. `10/h`.
* Make the first request at `00:00`; `reset_at` is expected to be `01:00`.
* Make 9 additional requests throughout that hour.
* Decrease the user's throttle from `10/h` to `5/h` before `01:00`.
* Make another request at `01:02`. In the API usage endpoint you can see: `used=9/5, remaining=0, reset_at=01:05`, while the API endpoint rejects the request with `Retry-After: 01:45`.

What's happening here is a mismatch between the `reset_at` value and the `Retry-After` header. The `Retry-After` header is the correct one.

The current `reset_at` represents: When will the oldest request fall out of the window?

While `Retry-After` represents: When will enough requests have expired for a new request to fit under the limit?

Because of that, the logic used to compute `reset_at` should match the logic used for `Retry-After`.

As I mentioned, this isn't an obvious issue during the normal workflow, but it can become a problem if a user's membership expires or if their manual throttle is reduced.

albertisfu (changes requested):
Thanks @ERosendo this looks great! I just have a few suggestions.

cl/api/views.py
ERosendo: Not crazy at all. I just added an api_usage scope with ["10/min", "120/hour"] and an ApiUsageRateThrottle class. Because the scope has its own cache key, requests here don't count against the caller's regular API quota and the endpoint stays reachable after that quota is exhausted, which was the reason it was unthrottled before.

I went with 120/hour instead of 60/hour because a flat 60/h is exactly one request per minute with no headroom, a user polling on a 60s timer would 429 themselves on any retry or extra refresh. The 10/min companion rate allows a dashboard burst while still capping sustained traffic.

cl/api/utils.py
ERosendo: I just addressed this comment in [fcca838](https://github.com/freelawproject/courtlistener/pull/7215/commits/fcca8381d8b8e0402d637ada33ccd06b32baa65f)

cl/api/views.py
ERosendo: I like the refactor i'll introduce the helper in this PR 👍

ERosendo: I just added the refactor in [03b78e0](https://github.com/freelawproject/courtlistener/pull/7215/commits/03b78e02dd1a40efc2480ff70eb8bc18363f9f22)

cl/api/views.py
```diff
@@ -412,3 +426,53 @@ async def wiki_data(request: HttpRequest) -> JsonResponse:
...
+    Returns current throttle usage, 14-day historical usage, and membership
+    information.
+
+    This endpoint is intentionally not rate-limited. Users must be able to
+    inspect their current usage, including ``reset_at``, even after being
+    throttled on other endpoints, when this information is most useful.
+    """
+
+    permission_classes = [IsAuthenticated]
+    throttle_classes = [ApiUsageRateThrottle]
```
albertisfu, line 455: This new `ApiUsageRateThrottle` is working as expected. I just have one suggestion to expose the `api_usage` scope in the `/api/rest/v4/api-usage/` endpoint so users can see their current limits directly from that endpoint.

Then we can document what the `api_usage` scope is for.

cl/api/utils.py
```diff
@@ -1466,6 +1655,51 @@ def update_user_counts(_user_id, _count, _date):
     return user_keyed_out
 
 
+def get_user_api_usage(
```
albertisfu, line 1677: It looks like this method can also replace `invert_user_logs` in `recent_api_usage`, since it only queries a single user. After that, we may be able to remove `invert_user_logs`, but let's do that cleanup in a separate PR.

Could you please open an issue for that once this gets merged?

albertisfu (changes requested):
Thanks for the changes, @ERosendo. This looks great.

I just have one suggestion regarding the `api_usage` scope.

Also, do we have any documentation planned for this new endpoint?

cl/api/views.py
ERosendo: addressed in [9858b60](https://github.com/freelawproject/courtlistener/pull/7215/commits/9858b608b4b79c8a3647c3a6517d392dd03aae34)

ERosendo:
@albertisfu Ready for another review!

albertisfu (approved):
Thanks @ERosendo this looks ready to be merged!

mlissner:
Reopening to restart CI — GitHub Actions didn't trigger on the latest push.

mlissner:
I went ahead and merged this without waiting for tests since GH CI is broken today.

PR #7225, opened 2026-04-16
Add OAuth app with DCR views

### LINKED ISSUE #7224: Add OAuth to CourtListener
### Describe the issue.

To integrate OAuth into CL we need to:
- Install Django OAuth Toolkit
- Create a new app for OAuth-related views
- Add a custom view for OAuth metadata
- Add a custom view for dynamic client registration

The Django OAuth Toolkit (DOT) will give us the following endpoints:

| Endpoint | What it does | Where it's used in the flow (see below) |
|---|---|---|
| `/o/authorize/` | Auth-code grant entry point; shows the consent screen | Step 7 (browser redirect) |
| `/o/token/` | Exchanges authorization codes and refresh tokens for access tokens | Step 9, step 15 (refresh) |
| `/o/revoke_token/` | RFC 7009 revocation endpoint | Not used yet |
| `/o/introspect/` | RFC 7662 token introspection | Not used in by our MCP |
| `/o/.well-known/jwks.json` | JWKS — public half of the OIDC signing key | Not used in by our MCP |
| Admin UI (`/o/applications/`, `/o/authorized_tokens/`) | User-facing CRUD for managing their own clients / tokens | Not in the Claude flow; nice for humans |

And our custom endpoints will give us these:

| Endpoint | Why custom | Notes |
|---|---|---|
| `/o/register/` | DOT doesn't ship RFC 7591 Dynamic Client Registration. | Needed because MCP clients (Claude, etc.) register themselves on the fly — you can't preregister every possible client. |
| `/.well-known/oauth-authorization-server` | DOT ships `/o/.well-known/openid-configuration` (OIDC discovery), but not the RFC 8414 OAuth authorization-server metadata at the root path that MCP clients look for. | This is the view that advertises `registration_endpoint`, `authorization_endpoint`, etc. as a flat JSON doc. |

Here is the flow of between MCP Clients, the MCP server, and CourtListener:

```
      Claude                   MCP                    CL (Django)
        │                       │                           │
   1 ───┤ POST /                                            
        │────────────────────▶  │                           │
   3    │ ◀──── 401 + WWW-Authenticate (resource_metadata)  │
   4    │ GET /.well-known/oauth-protected-resource        
        │────────────────────▶  │                           │
        │ ◀──── authorization_servers: [www.courtlistener.com]
   5    │ GET /.well-known/oauth-authorization-server       │
        │──────────────────────────────────────────────────▶│
        │ ◀─────── issuer + endpoint URLs ──────────────────│
   6    │ POST /o/register/ (DCR)                           │
        │──────────────────────────────────────────────────▶│
        │ ◀───────── client_id + client_secret ─────────────│
   7-9  │ (browser: authorize → consent → redirect → code → exchange)
        │ ◀─────── access_token + refresh_token ────────────│
  10    │ POST / Authorization: Bearer <token>              │
        │────────────────────▶  │                           │
  12                            │ GET /api/rest/v4/...      │
                                │ Authorization: Bearer ───▶│
                                │ ◀────── tool result ──────│
  14    │ ◀──── MCP response ── │                           │
```

### One-time connector setup

**1. User pastes URL into Claude.**
User adds `https://mcp.courtlistener.com` as a connector in Claude.

**2. Claude → MCP: blind POST.**
Claude tries to initialize the MCP session with no token.
```
POST https://mcp.courtlistener.com/
```

**3. MCP → Claude: 401 with discovery hint.**
FastMCP's auth middleware (the `RemoteAuthProvider` wrapping our `PassThroughTokenVerifier`) sees no bearer and returns:
```
401 Unauthorized
WWW-Authenticate: Bearer error="invalid_token",
                  resource_metadata="https://mcp.courtlistener.com/.well-known/oauth-protected-resource"
```

**4. Claude → MCP: fetch protected-resource metadata.**
```
GET https://mcp.courtlistener.com/.well-known/oauth-protected-resource
→ { "resource": "https://mcp.courtlistener.com/",
    "authorization_servers": ["https://www.courtlistener.com/"],
    "bearer_methods_supported": ["header"] }
```
Now Claude knows: to get a token, go talk to CL.

**5. Claude → CL: fetch authorization-server metadata.**
```
GET https://www.courtlistener.com/.well-known/oauth-authorization-server
→ { "issuer": "https://www.courtlistener.com",
    "registration_endpoint": ".../o/register/",
    "authorization_endpoint": ".../o/authorize/",
    "token_endpoint": ".../o/token/",
    "jwks_uri": ".../o/.well-known/jwks.json",
    ... }
```
Claude now has the full set of endpoints to talk to.

**6. Claude → CL: Dynamic Client Registration (RFC 7591).**
Claude has never registered itself with CL before, so it does it on the fly.
```
POST https://www.courtlistener.com/o/register/
{ "client_name": "Claude",
  "redirect_uris": ["http://127.0.0.1:<ephemeral>/callback"] }
```
Your DCR view creates an `Application` row, returns `{client_id, client_secret}`. Claude stores them.

**7. Claude opens the user's browser at the authorization endpoint (PKCE).**
```
https://www.courtlistener.com/o/authorize/
  ?response_type=code
  &client_id=<from step 6>
  &redirect_uri=http://127.0.0.1:<ephemeral>/callback
  &code_challenge=<SHA256 hash of a random verifier>
  &code_challenge_method=S256
  &state=<random>
```

**8. User logs into CL and clicks "Allow".**
CL redirects the browser back to the ephemeral callback with `?code=<auth_code>`. Claude's local callback server captures it.

**9. Claude → CL: exchange code for tokens.**
```
POST https://www.courtlistener.com/o/token/
grant_type=authorization_code
code=<from step 8>
code_verifier=<the plaintext whose SHA256 matches step 7's challenge>
client_id=...
client_secret=...
→ { "access_token": "<30-char opaque>",
    "refresh_token": "<30-char opaque>",
    "expires_in": 3600,
    "scope": "api",
    "token_type": "Bearer" }
```

### Making tool calls

**10. Claude → MCP: retry the original call with the bearer.**
```
POST https://mcp.courtlistener.com/
Authorization: Bearer <access_token>
```

**11. MCP: `PassThroughTokenVerifier.verify_token` accepts the non-empty token**, builds an `AccessToken` object, stashes it on the request. The tool dispatch middleware runs.

**12. MCP tool code forwards the bearer to CL's API.**
In `MCPTool.get_client()`, `get_access_token()` returns the `AccessToken` we stashed. That gets wrapped into a `CourtListener(access_token=...)` client, which sends:
```
GET https://www.courtlistener.com/api/rest/v4/courts/scotus/
Authorization: Bearer <access_token>
```

**13. CL's DRF `OAuth2Authentication` validates the token** — looks it up in `oauth2_provider_accesstoken`, checks expiry, resolves to the user that consented in step 8, authorizes the view. If valid, returns the data. If not, 401.

**14. Data flows back: CL → MCP → Claude → user.**

### When the access token expires (~1 hour later)

**15. Claude → CL: proactively refresh** (before sending the next MCP request, if `expires_in` says the token is close to expiring):
```
POST https://www.courtlistener.com/o/token/
grant_type=refresh_token
refresh_token=<from step 9>
client_id=... client_secret=...
→ { "access_token": "<new opaque>", "refresh_token": "<new>", ... }
```
Claude stores the new tokens and continues using them. Loop back to step 10. No user interaction.
After step 9, every subsequent tool call is just steps 10→12→13→14 plus the occasional refresh from step 15.

### Related Issue

_No response_

### Screenshots

_No response_

Task

nadahlberg:
## Fixes

Fixes #7224

## Summary

This PR implements the OAuth flow on the CourtListener side described in #7224. It installs Django OAuth Toolkit and adds an `oauth` app with two additional views for OAuth metadata and dynamic client registration.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [X] `skip-celery-deploy`
    
    
    - [X] `skip-cronjob-deploy`
    
    
    - [X] `skip-daemon-deploy`

1. We need to add a new secret `OIDC_RSA_PRIVATE_KEY`, generated with `openssl genrsa 4096`
2. The `oauth2_provider` installed app has models that need to be migrated.


nadahlberg:
@mlissner @ERosendo a few questions about this:

1. I added `ReplicaRoutingOAuth2Authentication` to follow the pattern for the other authentication classes. But I really don't understand what these ReplicaRouting classes do. Was this the correct move?
2. The `oauth_provider` plugin has built-in models that need migrations. Do we still need to generate `.sql` and `_customers.sql` files for these?
3. Should we make a separate issue for adding the `OIDC_RSA_PRIVATE_KEY` secret? We raise in production if it's not present, so that needs to be added first.

cl/urls.py
```diff
@@ -55,6 +55,8 @@
     path("", include("cl.api.urls")),
     path("", include("cl.visualizations.urls")),
     path("", include("cl.stats.urls")),
+    path("", include("cl.oauth.urls")),
+    path("o/", include("oauth2_provider.urls", namespace="oauth2_provider")),
```
ERosendo: I was reviewing the [getting started](https://django-oauth-toolkit.readthedocs.io/en/latest/getting_started.html) documentation for [django-oauth-toolkit](https://django-oauth-toolkit.readthedocs.io/en/latest/index.html) and noticed that when we register these URLs, it also adds an endpoint for application registration (`/o/applications/register/`), which renders the following form:

Image

I double-checked your plan for integrating the MCP, and it seems like this registration flow won’t be used. Do we actually need to include it?

I’d prefer to keep the scope limited to just what we need for the MCP, so I wanted to check before moving forward.

ERosendo: Also, since we have a dedicated app for OAuth, we could move the registration of these views into that app and keep all OAuth-related URLs in a single urls.py file.

cl/oauth/views.py
```diff
@@ -0,0 +1,255 @@
...
+                    "refresh_token",
+                ],
+                "token_endpoint_auth_methods_supported": [
+                    "client_secret_basic",
+                    "client_secret_post",
+                    "none",
+                ],
+                "code_challenge_methods_supported": ["S256"],
+                "scopes_supported": ["api"],
+                "service_documentation": (base + "/help/api/rest/"),
```
ERosendo:
```suggestion
                "service_documentation": (base + reverse("rest_docs")),
```

cl/oauth/views.py
```diff
@@ -0,0 +1,255 @@
...
+                "response_types_supported": ["code"],
+                "grant_types_supported": [
+                    "authorization_code",
+                    "refresh_token",
+                ],
+                "token_endpoint_auth_methods_supported": [
+                    "client_secret_basic",
+                    "client_secret_post",
+                    "none",
+                ],
```
ERosendo: I guess we can reuse constants defined at the top of the file: 


```suggestion
                "grant_types_supported": ALLOWED_GRANT_TYPES,
                "token_endpoint_auth_methods_supported": ALLOWED_TOKEN_AUTH_METHODS,
```

cl/oauth/views.py
```diff
@@ -0,0 +1,255 @@
...
+        {
+            "error": "invalid_client_metadata",
+            "error_description": description,
+        },
+        status=status.HTTP_400_BAD_REQUEST,
+    )
+
+
+@method_decorator(
+    ratelimit(key="ip", rate="10/h", block=True, method="POST"),
```
ERosendo: We’ve been using get_ip_for_ratelimiter instead of just "ip". You can find other examples in ratelimiter.py:

https://github.com/freelawproject/courtlistener/blob/main/cl/lib/ratelimiter.py

cl/oauth/views.py
```diff
@@ -0,0 +1,255 @@
...
+            ALLOWED_GRANT_TYPES
+        ):
+            return _invalid_metadata(
+                "grant_types must be a subset of "
+                f"{sorted(ALLOWED_GRANT_TYPES)}"
+            )
+        if "authorization_code" not in grant_types:
+            return _invalid_metadata(
+                "authorization_code grant type is required"
+            )
```
ERosendo: We should move this validation logic into a serializer to keep the view cleaner. I took a pass at it in a separate branch since the diff is fairly large:

https://github.com/freelawproject/courtlistener/compare/claude/oauth-dynamic-client-registration-IJWzE...oauth-dynamic-client-registration-serializer?expand=1

(The diff is just for reference and I used Claude to help with the refactor 😅)

ERosendo:
> 1. I added ReplicaRoutingOAuth2Authentication to follow the pattern for the other authentication classes. But I really don't understand what these ReplicaRouting classes do. Was this the correct move?

Yes, this is the right move. We previously saw that some API endpoints were not correctly routing requests to read replicas, and the fix was to wrap the authentication classes with the `ReplicaRouting` variants. I believe the same applies here for `OAuth2Authentication`, so adding ReplicaRoutingOAuth2Authentication is consistent with the pattern we’ve been using.

> 2. The oauth_provider plugin has built-in models that need migrations. Do we still need to generate .sql and _customers.sql files for these?

In our setup, we usually rely on generated SQL files for clients to apply schema changes consistently, and we still include table creation even for tables that aren’t actually replicated, mainly to keep schemas aligned.

In this case, since we don’t have control of the migration chain from the dependency, we still need to make sure the resulting tables are created in our replicas. Because of that, I think a single consolidated SQL file should be fine. The goal isn’t to mirror each migration step individually, but to make sure the final set of tables is consistent across production and client/replicas.

> 3. Should we make a separate issue for adding the OIDC_RSA_PRIVATE_KEY secret? We raise in production if it's not present, so that needs to be added first.

If adding the secret requires extra steps, validations, or coordination, then it probably makes sense to create a separate issue so it can be handled properly. However, if it’s just a matter of generating it (e.g., openssl genrsa 4096) and adding it to our env, then we should be fine to include it in this PR. In that case, we just need to make sure we don’t merge until the secret is actually available in our production env.

ERosendo (changes requested):

mlissner:
I only have time to take a quick look, but this seems pretty good to me structurally. I think we just want to make sure that: 

1. It works.
2. This doesn't have any security concerns. 

For 2, it might be a good idea to run the security check claude skill on this branch?

mlissner:
@claude review once

cl/oauth/views.py
```diff
@@ -0,0 +1,165 @@
...
+                "revocation_endpoint": base + reverse("oauth2_revoke_token"),
+                "introspection_endpoint": base + reverse("oauth2_introspect"),
+                "jwks_uri": base + reverse("oauth2_jwks"),
+                "response_types_supported": sorted(ALLOWED_RESPONSE_TYPES),
+                "grant_types_supported": sorted(ALLOWED_GRANT_TYPES),
+                "token_endpoint_auth_methods_supported": sorted(
+                    ALLOWED_TOKEN_AUTH_METHODS
+                ),
+                "code_challenge_methods_supported": ["S256"],
+                "scopes_supported": ["api"],
```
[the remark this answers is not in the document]
mlissner: I haven't seen so much Greek in a programming context in a long time. Wow, I can't make sense of this!

nadahlberg:
@ERosendo a bunch of questions to make sure I'm generating the sql files correctly:

1. You mentioned creating a single consolidated sql file. Would it make sense to stick with per-migration files just in case the plugin ever gets an update (so we know where we left off)? On that note, I will also pin the version of `django-oauth-toolkit==3.2.0` so we don't get an accidental update.

2. Here's what I've done so far. First get the list of migrations with `docker exec cl-django ls /opt/venv/lib/python3.13/site-packages/oauth2_provider/migrations/`:

```bash
0001_initial.py
0002_auto_20190406_1805.py
0003_auto_20201211_1314.py
0004_auto_20200902_2022.py
0005_auto_20211222_2352.py
0006_alter_application_client_secret.py
0007_application_post_logout_redirect_uris.py
0008_alter_accesstoken_token.py
0009_add_hash_client_secret.py
0010_application_allowed_origins.py
0011_refreshtoken_token_family.py
0012_add_token_checksum.py
0013_alter_application_authorization_grant_type_device.py
```

Then I ran `docker exec cl-django python manage.py sqlmigrate oauth2_provider 0001_initial 2>/dev/null > cl/oauth/migrations/0001_initial.sql` for each file.

3. That's all I've done so far. Do we want to clean these up, simplifying things like `DEFAULT '' ... DROP DEFAULT` or is there something different about this scenario?

4. Do we need separate `_customer.sql` files? I didn't see any `event` or `trigger` lines in the sql.

5. Migrations `0006` and `0012` include lines that say `Raw Python operation ... THIS OPERATION CANNOT BE WRITTEN AS SQL`. Looking at the Python migrations it seems like these we're just data backfill operations that I think we can ignore. Just wanted to flag this.

cl/oauth/views.py
```diff
@@ -0,0 +1,165 @@
...
+            "client_name": client_name,
+            "redirect_uris": data["redirect_uris"],
+            "grant_types": data["grant_types"],
+            "response_types": data["response_types"],
+            "token_endpoint_auth_method": token_endpoint_auth_method,
+            "client_id_issued_at": int(timezone.now().timestamp()),
+        }
+        if client_type == Application.CLIENT_CONFIDENTIAL:
+            response_data["client_secret"] = client_secret_plaintext
+        return Response(response_data, status=status.HTTP_201_CREATED)
```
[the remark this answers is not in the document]
nadahlberg: Added set client_secret_expires_at = 0 and a corresponding assert in tests.

mlissner:
@nadahlberg, you ask great questions. When you have it all figured out, can you please update the wiki too for those who come later (and for your future self)?

cl/settings/third_party/oauth2_provider.py
```diff
@@ -0,0 +1,33 @@
...
+        "api": "Access the CourtListener API on your behalf",
+        "openid": "OpenID Connect identity",
+    },
+    "DEFAULT_SCOPES": ["api"],
+    "PKCE_REQUIRED": True,
+    "ACCESS_TOKEN_EXPIRE_SECONDS": 60 * 60,  # 1 hour
+    "REFRESH_TOKEN_EXPIRE_SECONDS": 60 * 60 * 24 * 30,  # 30 days
+    "OIDC_ENABLED": bool(OIDC_RSA_PRIVATE_KEY),
+    "OIDC_RSA_PRIVATE_KEY": OIDC_RSA_PRIVATE_KEY,
+    "ALLOWED_REDIRECT_URI_SCHEMES": ["https", "http"],
```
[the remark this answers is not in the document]
nadahlberg: We block this in the DCR view and no longer serve the DOT register view, so this could only happen if someone created an app with an http redirect uri in the admin site. But I moved the validation logic from the DCR view to a pre-save signal, so these applications should get blocked no matter how they are created.

cl/oauth/views.py
```diff
@@ -0,0 +1,165 @@
...
+                "registration_endpoint": base + reverse("oauth2_dcr"),
+                "revocation_endpoint": base + reverse("oauth2_revoke_token"),
+                "introspection_endpoint": base + reverse("oauth2_introspect"),
+                "jwks_uri": base + reverse("oauth2_jwks"),
+                "response_types_supported": sorted(ALLOWED_RESPONSE_TYPES),
+                "grant_types_supported": sorted(ALLOWED_GRANT_TYPES),
+                "token_endpoint_auth_methods_supported": sorted(
+                    ALLOWED_TOKEN_AUTH_METHODS
+                ),
+                "code_challenge_methods_supported": ["S256"],
```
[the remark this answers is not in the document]
nadahlberg: Added a guard by patching `AuthorizationCodeGrant._code_challenge_methods`

ERosendo:
> You mentioned creating a single consolidated sql file. Would it make sense to stick with per-migration files just in case the plugin ever gets an update (so we know where we left off)? On that note, I will also pin the version of django-oauth-toolkit==3.2.0 so we don't get an accidental update.

That’s a great idea, pinning the dependency definitely helps avoid surprises. I’m also good with keeping per-migration files if that makes it easier to track changes over time.

> That's all I've done so far. Do we want to clean these up, simplifying things like DEFAULT '' ... DROP DEFAULT or is there something different about this scenario?

We usually don’t tweak the generated SQL much, aside from removing triggers or events when needed for clients. Otherwise, we usually keep it as-is.

> Do we need separate _customer.sql files? I didn't see any event or trigger lines in the sql.

Not really. If there’s nothing to remove for clients and the migrations match what we use for the replica, we just go with a single `.sql` file.

> Migrations 0006 and 0012 include lines that say Raw Python operation ... THIS OPERATION CANNOT BE WRITTEN AS SQL. Looking at the Python migrations it seems like these we're just data backfill operations that I think we can ignore. Just wanted to flag this.

I took another look, and we should try to generate SQL for those. Migration `0006` includes an AlterField, and `0012` adds the token_checksum field along with a couple of AlterField operations, so there’s still relevant schema work there we don’t want to miss.

ERosendo (approved):
LGTM :shipit: 

@chaco-fl we’ll need to apply the migrations from this PR to both clients and the replica. No data replication is required.

chaco-fl:
Customer RDS Updated

Screenshot 2026-04-21 at 13 37 26

chaco-fl:
CL Replica updated

Screenshot 2026-04-21 at 13 39 07

chaco-fl:
Done

mlissner:
@chaco-fl, you're killing it on the migrations. Thank you!

PR #7169, opened 2026-03-31
feat(components): add c-button component, migrate v2 buttons

elisa-a-v:
## Summary
Closes #7063.

- New `c-button` Cotton component with 5 variants (primary, outline, dark, ghost, grey), 2 sizes (default/xl), element flexibility (`<button>` or `<a>` via `href`), and optional touch target mode for dialog triggers
- Uses `{% with %}` pattern to define all Tailwind classes as complete strings visible to Tailwind's class scanner — no dynamic class construction
- When `variant` is omitted, no built-in styling is applied — caller controls all classes
- Migrates all buttons in cotton and search templates (11 files)
- Adds component library entry with demos for all variants, sizes, links, disabled states, and custom styling

**Not migrated**: v2 help pages (likely to be dropped) and `docket_alert_new.html` still use the CSS `btn-*` classes, which are kept for backwards compat.

### Note on hover states
The `btn-outline` hover (`hover:bg-greyscale-400`) was introduced as a contrast fix but was implemented without design input. The component restores the original designed hover (`hover:bg-greyscale-50`). That said, none of the current button hover states are fully contrast-compliant — this is a known gap that should be addressed with proper design review.

## Test plan
- [ ] `/components/` page: button section renders all variants/sizes/states correctly
- [ ] Homepage: search bar, advanced filters button, login/signup links
- [ ] Header (signed in): profile dropdown button hover correct
- [ ] Header (signed out): Login and Sign Up buttons
- [ ] Footer: Donate Now and Subscribe buttons
- [ ] Dialog trigger buttons (mobile search icon) still work with touch target
- [ ] Support plea banner button
- [ ] No visual regressions on any v2 page

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7227, opened 2026-04-16
feature(docket): Add pagination and filters components for the docket page

### LINKED ISSUE #7069: Implement: Docket entries filter & pagination bar
## Summary

Implement the filter bar and pagination controls for the docket entries tab. In the design these are part of the same toolbar above the entries, and pagination also appears below the entries. Should be added to the component library for proper documentation.

### Filter controls
- Search input for searching within the docket
- Date range pickers (Filed from/to)
- Document number range (from/to)
- Sort controls (Asc/Desc)
- Mobile: renders as a slide-in filter panel with an Apply button

### Pagination
- First/Last/Previous/Next controls in both the top toolbar and below the entries
- Responsive: tweak which buttons are visible depending on viewport width (e.g., hide First/Last labels on smaller screens, show icons only). All buttons should have `title` attributes so hovering reveals the full label.
- Disabled state when on first/last page

### General
- Accessible: proper form labels, `aria-label`s, keyboard navigation
- Add to the component library page (`v2_components.html`) with demo, props, and code example

## Open questions

- Should we display the current page range and total count? The current design does not include this but it could be useful.

## Figma

- [Filter bar design](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=7537-54722&m=dev)
- [Pagination design](https://www.figma.com/design/OEDp7KwNSpJDyJRiKrJmm4/Court-Listener?node-id=225-7288&m=dev)

## Design references

- [#5472](https://github.com/freelawproject/courtlistener/issues/5472) — Docket Page - Docket Entries Tab
- [#5902](https://github.com/freelawproject/courtlistener/issues/5902) — Design: Mobile for Docket Tabs

EstiShay:
## Fixes
<!-- What bugs does this fix? Use this syntax to auto-close the issue: -->
<!-- https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword -->
<!-- E.g.: "Fixes: #XYZ" -->
  Partially resolves #7069  

## Summary                                                                                                                                                                                                   
  - Add `<c-pagination>`, `<c-docket-filter>`, and `<c-filter-drawer>` components for the docket entries tab
  - Refactor all three components to use `<c-button>` after #7169 merged                                                                                                                                         
  - Add component library entries with demos and prop docs
  - A11y improvements: search inputs wrapped in own forms (Enter now submits search, not filters), focus returns to trigger on drawer close, descriptive labels on all filter fields, removed redundant ARIA semantics from sort toggle                                                                                                                                                                                     
                                                                                                                                                                                                                 
  These are standalone components not yet wired into the docket page — that work moves to #7068 alongside the entry row implementation.      

---
### Update (2026-05-07)
The wrapper template at `cl/opinion_page/test_assets/docket_filter_attr_propagation.html` lives outside Django's template search path and cotton's loader, so r`ender_to_string("tests/...")` no longer finds it. The test now reads the file from disk and runs `CottonCompiler().process(...)` explicitly before passing the result to `engines["django"].from_string(...)` — that reproduces the compile step cotton's loader would have done, so `<c-docket-filter>` still expands. Follow-up commit adds `encoding="utf-8"` on the open() call to satisfy a Semgrep finding.                                                                                    
                                                                                                                                                                   
  Trade-off worth flagging: the test now imports `django_cotton.compiler_regex.CottonCompiler,` which is a third-party internal. If Cotton restructures that module,  the test breaks. 

---

 ### Update (2026-05-06)                                                                                                                                        Addresses Elisa's CSP feedback on the mobile filter drawer.                                                                                                      
  
  - The inline `x-init="if (...) open()"` tripped Alpine's CSP build (`if` statements aren't allowed in expressions). Replaced with an `x-on:open-filter-drawer` listener; `docket_filter.js` now dispatches the open event when it sees `data-has-errors` in scope, per Elisa's suggestion to keep the logic in JS rather than a new file.                                                                                                                                                        
  - Discovered a related silent bug while fixing the CSP issue: `data-has-errors` was never reaching the rendered drawer at all (no `{{ attrs }}` passthrough in `cotton/filter_drawer.html`, and Cotton's tag-attribute parser doesn't handle `{% if %}` inside a `<c-…>` tag). The auto-open-on-error feature has likely never worked in production. Fixed by exposing `has_errors` as a c-var on `c-filter-drawer` and rendering `data-has-errors` conditionally inside the component.
  - Added `DocketFilterDrawerAttrPropagationTest` to lock in both pieces of plumbing — the flag reaching the drawer's root, and the `open-filter-drawer` listener wired to the same element.         
 
---

### Update 5/1: Feedback changes and other tweaks

**Dialog composable reuse**                           
Rewrote `<c-filter-drawer>` to use `x-data="dialog"` and the `x-dialog` / `x-dialog:overlay` / `x-dialog:panel` / `x-dialog:title` directives instead of a custom Alpine component. This gets us focus trap anchored on the panel, inert on background content, `aria-labelledby` wired to the title, and automatic focus restoration to the trigger on close — all from the existing infrastructure. Deleted `filter_drawer.js` entirely; the only custom behavior (auto-open on data-has-errors) is now a one-liner in `x-init`.                                                                                                                             
                                                                                                                                                                                                                 
Added slide/fade transitions. The overlay fades and the panel slides in from the right (200ms in, 150ms out).                                                                            

**DTL binding** 
Replaced `x-bind:aria-label="'{{ title }}'"` with plain `aria-label="{{ title }}"` on the drawer since title is a server-side DTL variable, not a JS one.

**Other a11y fixes**                                                                                                                                                                                              
  - Added `aria-label="Search"` to the icon-only search buttons (desktop + mobile)                                                                                                                                 
  - Added title to disabled prev/next buttons for parity with enabled variants                                                                                                                                   
                                                                                                                                                                                                               
**Housekeeping**                                                                                                                                                                                                   
  - Deduplicated the "Page shells" nav entry in the component library (merge artifact)
  - Synced uv.lock with main (juriscraper 3.0.14 → 3.0.15)

**On the Enter-key UX concern:**
The search input and the filter controls can't submit together because they're fundamentally different actions — search navigates away to RECAP search results (`action="/"` with `type=r`), while the filters refine the docket entries list on the current page (`action=""`). So the sibling form structure matches the actual behavior: Enter in the search field submits the search, and filter changes are applied via the dedicated submit button. Open to discussing this further.

---                                                                    
                                                                                                                                                                                                               
  ## Test plan
  - [ ] Visit `/components/` and verify pagination, docket filter, and filter drawer demos render correctly
  - [ ] Test filter drawer: opens/closes on button click, escape key, overlay click, and viewport resize past `md`                                                                                               
  - [ ] Test filter drawer: focus returns to trigger button after close (via close button, escape, or overlay click)
  - [ ] Test pagination: disabled states render with `aria-disabled`, enabled buttons link correctly                                                                                                             
  - [ ] Keyboard/screen reader: filter inputs labeled, drawer traps focus, pagination buttons have `title` attributes                                                                                            
  - [ ] Test search: typing in search field and pressing Enter navigates to RECAP search, not the filter form                                                                                   
                                                                                                                                                                                                               
## Deployment

**This PR should:**
<!-- The following labels control the deployment of this PR if they’re applied. -->
<!-- Please put an "X" in the box on ones that apply. -->
<!-- For more details on what pods are affected by each label, see the wiki -->
<!-- https://github.com/freelawproject/courtlistener/wiki/Pull-requests-%60skip%E2%80%90%7Btype%7D%E2%80%90deploy%60-labels -->

<!-- Check here if the entire deployment can be skipped -->
<!-- This might be the case for a small fix, a tweak to documentation or something like that. -->
    <!-- Check here if the web tier can be skipped -->
    <!-- This is the case if you're working on code that doesn't affect the front end, like management commands, tasks, or documentation. -->
    - [ ] `skip-web-deploy`
    <!-- Check here if the deployment to celery can be skipped -->
    <!--This is the case if you make no changes to tasks.py or the code that tasks rely on. -->
    - [x] `skip-celery-deploy`
    <!-- check this if deployment to cron jobs can be skipped -->
    <!-- This is the case if no changes are made that affect cronjobs. -->
    - [x] `skip-cronjob-deploy`
    <!-- Deployment of daemons can be skipped -->
    <!-- This is the case if you haven't updated daemons or the code they depend on. -->
    - [x] `skip-daemon-deploy`

<!-- DELETE this section if your PR doesn't require screenshots. -->
<!-- If this changes the front end, please include desktop and mobile screenshots or videos showing the new feature. -->
<details closed>
<details open>
<img width="792" height="546" alt="image" src="https://github.com/user-attachments/assets/03e89c4c-ab3a-40f1-ad45-7d849828f132" />
<img width="1094" height="864" alt="image" src="https://github.com/user-attachments/assets/136837de-3d28-4cd6-b62b-aa4731977c1d" />

</details>
<details open>
<img width="628" height="865" alt="image" src="https://github.com/user-attachments/assets/625d3a21-5ccd-4010-9f9d-64025fb4c8c2" />
<img width="628" height="865" alt="image" src="https://github.com/user-attachments/assets/28ddd36b-26b3-41dd-9a3e-2904055f20f0" />

</details>
<!-- END DELETE -->

<!-- Thank you for contributing and filling out this form! -->


cl/assets/templates/cotton/filter_drawer.html
```diff
@@ -0,0 +1,74 @@
...
+    {{ title }}
+  </c-button>
+
+  {# Drawer panel #}
+  <div
+    x-show="isOpen"
+    x-bind:id="drawerId"
+    role="dialog"
+    aria-modal="true"
+    x-bind:aria-label="'{{ title }}'"
```
elisa-a-v: `x-bind` is meant for bindings to JS variables, but `title` is in DTL. We can simplify this:

```suggestion
    aria-label="{{ title }}"
```

cl/opinion_page/templates/cotton/docket_filter.html
```diff
@@ -0,0 +1,297 @@
...
+      hx-swap="innerHTML"
+      hx-push-url="true"
+    {% endif %}
+  >
+    {% if form.errors %}
+      <p class="text-sm text-red-700 mb-2">There were errors applying your filters.</p>
+    {% endif %}
+
+    <div class="flex flex-wrap items-end gap-3">
+      {# Search — own form so Enter submits search, not filters #}
```
elisa-a-v: I see two issues with this, one in HTML validity, and one in UX:

1. Nested forms aren't valid HTML and browsers handle them differently (some even drop the inner form entirely)
2. Not submitting filters when pressing enter is confusing to me: if I change the filters and then change the input and press enter, I'd expect both the new search and the filters to be applied. Why would we split them?

This is the main blocker given the validity issue, and depending on what we settle for UX the fix could be either getting rid of the inner `form` (enter submits the whole thing), or using sibling forms and add a `form` attribute to the `input` (that way we can keep enter handling separate, since inputs can be associated with any form in the document by referencing its ID)

cl/assets/templates/cotton/filter_drawer.html
elisa-a-v, line 1: Could we use `x-dialog` here like `<c-dialog>` does? A drawer is essentially a side-positioned dialog, and we'd get a few things for free:

- Focus trap anchored to the right element (right now `x-trap` is on the panel but `role="dialog"` is on the outer wrapper)
- Focus restore to the trigger on close (not currently handled)
- Background `inert` so screen readers can't navigate to underlying content while the drawer is "modal"
- `x-dialog:title` / `x-dialog:description` auto-wire `aria-labelledby` / `aria-describedby`

Bigger question: could this just extend `<c-dialog>` with drawer-specific panel classes + a slide transition? That'd unify the two and let most of `filter_drawer.js` go away.

elisa-a-v (changes requested):
We're on the right track here, and I like the component library entries.

I just have one blocker, one minor comment, and one suggestion, left them all inline.

cl/assets/templates/cotton/filter_drawer.html
elisa-a-v: About the bigger question, I quickly tested unifying with `<c-dialog>` locally and it works well, so I opened #7318 so we can land this. Not worth blocking this PR any longer for this, but I think it's worth unifying so all the work of handling dialog behavior is kept in the dialog component only.

cl/assets/templates/cotton/filter_drawer.html
```diff
@@ -0,0 +1,91 @@
...
+
+<c-vars
+  title="Filters"
+  class=""
+/>
+
+<div
+  x-data="dialog"
+  x-id="dialogIdGroup"
+  x-init="if ($el.dataset.hasErrors !== undefined) open()"
```
elisa-a-v: This is the only blocker because it's not allowed by the CSP build.

Image

It's probably okay if this lives in the `docket_filter.js` file, I wouldn't add an entire new file just for this 🤔 what do you think?

elisa-a-v (changes requested):
Solid round of updates. Structural form fix is correct, and the `x-dialog` refactor is pretty nice: deleting `filter_drawer.js` while gaining a11y wins is the right outcome.

About the Enter-key UX point, I think it's worth revisiting whether that behavior is indeed ideal or if that was just the fastest path to enable users to search for docket entries; maybe the ideal UX was always to filter the entries in the same page. However, I wouldn't block this PR any longer for that as it seems it would now land in "UX improvement" territory, instead of baseline features we need to launch.

This is almost ready to be merged, I just have one blocker left inline.

cl/simple_pages/templates/v2_components.html
```diff
@@ -204,7 +207,7 @@ <h4 class="mt-3">Demo</h4>
       paragraph_text="If you use PACER, consider installing the RECAP extension in your browser. If you are a researcher, you can search millions of federal filings in the RECAP Archive. Journalists and court watchers will find docket alerts invaluable. If you get notifications from PACER, check out our NEF notification handler."
       :buttons="[
           {
-            'variant': 'primary',
+            'style': 'btn-primary',
```
elisa-a-v: 👀 (here and below)

cl/opinion_page/templates/tests/docket_filter_attr_propagation.html
```diff
@@ -0,0 +1,3 @@
+{# frontend-checks-skip: check_new_stack_leakage #}
```
elisa-a-v: If we move this file to `test_assets` from the `tests` directory, we can skip _all_ checks in this file, which makes sense to me since it's not user-facing.

```suggestion
```

elisa-a-v (changes requested):

elisa-a-v (approved):
Thanks for the updates, looks great!

PR #7234, opened 2026-04-20
Security: API auth token is embedded into documentation pages

tomaioo:
## Summary

Security: API auth token is embedded into documentation pages

## Problem

**Severity**: `Medium` | **File**: `cl/api/templates/cotton/docket_endpoint.html:L10`

The template renders `{{ user.auth_token }}` directly into example curl commands when a user is authenticated. Exposing long-lived API tokens in HTML increases leakage risk through browser extensions, screenshots, shared screens, client-side logs, or injected third-party scripts.

## Solution

Do not render real tokens into page HTML. Always use placeholders by default and provide an explicit, short-lived token reveal/copy flow (with re-auth) if necessary. Consider scoped/expiring tokens for docs UX.

## Changes

- `cl/api/templates/cotton/docket_endpoint.html` (modified)

### DIFF
--- a/cl/api/templates/cotton/docket_endpoint.html
+++ b/cl/api/templates/cotton/docket_endpoint.html
@@ -7,7 +7,7 @@
 <c-code>
 curl -v \
 -X OPTIONS \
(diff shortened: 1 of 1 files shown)


mlissner:
This is on purpose to make the API docs easier to read. We're moving these pages to a new system soon though which will remove the token, alas. 

Thanks though!

PR #7236, opened 2026-04-20
feat(api): Add DjangoModelPermissionsWithView to enable view-level permissions for EmailProcessingQueueViewSet

albertisfu:
## Fixes
No issue tracked.

## Summary

We need to grant read access to the `flp_staff_state_content` group for the `/api/rest/v4/recap-email/` endpoint for debugging purposes, without relying on the `has_recap_upload_access` permission.

To achieve this, this PR introduces `DjangoModelPermissionsWithView`, a subclass of `DjangoModelPermissions` that maps `GET` and `HEAD` requests to Django’s built-in `view_<model>` permission (instead of requiring no permission for safe methods, which is DRF’s default).

It also wires this into `EmailProcessingQueueViewSet` using DRF’s `|` operator, so requests are allowed if the caller satisfies **either** the existing `EmailProcessingQueueAPIUsers` rule **or** has the Django `view_emailprocessingqueue` permission. This allows internal staff with read-only Django model permissions to inspect the queue without needing the RECAP uploader group.

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/api/utils.py
+++ b/cl/api/utils.py
@@ -980,6 +980,14 @@ class EmailProcessingQueueAPIUsers(DjangoModelPermissions):
     }
 
 
+class DjangoModelPermissionsWithView(DjangoModelPermissions):
+    perms_map = {
+        **DjangoModelPermissions.perms_map,
+        "GET": ["%(app_label)s.view_%(model_name)s"],
+        "HEAD": ["%(app_label)s.view_%(model_name)s"],
+    }
+
+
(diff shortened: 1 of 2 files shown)


ERosendo (approved):
LGTM :shipit:

Issue #6320, opened 2025-09-20
Update Opinion ES Mapping to use `html_with_citations` in `prepare_text`


ERosendo:
### Describe the issue.

We need to update the Elasticsearch mapping for Opinion documents to correctly index content in the text field using the new `html_with_citations` attribute.

This requires:

- Updating the `prepare_text` method to reference `html_with_citations` instead of the old logic.

- Reindexing existing data to make sure the new field is properly indexed.

### What type of issue is this?

To Do

mlissner:
Sorry, I don't understand this one well enough. What happens if we never do this? What's wrong about our current approach?

ERosendo:
@mlissner The current `prepare_text` method used for Opinion mappings is defined here:

https://github.com/freelawproject/courtlistener/blob/05d7f9fc3455043eda1dacb896278acdf9e5b4b5/cl/search/documents.py#L2061-L2081

As you can see, the logic does not take the `html_with_citations` field into account. Since this field is used throughout the codebase as the standardized source of truth for Opinion content, we opened this issue to update the ES mapping and keep it consistent with other helpers:

https://github.com/freelawproject/courtlistener/blob/05d7f9fc3455043eda1dacb896278acdf9e5b4b5/cl/search/models.py#L3116-L3163

Below are some observations from a recent PR related to embedding generation:

> Embedding generation would be only triggered when the `html_with_citations` field is updated. This decision is based on the following observations:
> 
> - Nearly all Opinion records already have this field populated (only 513 records are missing content here, based on a recent query).
> 
> - `html_with_citations` has the highest priority in the `with_best_text` method, meaning there’s a very high likelihood this field was used when generating the embeddings currently stored in S3.
> 
> - The `html_with_citations` field is actively updated in methods such as `recap_document_into_opinions` and `store_opinion_citations_and_update_parentheticals`.
> 
> - Given its coverage, it looks like this field should be considered the standardized source of truth for Opinion content.
> 
> _source_: https://github.com/freelawproject/courtlistener/pull/6311

PR #7238, opened 2026-04-21
feat(search): Add semantic search toggle to the frontend

### LINKED ISSUE #4345: Set up html front-end behind a waffle (htmx)

rachlllg:
## Fixes

Fixes: #4345

## Summary

Adds a semantic search toggle to the legacy frontend search bar, enabling users to switch between keyword and semantic (natural language) search modes for case law opinions.

**Backend:**
- Updated `SearchForm.clean()` so `semantic=true` works from the frontend `do_es_search` path. Semantic search on the frontend is gated by `KNN_SEARCH_ENABLED` and the `semantic_search_frontend` waffle flag (resolved via `flag_is_active` in the view and passed through as `is_semantic_frontend_active`). API requests are only gated by `KNN_SEARCH_ENABLED`.
- Extracted `get_query_embedding()` helper and pre-generates the embedding once in `do_es_search()` so both the main query and facets query reuse it (avoids duplicate inception calls).
- Catches `InputTooLongError` in `do_es_search` to show a user-friendly message when queries exceed `MAX_EMBEDDING_CHAR_LENGTH`, instead of HTTP 500.

**Frontend:**
- Clickable toggle icon on the left side of the search input: `&` for keyword, `?` for semantic. Brand red (#B53C2C) background with white text.
- Placeholder text updates to guide users: "Keyword(s) search with operators" vs "Natural language search, use quotes for exact terms".
- User preference cached in `localStorage` across page loads. Inline script restores state before paint to avoid visual flash.
- Alert bell icon hidden in semantic mode since search alerts don't support semantic queries.
- Gated by the `semantic_search_frontend` waffle flag for controlled rollout.
- Mobile-responsive: flex layout scoped to homepage and advanced search pages.
- Accessibility: `role="button"`, `tabindex="0"`, `aria-label`, `title`, keyboard Enter/Space support.

**Tests:**
- Frontend integration tests mirroring the API semantic search tests: basic search, court filter, docket number filter, citation count sorting, hybrid search.
- Integration test verifying the waffle flag properly gates semantic search on the frontend.
- Form clean tests verifying `KNN_SEARCH_ENABLED` gating and `is_semantic_frontend_active` flag gating.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

1. Create the `semantic_search_frontend` waffle flag via Django admin or shell and enable for testers.
2. Ensure `KNN_SEARCH_ENABLED=true` and the inception microservice is reachable.

<details closed>
<details open>

Home:
home-desktop

Advanced Search:
advanced-desktop

Search Results:
results-keyword-desktop

results-semantic-desktop

</details>
<details open>

Home:
home-mobile

Advanced Search:
advanced-mobile

Search Results:
results-keyword-mobile

results-semantic-mobile

</details>
<details open>
<summary><h4>ipad mini</h4></summary>

Home:
home-ipad

Advanced Search:
advanced-ipad

Search Results:
results-keyword-ipad

results-semantic-ipad

</details>


rachlllg:
The frontend error is unrelated to my changes, let me know if I should open a PR to fix, or if it's already on someone's plate to be fixed. 

Claude helped me through a good chunk of the frontend work, I must admit my frontend is rusty, so do let me know what I can do to improve it.

mlissner:
Looks good to me. I'll leave it to others to do the review and I'll loop in Claude to see what it thinks.

mlissner:
@claude review once

rachlllg:
Sorry, I noticed a small bug with the placeholder text not updating properly when switching mode so I added a fix to it.

mlissner:
@claude review once

mlissner (approved):
LGTM at a product-level. I'd be happy to see this land. 

It might help to make the waffle more specific. Perhaps `semantic_search_frontend`, say? 


I flagged it for Claude review too.

rachlllg:
I will address Claude's comments together with Elisa and Eduardo's comments after their review.

cl/lib/search_utils.py
albertisfu, line 791: We need to catch `InputTooLongError` here so that when users send a query larger than `MAX_EMBEDDING_CHAR_LENGTH`, we can show a customized message in the frontend instead of a generic 500 error page.

For that, we can do:

```suggestion
      except InputTooLongError:
            error = True
            error_message = "input_too_long_error"
```

Then we can add the customized message in `no_results.html` using the "input_too_long_error" `error_message`.

cl/lib/search_utils.py
```diff
@@ -712,6 +713,13 @@ def do_es_search(
                         remove_missing_citations(missing_citations, cd)
                     )
                     cd["q"] = suggested_query if suggested_query else cd["q"]
+
+            # For semantic searches, generate the embedding once so both
+            # the main query and the facets query reuse it instead of
+            # making duplicate calls to the inception microservice.
+            if has_semantic_params(cd) and not cd.get("embedding"):
+                cd["embedding"] = get_query_embedding(cd["q"])
```
albertisfu, line 727: We have a micro-cache enabled in the search frontend that stores search results for 10 minutes (`SEARCH_RESULTS_MICRO_CACHE`). You can tell a request was served by the cache when the search response time is 1 ms. However, in cases where results are returned by the cache instead of ES, we are still making a request to Inception to get the query embedding. My suggestion is to first check whether the query is cached (`retrieve_cached_search_results`) before requesting the embedding.

This is just a performance improvement, so if it is easy to do in this PR, great; otherwise, it can be done in a follow-up PR whatever seems best.

cl/search/forms.py
```diff
@@ -700,10 +700,7 @@ def clean(self):
             if isinstance(v, str):
                 cleaned_data[k] = v.strip()
 
-        should_disable_knn_search = (
-            not settings.KNN_SEARCH_ENABLED or not self.request
-        )
-        if should_disable_knn_search:
+        if not settings.KNN_SEARCH_ENABLED:
```
albertisfu: I'd suggest also checking the `semantic_search` flag here when the response is coming from the frontend (`not self.request`). That way, we can properly control the rollout of this feature. Otherwise, if users or bots realize that passing `semantic=true` to a frontend request triggers a semantic response even when the frontend toggle is disabled, they will still be able to make those requests.

If we need to roll back this feature for the frontend, we would otherwise also need to disable it for the API too by switching off `KNN_SEARCH_ENABLED`.

albertisfu (changes requested):
Thanks @rachlllg, this looks pretty good. I just have a couple of suggestions.

I also found a potential issue that could be confusing for users:

When we rolled out semantic search for the API, we decided that `html_with_citations` is the correct field to use for generating embeddings. Currently, we also trigger embedding generation and re-indexing whenever `html_with_citations` changes on an opinion.

However, this field is not currently indexed in Elasticsearch for keyword search. We already have an open issue for that:
https://github.com/freelawproject/courtlistener/issues/6320

The problem I see is that users may notice mismatched results between semantic and keyword searches while using the frontend. For example, if a semantic chunk contains the word `copyright` and the user then toggles to keyword search, they may not see the same opinion in results because that content is not currently indexed in Elasticsearch.

This is not an error in this PR, but something I wanted to point out in case you think it is relevant for launch or post-launch planning.

Once `html_with_citations` is included in keyword search indexing in Elasticsearch, we will need to trigger a full re-index of Opinions.

albertisfu:
Also, we’ve had a few `ReadTimeout` issues from Inception when generating embedding queries.

I just wanted to flag this in case we need to adjust something related to infrastructure, since we expect the number of requests to increase once this is enabled on the frontend, which could make this issue worse.

https://freelawproject.sentry.io/issues/7303968001/events/aef72068c3b941aca9dbb824d39afdf9/

cc @ERosendo

elisa-a-v (commented):
I have no additional notes other than what has been flagged.

I've opened #7271 to address the frontend check tabnabbing error; for future reference in case you run into it again, you can simply add `rel="noopener noreferrer"` to the `<a>` flagged by the check and that should do it 👍🏽

cl/lib/search_utils.py
rachlllg: I will open a follow-up PR shortly so we can get this one merged first.

cl/search/forms.py
```diff
@@ -700,10 +700,7 @@ def clean(self):
...
             cleaned_data["semantic"] = False
```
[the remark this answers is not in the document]
rachlllg: No longer applicable after addressing Alberto's comment to use self.request to distinguish frontend and api requests

rachlllg:
Thank you @albertisfu, your comments have been addressed, also addressed Claude & Mike's suggestions

cl/search/forms.py
```diff
@@ -700,7 +701,15 @@ def clean(self):
...
 
-        if not settings.KNN_SEARCH_ENABLED:
+        # Disable semantic search if KNN is off globally, or if the
+        # request is from the frontend and the waffle flag is not active.
+        # API requests pass self.request and are only gated by KNN_SEARCH_ENABLED.
+        if not settings.KNN_SEARCH_ENABLED or (
+            not self.request
+            and not Flag.objects.filter(
+                name="semantic_search_frontend", everyone=True
+            ).exists()
```
albertisfu: There is a performance issue with doing this, as it would perform an extra DB query on each search request. It would also act as a global switch, meaning the flag would need to be enabled for everyone in order to test it.

Using `waffle.flag_is_active` instead has a couple of advantages. It includes a cache layer that prevents hitting the DB on each request (it uses Redis as a cache instead). Another advantage is that we can enable the flag only for staff or specific users, which would allow us to test the feature as soon as this is merged without enabling it for everyone.

The `flag_is_active` method requires a request object, which is not available in the frontend form. In the past, we solved this issue by passing the flag status from the caller (in this case, frontend search) as an argument to the form initialization. 

As an example, here is the code from that approach (it belongs to old code already removed from production).

https://github.com/freelawproject/courtlistener/blob/2955b0ba5d7426fd62671e6d00c044818f2773c2/cl/search/api_views.py#L202

https://github.com/freelawproject/courtlistener/blob/2955b0ba5d7426fd62671e6d00c044818f2773c2/cl/search/forms.py#L459

albertisfu (changes requested):
Thanks @rachlllg, the changes look good. I only have one additional suggestion.

cl/search/forms.py
rachlllg: Thank you @albertisfu, this makes sense, I made the updates, let me know if further updates are needed.

cl/search/views.py
```diff
@@ -146,7 +147,12 @@ def show_results(request: HttpRequest) -> HttpResponse:
         if not alert_form.is_valid():
             # Invalid form. Do the search again and show them the alert form
             # with the errors
-            render_dict.update(do_es_search(request.GET.copy()))
+            render_dict.update(
+                do_es_search(
+                    request.GET.copy(),
+                    is_semantic_frontend_active=is_semantic_active,
```
albertisfu: Passing `is_semantic_active` might not be required here, since this can only be rendered when the alert form has an error and we’re not supporting Search Alerts for semantic queries because the percolator does not support them:
https://github.com/elastic/elasticsearch/issues/109344

Also, to prevent confusion for users, I think we should hide the Alert icon in the UI for semantic search.

Image

We could also document in the Search Alerts documentation that semantic queries are not supported for alerts. If a user creates an alert using the `semantic=true` param, it would be ignored and behave as a keyword search alert instead.

Hiding the alert icon and updating the documentation could probably be done in a separate PR if you’d like.

albertisfu (changes requested):
Thanks @rachlllg the flag is now working as expected.

I only left one comment regarding semantic search and Search Alerts.

cl/search/views.py
rachlllg: Thank you @albertisfu, this has been updated and I updated the screenshots in the PR description to reflect the latest version.

I also noticed a very minor existing bug where the bell icon is not visible when entered via homepage with an empty query, I filed a separate issue for that. 

I opened a separate issue to update the alerts documentation for what you mentioned above.

albertisfu:
Thank you @rachlllg !

I did a test and found that users would still be able to save semantic queries as search alerts. This happens because switching between semantic and keyword search does not refresh the search results. Here's a video showing an example:

https://github.com/user-attachments/assets/a9ce2207-fe22-47c9-b732-7dd7d2c20324

I'm not sure what the best approach would be here. We could either keep the Search Alerts documentation you added in https://github.com/freelawproject/courtlistener/pull/7287 stating that search alerts won't work, or refresh the search when toggling between keyword and semantic search.

rachlllg:
@albertisfu when I was working on hiding the bell, there was an earlier iteration which hid the bell icon upon toggling from semantic to keyword until the search was fired, this would prevent the user from clicking on the bell icon while the URL is still in semantic mode. However, I felt this was not very intuitive as the user would expect the bell to show up as soon as they clicked the toggle to switch to keyword mode, so I intentionally made the bell icon show up as soon as the mode was switched.

The alternative is to fire the search as soon as the toggle was clicked, which I also don't love because now the toggle functions not only as a mode switch but also as a search button, I felt it would be better to keep the mode switch and search button separate.

I'd like to propose leaving the current design as is, it would be apparent to the user that the bell icon is designed to only be available in keyword mode. I would hope they understand that they need to click on the search button to actually reflect the search results of the new mode they selected, and that the fact the bell only shows up in the keyword mode means the alerts are only available for keyword results.

We could also add one sentence in the alert pop-up to note that alerts are only available for keyword results (and keep the documentation PR). What do you think?

rachlllg:
Mike pinged me and suggested to simply hide the bell on toggle switch to keyword until the search has been fired. Let me know if this works!


https://github.com/user-attachments/assets/27cc6ed9-ef96-4b48-bc03-8861f7e22a24

albertisfu (approved):
> I'd like to propose leaving the current design as is, it would be apparent to the user that the bell icon is designed to only be available in keyword mode. I would hope they understand that they need to click on the search button to actually reflect the search results of the new mode they selected, and that the fact the bell only shows up in the keyword mode means the alerts are only available for keyword results.
> 
> We could also add one sentence in the alert pop-up to note that alerts are only available for keyword results (and keep the documentation PR). What do you think?

Yeah. I also don't like triggering the search as soon as the user switches the search mode.

So yes, I believe we can keep the current design, and adding the sentence that alerts are only available in keyword search will help prevent confusion for users.

So I think this PR is ready to go. 
We can merge it whenever you are ready. I'd recommend creating the flag before we merge it and setting it as disabled.

PR #7253, opened 2026-04-22
feat(api): Supports multiple throttle rates per user

### LINKED ISSUE #7194: Extend `APIThrottle` model to support multiple rate limits per user
The `APIThrottle` model ([`cl/api/models.py`](cl/api/models.py)) allows admins to override a user's rate limit. It stores one `rate` string (e.g. `'100/hour'`) per `(user, throttle_type)` pair, enforced by a `UniqueConstraint` on those two fields:

```python
class ThrottleType(models.IntegerChoices):
    API = 1, "API"
    CITATION_LOOKUP = 2, "Citation Lookup"

class APIThrottle(AbstractDateTimeModel):
    user = models.ForeignKey(User, ...)
    throttle_type = models.SmallIntegerField(choices=ThrottleType.choices)
    rate = models.CharField(...)  # e.g. "100/hour"

    class Meta:
        constraints = [
            models.UniqueConstraint(
                fields=["user", "throttle_type"],
                name="unique_user_throttle_type",
            )
        ]
```

The new membership tiers define **simultaneous limits** per user: per-minute, per-hour, and per-day. For example:

- 5/minute
- 50/hour
- 125/day

The current model cannot represent this: a user can only have one `APIThrottle` row for `ThrottleType.API`, holding a single rate string.

ERosendo:
## Fixes

This PR fixes #7194 

## Summary

This PR instroduce logic and migrations to lets a single user carry several simultaneous rate limits (e.g. `5/min`, `50/hour`, `125/day`)

Key changes:

- Updates `APIThrottle` to support multiple rate rows per user by relaxing the unique constraint from (user, throttle_type) to (user, throttle_type, rate). This allows one user to carry multiple rate limits. A clean() validation was also added to prevent multiple rows with the same time unit, so the admin returns a clear validation error.
- Updates `ExceptionalUserRateThrottle` and `CitationCountRateThrottle` to accept either a single rate string or a list of rates. Each rate window is evaluated from the tightest to the broadest for faster rejection. Retry-After now reflects the window that was actually exceeded. Cache TTL is set using the longest window so longer-range checks are not evicted too early.
- A 0/min rate now effectively blocks the user since count >= 0 is always true. This removes the need to rely on a separate blocked boolean in the request enforcement path. The blocked column remains in the schema for this PR so existing rows continue to work during rollout, and it can be removed in a follow-up cleanup PR.
- Updates `DEFAULT_THROTTLE_RATES["user"]` from a single "5000/hour" value to ["5/min", "50/hour", "125/day"].
- Updates the citation lookup API help page to iterate over the configured rate list and render a more natural sentence such as: “5 requests per minute, 50 per hour, and 125 per day.”
- Removes the blocked column from list_display and list_filter in the admin panel, since blocking is now represented directly through rate rows.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### Additional notes

This PR changes the default authenticated user rate from 5000/hour to:

- 5/min
- 50/hour
- 125/day

This means any authenticated user without an explicit APIThrottle override will begin using the new default tier as soon as this ships.

Before merging, we should:

- Identify users who may be negatively affected by the new defaults, starting with paying members and the highest-volume API consumers based on recent request data.
- Grandfather those users by adding a 5000/hour override row in APIThrottle so their current effective limit remains unchanged.


cl/api/utils.py
```diff
@@ -721,30 +721,55 @@ def _save_page_in_cache(
...
+            self.rate = raw
+            self.default_rates = [raw]
+        else:
+            self.default_rates = list(raw)
+            # Set self.rate so DRF base-class attrs have valid defaults.
+            self.rate = self.default_rates[0]
+
+        if self.rate is not None:
+            self.num_requests, self.duration = self.parse_rate(self.rate)
+
```
albertisfu, line 771: When hitting a throttle limit, if the user has multiple custom throttles defined and the number of timestamps stored in the window is larger than the throttle being checked, the wait time is not computed properly. Here is what I’m currently seeing:

```
HTTP 429 Too Many Requests
Allow: GET, PUT, PATCH, DELETE, HEAD, OPTIONS
Content-Type: application/json
Vary: Accept

{
    "detail": "Request was throttled."
}
```

Instead of the regular response with the Retry-After header:

```
HTTP 429 Too Many Requests
Allow: GET, PUT, PATCH, DELETE, HEAD, OPTIONS
Content-Type: application/json
Retry-After: 48
Vary: Accept

{
    "detail": "Request was throttled. Expected available in 48 seconds."
}
```

It seems that the problem is the following:

In my example, I set up two different rates:

`1/m` and `5/h`.

The first time I hit the `1/m` limit, I saw the Retry-After header properly.

Then I waited for 1 minute and hit the rate again.

This time, the Retry-After header was not shown.

So it seems the issue is that the wait time is computed based on all the timestamps stored in the window, which belong to the larger rate (in this case, 1 hour). However, the wait calculation expects the number of requests in the window to be equal to or smaller than the limit of the rate being checked. As a result, the computed wait time becomes negative, which translates to a None value.

The solution seems to be to override the `wait()` method to filter out timestamps outside the rate being checked.

cl/api/utils.py
```diff
@@ -755,24 +780,70 @@ def allow_request(self, request, view):
...
+            return None
+
+        cutoff = self.now - self.duration
+        relevant = [ts for ts in self.history if ts > cutoff]
+        if relevant:
+            remaining_duration = self.duration - (self.now - relevant[-1])
+        else:
+            remaining_duration = self.duration
+
+        available_requests = self.num_requests - len(relevant) + 1
```
albertisfu: This `wait()` method is working now in most of the cases however I still found an edge case where the wait is not computed properly.

To reproduce:

- Set a 10/h rate
- Reach out the limit in less than a minute and get throttled.
- Create a new 5/m rate.
- Hit the API and. You'll get trhottled but not Retry-After header will be shown.

This is happening because the 10 requests fall within the 5/m window, and `available_requests` becomes negative.

This issue is more likely to occur if the user originally only had a daily rate and then you add an hourly rate. Requests within that hour will be throttled without a `Retry-After` header.

To solve this, we can compute how long it will take for enough old requests to age out of the window so that the count falls back below the limit.

albertisfu (changes requested):
Thanks @ERosendo this looks good. Just an additional edge case I found.

cl/api/utils.py
```diff
@@ -755,24 +780,80 @@ def allow_request(self, request, view):
...
+        active window.
+
+        Instead, we compute how many requests must expire before a new request fits,
+        then return the time until the most recent of those expiring requests leaves
+        the window.
+        """
+        if self.duration is None or self.num_requests is None:
+            return None
+
+        window_start = self.now - self.duration
```
albertisfu, line 842: I tested this new approach and found that it can still be a little inconsistent. Since `self.duratio`n is defined by the tightest rejected window, there may still be timestamps within that window (for example, the minute window) when the user hits the API. If the next request that needs to expire from that window will expire in 10 seconds, the user will see a `Retry-After` of 10 seconds. The user will then retry in 10 seconds, but the request will still be rejected by the larger window (for example, the `1-hour` window). This is not too bad; it just means the user would have to wait longer than we initially indicated.

Would it be possible to check whether there is a larger window and compute the wait time based on that window, even if a smaller window is still active?

ERosendo: can you provide an example of this case?

albertisfu: Sure. It's the same case described previously. 
- Set a 10/h rate
- Reach out the limit (10 requests) in less than a minute and get throttled.
- Create a new 5/m rate.
- Hit the API (you still need to be in the same minute period) and you'll get throttled and the Retry-After header will be shown close to the smallest window (5/m) once the request within the 5/m window expires. You'll still be throttled but this time by the larger window (10/h)

For this test you'll need to reduce the `get_all_throttle_overrides` cache to something really small like 1 second.

So this specific scenario will not happen in reality due to the cache but it still can happen with larger rates like 1000/day and 100/hour rates.

ERosendo: Thanks for the example @albertisfu.

We don’t even need to introduce a new rate to reproduce the issue. Let’s assume we start with both limits set from the beginning: 5/min and 30/hour.

In the first 6 minutes, we can fully exhaust the API quota.

For example:

Minute 1: 5 requests -> gets rate limited (5/minute)
Minute 2: 5 requests -> gets rate limited (5/minute)
Minute 3: 5 requests -> gets rate limited (5/minute)
Minute 4: 5 requests -> gets rate limited (5/minute)
Minute 5: 5 requests -> gets rate limited (5/minute)
Minute 6: 5 requests -> gets rate limited (5/minute)

Then:

Minute 7: 1 request ->  rate-limited by the hourly window

I’ll try to tweak the error message. Specifically, improve the message shown when a throttle happens so the user has context about which rate limit they hit. If it turns out to be too much work, I’ll open an issue instead.

ERosendo: Now, when the user is throttled, the message clearly shows the specific rate limit that was hit:

image
 
And when the user is blocked, it explicitly states that as well, so there’s no ambiguity between being blocked and being rate-limited:

image

cl/api/utils.py
```diff
@@ -814,118 +912,111 @@ def get_cache_key_for_citations(self, request, view):
...
+        rates = self.get_citations_rate(request)
+        parsed: list[tuple[int, int]] = [
+            (n, d)
+            for r in rates
+            for n, d in [self.parse_rate(r)]
+            if n is not None and d is not None
+        ]
+        max_duration = max(d for _, d in parsed)
 
         self.key = self.get_cache_key_for_citations(request, view)
```
albertisfu, line 946: Not sure if this is a pre-existing issue or if I'm setting something up wrong.

But I set a Citation Lookup rate, and it seems to be ignored. Instead, it falls back to the general API rate.

After inspecting the keys, I found that Citation Lookup requests are being stored in the `throttle_user_ID` key, while we're checking the history in `throttle_citations_ID`, which seems to be a mismatch and the reason the Citation Lookup rate is not being enforced.

ERosendo: Citation Lookup is throttled based on the number of citations processed. If a request has no citations, it won’t count toward the citation rate limit.

The citation rate represents how many valid citations can be processed per time unit (e.g., 60 citations/minute by default). So:

- You could send 2 requests with 30 citations each, allowed
- Or 60 requests with 1 citation each, also allowed

However, the general API rate limit still applies independently. If your API throttle is lower (e.g., < 60 requests/min), it will kick in first, even if you’re within the citation limit.

So what you’re seeing is the API throttle being enforced before the citation limit, rather than the citation rate being ignored.

albertisfu: Thanks. Yeah, I was sending requests that didn't return any citation results, so the citations throttle was not being triggered.

albertisfu (approved):
Thanks @ERosendo this looks good now.

Let's merge it.

PR #7263, opened 2026-04-24
feat(api): Add multi-rate user throttle defaults behind a switch

ERosendo:
## Fixes
Fixes: #7196

## Summary
Issue #7196 originally proposed creating an `APIThrottle` row per new account at signup. Instead, this PR takes a simpler, cheaper approach: ship the new defaults in settings and let regular authenticated users pick them up automatically.

To make the rollout safe, the change is gated behind a new `use_new_throttle_defaults` waffle switch in `ExceptionalUserRateThrottle`:

- **Switch off (default):** the `"user"` scope falls back to the historical `5000/hour` rate, regardless of what `DEFAULT_THROTTLE_RATES` contains. This is the production behavior on day one.
- **Switch on (admin-controlled):** the `"user"` scope reads `DEFAULT_THROTTLE_RATES["user"]`, which this PR sets to the multi-rate list `["5/min", "50/hour", "125/day"]`. The list semantics rely on the multi-rate enforcement landed in #7194.

### Things to know
- Production behavior is unchanged on merge. The new defaults only take effect once an the switch is updated in the admin panel, which lets us roll out and roll back cheaply.
- Users who already have explicit `APIThrottle` overrides keep them in either state. Overrides take precedence over the default.
- The legacy fallback string is hard-coded (`LEGACY_USER_DEFAULT_RATE = "5000/hour"`) so the off-state can't drift if someone tweaks settings.

## Deployment

**This PR should:**
- [ ] `skip-deploy` (skips everything below)
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

**Special steps:**

1. In Django admin (`/admin/waffle/switch/`), create a new switch named `use_new_throttle_defaults` with `active=False`. 
2. Once we're ready to roll the new limits out, flip the switch to active and monitor distribution before announcing the change.
3. To roll back: flip the switch off. No data migration is needed.

PR #7265, opened 2026-04-27
feat(api): Sync APIThrottle with Neon membership lifecycle events

### LINKED ISSUE #7199: Update `APIThrottle` in response to Neon membership lifecycle events
Neon CRM sends webhooks to `MembershipWebhookViewSet` ([`cl/donate/api_views.py`](cl/donate/api_views.py)), which routes to three handlers:

| Webhook event | Handler |
|---|---|
| `createMembership` | `_handle_membership_creation()` |
| `editMembership` / `updateMembership` | `_handle_membership_update()` |
| `deleteMembership` | `_handle_membership_deletion()` | 

For each lifecycle event, add throttle assignment logic after the membership record is saved/deleted.

ERosendo:
## Fixes

This fixes #7199, it also fixes https://github.com/freelawproject/internal/issues/925

## Summary

This PR connects Neon membership lifecycle events (`createMembership`, `editMembership / updateMembership`, `deleteMembership`) to API rate-limit throttles so that users’ `APIThrottle` rules stay in sync with their membership tier in real time. Nothing changes in production until the `sync_membership_throttles` waffle switch is enabled.

The implementation is split into four parts:

### 1. Source tracking for throttles

Adds a source field to `APIThrottle` to distinguish between `MANUAL` and `MEMBERSHIP` rate-limit overrides. Existing rows default to `MANUAL`. Updates validation so `MANUAL` and `MEMBERSHIP` throttles can coexist for the same user record.

### 2. Effective override rules

Reworks `get_all_throttle_overrides` to define how throttles are applied:

- `MANUAL` overrides are always applied
- `MEMBERSHIP` overrides are only applied when the user’s Neon membership is active (successful payment and not expired)
- if any `MANUAL` override exists, all `MEMBERSHIP` overrides are ignored

This keeps admin overrides simple and ensures they fully take precedence over membership-based limits.

Also updates the admin so source is visible and prevents editing of `MEMBERSHIP` records:

- change form uses `get_readonly_fields`
- user inline uses a custom form that disables fields

### 3. Membership sync helpers + feature flag

Adds two helpers:

- `apply_membership_throttles(user, level)` — applies rate limits for a given membership tier
- `clear_membership_throttles(user)` — removes membership-based throttles

Both are gated behind the `sync_membership_throttles` waffle switch (off by default).

Unknown or legacy membership levels (e.g. BASIC, Commercial) are ignored with a log entry. Supported tiers map directly to predefined API rate limits. Constants live in cl/api/constants.py for reuse.

### 4. Webhook integration

Hooks the helpers into `MembershipWebhookViewSet` so throttles update automatically:

- `createMembership`: applies tier throttles after the membership row is created
- `update/editMembership`: replaces previous membership throttles when tier changes
- `deleteMembership`: clears membership throttles before deletion

All calls are gated by the waffle switch, so this has no effect until explicitly enabled.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


cl/api/constants.py
```diff
@@ -0,0 +1,65 @@
...
+
+# Rate sets per tier. The "size" dimension on group memberships
+# (Smallest/Small/Medium/Large/Unlimited) affects seat count, not
+# per-user throughput, so all sizes within a tier share the same
+# per-user rates. HeyCounsel tiers mirror the matching individual
+# tier.
+TIER_1_RATES = ["10/min", "75/hour", "300/day"]
+TIER_2_RATES = ["15/min", "150/hour", "600/day"]
+TIER_3_RATES = ["20/min", "250/hour", "1000/day"]
+TIER_4_RATES = ["25/min", "300/hour", "1400/day"]
```
Brennan-Chesley-FLP, line 22: These make more contextual sense when paired with [this PR](https://github.com/freelawproject/courtlistener/pull/7263/changes). Mostly noting this for other readers.

Brennan-Chesley-FLP:
This PR looks good to me. It has reasonable coverage of the path's that I'd expect, I don't see any obvious race conditions, there's clearly some planning in it for future development. I will just note that I suspect it's likely that we'll get some drift between our records of membership and neons. and that it would be nice to make a button in admin to manually force a refresh of the current membership level from neon. Ideally when we make token refreshes self-service for users, we could have such an option show up there too.

ERosendo:
@Brennan-Chesley-FLP thanks for the review! I resolved a merge conflict, but I’ll hold off on merging this PR so I can coordinate with @chaco-fl, since it includes migrations.

chaco-fl (approved):
LGTM

PR #7269, opened 2026-04-27
SCOTUS CAPTCHA solving fix

### LINKED ISSUE #7266: Review SCOTUS CAPTCHAs with less than 5 words
### Describe the issue.

We have enabled the SCOTUS ongoing scraper, which also subscribes to new cases.
During the subscription to new cases, we encountered this error related to the subscription process.

```
ValueError
Expected 5 words, got 3
```

[COURTLISTENER-CVF](https://freelawproject.sentry.io/issues/7443796032/events/c223ec0b6c16437b92af70cfd49fb6ae/)

We should review it and fix it so we can subscribe to cases from the last year.

### Related Issue

_No response_

### Screenshots

_No response_

Research

MorganBennetDev:
## Fixes

Fixes #7266 

## Summary

Splits CAPTCHA transcriptions on runs of non-alphanumeric characters instead of just spaces to account for edge cases like "delta-2".

Switches transcription model to `whisper-1` for improved accuracy.

Adds `language="en"` parameter to LLM transcription call for improved accuracy.

Moves CAPTCHA solution candidate generation to `get_scotus_captcha_solution`, which has retry logic independent of the Celery task.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/scrapers/tasks.py
+++ b/cl/scrapers/tasks.py
@@ -799,6 +799,75 @@ def process_scotus_captcha_transcription(transcription: str) -> str:
     return "".join(characters)
 
 
+def get_scotus_captcha_solution(
+    session: requests.Session,
+    base_url: str,
+    form_url: str,
+    anti_forgery_token: str,
+    n_tries: int = 3,
+) -> tuple[str, str]:
+    """Get the solution to the SCOTUS audio CAPTCHA with retries.
+
+    :param session: The requests session to use.
+    :param base_url: The base URL of the SCOTUS website.
+    :param form_url: The URL for the subscription form.
+    :param anti_forgery_token: The anti-forgery token to pass in requests.
+    :param n_tries: The maximum number of times to try generating CAPTCHAs.
+        Defaults to 3, which given the 4% failure rate observed in whisper-1,
+        should produce an overall failure rate of 0.0064%.
+
+    :return: A tuple containing the solution and the CAPTCHA ID."""
+    captcha_reset_url = f"{base_url}/Captcha/Reset"
+    captcha_payload = {"__RequestVerificationToken": anti_forgery_token}
+
+    for attempt in range(n_tries):
+        reset_response = session.post(
+            captcha_reset_url,
+            data=captcha_payload,
+            headers={
+                "Referer": form_url,
+                "X-Requested-With": "XMLHttpRequest",
+            },
+            timeout=10,
+        )
+        reset_response.raise_for_status()
+        reset_data = reset_response.json()
+        captcha_id = reset_data.get("captchaId")
+        if not captcha_id:
+            raise ScrapeFailed(
+                f"Failed to get captchaId from /Captcha/Reset. Response: {reset_response.text}"
+            )
+
+        # Fetch the Audio
+        audio_url = f"{base_url}/Captcha/audio?captchaId={captcha_id}"
+        audio_response = session.get(
+            audio_url, headers={"Referer": form_url}, timeout=10
+        )
+        audio_response.raise_for_status()
+
+        # Solve the captcha
+        audio_file = BytesIO(audio_response.content)
+        transcription = call_llm_transcription(
+            ("captcha.wav", audio_file),
+            api_key=settings.OPENAI_TRANSCRIPTION_KEY,
+            model="whisper-1",
+        )
+        try:
+            solution = process_scotus_captcha_transcription(transcription)
+        except ScrapeFailed as _:
+            ...
+        else:
+            if attempt > 0:
+                logger.warning(
+                    "Generated valid CAPTCHA solution in %d attempts",
+                    attempt + 1,
+                )
+            return solution, captcha_id
+    raise ScrapeFailed(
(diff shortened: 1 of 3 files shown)


cl/scrapers/tasks.py
```diff
@@ -799,6 +799,75 @@ def process_scotus_captcha_transcription(transcription: str) -> str:
...
+    :param anti_forgery_token: The anti-forgery token to pass in requests.
+    :param n_tries: The maximum number of times to try generating CAPTCHAs.
+        Defaults to 3, which given the 4% failure rate observed in whisper-1,
+        should produce an overall failure rate of 0.0064%.
+
+    :return: A tuple containing the solution and the CAPTCHA ID."""
+    captcha_reset_url = f"{base_url}/Captcha/Reset"
+    captcha_payload = {"__RequestVerificationToken": anti_forgery_token}
+
+    for attempt in range(n_tries):
```
albertisfu, line 827: Wouldn't this retry method compound with the Celery retries in `subscribe_to_scotus_updates`?

Does that mean `process_scotus_captcha_transcription` could be executed up to `3 * celery_retries `times?

I saw that `process_scotus_captcha_transcription` was already raising `ScrapeFailed`, and the `subscribe_to_scotus_updates` task was already automatically retrying.

Would it be possible to move the retry logic to either the task or `get_scotus_captcha_solution` so it's easier to keep track of the retry behavior and avoid compounded retries (maybe task retries is enough?), which could lead to more attempts than expected?

albertisfu (changes requested):
Thanks @MorganBennetDev  I just have a suggestion.

Also, were you able to test the new model with the production key that Mike shared with us in Bitwarden? I tested locally with my own key and got a permissions issue, so I want to confirm that we don’t need to adjust any permissions for our production key.

cl/scrapers/tasks.py
MorganBennetDev: This is correct. I initially just had it raise `ScrapeFailed` but I didn't like that the invalid transcription would be compounding with the other reasons for retry. I can switch it back to that approach if you think that's better. Maybe up the retry count to 5?

MorganBennetDev:
@albertisfu I used the key in Bitwarden and didn't encounter any issues using the other model.

cl/scrapers/tasks.py
albertisfu: Got it. Correct, but if I’m not wrong, with the current approach, once the n_tries are exhausted, a ScrapedFailed exception will be raised and the task retried again. So, in the worst-case scenario, will the transcription be retried up to 9 times?

If we only want this to retry up to 3 times, maybe we can raise a different error that isn’t automatically retried by the task?

Also, if this works on the first attempt—as we expect it will most of the time—will the warning log still be emitted? What if we only log it on each failure instead?

Finally, can we add a small delay before each attempt to avoid hitting the OpenAI endpoint too quickly?

MorganBennetDev: Oh, I see what you're saying. That's a mistake on my part; I only want it to retry at most 3 times.

> Also, if this works on the first attempt—as we expect it will most of the time—will the warning log still be emitted?

The warning won't be logged if we generate a valid candidate on the first attempt.

> What if we only log it on each failure instead?

Each overall failure or each failed attempt in the loop? If we log each failed attempt, I think we should still keep the summary log since it's giving us useful data.

albertisfu: Ok sounds good. If the log is not logged on the first successful attempt this is ok

MorganBennetDev: Implemented changes

cl/scrapers/tasks.py
```diff
@@ -799,6 +800,78 @@ def process_scotus_captcha_transcription(transcription: str) -> str:
...
+        audio_file = BytesIO(audio_response.content)
+        transcription = call_llm_transcription(
+            ("captcha.wav", audio_file),
+            api_key=settings.OPENAI_TRANSCRIPTION_KEY,
+            model="whisper-1",
+        )
+        try:
+            solution = process_scotus_captcha_transcription(transcription)
+        except ScrapeFailed as _:
+            ...
```
albertisfu: As we are swallowing this exception, after the 3 retries we’ll lose the context of the failure, such as:

`Expected 5 words, got 4 (['2', 'lima', '4', '4'])`

when raising the `ValueError`.

So I suggest preserving the error message:

```suggestion
        except ScrapeFailed as e:
            error_message = str(e)
```

and then doing this when raising the ValueError:

```
raise ValueError(
        f"Failed to generate valid CAPTCHA solution in {n_tries} attempts: {error_message}"
    )

```

MorganBennetDev: Implemented this with the minor change of logging _all_ failed attempts, since that might be useful information.

albertisfu (approved):
Thanks @MorganBennetDev this is ready to go.

Merging.

PR #7272, opened 2026-04-28
ci(front-end-lint) line level skipping for lints

Brennan-Chesley-FLP:
## Fixes

This adds a line-ignore to the front-end lint check. This is primarily for the case where we want to identify specific exceptions to our tabindex rules, but would still like the rest of the file to be checked. 

## Deployment

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/.github/scripts/frontend_checks.py
+++ b/.github/scripts/frontend_checks.py
@@ -5,6 +5,18 @@ Runs checks on changed template and CSS files to enforce frontend
 conventions. Called by the frontend-lint GitHub Actions workflow.
 
 Input file must be in git --name-status format (STATUS\\tPATH per line).
+
+Two directives are recognised in template/CSS files:
+
+* ``{# frontend-checks-skip: <names> #}`` — file-level skip for the
+  listed checks. Restricted to ``SKIPPABLE_CHECKS`` so security and
+  a11y rules cannot be wholesale bypassed.
+* ``{# frontend-checks-ignore-line[: <names>] #}`` — silences findings
+  on the line where it appears. With no names, ignores all checks on
+  that line; otherwise only the named checks. Any check is bypassable
+  here because the scope (one line) makes the suppression reviewable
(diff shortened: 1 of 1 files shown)


Brennan-Chesley-FLP:
This is used in PR https://github.com/freelawproject/courtlistener/pull/7275 .

elisa-a-v (changes requested):
Since we already have a way to skip files that aren't meant to be user-facing, line-based skips should only be allowed to a subset of all checks: I don't think it's ever okay to skip security or critical a11y checks.

Right now this is only used to suppress the tabindex issues in `base.html`, right? In that case, I think the proper fix would be to remove the explicit tabindex values and rely on doc order instead of bypassing the check entirely, so we don't really need to allow line level skipping like this.

If you still think we should allow line level skips, we should at least make an explicit allowlist like we have for file-level skips.

Brennan-Chesley-FLP:
Yup, so long as [the natural tabindexes](https://github.com/freelawproject/courtlistener/pull/7275) look good to you, I like that solution more, and we can drop this PR.

elisa-a-v:
Yes, I also like that more! Feels more natural and goes with the docs which explicitly discourage positive tabindex values:

> Warning: You are recommended to only use 0 and -1 as tabindex values. Avoid using tabindex values greater than 0 and CSS properties that can change the order of focusable HTML elements ([Ordering flex items](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Flexible_box_layout/Ordering_items)). Doing so makes it difficult for people who rely on using keyboard for navigation or assistive technology to navigate and operate page content. Instead, write the document with the elements in a logical sequence.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex

PR #7277, opened 2026-04-28
build(deps): upgrade Django to 6.0.4

albertisfu:
## Summary
Bumps Django from 6.0.2 → 6.0.4 and fixes test failures that the upgrade triggered in several API tests.

The tests broke because Django 6.0.4 backports the **CVE-2026-3902** security fix ([upstream commit `a623c39`](https://github.com/django/django/commit/a623c3982857e80324448f85c7faf9a6710330ef)), which closes an ASGI header-spoofing vector by:

1. Making `ASGIRequest` ignore any incoming header whose on-the-wire name contains `_` (the actual security fix).
2. Updating `AsyncRequestFactory.generic` to normalize underscored keys in the test client's `**extra` kwargs (`key.lower().replace("_", "-")`) so test code that has historically passed `CONTENT_TYPE=` still works after change (1).

Our `cl.tests.utils.AsyncAPIClient.generic` was setting `extra["CONTENT_TYPE"]` *and* forwarding `content_type` as a positional argument to upstream `generic`. The async upstream stores headers as a list of tuples, no dict-based deduplication, unlike the sync `RequestFactory.generic` from which this pattern was originally copied. Pre-6.0.4 the underscored key was silently dropped at the ASGI layer, so the duplication was invisible. Post-6.0.4 the normalization adds a second `(b"content-type", ...)` tuple, the two entries get joined per ASGI/HTTP semantics, and the request reaches DRF as `Content-Type: application/json,application/json`. DRF returns **HTTP 415** failed request 

### Fix

Set `extra["CONTENT_TYPE"]` only when the body is empty. When `data` is truthy, `AsyncRequestFactory.generic` already adds the content-type header from the positional argument, so feeding it back through `extra` would duplicate. When `data` is empty the upstream skips that branch, and the explicit assignment preserves the original "header is present even for empty bodies" intent (matching how DRF's sync `APIRequestFactory.generic` behaves, where dict semantics naturally dedupe).

### Production-side audit

The other half of the CVE patch (ASGIRequest dropping underscored headers) does run in production: gunicorn + UvicornWorker against `cl.asgi`. Audited every incoming-header read in the codebase (`request.headers[*]`, `request.META["HTTP_*"]`, custom auth/webhook signature headers): all use standard hyphenated names (`hx-request`, `user-agent`, `x-requested-with`, `CloudFront-Viewer-Address`, `SECURE_PROXY_SSL_HEADER = HTTP_X_FORWARDED_PROTO`). 

No production code reads a header that a sender would plausibly transmit underscored, so production behavior is unchanged.

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [ ] `skip-cronjob-deploy`
    - [ ] `skip-daemon-deploy`

The Django dependency bump rolls out through the normal image-rebuild path; the test util change has no production effect.

### DIFF
--- a/cl/tests/utils.py
+++ b/cl/tests/utils.py
@@ -83,8 +83,14 @@ class AsyncAPIClient(AsyncClient, APIRequestFactory):
         secure=False,
         **extra,
     ):
-        # Include the CONTENT_TYPE, regardless of whether or not data is empty.
-        if content_type is not None:
+        # Only set CONTENT_TYPE in extra when the body is empty. When data is
+        # non-empty, AsyncRequestFactory.generic already adds the content-type
+        # header from the positional argument; passing it again via extra
+        # would duplicate the header since Django 6.0.4 normalizes
+        # ``CONTENT_TYPE`` → ``content-type`` (CVE-2026-3902 patch), producing
+        # ``application/json,application/json`` and a DRF 415 response and some
+        # tests to fail.
+        if content_type is not None and not data:
             extra["CONTENT_TYPE"] = str(content_type)
 
         return await super().generic(
--- a/uv.lock
+++ b/uv.lock
@@ -805,16 +805,16 @@ wheels = [
 
 [[package]]
 name = "django"
-version = "6.0.2"
+version = "6.0.4"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "asgiref" },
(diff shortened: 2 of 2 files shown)


ERosendo (approved):
LGTM :shipit:

PR #7282, opened 2026-04-28
fix(postgres) strip nuls from plain_text

Brennan-Chesley-FLP:
## Fixes

This strips nulls from the ocred/extracted text of PDFs. Postgres doesn't like em.

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/scrapers/tasks.py
+++ b/cl/scrapers/tasks.py
@@ -608,7 +608,7 @@ async def extract_formatted_text_document_base(
                 rd.ocr_status = AbstractPDF.OCR_NEEDED
 
         rd.plain_text, _ = anonymize(content)
-        rd.plain_text = rd.plain_text.replace("\0","")
+        rd.plain_text = rd.plain_text.replace("\0", "")
         # Kludgey fix to handle RECAPDocument's custom save logic.
         if isinstance(rd, RECAPDocument):
             await rd.asave(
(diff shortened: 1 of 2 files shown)


Brennan-Chesley-FLP:
Should fix https://github.com/freelawproject/courtlistener/issues/7283

MorganBennetDev (approved):
Small change. Seems good.

cl/lib/privacy_tools.py
```diff
@@ -132,4 +132,4 @@ def anonymize(s: str) -> tuple[str, bool]:
     s, ien_count = re.subn(ein_re, r"XX-XXXXXXX", s)
     s, a_number_count = re.subn(a_number_re, r"AXXXXXXXX", s)
     modified = bool(ssn_count + ien_count + a_number_count)
-    return s, modified
+    return s.replace("\0", ""), modified
```
albertisfu: This seems like a safe assumption for content being extracted.

However, this `anonymize()` method is also used [here](https://github.com/freelawproject/courtlistener/blob/5f9d2065497d596c99b2fdf2e6b9b2f0b9d5edaf/cl/recap/mergers.py#L537) for checking `docket_numbers`. This might be quite rare, but if a NUL value is returned in the `docket_number` during parsing, it could indicate that something is wrong with the parser and that issue would be silently ignored after this fix.

I'd recommend using this replacement only for document extraction.

albertisfu (approved):
Thanks, looks good. Merging.

Issue #7105, opened 2026-03-13
Fill out the 2025-2026 SCOTUS cases gap.


albertisfu:
### Describe the issue.

We've ingested the cases scraped in https://github.com/freelawproject/courtlistener/issues/6109.

We should scrape SCOTUS historical cases from 2025 to the present to fill the gap and ingest them. Once we enable the [ongoing](https://github.com/freelawproject/courtlistener/issues/6110)(https://github.com/freelawproject/courtlistener/issues/6110) scraper, we will have both the historical cases and the new ones as they are published.

### What type of issue is this?

To Do

MorganBennetDev:
I used the most recent SCOTUS session journal ([October, 2025](https://www.supremecourt.gov/orders/journal/Jnl25.pdf)) along with the below script to extract docket numbers from the 2026 session.

```python
import pypdf
import re

reader = pypdf.PdfReader("scotus/Jnl25.pdf")

dns = []

dn_re = re.compile(r"No\. \d{2}(?:(?:(?:–|A)\d+)|(?:M\d+))\.")

for page in reader.pages:
    text = page.extract_text()
    dns.extend(dn_re.findall(text))

dn_norm_re = re.compile(r"No\. (.+)\.")

dns_norm = [dn_norm_re.sub(r"\1", dn) for dn in dns]
```

I've attached the list of docket numbers in this file: [dns-25.txt](https://github.com/user-attachments/files/26689255/dns-25.txt).

This file can be passed as input to the following command to download all the dockets for the 2025 session.

```python

import time

import httpx

from cl.lib.command_utils import VerboseCommand
from cl.lib.storage import AWSMediaStorage


class Command(VerboseCommand):
    help = "Save SCOTUS JSON dockets to S3 using docket numbers from a file."
    
    def add_arguments(self, parser):
        parser.add_argument(
            "--file",
            required=True,
            help="Path to file containing SCOTUS docket numbers to import.",
        )
        parser.add_argument(
            "--delay",
            type=float,
            default=1.0,
            help="Seconds to sleep between scheduling tasks.",
        )
        parser.add_argument(
            "--bucket",
            type=str,
            required=True,
            help="S3 bucket to save files to."
        )
        parser.add_argument(
            "--prefix",
            type=str,
            required=True,
            help="Prefix for S3 keys."
        )

    def handle(self, *args, **options):
        filepath = options["file"]
        delay = options["delay"]
        bucket = options["bucket"]
        prefix = options["prefix"]
[...]

PR #7289, opened 2026-04-29
feat(corpus_importer): sweep mode for SCOTUS docket daemon

albertisfu:
## Fixes
Progresses #7105  adds the mechanism we'll use to fill the 2025–2026 SCOTUS docket gap without hammering SCOTUS.

## Summary
The `probe_scotus_dockets_daemon` already maintained two watermarks per (sequence, term): `HIGHEST_SCOTUS_KNOWN_SERIAL` (last serial fully ingested) and `HIGHEST_SCOTUS_OBSERVED_SERIAL` (highest serial the probe has confirmed exists). When `observed > ingested`, the daemon would drain the entire gap in one iteration. That's fine for a 2–3 serial gap left by an interrupted run, but it is not safe when the gap is thousands of serials which is exactly the situation we're in for the 2025–2026 backfill.

This PR introduces a **sweep mode** that caps per-iteration ingestion at `SCOTUS_FIXED_SWEEP` (default `100`) serials. The change is small: it modifies the existing `if highest_observed > highest_ingested:` branch in `_probe_scotus_sequence` to compute `to_serial = min(highest_observed, highest_ingested + SCOTUS_FIXED_SWEEP)` instead of using `highest_observed` directly. Each daemon iteration (currently set to one hour `SCOTUS_PROBE_WAIT=3600`) ingests up to 100 serials, advances `HIGHEST_SCOTUS_KNOWN_SERIAL`, and the daemon naturally re-enters the same branch on the next iteration. Once `highest_ingested == highest_observed`, the recovery branch is skipped and the daemon falls back to the regular forward probe.

A second, related change: `compute_binary_probe_jitter` and `compute_next_binary_probe` previously hard-coded `settings.IQUERY_MAX_PROBE` for both the geometric step cap and the jitter range. SCOTUS docket-number ranges per term are much narrower than PACER's (low ~3000, high ~7000, applications ~2000), so the iquery cap of 32 risks overshooting the active region. Both helpers now accept an optional `max_probe` override (defaulting to `IQUERY_MAX_PROBE` for back-compat); the SCOTUS daemon passes `SCOTUS_MAX_PROBE=16`. Jitter range becomes `[1, 8]` and the linear phase steps in 16-wide chunks instead of 32, which keeps backfill ranges tight while still leaping past sealed/missing serials.

### New settings
- `SCOTUS_FIXED_SWEEP` (default `100`) — per-iteration cap for the sweep/recovery branch.
- `SCOTUS_MAX_PROBE` (default `16`) — geometric step cap and jitter base for the SCOTUS forward probe.

## Deployment

**This PR should:**
    - [x] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [ ] `skip-daemon-deploy`

**Post-deploy, to start filling the 2025–2026 gap:**
1. Determine the highest known serial per sequence/term from supremecourt.gov.
2. `HSET scotus:highest_observed_serial <sequence>:<term> <serial>` for each `(sequence, term)` pair to backfill.
3. The daemon will trickle-ingest at ~100 serials/hour/sequence until `highest_known_serial` catches up, then resume normal probing automatically.

### DIFF
--- a/cl/corpus_importer/management/commands/probe_scotus_dockets_daemon.py
+++ b/cl/corpus_importer/management/commands/probe_scotus_dockets_daemon.py
@@ -220,14 +220,24 @@ def _probe_scotus_sequence(
         int(raw_observed) if raw_observed is not None else highest_ingested
     )
 
-    # Recovery path: a previous run completed the probe but was interrupted
-    # mid-ingestion.  Resume from where we left off without re-probing.
+    # Recovery / sweep path: ``highest_observed > highest_ingested`` either
+    # because a previous run was interrupted mid-ingestion, or because the
+    # operator seeded ``HIGHEST_SCOTUS_OBSERVED_SERIAL`` ahead of the ingested
+    # watermark to backfill a known gap. Resume ingestion without re-probing,
+    # but cap each iteration at ``SCOTUS_FIXED_SWEEP`` serials so a large
+    # backlog (thousands of cases) is processed slowly across many iterations
+    # rather than hammering supremecourt.gov in one burst. The daemon falls
+    # back to forward probing once ``highest_ingested`` catches up.
     if highest_observed > highest_ingested:
+        sweep_cap = highest_ingested + settings.SCOTUS_FIXED_SWEEP  # type: ignore[misc]
+        to_serial = min(highest_observed, sweep_cap)
         logger.info(
-            "SCOTUS %s %s: resuming interrupted ingestion (%s..%s].",
+            "SCOTUS %s %s: %s ingestion (%s..%s] (highest observed: %s).",
             sequence,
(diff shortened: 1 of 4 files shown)


cl/corpus_importer/management/commands/probe_scotus_dockets_daemon.py
MorganBennetDev: This needs to be updated.

MorganBennetDev (changes requested):
Looks good aside from a docstring update. After that, we should be ready to merge.

MorganBennetDev (approved):
Looks good. Thanks for the doc update.

PR #7291, opened 2026-04-29
Show semantic chunks with keyword highlighting in results

### LINKED ISSUE #7290: Show semantic search relevant chunks in frontend results
Parent: #7252

## What
When semantic search returns results on the frontend, display the most relevant text chunk from each opinion instead of the beginning of the opinion text. Reduce the per-page result count for semantic search from 20 to 10 so the longer chunk previews are less overwhelming to scan.

## Why
Semantic search matches against individual chunks of an opinion. Currently, the frontend shows the start of the opinion, which may not contain the text that actually matched the user's query. Showing the relevant chunk gives users immediate context for why the result was returned. Semantic chunks are also more visually dense than the keyword preview text, so a smaller page size keeps the results page readable.

## Acceptance Criteria
- [ ] Semantic search results on the frontend show the matched chunk text
- [ ] Semantic search uses a smaller page size (10) than keyword search (20)
- [ ] Keyword search results are unaffected
- [ ] API behavior is unchanged

rachlllg:
## Fixes
Closes: #7290

## Summary
When semantic search is active, display the most relevant semantic chunk in search results instead of the default first few sentences of the opinion. For hybrid queries (semantic + quoted keywords), quoted phrases are highlighted with `<mark>` tags for visual consistency with keyword search highlighting. Semantic results also use a smaller page size (10 vs 20) since the longer chunk previews are more visually dense than the keyword preview text.

**Changes:**
- Call `merge_semantic_relevant_chunks` in `do_es_search` when semantic params are present
- Branch the search result template to use `linebreaksbr|read_more` for semantic chunks vs the existing `|safe` path for keyword results
- New `highlight_query` template filter in `extras.py` that wraps quoted phrases in `<mark>` tags, running after `read_more` to avoid breaking word-level truncation. Only highlights within text content — HTML tags are left untouched.
- New `SEMANTIC_SEARCH_PAGE_SIZE` setting (default 10), applied in `do_es_search` when the query has semantic params. Keyword search remains at `SEARCH_PAGE_SIZE` (20).
- 10 unit tests for the new filter

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

<details closed>
<details open>
I think a screen recording here might be better:

https://github.com/user-attachments/assets/e8643ba2-defc-4a12-b7a8-509b19222c18

</details>
<details open>

Screenshot 2026-04-29 at 3 38 43 PM
Screenshot 2026-04-29 at 3 38 53 PM

</details>


mlissner:
Looks right to me! One other thing I think we can do it switch to 10 results/page. It should help performance (I think?) and it'll make the results less overwhelming. 

Thanks for working on this!

rachlllg:
Thanks for the reminder @mlissner , added the change to 10 results/page for semantic search mode only (keyword search remains 20).

mlissner:
Thank you, Rachel. This is great.

cl/custom_filters/templatetags/extras.py
```diff
@@ -217,6 +217,53 @@ def render_string_or_list(value: any) -> any:
...
+        return text
+
+    # HTML-escape each phrase so it matches against the already-escaped
+    # text (e.g. & → &amp;), then regex-escape for safe pattern use.
+    pattern = "|".join(re.escape(escape(p)) for p in phrases)
+
+    # Split into HTML tags and text runs; only highlight in text runs
+    parts = _HTML_TAG_RE.split(text)
+    for i, part in enumerate(parts):
+        if not part.startswith("<"):
```
albertisfu, line 263: When a keyword search snippet appears in a hybrid query, it is already highlighted and escaped. This results in broken highlighting in the final output.

It seems the issue is caused by the `linebreaksbr` filter, which escapes the original `<mark>` tags.

Image

After fixing the escaping issue, the word gets highlighted twice:
`<mark><mark>married</mark></mark>`

This duplication is not visible to users, but it’s still not ideal.

cl/custom_filters/templatetags/extras.py
```diff
@@ -217,6 +217,53 @@ def render_string_or_list(value: any) -> any:
...
+        ``<mark>``.
+    """
+    if not query or not text:
+        return text
+
+    text = str(text)
+
+    # Only extract quoted phrases — these are the keyword part of a
+    # hybrid search and the only terms we want to visually highlight.
+    phrases = [p for p in re.findall(r'"([^"]*)"', query) if p]
```
albertisfu: If a user searches for an empty phrase like `copyright " ",` all whitespace gets highlighted.

We should ignore empty phrases during highlighting to prevent this.

Image

cl/custom_filters/templatetags/extras.py
```diff
@@ -217,6 +217,53 @@ def render_string_or_list(value: any) -> any:
...
+    # HTML-escape each phrase so it matches against the already-escaped
+    # text (e.g. & → &amp;), then regex-escape for safe pattern use.
+    pattern = "|".join(re.escape(escape(p)) for p in phrases)
+
+    # Split into HTML tags and text runs; only highlight in text runs
+    parts = _HTML_TAG_RE.split(text)
+    for i, part in enumerate(parts):
+        if not part.startswith("<"):
+            parts[i] = re.sub(
+                f"({pattern})", r"<mark>\1</mark>", part, flags=re.IGNORECASE
```
albertisfu: The current method can match phrases inside other words, which is not desirable.

For example, searching for `"per"` partially highlights `"experience"`.

Image

This happens because word boundaries are not enforced in the regex. A possible fix is:
` re.sub(rf"\b({pattern})\b", r"<mark>\1</mark>", part, flags=re.IGNORECASE)`

cl/custom_filters/templatetags/extras.py
```diff
@@ -217,6 +217,53 @@ def render_string_or_list(value: any) -> any:
...
+
+    # Only extract quoted phrases — these are the keyword part of a
+    # hybrid search and the only terms we want to visually highlight.
+    phrases = [p for p in re.findall(r'"([^"]*)"', query) if p]
+    if not phrases:
+        return text
+
+    # HTML-escape each phrase so it matches against the already-escaped
+    # text (e.g. & → &amp;), then regex-escape for safe pattern use.
+    pattern = "|".join(re.escape(escape(p)) for p in phrases)
```
albertisfu: When searching for two phrases where one is contained within the other (e.g., "married" "married couple"), the longer phrase gets obscured by the shorter one.

This happens because regex alternation matches left-to-right and stops at the first match, not the longest one.

Image
    
A simple solution would be to sort phrases by length (descending):
      ```
      pattern = "|".join(
            re.escape(escape(p)) for p in sorted(phrases, key=len, reverse=True)
        )
      ```

albertisfu (changes requested):
Thanks @rachlllg,  this looks good. I only found a few edge cases in the `highlight_query `filter.

Also here are a couple of potential issues that aren’t directly related but are worth addressing:

If a user submits empty phrases, they’ll still be sent to Inception. Since Inception cleans up phrases, this can result in no text being left to embed, causing the request to fail.
Maybe before sending the query to inception we can validate that, after removing all quotes from the query, there is still some text remaining.

Image

Image


______
- Unbalanced quotes and other syntax errors that are validated in keyword search are currently being bypassed in semantic search. For example:

`"foo bar" baz"` raises `UnbalancedQuotesQuery` in keyword search, and a message is shown to users.

In semantic search, this error is not raised, and the query is sent as-is. This doesn’t seem to break semantic search itself, but highlighting might fail due to unbalanced quotes, since no properly formed phrases are available to highlight.

The issue appears to be in `build_full_join_es_queries`, specifically within:

```
            if has_valid_semantic_query:
                keyword_text_query, child_text_query = build_semantic_query(
                    string_query, child_filters, embeddings
                )
                if not keyword_text_query:
                    child_filters = []
            else:
                child_text_query = build_fulltext_query(
                    child_fields, string_query, only_queries=True
                )
```

`build_fulltext_query` calls this method:

`validate_query_syntax(value, QueryType.QUERY_STRING)`

That validates different syntax errors but `build_semantic_query` doesn't call it.

cl/custom_filters/templatetags/extras.py
rachlllg: This is now caught by checking if there is already a highlight. Only apply highlight if there isn't already a highlight.

cl/custom_filters/templatetags/extras.py
rachlllg: Updated behavior:
whitespace

cl/custom_filters/templatetags/extras.py
rachlllg: Updated behavior:
partial

cl/custom_filters/templatetags/extras.py
rachlllg: Updated behavior:
sort

rachlllg:
Thanks @albertisfu the comments have been addressed. Below is the updated behavior on handling empty quotation and unmatched quotations.

For empty quotation, if the user only passes an empty quotation to the semantic query, it's considered not having a semantic param and treated as searching for a blank keyword.
empty


For unmatched quotations, it now raises the same error as keyword.
unmatched

albertisfu (approved):
Thanks @rachlllg changes looks great and ready to be merged :shipit:

PR #7299, opened 2026-04-30
Fix semantic banner showing docket case name

### LINKED ISSUE #7298: Semantic search banner shows docket case name on docket pages
## What
On docket pages, the semantic search announcement banner displays the docket's case name above the banner text. The banner should not include page-specific context.

## Why
The case name leaking into the banner is confusing — it looks like it's part of the announcement rather than the page title.

## Approach
- Investigate how the banner in `base.html` sources its content
- Ensure the banner renders only its own static content regardless of which page it appears on

Example:
Image

## Acceptance Criteria
- [ ] The semantic search banner does not display any page-specific text (case names, docket numbers, etc.)
- [ ] Banner renders identically on search, docket, opinion, and homepage

rachlllg:
## Fixes
Fixes: #7298

## Summary
The semantic search announcement banner was picking up the page's `title` context variable on docket pages, displaying the case name above the banner text. Added `only` to the `{% include %}` tag to isolate the banner's context, and explicitly pass `request` since the banner needs it for the cookie dismissal check.

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

Screenshot confirming the leaked content is no longer present:
Screenshot 2026-04-30 at 1 47 06 PM

### DIFF
--- a/cl/assets/templates/base.html
+++ b/cl/assets/templates/base.html
@@ -88,7 +88,7 @@
     {% endflag %}
     <!-- Semantic Search Banner -->
     {% flag "semantic_search_banner" %}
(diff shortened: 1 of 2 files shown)


elisa-a-v (approved):
Looks good! Might be worth adding a note about this in the dismissible nav banner comment so other devs know to use `only` to prevent this issue from happening again, but I wouldn't block merging.

rachlllg:
Good idea! Added the comment to the dismissible_nav_banner.html. Will merge after the checks pass.

PR #7303, opened 2026-04-30
feat(user): Shows total api usage and access level

### LINKED ISSUE #7112: Show user's their total API usage and access level
### Describe the issue.

On the API usage page, we just show the last few days for that user: 

https://www.courtlistener.com/profile/api-usage/

We also keep a tally of their total usage. Let's show that to the users too, and if we're feeling clever, let's put a little donate link next to it!

While we're working on this, we should start showing their access level too. They may have several throttle limits (across multiple APIs or within a single API) and we should show them all.

ERosendo:
## Fixes

This fixes #7112 and close https://github.com/freelawproject/internal/issues/914

> [!NOTE]
> This PR is based on `7199-feat-update-throttles-on-membership-events` because it relies on the throttle filtering added there. It should be retargeted to `main` once #7265  merges.

## Summary

Adds two new sections to the `/profile/api-usage/` page, gated by the
`sync_membership_throttles` waffle switch:

- **Your Access Level**: lists the user's effective rate limits (per-user `APIThrottle` overrides, falling back to `DEFAULT_THROTTLE_RATES["user"]`). Layout uses `widthratio` so 1, 2, or 3 rates all lay out cleanly. Blocked users (`0/min` rows) are filtered, we don't advertise blocked as a throughput limit.
- **Your Total API Usage**: lifetime request count from the cumulative `api:v{3,4}.user.counts` Redis sets, paired with a donate CTA on the same row.

The lifetime count is exposed as a new `UserProfile.total_api_usage` property; `UserAdmin.api_calls_count` was refactored to delegate to it so both surfaces share one source of truth.

The existing 14-day "Recent Usage" table is unchanged. New sections are hidden when the waffle switch is off.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

<details closed>
<details open>

- Three rate limits (Default values) layout
image

- Two rate limits layout
image

- Single rate limit layout
image

</details>


mlissner:
Looks amazing! 🤩 

I'll see if the wiki needs updates once this lands.

ERosendo:
@mlissner thanks! I was waiting for the tests to pass before marking it ready for review, but got distracted 😄

cl/users/views.py
```diff
@@ -263,10 +272,41 @@ def view_api_token(request: AuthenticatedHttpRequest) -> HttpResponse:
...
+    show_membership_features = switch_is_active(
+        SYNC_MEMBERSHIP_THROTTLES_SWITCH
+    )
+
+    throttle_rates: list[tuple[int, str]] = []
+    if show_membership_features:
+        overrides = get_all_throttle_overrides(ThrottleType.API)
+        user_rates = overrides.get(request.user.username) or []
+        if not user_rates:
+            raw_default = settings.REST_FRAMEWORK["DEFAULT_THROTTLE_RATES"][
```
albertisfu: This defaults to `DEFAULT_THROTTLE_RATES`. It is only shown if `SYNC_MEMBERSHIP_THROTTLES_SWITCH` is enabled. However, if this flag is not in sync with `USE_NEW_THROTTLE_DEFAULTS_SWITCH`, it could cause confusion for users.
To prevent this, we should use the same `LEGACY_USER_DEFAULT_RATE` until `USE_NEW_THROTTLE_DEFAULTS_SWITCH` is active.

cl/users/models.py
```diff
@@ -216,6 +217,17 @@ def recent_api_usage(self) -> dict[str, int]:
...
+    @property
+    def total_api_usage(self) -> int:
+        """Lifetime count of API requests for this user across v3 + v4."""
+        r = get_redis_interface("STATS")
+        total = 0
+        for api_prefix in ("v3", "v4"):
+            count = r.zscore(f"api:{api_prefix}.user.counts", self.user_id)
+            if count:
+                total += int(count)
+        return total
```
albertisfu: What about using a pipeline to return both counts in a single Redis request?

```suggestion
        r = get_redis_interface("STATS")
        pipe = r.pipeline()
        for api_prefix in ("v3", "v4"):
            pipe.zscore(f"api:{api_prefix}.user.counts", self.user_id)
        return sum(int(count) for count in pipe.execute() if count)
```

albertisfu (changes requested):
Thanks @ERosendo this looks great. I only have a couple of suggestions.

ERosendo:
@albertisfu Thanks for the feedback. I addressed both of your comments. Here’s a screenshot of the updated page with `USE_NEW_THROTTLE_DEFAULTS_SWITCH` turned off:

image

cl/users/tests.py
```diff
@@ -4035,26 +4043,156 @@ def test_api_calls_count_returns_zero_for_unsaved_user(self) -> None:
...
+    ) -> None:
+        """Lifetime total renders with a donate link when count > 0."""
+        mock_redis = MagicMock()
+        mock_redis.zscore.side_effect = lambda key, _: (
+            42.0 if "v3" in key else 0.0
+        )
+        mock_get_redis.return_value = mock_redis
+
+        response = self.client.get(self.url)
+        self.assertContains(response, "Your Total API Usage")
```
Brennan-Chesley-FLP, line 4190:
```suggestion
        self.assertContains(response, "Your Total API Usage")
        self.assertNotContains(response, "No API usage yet")
```

I think this may need to be updated after the pipelining work.

Brennan-Chesley-FLP:
One small test-tightening suggestion. Looks like the changes Alberto asked for are in place. The rest of this looks good, though I am a little leery about having a sole-source-of-truth in Redis, since I've only used it as ephemeral storage before. Not a blocker, but for my own edification, what's our replication setup for it?

mlissner:
>  for my own edification, what's our replication setup for it?

It's a 2-node cluster, I believe. Never had a problem with it. It uses AWS Elasticache?

albertisfu (approved):

Issue #7044, opened 2026-03-02
Expand docket_number_cleaning logic in general


rachlllg:
### Describe the issue.

The docket_number_cleaning logic is currently designed to clean docket_number_raw for case law, but we plan to utilize cleaned docket_number for linking not just case law appellate chain but also RECAP & state content, so we probably should expand the cleaning logic for the different use cases.

### Related Issue

Perhaps it would be valuable to have a discussion first and then we can add more detail, but here is the background context: https://github.com/freelawproject/courtlistener/issues/6361#issuecomment-3974172302

### What type of issue is this?

Research

Issue #6998, opened 2026-02-19
Update the docket_number cleaning logic to update the docket_number_source


rachlllg:
### Describe the issue.

To keep track of which `docket_number` was cleaned, we added a new field `docket_number_source`. The next step is to make sure the field is actually updated when the `docket_number` is cleaned. This means a PR to tweak the docket_number_cleaning logic to also modify the `docket_number_source` field.

### What type of issue is this?

To Do

Issue #6985, opened 2026-02-18
Add "docket_number_source" field to tag the source of a docket_number


rachlllg:
### Describe the issue.

Before turning on the `docket_number` cleaning feature https://github.com/freelawproject/courtlistener/issues/6468 , let's add a field "docket_number_source" with below value choices to track whether the `docket_number` was cleaned by AI or human.

- 0: Original from source (all records will be auto-assigned this value when it first comes in)
- 1: Automated cleaning via heuristics + LLM (the cleaning algorithm will change the value from O to A when automated cleaning was performed)
- 2: Manual cleaning by a human (the field can only be flagged as "M" if manual edits were made to correct for errors from the cleaning algorithm)

This will make it more transparent to a user if the value is cleaned (or original) and the human-cleaned/reviewed docket numbers can serve as datapoints for periodic evaluation of the docket_number cleaning algorithm.

### What type of issue is this?

To Do

PR #7315, opened 2026-05-01
feat(docket cleaning) OCI docket_number_raw introduction

Brennan-Chesley-FLP:
## Summary
This prepares OriginatingCourtInformation and its usages for the docket_number/docket_number_raw split. It populates both docket_number and docket_number_raw from incoming sources until the DOCKET_NUMBER_CLEANING_ENABLED flag is set. Once that happens, only the docket_number_raw will be populated. At that point in time, we expect the docket_number to be populated [via a save hook](https://github.com/freelawproject/courtlistener/issues/7044). 

After that flag is set and we're happy with the results/confident it won't be switched back, we can make another PR to remove the old code and the gates in this PR. 

This PR doesn't touch the template file where [docket_number is referenced](https://github.com/freelawproject/courtlistener/blob/0751a25614748e7ccf999fc0e3f9766f9f75fee8/cl/opinion_page/templates/docket_tabs.html#L354-L366), it looks like it will have the correct behavior throughout without any changes. 

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/scrapers/tests.py
+++ b/cl/scrapers/tests.py
@@ -14,7 +14,7 @@ from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
 from django.core.files.base import ContentFile
-from django.test import SimpleTestCase
+from django.test import SimpleTestCase, override_settings
 from django.utils.timezone import now
 from juriscraper.AbstractSite import logger
 from juriscraper.lib.exceptions import UnexpectedContentTypeError
@@ -41,6 +41,7 @@ from cl.lib.microservice_utils import microservice
 from cl.lib.model_helpers import make_texas_docket_number_core
 from cl.lib.test_helpers import generate_docket_target_sources
 from cl.people_db.factories import PersonFactory
+from cl.recap.mergers import update_docket_appellate_metadata
 from cl.recap.models import (
     PROCESSING_STATUS,
     EmailProcessingQueue,
@@ -80,6 +81,7 @@ from cl.scrapers.utils import (
     get_existing_docket,
     get_extension,
     update_or_create_docket,
+    update_or_create_originating_court_information,
 )
 from cl.search.cluster_sources import ClusterSources
 from cl.search.documents import (
@@ -228,8 +230,14 @@ class ScraperIngestionTest(ESIndexTestCase, TestCase):
         oci = Docket.objects.get(
             case_name="In Re Motion for Consent to Disclosure of Court Records"
         ).originating_court_information
+        # DOCKET_NUMBER_CLEANING_ENABLED should change this behavior
         self.assertEqual(
             oci.docket_number, "09-2222", "New OCI.docket_number was not saved"
+        )
+        self.assertEqual(
+            oci.docket_number_raw,
+            "09-2222",
+            "New OCI.docket_number was not saved",
         )
         self.assertEqual(
             oci.assigned_to_str,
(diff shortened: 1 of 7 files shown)


rachlllg:
> It populates both docket_number and docket_number_raw from incoming sources until the DOCKET_NUMBER_CLEANING_ENABLED flag is set. Once that happens, only the docket_number_raw will be populated. At that point in time, we expect the docket_number to be populated https://github.com/freelawproject/courtlistener/issues/7044.

I just want to point out that, even after `DOCKET_NUMBER_CLEANING_ENABLED` flag is set, we should continue to populate both `docket_number` and `docket_number_raw` from source. The cleaning pipeline will *replace* the existing `docket_number` with the cleaned version. This is because `docket_number` is the only field indexed and is what drives the `docket_number` search, this ensure `docket_number` is always populated and ready for search, even if the cleaning pipeline fails (and it could, because it is LLM driven). This also allows us to roll out the `docket_number` cleaning pipeline in stages (a collection of courts at a time) without needing to handle the `docket_number` save logic in stages as well.

Earlier this year, I added a new field `docket_number_source` which shows where the `docket_number` came from. https://github.com/freelawproject/courtlistener/issues/6985. The intention is this field can be used to identify whether the `docket_number` is original, LLM-cleaned, or manually-cleaned. I still need to implement this https://github.com/freelawproject/courtlistener/issues/6998.

Brennan-Chesley-FLP:
> I just want to point out that, even after DOCKET_NUMBER_CLEANING_ENABLED flag is set, we should continue to populate both docket_number and docket_number_raw from source.

PR is updated to reflect that, which removed some tests and gated behavior.

cl/scrapers/tests.py
```diff
@@ -242,6 +247,11 @@ def test_ingest_opinions_from_scraper(self) -> None:
             "09-1111",
             "Existing OCI.docket_number number changed",
         )
+        self.assertEqual(
+            d_1.originating_court_information.docket_number_raw,
+            "09-1111",
+            "Existing OCI.docket_number number changed",
```
grossir: Message should be
"Existing OCI.docket_number_raw changed"

?

cl/scrapers/tests.py
```diff
@@ -231,6 +231,11 @@ def test_ingest_opinions_from_scraper(self) -> None:
         self.assertEqual(
             oci.docket_number, "09-2222", "New OCI.docket_number was not saved"
         )
+        self.assertEqual(
+            oci.docket_number_raw,
+            "09-2222",
+            "New OCI.docket_number was not saved",
```
grossir: OCI.docket_number -> OCI.docket_number_raw ?

grossir (approved):
Looks good @Brennan-Chesley-FLP . Maybe test messages can be improved

cl/scrapers/utils.py
```diff
@@ -543,6 +543,9 @@ def update_or_create_originating_court_information(
         if not existing_oci.docket_number and lower_court_number:
             existing_oci.docket_number = lower_court_number
             update = True
+        if not existing_oci.docket_number_raw and lower_court_number:
+            existing_oci.docket_number_raw = lower_court_number
+            update = True
```
grossir, line 548: this is called in `cl_scrape_opinions`, so we should remove the `skip-cronjob` label

grossir (commented):
About the skip-cronjob-deploy label

Brennan-Chesley-FLP:
Fixed the label and assertion issues that @grossir ided.

PR #7320, opened 2026-05-05
feat(zoho): Tag records on membership purchase

### LINKED ISSUE #7201: Sync membership data to Zoho and tag existing lead records
When a membership is created, updated, or cancelled, Zoho needs to receive that information so that automated cadences and flows have enough context about each lead.

This issue has two parts:
1. New logic in CourtListener that sends membership data to Zoho whenever a membership lifecycle event occurs
2. A one-time script that backfills membership tier tags onto existing Zoho Lead and Contact records

ERosendo:
## Fixes

Fixes: #7201

## Summary

This PR wires up Zoho tagging when a user purchases a membership. It ensures users are properly tagged whether their Zoho record already exists or is created later.

Key changes: 

- Adds `AddTagsMixin`: a lightweight helper around the Zoho SDK’s TagsOperations.add_tags. It builds the required Tag objects, wraps them in a NewTagRequestWrapper, and sends the request to `/{module}/{record_id}/actions/add_tags`. This follows the [official Zoho SDK example](https://github.com/zoho/zohocrm-python-sdk-8.0/blob/eef7ddbef88a2e5907851bc3f51771a2d47d0cb5/samples/tags/add_tags_to_record.py) and is mixed into both `LeadsModule` and `ContactsModule`.

- Introduces `tag_zoho_record_for_membership` (Celery task): Finds the user in Zoho using CourtListener ID and email. Since each account in our CRM is represented as either a Lead or a Contact, the task checks the Leads module first and then falls back to Contacts to make sure we don’t miss the record. Applies:
   - "Student" for EDU memberships
   - "CL Membership" for all others
   
   If no Zoho record exists yet, the task safely does nothing—this case is handled later by the milestone flow.

- In `cl/donate/api_views.py`, `_handle_membership_creation` now queues the tagging task right after `apply_membership_throttles`. This means every successful `createMembership` webhook will tag the user in Zoho.

- In `LoggingMixin._handle_events`, when an API v4 request hits a milestone and the user has an active `NeonMembership`, we now chain:

  - `create_or_update_zoho_account`
  - `tag_zoho_record_for_membership`

   This covers users who purchased a membership before having a Zoho record. They’re skipped at purchase time, but tagged later when their **Lead** is created.

- Extends response handling in `ZohoModule.handle_api_response` to support `RecordActionWrapper`, which is returned by `add_tags` but wasn’t previously accounted for. This ensures responses from the new mixin are properly parsed without errors.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/api/tests.py
+++ b/cl/api/tests.py
@@ -598,7 +598,7 @@ class ApiEventCreationTestCase(TestCase):
         self.assertEqual(event_descriptions, expected_descriptions)
         mock_zoho_task.delay.assert_not_called()
 
-    @mock.patch("cl.api.utils.tag_zoho_record_for_membership")
+    @mock.patch("cl.api.utils.tag_zoho_record")
     @mock.patch("cl.api.utils.create_or_update_zoho_account")
     @mock.patch(
         "cl.api.utils.get_logging_prefix",
@@ -626,10 +626,10 @@ class ApiEventCreationTestCase(TestCase):
         # alone — no chain, no tag task.
         mock_zoho_task.si.assert_called_once_with(self.user.pk, 1)
         mock_zoho_task.si.return_value.apply_async.assert_called_once()
-        mock_tag_task.si.assert_not_called()
+        mock_tag_task.s.assert_not_called()
 
     @mock.patch("cl.api.utils.celery_chain")
-    @mock.patch("cl.api.utils.tag_zoho_record_for_membership")
+    @mock.patch("cl.api.utils.tag_zoho_record")
     @mock.patch("cl.api.utils.create_or_update_zoho_account")
     @mock.patch(
         "cl.api.utils.get_logging_prefix",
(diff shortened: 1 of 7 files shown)


Brennan-Chesley-FLP:
I think this mostly looks good to me. Based on [this zoho blogpost](https://help.zoho.com/portal/en/community/topic/kaizen-60-get-records-search-records-query-and-bulk-read-apis#:~:text=Due%20to%20the%20indexing%20delay%2C%20it%20is%20advised%20not%20to%20use%20Search%20API%20to%20check%20for%20record%20duplication%20while%20creating%20a%20new%20record) , it looks like the search index has a delay in indexing newly created items. If this is true, we should probably get the id of newly created/updated Zoho Leads/Contacts and pass them along through the celery task chain, instead of doing the follow up searches in `tag_zoho_record_for_membership`. It seems likely that it will fail on the first API milestone that users hit, and only really be effective on the second if the indexing delay is more than a few seconds.

Do we have any Zoho price alerts set up, or anything else that would alarm us if we get thousands of milestone hits a second?

ERosendo:
@Brennan-Chesley-FLP Thanks for the catch on the indexing delay.

I refactored the chain so we don't search at all in the tag step:

- `create_or_update_zoho_account` now returns `(module_name, record_id)` for whichever record it touched (Lead match, Contact match, or newly created Lead).
- Added a new `tag_zoho_record` task that takes that tuple directly and tags by id (no search)
- Rewired the milestone chain in `LoggingMixin` from `.si() -> .s()` so the return value flows through.
- Kept `tag_zoho_record_for_membership` (the search-based task) for webhooks, since there's no upstream task there to pass an id from.

Brennan-Chesley-FLP (approved):
TipTop

PR #7321, opened 2026-05-05
feat(api): Allow any authenticated user to read RECAP Docs API endpoints

ERosendo:
## Summary

This PR removes the `has_recap_api_access` permission gate from the RECAP API endpoints so any authenticated user can read them. Previously, users had to request access and be granted the `has_recap_api_access` permission manually; that workflow is being retired.

The viewset `permission_classes` are switched from `RECAPUsersReadOnly` to DRF's stock `DjangoModelPermissions`, which preserves the existing write protection (POST/PUT/PATCH/DELETE still require `add/change/delete` model perms) while allowing any authenticated user to perform safe-method requests.

Affected viewsets:

- `DocketEntryViewSet`
- `RECAPDocumentViewSet`
- `TagViewSet`
- `PartyViewSet`
- `AttorneyViewSet`
- `PacerDocIdLookupViewSet`

`RECAPUsersReadOnly` is left in `cl/api/utils.py` for now and can be removed in a follow-up.

`DRFRecapPermissionTest` is updated to reflect the new contract: authenticated users reach all endpoints; anonymous users are rejected.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/api/tests.py
+++ b/cl/api/tests.py
@@ -2984,15 +2984,8 @@ class DRFRecapPermissionTest(TestCase):
     @classmethod
     def setUpTestData(cls) -> None:
         # Add the permissions to the user.
-        up = UserProfileWithParentsFactory.create(
-            user__username="recap-user",
-            user__password=make_password("password"),
-        )
-        ps = Permission.objects.filter(codename="has_recap_api_access")
-        up.user.user_permissions.add(*ps)
-
         UserProfileWithParentsFactory.create(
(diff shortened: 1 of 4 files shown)


mlissner (approved):
Looks great, thanks.

PR #7323, opened 2026-05-06
refactor(lib): Adds tiered cache mockable prefix

ERosendo:
## Fixes

Prevents `tiered_cache` test contamination when multiple tests share a Redis instance. Previously, entries persisted across tests due to the hardcoded tiered: key prefix.

## Summary

This PR extracts the Redis key prefix used by tiered_cache into a small, mockable helper: get_tiered_cache_prefix(). This follows the existing pattern used by cl.alerts.utils.get_alerts_set_prefix.

With this change, tests can isolate their Redis namespace by overriding the prefix. As a result, clear_tiered_cache() only removes keys created by the current test class, avoiding cross-test interference and improving reliability in concurrent runs.

Changes:
- `cl/lib/decorators.py`: new `get_tiered_cache_prefix()` returning the default `"tiered"`. Both `tiered_cache`'s key construction and `clear_tiered_cache()`'s Redis `scan_iter` call read the prefix through the helper instead of hardcoding it.

- `cl/api/tests.py`: `ThrottleOverrideIntegrationTest` patches the helper to a class-specific prefix (`tiered_throttle_override_integration_test`) so its `setUp`/`tearDown` calls to `clear_tiered_cache()` clean up only the keys this class produced.

The prefix override is started in `setUpClass`, not via a `@mock.patch` class decorator. Class decorators only wrap `test_*` methods, which means `setUp/tearDown` would still use the unpatched prefix. This caused cleanup to miss test-generated keys during development, prompting the switch to setUpClass with addClassCleanup.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/api/tests.py
+++ b/cl/api/tests.py
@@ -4735,6 +4735,19 @@ class APIThrottleConstraintTest(TestCase):
 class ThrottleOverrideIntegrationTest(TestCase):
     """Integration tests for throttle overrides using the APIThrottle model."""
 
+    @classmethod
+    def setUpClass(cls) -> None:
+        super().setUpClass()
+        # Start at setUpClass so the patched prefix is active during
+        # setUp/tearDown calls to clear_tiered_cache(). A class decorator
+        # would only wrap test_* methods, leaving setUp unpatched.
+        patcher = mock.patch(
+            "cl.lib.decorators.get_tiered_cache_prefix",
+            new=lambda: "tiered_throttle_override_integration_test",
+        )
+        patcher.start()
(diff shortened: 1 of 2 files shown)


Brennan-Chesley-FLP (approved):

Brennan-Chesley-FLP:
A more elegant solution from a more civilized age.

PR #7327, opened 2026-05-06
fix(recap): Retry transient Texas email parse failures (#7324)

MorganBennetDev:
## Fixes

Fixes (hopefully) #7324 

## Summary

Written by Claude

When `_parse_text` fails on a Texas CaseMail docket page (e.g. a transient ASP.NET app pool recycle, mid-write race, or intermediate network device briefly mangling the 302 redirect), the task previously raised ValueError uncaught, leaving the EmailProcessingQueue stuck in IN_PROGRESS while Sentry caught the crash. Wrap the parse step with two short retries (60s countdown) so transients self-heal, and on final exhaustion log a diagnostic with URL / status / length / redirect chain / body preview before marking the EPQ FAILED.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/recap/tasks.py
+++ b/cl/recap/tasks.py
@@ -3786,8 +3786,46 @@ def process_texas_email(self: Task, epq_pk: int) -> None:
     )
     res.raise_for_status()
 
-    docket_parser._parse_text(res.text)
-    docket_data = docket_parser.data
+    try:
+        docket_parser._parse_text(res.text)
+        docket_data = docket_parser.data
+    except ValueError as parse_exc:
+        redirects = [
+            (h.status_code, h.headers.get("location", "")) for h in res.history
+        ]
+        if self.request.retries < 2:
+            logger.warning(
+                "Texas docket parse failed for EPQ %s (retry %s/2): %s | "
+                "url=%s status=%s len=%s redirects=%s",
+                epq.pk,
+                self.request.retries + 1,
+                parse_exc,
+                res.url,
+                res.status_code,
+                len(res.text),
+                redirects,
+            )
+            raise self.retry(exc=parse_exc, countdown=60, max_retries=2)
+
+        logger.error(
+            "Texas docket parse failed for EPQ %s after retries: %s | "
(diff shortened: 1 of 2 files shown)


cl/recap/tasks.py
```diff
@@ -3786,8 +3786,46 @@ def process_texas_email(self: Task, epq_pk: int) -> None:
...
-    docket_parser._parse_text(res.text)
-    docket_data = docket_parser.data
+    try:
+        docket_parser._parse_text(res.text)
+        docket_data = docket_parser.data
+    except ValueError as parse_exc:
+        redirects = [
+            (h.status_code, h.headers.get("location", "")) for h in res.history
+        ]
+        if self.request.retries < 2:
```
Brennan-Chesley-FLP: Is there a reason we have special retry handling for this instead of raising something the auto retry will work with? 

If we timeout twice and then get a redirect, should we not try again?

MorganBennetDev: Ah, this is an oversight on my part. I want to separate the failed 302 retries from the other network errors. Maybe just sleep for 60 seconds and then try one more time within the task instead of using Celery retries?

Brennan-Chesley-FLP: We can. Aside from logging, what do we gain from separating this out from the other types of network errors? We suspect these are transient errors from Texas, right?

MorganBennetDev: It's a rare enough failure case (we'd only encountered it once when I checked sentry this morning) that I don't want it counting against our retry count for more common failures.

Brennan-Chesley-FLP: We can certainly give it an inline retry if you'd like to do that. I suspect there were about 100 of these across the whole backfill, which is rare enough for me.

Brennan-Chesley-FLP (approved):

PR #7332, opened 2026-05-08
fix(audio): Fix feeds

flooie:
Audios indexed before `process_audio_file` finishes (or whose `local_path` ends up None at index time) must not appear in the feed.

## Fixes
Fixes RSS Feeds showing oral arguments that are not yet available.

## Summary

This PR fixes the oral-argument RSS/podcast feeds so they no longer include audios that haven't finished processing yet.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/audio/feeds.py
+++ b/cl/audio/feeds.py
@@ -11,15 +11,6 @@ from cl.search.forms import SearchForm
 from cl.search.models import SEARCH_TYPES, Court
 
 
-def _has_processed_mp3(item) -> bool:
-    """Has the audio been processed yet
-
-    :param item: Audio
-    :return: True or False
-    """
-    return bool(getattr(get_item(item), "local_path", None))
-
-
 class JurisdictionPodcast(JurisdictionFeed):
     feed_type = iTunesPodcastsFeedGenerator
     description = (
@@ -53,8 +44,12 @@ class JurisdictionPodcast(JurisdictionFeed):
             "type": SEARCH_TYPES.ORAL_ARGUMENT,
         }
         search_query = AudioDocument.search()
-        items = do_es_feed_query(search_query, cd, rows=20)
-        return [i for i in items if _has_processed_mp3(i)]
+        return do_es_feed_query(
+            search_query,
+            cd,
+            rows=20,
+            exclude_docs_for_empty_field="local_path",
+        )
 
     def feed_extra_kwargs(self, obj):
         extra_args = {
@@ -118,8 +113,12 @@ class AllJurisdictionsPodcast(JurisdictionPodcast):
             "type": SEARCH_TYPES.ORAL_ARGUMENT,
         }
         search_query = AudioDocument.search()
-        items = do_es_feed_query(search_query, cd, rows=20)
-        return [i for i in items if _has_processed_mp3(i)]
+        return do_es_feed_query(
+            search_query,
+            cd,
+            rows=20,
+            exclude_docs_for_empty_field="local_path",
+        )
 
 
 class SearchPodcast(JurisdictionPodcast):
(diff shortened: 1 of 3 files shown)


cl/audio/feeds.py
```diff
@@ -11,6 +11,15 @@
 from cl.search.models import SEARCH_TYPES, Court
 
 
+def _has_processed_mp3(item) -> bool:
+    """Has the audio been processed yet
+
+    :param item: Audio
```
grossir: its an AudioDocument, not precisely an audio

cl/audio/feeds.py
```diff
@@ -45,7 +54,7 @@ def items(self, obj):
         }
         search_query = AudioDocument.search()
         items = do_es_feed_query(search_query, cd, rows=20)
```
grossir: it would be better to pass `exclude_docs_for_empty_field="local_path"` to `do_es_feed_query. 

This is slightly confusing because local_path could mean any of both fields

https://github.com/freelawproject/courtlistener/blob/6ed7f5cd31b0fd109d5d5350ad1fa7cda9112975/cl/audio/models.py#L95-L109

but it is actually the field we care for

https://github.com/freelawproject/courtlistener/blob/6ed7f5cd31b0fd109d5d5350ad1fa7cda9112975/cl/search/documents.py#L451-L461

grossir (commented):
solution may be easier using  `do_es_feed_query` argument `exclude_docs_for_empty_field`

You can compare the lists on a shell and see they are different


```python
from cl.lib.elasticsearch_utils import do_es_feed_query
from cl.search.models import SEARCH_TYPES, Court

cd = {
            "q": "*",
            "order_by": "dateArgued desc",
            "type": SEARCH_TYPES.ORAL_ARGUMENT,
        }

items_exclude = list(do_es_feed_query(search_query, cd, rows=20, exclude_docs_for_empty_field="local_path"))

items = list(do_es_feed_query(search_query, cd, rows=20))


```

I actually looked a bit deeper into the code and the `exclude_docs_for_empty_field` is not always applied downstream. Not sure why the results list is different.... However, it seems that modifying the filters on the search query generation should work `search_query = AudioDocument.search().filter(Q("exists", field="local_path"))`

https://github.com/freelawproject/courtlistener/blob/6ed7f5cd31b0fd109d5d5350ad1fa7cda9112975/cl/lib/elasticsearch_utils.py#L2629-L2637

cl/audio/feeds.py
flooie: Yeah, that makes sense.  we need to tweak the building helper function because it doesnt apply outside of recap I think so passing in the flag wouldnt work otherwise .

cl/lib/elasticsearch_utils.py
```diff
@@ -2637,6 +2637,8 @@ def build_search_feed_query(
         )
         s = search_query.query(s)
         s = add_highlighting_for_feed_query(s, hl_field)
+    elif exclude_docs_for_empty_field:
+        s = s.filter("exists", field=exclude_docs_for_empty_field)
```
quevon24, line 2641: As a comment: the new elif in elasticsearch_utils.py also affects opinions. Before this PR, the exclude_docs_for_empty_field kwarg was ignored on the non-jurisdiction/non-RECAP path. Now it actually runs, so SearchFeed for OPINION (cl/search/feeds.py:111, where it resolves to "dateFiled") will start dropping opinions without a dateFiled from the feed. Probably what was originally intended, but worth a note

quevon24 (commented):
the changes look good, just one comment

cl/lib/elasticsearch_utils.py
flooie: Yes, seems like it should be a bug fix.  Not sure if we have any opinion clusters without date filed.

quevon24 (approved):
I checked the dev db and there are not records without date filed, so the change shouldn't affect the opinion results.

PR #7336, opened 2026-05-09
feat(audio): replace task throttle with single-dispatcher daemon

### LINKED ISSUE #7270: Transcription throttling is not working as expected
`cl.audio.tasks.transcribe_from_open_ai_api` is decorated with `@throttle_task("1/3m")` (defined in `cl/lib/celery_utils.py:137-186`). The throttle was added in  #6123 to absorb bursts of transcription requests that were causing OpenAI errors. The decorator works by calling `task.retry(countdown=delay)` whenever the rate window is full, where `delay` is computed from a Redis-backed schedule.

I periodically run a script to re-process failed audio standardization and transcription. I noticed bigger than usual numbers (which may be caused by 1) [processing](https://github.com/freelawproject/juriscraper/issues/1877) of bulk ca9 oral arguments and 2) doctor seems to be under pressure from state courts document extraction). Checking the issue inmore depth, I noticed that the transcription throttle is unbounded and that tasks waiting for so long are sure to be killed by a worker restart. 

Currently the throttle key `celery_throttle:schedule:cl.audio.tasks.transcribe_from_open_ai_api` has as value `2026-05-29 05:39:20`. That's a month of ETA. I think this happened because of the combination of 1) and 2) and a celery restart (where the tasks were forgotten, but the ETA was not reset), which the system hasn't seen since the rate limits were put in place. The quick solution is to manually delete the throttle key, which will restart the throttle counter. Anyway, we should prevent this from happening in the future. 

## Problem

Three structural issues in `cl/lib/celery_utils.py:208-336` (`get_task_wait` / `set_for_next_window`) compound:

1. **`schedule_key` has no TTL.** It is set with `r.set(schedule_key, str(new_time))` — never with `ex=`. The only path that resets it is the `delay < n` branch at `cl/lib/celery_utils.py:329`, which is taken only when a new task hits a *full* throttle window *and* finds the schedule already in the past. If the throttle counter ever expires while `schedule_key` is still in the future, the stale value persists indefinitely.

2. **New tasks supplement the stale value instead of starting fresh.** When a task hits the throttle and reads a future-dated `schedule_key`, it adds 180s and writes back. There is no upper bound on the resulting countdown. We have observed tasks scheduled 31 days into the future from a single backlog event.

3. **Eta tasks live in worker process memory.** `task.retry(countdown=N)` republishes the message with `eta=now+N`. With Celery's Redis broker and the default `acks_late=False`, a worker pulls the message, acks it (removing it from the broker), and parks it in an internal timer until the eta fires. A pod restart, OOM kill, or graceful shutdown drops these silently — they are not recoverable from the broker.

### How the failure mode reproduces

1. A burst of N tasks is enqueued faster than the throttle can drain 
2. The first task runs; the next N-1 chain countdowns of 180s, 360s, 540s, …
3. Workers pull each retry message and park it in their in-memory timer at the assigned eta.
4. Some workers restart before the etas fire. Those messages are lost — the audio rows stay at `STT_NEEDED`.
5. `schedule_key` keeps the cumulative future timestamp. The next task that hits a full throttle window inherits that base and gets a multi-day countdown of its own.

### Observed symptoms

- `celery_throttle:schedule:cl.audio.tasks.transcribe_from_open_ai_api` showing a value 31 days in the future while the broker queue list is empty.
- `app.control.inspect().scheduled()` showing transcribe tasks held in worker memory with etas weeks out.
- Duplicate `audio_pk` across multiple held messages (the task's `STT_COMPLETE` early-exit at `cl/audio/tasks.py:133` covers this for API cost, but each duplicate still consumes a throttle slot).
- Audios stuck at `stt_status=STT_NEEDED` after the apparent successful completion of upstream tasks, with no exception in Sentry — they are simply lost from worker memory.

The pattern in #6123's later comments (sporadic `STT_NEEDED` rows clustered around scraper completion times, ~5/day) is consistent with worker-restart loss, not with API errors.

## Solution 1 — Replace the throttle with a DB-backed dispatcher (recommended)

The `Audio.stt_status` field already encodes a work queue. Use it as one.

### Changes

1. Add a transient state on `Audio.STT_STATUSES` (e.g. `STT_QUEUED = 6`) with a `stt_queued_at` timestamp column.
2. **Scraper:** stop chaining `transcribe_from_open_ai_api.si()` in `cl/scrapers/management/commands/cl_scrape_oral_arguments.py:153`. The scraper leaves the audio at `STT_NEEDED`.
3. **Dispatcher:** new periodic job (Celery Beat or a cron-driven management command) that runs every N seconds and:
   - Selects up to K audios with `stt_status=STT_NEEDED` and non-null `local_path_mp3`, oldest first, `SELECT … FOR UPDATE SKIP LOCKED`.
   - Atomically transitions them to `STT_QUEUED` with `stt_queued_at = now()`.
   - Dispatches `transcribe_from_open_ai_api.delay(audio_pk)` for each.
   - Reaps any `STT_QUEUED` row older than a configurable threshold (e.g. 1 hour) back to `STT_NEEDED`.
4. **Task:** remove the `@throttle_task` decorator. On success, set `STT_COMPLETE`. On retryable failure (e.g. `RateLimitError`), let the existing celery retry handle it; on terminal failure, the task already sets `STT_FAILED`/etc. If the task dies without transitioning, the reaper returns the row to `STT_NEEDED`.
5. Keep the `transcribe` management command for manual backfills, but have it call the dispatcher in a loop instead of bypassing it.

### Properties

- **Restart-safe.** All state is in Postgres. Worker death loses at most the in-flight batch; the reaper recovers it.
- **Burst-safe.** Dispatcher tick rate × batch size sets a hard ceiling on dispatch rate.
- **Observable.** Backlog is `SELECT count(*) FROM audio_audio WHERE stt_status=0`.
- **Tunable.** Two knobs (K, N) in settings; no Redis math, no decorator state.
- **Simpler.** Retires `cl/lib/celery_utils.py:208-336` for this use case.

## Solution 2 — Patch the existing throttle

If we want the smallest possible change while keeping the current decorator-driven flow.

### Changes

1. **Bound the countdown.** In `get_task_wait` (`cl/lib/celery_utils.py:208`), cap the returned delay at a configurable maximum (e.g. 30 min). If the computed delay would exceed the cap, raise (so it surfaces in Sentry) or mark the audio as `STT_NEEDED` for later pickup, rather than scheduling weeks out.
2. **Give `schedule_key` a TTL.** On every `r.set(schedule_key, ...)` in `set_for_next_window` (`cl/lib/celery_utils.py:203`) and the supplement branch (`cl/lib/celery_utils.py:335`), also set `ex=int((new_time - now).total_seconds()) + buffer`. The key self-cleans when the schedule it represents has elapsed; stale state cannot accumulate.
3. **Set `acks_late=True`** on `transcribe_from_open_ai_api`. With (1) bounding the countdown to under the broker's `visibility_timeout` (default 1 hour), worker restarts return the message to the broker instead of dropping it. Combined with the existing `STT_COMPLETE` early-exit, occasional duplicate delivery is harmless.

### Properties

- Preserves the current scraper → task chain shape.
- Eliminates the 31-day-stale-key class of bugs.
- Restart safety improves but is not as clean as Solution 1: visibility-timeout redelivery can still cause occasional duplicates, and the throttle still consumes slots for failed/retried tasks.
- Does not address the burstiness on the seeder side — `transcribe.py`'s default `--rpm 50` is still 150× the throttle's drain rate. The seeder's RPM should be lowered to match or a sleep should be added between batches sized to the throttle window.

## Claude's Recommendation

Solution 1. The workload (a few dozen audios per day, ~5 errors/day per #6123 comments) is small enough that a Beat-driven dispatcher is genuinely simpler than the current throttle, and it removes whole classes of failure modes (worker-restart loss, stale schedule, unbounded countdowns) instead of patching them. Solution 2 is a reasonable interim if a larger refactor isn't in scope this sprint, but it spends complexity on a problem the DB-as-queue pattern doesn't have.

grossir:
Remove @throttle_task("1/3m") from process_audio_file and transcribe_from_open_ai_api. Add per-pk Redis dispatch locks in cl/audio/dispatch.py for dedup. Add reenqueue_pending_audio_tasks (one-shot, for manual ops) and reenqueue_pending_audio_tasks_daemon (long-running, single-replica, the only path that enqueues transcription). The daemon caps dispatches per cycle so the cluster- wide transcription rate stays below OpenAI's safe concurrency.

process_audio_file no longer hands off to dispatch_transcribe; the daemon paces it on the next cycle. Trim retries on transcribe_from_open_ai_api: drop RateLimitError, InternalServerError, and ConflictError retries (the daemon reaps them when the dispatch lock expires), keep a short APIConnectionError retry. New retry countdown stays well below the 6h Redis broker visibility_timeout.

Solves #7270

This will need 2 follow up PRs
- infrastructure issue to set the variables
- update the skip daemons deploy logic to add this daemon


cl/audio/management/commands/reenqueue_pending_audio_tasks.py
```diff
@@ -0,0 +1,224 @@
...
+        older_than_hours,
+        newer_than_days,
+    )
+
+    # ---------- Stage 1: process_audio_file ----------
+    process_pks: list[int] = []
+    if not skip_processing:
+        qs = Audio.objects.filter(
+            duration__isnull=True,
+            local_path_original_file__isnull=False,
```
quevon24: `local_path_original_file` is declared without null=True, so the DB column is NOT NULL and this filter is a no-op. If a row exists with duration=None and local_path_original_file="", it'll be dispatched to process_audio_file

cl/audio/management/commands/reenqueue_pending_audio_tasks_daemon.py
```diff
@@ -0,0 +1,148 @@
...
+caps the number of dispatches at ``AUDIO_REENQUEUE_MAX_PER_SWEEP`` so the
+cluster-wide transcription rate stays under OpenAI's safe concurrency:
+
+    rate ≈ AUDIO_REENQUEUE_MAX_PER_SWEEP / AUDIO_REENQUEUE_WAIT
+
+The dispatch helpers in ``cl.audio.dispatch`` enforce per-pk dedup so
+two daemon cycles overlapping (or a manual one-shot run alongside) never
+double-fire for the same audio.
+
+``settings.AUDIO_REENQUEUE_DAEMON_ENABLED`` is a kill switch checked each
```
quevon24: I think settings.AUDIO_REENQUEUE_DAEMON_ENABLED is evaluated once at Django startup, changing the env var on a running pod doesn't update it

quevon24 (changes requested):
The daemon isn't wired into docker/django/docker-entrypoint.sh, so it can't actually be deployed as a container

cl/audio/management/commands/reenqueue_pending_audio_tasks_daemon.py
```diff
@@ -0,0 +1,148 @@
...
+    )
+    shutdown_requested = True
+
+
+def _interruptible_sleep(total_seconds: int) -> None:
+    """Sleep in 1s ticks so SIGTERM doesn't have to wait an hour."""
+    for _ in range(total_seconds):
+        if shutdown_requested:
+            return
+        time.sleep(1)
```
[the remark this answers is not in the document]
grossir: /fp yes the sleep was meant to be there

cl/audio/management/commands/reenqueue_pending_audio_tasks.py
grossir: thanks, changed

cl/audio/management/commands/reenqueue_pending_audio_tasks_daemon.py
grossir: noted in the docstring

quevon24 (approved):
Thanks, LGTM

PR #7340, opened 2026-05-12
feat(oauth): add branded oath consent page template

### LINKED ISSUE #7228: Add branded template to OAuth consent page
Django OAuth Toolkit adds a minimal default view for the OAuth consent page. Create a custom template for this with CourtListener branding.

nadahlberg:
## Fixes

Fixes #7228

## Summary

This PR adds a template to render the `/authorize` route for the Django OAuth toolkit page. This is the page where user's consent when registering a new OAuth application.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

<details open>
<details open>
Screenshot 2026-05-11 at 8 44 58 PM
</details>
<details open>
Screenshot 2026-05-11 at 8 45 05 PM
</details>

### DIFF
--- a/cl/users/views.py
+++ b/cl/users/views.py
@@ -33,6 +33,10 @@ from django.views.decorators.debug import (
     sensitive_variables,
 )
 from django.views.decorators.http import require_http_methods
+from oauth2_provider.models import (
+    get_access_token_model,
+    get_refresh_token_model,
+)
 from rest_framework.renderers import JSONRenderer
 from waffle import switch_is_active
 
@@ -782,6 +786,73 @@ def view_webhooks(request: AuthenticatedHttpRequest) -> HttpResponse:
             "private": True,
             "page": "api_webhooks",
             "sub_page": "webhooks",
+        },
+    )
+
+
+@login_required
+@never_cache
+@require_http_methods(["GET", "POST"])
+def view_oauth_authorizations(
+    request: AuthenticatedHttpRequest,
+) -> HttpResponse:
+    """List the OAuth applications the user has granted access to."""
+    AccessToken = get_access_token_model()
+    RefreshToken = get_refresh_token_model()
+
+    if request.method == "POST":
+        app_id = request.POST.get("application_id")
+        if app_id:
+            AccessToken.objects.filter(
+                user=request.user, application_id=app_id
+            ).delete()
+            RefreshToken.objects.filter(
+                user=request.user, application_id=app_id
+            ).delete()
+            messages.success(request, "Access revoked.")
+        return HttpResponseRedirect(reverse("view_oauth_authorizations"))
+
+    scopes_settings = settings.OAUTH2_PROVIDER.get("SCOPES", {})
+    tokens = (
+        AccessToken.objects.filter(
+            user=request.user, application__isnull=False
+        )
+        .select_related("application")
+        .order_by("-updated")
+    )
+    by_app: dict[int, dict] = {}
+    for tok in tokens:
+        entry = by_app.setdefault(
+            tok.application_id,
+            {
+                "application": tok.application,
+                "first_granted": tok.created,
+                "last_used": tok.updated,
+                "scope_names": set(),
+            },
+        )
+        entry["first_granted"] = min(entry["first_granted"], tok.created)
+        entry["last_used"] = max(entry["last_used"], tok.updated)
+        entry["scope_names"].update(tok.scope.split())
+
+    # One row per Application, aggregated across any access/refresh
+    # tokens the user holds for that application.
+    authorizations = []
+    for entry in by_app.values():
+        entry["scopes"] = [
+            scopes_settings.get(s, s) for s in sorted(entry["scope_names"])
+        ]
+        authorizations.append(entry)
+    authorizations.sort(key=lambda e: e["last_used"], reverse=True)
+
+    return TemplateResponse(
+        request,
+        "profile/oauth_authorizations.html",
+        {
+            "private": True,
+            "page": "api_oauth",
+            "page_title": "Authorized Apps",
+            "authorizations": authorizations,
         },
     )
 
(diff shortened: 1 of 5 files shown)


mlissner:
@claude review once

mlissner:
The "revoke access" link sort of answers a question I've been wondering: "How do I yank Claude's access to my API?" According to your screenshot above, the following URL should make that possible: 

https://www.courtlistener.com/profile/api/

But it doesn't? Do we need another issue for this or is this something I'm not understanding?

nadahlberg:
@mlissner no, you're right, we need a way for users to revoke apps. Maybe a little out of scope for this PR, but can we squeeze it in?

I added another tab to the user profile developer settings:

### With an app:
Screenshot 2026-05-12 at 7 47 48 AM

### Success on revokation state:
Screenshot 2026-05-12 at 7 48 04 AM

### Mobile:
Screenshot 2026-05-12 at 7 47 39 AM

Also fixed the other two smaller issues Claude raised.

mlissner:
We have a bunch of user experiences around deletion in our profiles:
 - Alerts
 - Webhooks
 - This
 - Notes?
 - Visualizations?

If we're going to add this, we should at least make it follow *one* of these, but I'm not sure which and don't have time today to figure this out. If you want to take a look and try to suggest one, that's great, but I don't want to add another UX since we already have, like, five!

nadahlberg:
@mlissner I'm trying to conform this to webhooks but noticing another issue: Webhooks gets its data from a DRF endpoint which counts against the user's rate limit. So if a user has hit their limit for any reason then they'll get an error when visiting the webhooks tab. If they've hit their hourly limit, this page could be blocked for the whole hour.

One workaround is to add `throttle_classes = []` to `WebhooksViewSet` and `WebhookEventViewSet` (and the analogous `OAuthAuthorizationsViewSet` we create). Does this seem like the right approach? Is there any risk removing these API views from the rate limit system given that they are internal / for UI only?

mlissner:
I'm not sure, honestly. I'd have to dig in, but maybe we can ask Alberto when he gets back?

vasilyslevin:
> @mlissner I'm trying to conform this to webhooks but noticing another issue: Webhooks gets its data from a DRF endpoint which counts against the user's rate limit. So if a user has hit their limit for any reason then they'll get an error when visiting the webhooks tab. If they've hit their hourly limit, this page could be blocked for the whole hour.
> 
> One workaround is to add `throttle_classes = []` to `WebhooksViewSet` and `WebhookEventViewSet` (and the analogous `OAuthAuthorizationsViewSet` we create). Does this seem like the right approach? Is there any risk removing these API views from the rate limit system given that they are internal / for UI only?

Agreed on throttle_classes = []. These viewsets are session-authenticated 
and serve only the user's own data. There is no abuse vector. A user cannot 
enumerate other users' OAuth apps. 

Webhooks: Maybe add a comment on the viewset explaining why throttling 
is off, so future devs don't restore it by accident.

On UX: If the rate-limit problem also affects the existing 
webhooks tab,  maybe opening a separate issue would be better.

mlissner:
Alberto, I think we just need your thoughts here, when you have a moment.

cl/users/views.py
```diff
@@ -786,6 +790,73 @@ def view_webhooks(request: AuthenticatedHttpRequest) -> HttpResponse:
...
+        .select_related("application")
+        .order_by("-updated")
+    )
+    by_app: dict[int, dict] = {}
+    for tok in tokens:
+        entry = by_app.setdefault(
+            tok.application_id,
+            {
+                "application": tok.application,
+                "first_granted": tok.created,
```
albertisfu, line 829: Is this label accurate for the value being displayed?

I'm not very familiar with OAuth, but I see we have `ACCESS_TOKEN_EXPIRE_SECONDS` set to 1 hour and `REFRESH_TOKEN_EXPIRE_SECONDS` set to 30 days. Does that mean the token displayed here is reissued every hour?

If so, would it be better to label this as something `Last Issued` instead?

cl/users/views.py
```diff
@@ -786,6 +790,73 @@ def view_webhooks(request: AuthenticatedHttpRequest) -> HttpResponse:
...
+                "last_used": tok.updated,
```
albertisfu, line 830: Is the token `updated` field refreshed on each call, or does this refer to the last time the token was  last refreshed?

albertisfu (changes requested):
This looks great, @nadahlberg. I just left a couple of questions/suggestions.


Regarding webhooks, yeah, right now the new UI throttles are causing an issue. The solution is indeed to add `throttle_classes = []` to `WebhooksViewSet` and `WebhookEventViewSet`.

However, I'm worried that a user could figure out the webhooks endpoint URL (it's not currently listed in `/api/rest/v4/`, but they could still find it by inspecting the code) and start making automated requests and if this endpoint is not throttled, it could be abused.

So I'd also suggest removing token access for this endpoint so that only session auth (via the website) is allowed. For that, you can override `authentication_classes` in `WebhooksViewSet` and `WebhookEventViewSet` with:

`authentication_classes = [ReplicaRoutingSessionAuthentication]`

Considering that `ReplicaRoutingSessionAuthentication` is currently the default session auth class.

I found this is also a problem with the Tags UI. However, since that endpoint is publicly available, the solution might be different. I'll open a separate issue to review it independently.

mlissner:
@nadahlberg, we're getting a few errors related to the webhooks website failing. Can we get this one merged?

nadahlberg:
Hi all, this one is blocked because we want to diagnose why so many oauth apps are being created per user before we add the applications tab to the developer tools page. Instead I'm going to close this and break it up accordingly:

- See PR #7453 for the fix to the webhook api issue
- See PR #7454 for the oauth consent template
- See issue #7455 for adding the applications tab to dev tools page

PR #7341, opened 2026-05-12
feat(oauth): relax limit in DCR to see if fixes

nadahlberg:
Is the rate limit being hit on the DCR endpoint if all of the registrations are coming from a common IP (i.e. anthropic)


chaco-fl (approved):
LGTM

PR #7343, opened 2026-05-13
Fixing Lag Replica monitor

chaco-fl:
The new RDS read replica added a physical replication slot whose NULL confirmed_flush_lsn crashed the monitor, since the code only expected logical slots.


ERosendo (approved):
Thanks @chaco-fl. I tests the new query and it's working as expected. :shipit:

mlissner:
Grooooan. Nice fix.

PR #7348, opened 2026-05-15
feat(opinion page): Add filters & pagination components to docket shell

EstiShay:
## Fixes                                                                                                                                                            
Part of #7068.  

## Summary                                                                                                                                                          
Wires `<c-docket-filter>` and `<c-pagination>` into the v2 docket shell. The filter form's GET params already flow through `view_docket`'s existing form pipeline — this PR makes them visible and adds the bottom pagination control. Adds tests pinning down the wiring contract and tightens two pre-existing `DocketPageV2TemplateTest` assertions that were giving false positives.

**What to review:**
- `cl/opinion_page/templates/v2_docket.html` — the Cotton call sites, especially the c-vars on `<c-docket-filter>` (`:page_obj="docket_entries"` since the view names its page object `docket_entries`) and `<c-pagination>` (each `show_*` toggle is a bare HTML-style boolean attribute — see "Non-obvious decisions" for the reasoning).
- `cl/simple_pages/templates/v2_components.html` — the `<c-pagination>` library demos and inline code examples use the same bare-attribute pattern.
- `cl/opinion_page/tests.py::DocketFilterPaginationWiringTest` — five tests. All call a shared `_get_docket_and_verify_v2()` helper that asserts the waffle flag is enabled in the DB and that the middleware swapped to `v2_docket.html` before each test's own assertions run. Each test's docstring explains the contract it pins.
- `cl/opinion_page/tests.py::DocketPageV2TemplateTest` — newly tightened: `self.assertTemplateUsed(r, "v2_docket.html")` replaces the prior hand-rolled `assertIn` against `r.templates`. Class-level `WAFFLE_CACHE_PREFIX` decorator stays.

**Intentionally not wired up yet:**
- **Entry rows.** The `<div id="docket-entries">` is a placeholder; entry rendering arrives in PR 2 (branch `7068-docket-entries-2`).
- **HTMX partials.** Filter + pagination submit as full-page GETs; htmx swap-in-place is PR 3.
- **`htmx_target` c-var on `<c-docket-filter>`** — left blank in this PR; PR 3 will set it.

**Non-obvious decisions:**
- `test_filter_params_narrow_queryset` reads `r.context["docket_entries"]` instead of HTML. The filter→queryset contract lives in `view_docket` (predates PR 1) and the test guards against a future refactor silently breaking it. The other tests assert against HTML to verify PR 1's template changes specifically.
- Pagination test bulk-creates 211 entries because `Paginator(200, orphans=10)` only spills onto a 2nd page at ≥211 entries.
- One `# type: ignore[misc]` on `docket=cls.docket` in `setUpTestData` — same factory-vs-model pattern already present in ~14 unsuppressed instances in this file. Per-line ignore rather than widening the file-level pragma.
- `v2_docket.html` is asserted via `self.assertTemplateUsed(r, "v2_docket.html")` in `DocketPageV2TemplateTest`'s two tests. Their existing string and context-key assertions all pass against v1 too — "Docket Entries", "28:1331", "Contract" appear in v1's docket template, and `metadata`/`tabs` are set unconditionally in `view_docket` regardless of which template renders. The template-name assertion is what actually distinguishes v2 from v1; without it these tests gave false positives.
- `@override_settings(WAFFLE_CACHE_PREFIX=...)` wraps both v2 test classes, above `@override_flag`. Waffle caches flag lookups in Redis, which parallel test workers share even though each has its own test DB. A worker without `use_new_design` in its DB that calls `flag_is_active("use_new_design")` writes `CACHE_EMPTY` to the unprefixed global cache key; any other worker that reads it then sees the flag as inactive and renders v1. Per-class prefix isolates the namespace. Decorator order matters — outer setting + inner flag, so `@override_flag`'s setUp/tearDown flushes go through the prefixed key. Full mechanism in `DocketPageV2TemplateTest`'s docstring.
- `_get_docket_and_verify_v2()` does two diagnostic assertions before each test's content checks — `Flag.everyone == True` in the DB (the override worked) and `v2_docket.html` rendered (the middleware swapped). If either fails, the message says exactly what broke instead of leaving the reviewer to debug content-assertion failures buried in a multi-KB v1 HTML response. Used by all tests in `DocketFilterPaginationWiringTest` so the diagnostic applies uniformly.
- `<c-pagination>` toggles use bare HTML-style boolean attributes (`show_first`) rather than string form (`show_first="True"`). The component gates rendering with `{% if show_first %}` and any non-empty string is truthy — `show_first="False"` would render identically to `"True"`. Bare attributes bind the c-var to actual `True` at render time, so flipping the attribute is the only way to toggle the control off. Same pattern in `v2_components.html`'s demos so the canonical reference doesn't teach the broken form.
- One `# type: ignore[return-value]` on the helper's `return r`. `self.async_client.get()` returns a type mypy sees as `_MonkeyPatchedASGIResponse` (a private Django test-stub type), functionally `HttpResponse`-compatible but not nominally so. Matches the file's existing convention for unfixable mypy edge cases.

## Test plan
1. **Local preview:** with the `use_new_design` waffle flag on, visit a docket with >210 entries — you should see the filter row up top, an empty entries area, and pagination at the bottom (First / Prev / Page indicator / Next / Last).
2. **Verify filter querystring:** append `?entry_gte=3` to the URL and check `context["docket_entries"]` reflects the filter. (No visible entry rows yet — that's PR 2.)
3. **Verify pagination preserves filters:** with a filter param in the URL, click Next / Page 2 / Last — the querystring should carry forward (not reset to unfiltered).
4. Run the tests:
   ```
   docker exec cl-django python manage.py test cl.opinion_page.tests.DocketFilterPaginationWiringTest cl.opinion_page.tests.DocketPageV2TemplateTest --keepdb
   ```

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

  _The component, tests, and this PR description were drafted with Claude Opus 4.7._

<details closed>
<details open>
image
</details>
<details open>
image
image
image

</details>


cl/opinion_page/tests.py
```diff
@@ -2746,6 +2758,11 @@ async def test_v2_docket_page_renders(self) -> None:
             )
         )
         self.assertEqual(r.status_code, HTTPStatus.OK)
+        # Required: the v2 template was actually used. Strings like
+        # "Docket Entries" / "28:1331" / "Contract" appear in v1 too,
+        # so without this assertion the test passes against either
+        # template — masking a broken `use_new_design` flag override.
```
elisa-a-v: We should probably assert template used in all redesign-related tests going forward, so I'd remove this comment.

cl/opinion_page/tests.py
```diff
@@ -2759,6 +2776,10 @@ async def test_v2_docket_metadata_in_context(self) -> None:
                 args=[self.docket.pk, self.docket.slug],
             )
         )
+        # `metadata` and `tabs` are set unconditionally in view_docket,
+        # so context-key assertions alone don't prove v2 rendered.
+        # Pin the template too.
```
elisa-a-v: I don't think we need this comment either, it should be our standard moving forward that we assert the template used in tests.

elisa-a-v (approved):
Looks good and manual testing confirms everything works (as much as expected given no HTMX yet), so approving!

One thing before merging: could you do a pass over the docstrings in this PR? Some reference "PR 2" which will go stale, and others could be stronger. Focus on describing what the code does and why, so a future reader can understand it without needing the PR history.

Issue #7345, opened 2026-05-14
`recap_documents.is_available` stale when PDF exists in storage bucket under a different `pacer_case_id`


seanthegeek:
### What happened?

On some dockets, documents are available via the RECAP browser extension's "Get this document for free from the RECAP Archive" link when viewing the entry directly in CM/ECF, but on CourtListener those same documents show a "Buy on PACER" button instead of a "Download document" link — and stay that way until someone purchases the document from PACER a second time and the extension re-uploads it to RECAP.

The bug extends to the v4 API: `recap_documents.is_available` is `false` and `filepath_local` is `null`, even though the corresponding PDF is already present in `storage.courtlistener.com`. The file appears to have been uploaded under a `pacer_case_id` that the docket has since been renumbered away from, and whatever process reconciles uploads to recap_document rows does not seem to re-attach it.

The user-visible consequence is that someone has to buy the document on PACER a second time and let the RECAP extension re-upload it before the CL UI / API will admit that CL has it — even though CL has had it since 2026-05-01.

### Expected behavior

Either:

1. `is_available: true` with a populated `filepath_local` (or `filepath_ia`) on `recap-documents/479147010/`, so the UI shows "Download document" and the API returns a fetchable file path; or
2. The recap_document row should track the historical `pacer_case_id` it was uploaded under (e.g. an alias / `pacer_case_id_history` field), so downstream reconcilers can match against either.

### Actual behavior

- UI: the entry on the docket page renders a "Buy on PACER" button rather than a "Download document" link, on both the docket page and the recap_document permalink.
- API: `recap_documents.is_available` is `false`, `filepath_local` is `null`, `filepath_ia` is `""`, `sha1` is `""`, `file_size` / `page_count` / `date_upload` are `null`, and `plain_text` is empty.
- The RECAP browser extension, viewing the same entry directly on CM/ECF, surfaces a "Get this document for free from the RECAP Archive" link pointing at `https://storage.courtlistener.com/recap/gov.uscourts.txsd.1940635/gov.uscourts.txsd.1940635.20.0_1.pdf` — i.e. the extension knows the file is in the archive even though the CL UI does not.
- The file at that storage URL exists and is reachable (HTTP 200, 169,367 bytes, last-modified 2026-05-01).
- API consumers that gate fetches on `is_available` + `filepath_local`/`filepath_ia` (we hit this in a docket-watcher) silently skip the document. The metadata says the file isn't there; the bucket says it is.

### Steps to reproduce

[...]

PR #7351, opened 2026-05-15
feat(api): surface doppelganger sibling PDFs via available_via

seanthegeek:
## Fixes

Read-time mitigation for #7345. Symptom of the long-standing master/sub-docket data-model gap in #2185 (open since 2017). Complementary to — but does not replace — the backward-replication work in #4864.

## Summary

When PACER renumbers a docket's `pacer_case_id` — common in criminal master/sub-docket structures in #2185, where each defendant gets a separate `pacer_case_id` but the same set of `pacer_doc_id`s — the PDF stays in our bucket under the original `pacer_case_id` while the `RECAPDocument` row on the current docket shows `is_available=false`. Users hit "Buy on PACER" for files we already have. Concretely on the reported case (#7345): docket `70789744` has `pacer_case_id=1940634`; the PDF for entry 20 was uploaded under sibling `pacer_case_id=1940635` and is reachable in the bucket but not exposed by the API.

This PR adds a read-only `available_via` field on the v4 `RECAPDocument` serializer. When the row's own `filepath_local` is empty, it surfaces a nested object containing the complete metadata of an available, unsealed sibling `RECAPDocument` sharing the same `(court_id, pacer_doc_id, attachment_number)` but living on a docket with a different `pacer_case_id`. Returns `null` when no such sibling exists or the row already has its own file.

The `available_via` object includes:
- `filepath_local` — the URL string for fetching the file
- `sha1` — the document hash (useful for deduplication)
- `file_size` — bytes
- `page_count` — pages extracted
- `recap_document_id` — the PK of the sibling, in case consumers want to follow the link

This nested shape is forward-compatible with the eventual M-to-M relationship in #2185 (one document, multiple dockets) and eliminates the need for a breaking change later when that model lands.

The matching helper lives in `cl/recap/utils.py` next to `get_main_rds` so it can be reused by the upcoming backward-replication backfill (#4864), which will eventually copy each PDF to all sibling paths and retire this fallback.

### Disclosure

This PR was authored by Claude (Anthropic). The design and code were produced by an AI assistant; a human reviewed and is opening the PR.

### Design: why not overload `filepath_local`?

An alternative was to transparently return the sibling's path when this row is stranded, reusing the existing `filepath_local` field. That would have fixed every existing consumer for free, but at significant cost:

- The serialized path would appear on a row whose own `pacer_case_id` doesn't match — internally inconsistent and a footgun for any code (including FLP's reconcilers) that parses `pacer_case_id` out of paths.
- It would force `is_available` to be flipped to `true` at serialization time while the stored boolean stayed `false`. Bulk DB dumps would disagree with the API on the same row.
- Sibling rows can be sealed or carry mismatched metadata. Inheriting them silently is harder to retract than opting in.
- When backward replication (#4864) lands and gives each sibling its own copy, the path value would silently change (different `pacer_case_id` in it). Consumers that cached or logged paths would see drift.

A separate, opt-in `available_via` field avoids all these pitfalls.

### Performance: avoiding N+1 on nested/list responses

`RECAPDocumentSerializer` is reused by `DocketEntrySerializer.recap_documents` and (transitively) by full-docket payloads. A naive per-row `find_available_sibling_rd` call would fan out to one query per stranded row — fine for `retrieve`, but O(entries × rds) for a full docket render.

To keep this cheap:

- `find_available_sibling_rd_map(rds)` in `cl/recap/utils.py` pools the inputs by `pacer_doc_id` and runs a single `SELECT` to fetch every viable candidate, then matches them back to the inputs in Python.
- `RECAPDocumentListSerializer` and `DocketEntryListSerializer` (wired via `Meta.list_serializer_class`) call the bulk helper once per batch and stash the result in `serializer.context["available_via_map"]`. The entry-level list serializer aggregates RDs across all entries first, so a full-docket render is still one sibling query.
- `get_available_via` reads from the context map when present and falls back to the per-row helper only for single-object retrieves.

Both helpers `order_by("pacer_case_id", "pk")`, so single and bulk paths pick the same sibling deterministically when multiple qualify.

### Scope

- Adds `find_available_sibling_rd` and `find_available_sibling_rd_map` helpers in `cl/recap/utils.py`.
- Adds `available_via` `SerializerMethodField` on `RECAPDocumentSerializer` returning a rich object (not just a string).
- Adds `RECAPDocumentListSerializer` / `DocketEntryListSerializer` for batched sibling lookup, wired via `Meta.list_serializer_class`.
- Adds `DoppelgangerAvailableViaTest` covering helper edge cases, deterministic tie-breaking, ACMS-style hyphenated `pacer_doc_id`s, the bulk-grouping path, an `assertNumQueries(1)` regression guard for the candidate fetch, and `assertNumQueries(0)` on the field method when the context map is preloaded.
- Does **not** change templates (`rd_download_button.html` etc.) or the search-result UI. Those are separate concerns that warrant UI testing; the same helper can be reused in a follow-up PR.
- Does **not** introduce schema changes or any backfill — this is purely a serializer-time fallback.

### Verification

- `pre-commit` passes (ruff + format).
- `mypy --follow-imports=skip cl/recap/utils.py cl/search/api_serializers.py` introduces no new errors (existing errors on unrelated lines unchanged).
- New tests in `cl/recap/tests/tests.py::DoppelgangerAvailableViaTest` cover: sibling found; no sibling exists; sealed sibling skipped; sibling without file skipped; same-`pacer_case_id` self-match excluded; cross-court excluded; mismatched `attachment_number` excluded (both directions); blank `pacer_doc_id` handled; deterministic pick when multiple siblings qualify; ACMS-style hyphenated `pacer_doc_id`s; bulk helper returns the right map for multiple inputs; bulk helper issues a single candidate query; serializer reads from the context map without a query.
- Test suite not executed locally (Docker dev environment not configured). CI will run it.

## Deployment

**This PR should:**
- [x] `skip-celery-deploy`
- [x] `skip-cronjob-deploy`
- [x] `skip-daemon-deploy`

Web tier deploy required (API serializer change). No migrations. No special steps.

### DIFF
--- a/cl/recap/tests/tests.py
+++ b/cl/recap/tests/tests.py
@@ -131,6 +131,7 @@ from cl.recap.tasks import (
 )
 from cl.recap.utils import (
     find_available_sibling_rd,
+    find_available_sibling_rd_map,
     get_court_id_from_fetch_queue,
     send_bad_redaction_email,
 )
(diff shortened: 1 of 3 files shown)


mlissner:
This is set as Draft. Please let me know when ready.

seanthegeek:
Set to ready for review with the disclosure that I don't know much about the project code to attest it's suitably for merging. I simply fed Claude the issues, instructed it to read every comment and related issue, and to present a solution. We discussed a couple possibilities before I decided on this one and instructed it to open this PR.

seanthegeek:
Switched Bach to draft because I prefer the nested approach that was deferred. I'll update this shortly to include that change, then mark it ready for review again.

seanthegeek:
Ready now

mlissner:
Just read through the proposed solution. I hope you didn't take too much time with it. It's not the way we've decided to go. The solution we're doing is to copy the item to the correct rows since storage is cheap, and it doesn't require a DB change. 

You can find a handful of issues related to this. We do some of the copying already, but there are a few more places we need to catch unfortunately.

seanthegeek:
I didn't Just a few AI prompts.

PR #7352, opened 2026-05-16
feat(users): Allow users to reset their own API token

### LINKED ISSUE #7189: Make it possible to change your own API token
Suddenly, people are messing this up every few days. I think it's because we've gotten more popular again. The current solution is that people have to write in, and we have to do some custom Python code to fix it. It'd be better if people could reset their own API tokens. 

We don't want them setting their tokens to whatever they want, so I think we just need a button in the profile pages somewhere that they can click. When they do, it does a POST, and their token is changed. 

We'll want a confirmation step since this is a big deal and can screw up production code. 

I don't think this is urgent, but it's probably worth doing eventually. We could also make a management command for this, so that we could do something like `update_api_token --user 1` and it'd do the right thing.

ERosendo:
## Fixes

This PR fixes #7189 

## Summary

Currently, users who accidentally expose their API token (for example, by committing it to GitHub or leaking it in a screenshot) must contact support and wait for someone to manually rotate the token in production using a Python script. This PR enables users to reset their API token directly from their profile instead.

Key changes: 
- Adds a “Reset API Token” button to the API Token section of the user profile.
- Clicking the button routes the user to a password confirmation page modeled after the existing account deletion flow.
- Upon confirmation, the user’s DRF token is atomically rotated (deleted and recreated), and the user is redirected back to the token page with a success message.
- This PR refactors the existing `AccountDeleteForm` into a reusable `PasswordConfirmForm` so both flows can share the same password confirmation logic.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

<details closed>
<details open>
image
image

</details>

### DIFF
--- a/cl/users/templates/profile/api_token_reset.html
+++ b/cl/users/templates/profile/api_token_reset.html
@@ -17,6 +17,12 @@
        use the old token will fail until you update them with the new
        value.</p>
 
+    <p class="alert alert-info">
+      Your account has made <strong>{{ recent_api_count }}</strong>
+      authenticated API request{{ recent_api_count|pluralize }} in the last
+      five minutes.
+    </p>
+
     <p>Resetting your token is the right move if you suspect it has been
        exposed (for example, committed to a public repository). Otherwise,
        you may want to <a href="{% url "view_api_token" %}">go back</a>.</p>
--- a/cl/users/templates/profile/api_token_reset.html
+++ b/cl/users/templates/profile/api_token_reset.html
@@ -34,7 +34,7 @@
        exposed (for example, committed to a public repository). Otherwise,
        you may want to <a href="{% url "view_api_token" %}">go back</a>.</p>
 
-    <form action="" method="post" class="v-offset-above-3">{% csrf_token %}
+    <form id="reset-token-form" action="" method="post" class="v-offset-above-3">{% csrf_token %}
       <div class="form-group">
         {{ reset_form.password.label_tag }}
         {{ reset_form.password }}
@@ -46,9 +46,17 @@
           </p>
         {% endif %}
       </div>
-      <button type="submit" class="btn btn-danger btn-lg">Reset My API Token</button>
+      <button id="reset-token-btn" type="submit" class="btn btn-danger btn-lg">Reset My API Token</button>
       <a href="{% url "view_api_token" %}" class="btn btn-default btn-lg">Cancel</a>
     </form>
   </div>
   <div class="hidden-xs col-sm-1 col-md-3"></div>
+{% endblock %}
+
+{% block footer-scripts %}
+<script type="text/javascript" nonce="{{ request.csp_nonce }}">
+  $("#reset-token-form").on("submit", function () {
+    $("#reset-token-btn").prop("disabled", true).text("Resetting…");
+  });
+</script>
 {% endblock %}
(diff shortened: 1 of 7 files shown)


mlissner:
This looks great. One random thought: Could you include "Your token has been used XX times in the last five minutes" or something like that in the confirm message? If it's easy, great. If not, pass!

mlissner (commented):
LGTM, but, as ever, I defer to others for final approval. :)

ERosendo:
@mlissner here’s a first attempt at including recent API token activity on the new page:

image

Let me know what you think.

cl/users/templates/profile/api_token_reset.html
```diff
@@ -0,0 +1,47 @@
...
+    <p>After you press the button below, your existing API token will be
+       replaced with a new one. Any scripts, integrations, or services that
+       use the old token will fail until you update them with the new
+       value.</p>
+
+    <p class="alert alert-info v-offset-above-2">
+      Your account has made <strong>{{ recent_api_count }}</strong>
+      authenticated API request{{ recent_api_count|pluralize }} in the last
+      five minutes.
+    </p>
```
mlissner: Can we make this a bit more dynamic? Green or blue for 0 with a check icon and yellow or red for anything > 0?

ERosendo: Sure!

ERosendo: Here’s the updated design when the user has made 0 requests in the last five minutes:

image

And here’s how it looks when the user has made more than 0 requests:

image

cl/users/views.py
```diff
@@ -273,6 +275,41 @@ def view_api_token(request: AuthenticatedHttpRequest) -> HttpResponse:
     )
 
 
+@sensitive_variables("password")
```
albertisfu: Shouldn't this be `sensitive_post_parameters` like in delete_account view?

```suggestion
@sensitive_post_parameters("password")
```

cl/users/templates/profile/api_token_reset.html
```diff
@@ -0,0 +1,54 @@
...
+        {{ reset_form.password }}
+        {% if reset_form.password.errors %}
+          <p class="help-block">
+            {% for error in reset_form.password.errors %}
+              {{ error|escape }}
+            {% endfor %}
+          </p>
+        {% endif %}
+      </div>
+      <button type="submit" class="btn btn-danger btn-lg">Reset My API Token</button>
```
albertisfu: would it be possible to disable this button once it's clicked and the request is in progress? Currently, it's possible to click it multiple times. So if the request is slow, a user can click it multiple times, ending up with multiple `POST` requests being handled.

albertisfu (changes requested):
Thanks @ERosendo this looks great. I just left a couple of small suggestions.

cl/users/templates/profile/api_token_reset.html
ERosendo: Sure! i think i can add a small script for that

albertisfu (approved):
Thanks! This looks good and ready to be merged.

Issue #5915, opened 2025-07-08
Remove/Reuse Recap Alert Waffle Flags and Switch


ERosendo:
As part of the recap alert feature launch, we created waffle flags and a switch to control promotional elements like the banner and tooltip.

Now that the launch is complete, we should consider either removing or reusing the following:

- Flags:
   - [recap_alert_tooltip_enabled](https://www.courtlistener.com/admin/waffle/flag/23/change/)
   - [recap-alerts-home-banner](https://www.courtlistener.com/admin/waffle/flag/22/change/)

- Switch: 
  - [recap-alerts-email-banner](https://www.courtlistener.com/admin/waffle/switch/5/change/)

mlissner:
Seems like switching these is easiest, right, @ERosendo?

rachlllg:
Noticed the recap alert tooltip is still around, should we remove it from the codebase?

<img width="600" height="323" alt="Image" src="https://github.com/user-attachments/assets/757be510-fe00-47c5-b675-1b7ac1ad1281" />
<img width="945" height="486" alt="Image" src="https://github.com/user-attachments/assets/48c12afd-02dd-4a23-8811-f29898bb7257" />

mlissner:
Yep. let's do this. One tooltip at a time!

PR #7358, opened 2026-05-19
chore(search): remove legacy RECAP alert tooltip

rachlllg:
## Fixes
Refs: #5915

This addresses the tooltip portion of #5915. The `recap-alerts-email-banner` switch in that issue is out of scope here; the home banner was already removed previously.

## Summary
The "RECAP Search Alerts" promotional tooltip that surfaced near the alert bell on RECAP search results has run its course. Removes:

- **Template block** in `cl/search/templates/search.html` — the `{% flag "recap_alert_tooltip_enabled" %}` block around the tooltip HTML and the matching script include at the bottom of the file.
- **CSS** in `override.css` — the `.recap-tooltip`, `.recap-tooltip-body`, `.tooltip-header`, and `.recap-tooltip-close` rules. The generic `.tooltip-arrow` and `.tooltip-arrow::after` rules are left in place since they may affect Bootstrap tooltips elsewhere; they can be cleaned up separately if confirmed unused.
- **JS** — `cl/assets/static-global/js/recap_dismissible_tooltip.js` deleted.

The `recap_alert_tooltip_enabled` flag becomes inert after merge; its DB row can be deleted from the Django admin.

Verified locally with the flag ON: tooltip no longer renders, alert bell and search form unchanged.

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

Templates + static CSS/JS only — web deploy required, other tiers can be skipped.

After merge: delete the `recap_alert_tooltip_enabled` waffle flag row via Django admin.

Screenshot 2026-05-18 at 4 54 25 PM


ERosendo (approved):
Thanks, @rachlllg! Looks good to me. 

I compared it against the original [PR](https://github.com/freelawproject/courtlistener/pull/5882) to confirm that none of the changes introduced there remain. I also double-checked that the tooltip-arrow CSS class is only used by this tooltip, so I think it can be safely removed as part of this cleanup. 👍

rachlllg:
Thank you! Merging.

PR #7360, opened 2026-05-19
fix(recap): Drop class_action__in to avoid 500 on FJC API

### LINKED ISSUE #7359: fjc-integrated-database AttributeError: 'bool' object has no attribute 'split'
### Describe the issue.

We have received a couple of sentry errors [COURTLISTENER-D1X](https://freelawproject.sentry.io/issues/7489998871/events/600d85127bdd41678c1bc42683a54286/) related to the  `/api/rest/v4/fjc-integrated-database/` endpoint.  It crashes with a 500 error when a request uses `class_action__in=True` the problem is that `django`-filter can't handle the combination of a BooleanField with an `__in` lookup: under the hood it converts "True" into a boolean before trying to split it on commas which raises up with `AttributeError: 'bool' object` has no attribute 'split'. 

The fix is to simply to drop the `"in"` lookup from `class_action` in `FjcIntegratedDatabaseFilter` leaving  just `"class_action": ["exact"]` since a boolean only has two meaningful values `True` or `False`.

### Related Issue

_No response_

### Screenshots

_No response_

Research

albertisfu:
## Fixes
Fixes #7359

## Summary
`GET /api/rest/v4/fjc-integrated-database/?class_action__in=…` was returning a 500 (`AttributeError: 'bool' object has no attribute 'split'`) for any request that included `class_action__in`. The crash originates inside django-filter: when an `__in` lookup is registered for a `BooleanField`, the dynamically composed CSV widget combines `BaseCSVWidget` with `NullBooleanSelect`. `NullBooleanSelect.value_from_datadict` converts the raw query value (`"True"`, `"False"`, `"1"`, …) to a Python `bool` *before* `BaseCSVWidget` tries to call `.split(",")` on it, raising up with an unhandled `AttributeError`.

`__in` on a boolean adds no real query meaning as only `True`/`False`/`None` are possible, so the simplest fix is to drop the `"in"` lookup from the `class_action` field in `FjcIntegratedDatabaseFilter` (`cl/recap/filters.py`). With `"in"` gone, `class_action__in` is no longer a recognized filter parameter and `UnknownFilterParamValidationBackend` turns the request into a clean **400** with an `"unknown_params"` payload instead of an unhandled 500:

```json
{
    "detail": "Unknown filter parameters are not allowed.",
    "unknown_params": ["class_action__in"]
}
```

A regression test (`UnknownFilterParameterBlockingTests.test_fjc_class_action_in_lookup_returns_400`) was added in `cl/api/tests.py` to lock this behavior in.

`class_action` is the only `BooleanField` in our filtersets that exposed the `"in"` lookup, so no other endpoint is affected.

## Deployment

**This PR should:**
- [ ] \`skip-deploy\` (skips everything below)
    - [ ] \`skip-web-deploy\`
    - [x] \`skip-celery-deploy\`
    - [x] \`skip-cronjob-deploy\`
    - [x] \`skip-daemon-deploy\`

No special deploy steps required.

### DIFF
--- a/cl/api/tests.py
+++ b/cl/api/tests.py
@@ -113,7 +113,10 @@ from cl.people_db.factories import (
     RoleFactory,
 )
 from cl.people_db.models import Attorney
-from cl.recap.factories import ProcessingQueueFactory
+from cl.recap.factories import (
+    FjcIntegratedDatabaseFactory,
+    ProcessingQueueFactory,
+)
 from cl.recap.views import (
     EmailProcessingQueueViewSet,
     FjcIntegratedDatabaseViewSet,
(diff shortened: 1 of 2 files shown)


ERosendo (approved):
LGTM :shipit:

PR #7371, opened 2026-05-20
feat(simple_pages): include account info on Zoho tickets, gate email field

### LINKED ISSUE #7137: Include whether a confirmed account when creating Zoho Desk issue
People use our contact form to send in their support needs, then we spend cycles confirming their accounts: 

https://wiki.free.law/c/courtlistener/support/how-to-confirm-a-users-account

This isn't a good use of time. If they're logged into CL when they send the message, include that fact in the Zoho Desk ticket, and then we don't have to confirm it again. Easy-peasy!

mlissner:
## Fixes
Fixes: #7137

## Summary
This adds information about whether a zoho desk ticket was created by a logged-in user. If so, we don't have to confirm their email address, which makes life a lot easier and we can do the [account confirmation flow](https://wiki.free.law/c/courtlistener/support/how-to-confirm-a-users-account) less frequently. 

Also:
 - removes the old jira code
 - fixes line returns in Zoho Desk tickets
 - stops asking logged-in users for their email on the contact form (we already know it, and it can be confusing if somebody put in a different value than the one on their account).

The following is Claude's explainer of the commits. Skip reading this if you want to just read the code: 

1. **`refactor(simple_pages): consolidate contact form on Zoho Desk`** — drops the `zoho-desk-tickets` waffle gate, the `EmailMessage` fallback to `support@freelawproject.atlassian.net`, the unused `ContactForm.email_subject` helper, and the `target` parameter on `render_email_body`. Tests stop asserting on `mail.outbox` and assert on the Zoho task mock instead.

2. **`feat(simple_pages): tag Zoho tickets with account info, gate email field`** (this is the #7137 work):
   - `ContactForm.__init__` accepts `is_authenticated` and `account_email`. `clean()` rejects an emailed payload from authenticated users and requires email from anonymous users.
   - `_is_sealing_order` consults a new `_effective_email()` helper so government-domain detection still works for logged-in clerks (their account email, not the form's empty field).
   - The Zoho Desk description now includes a `Logged In As: <username> (<email>)` line and an `Email Confirmed: yes/no` line whenever the submitter is authenticated. The redundant `User Email` line is dropped for logged-in tickets.
   - The view passes `user.email` directly to `create_zoho_desk_ticket.delay` for logged-in users — the form's email is never used.
   - `contact_form.html` wraps the email `<input>` in `{% if not user.is_authenticated %}`.
   - Cosmetic side improvements that came with this work: blank line between every rendered field (one paragraph per field) and `\n` → `<br>` inside multi-line user inputs (Message, Technical Description) so user line breaks survive into Zoho's HTML description.

## Deployment

**This PR should:**
- [x] `skip-celery-deploy`
- [x] `skip-cronjob-deploy`
- [x] `skip-daemon-deploy`

Web tier deploy is required (contact form view + template).

Delete `zoho-desk-tickets` waffle flag.

## Test plan
- [x] Verify that it still works locally
- [ ] Test issues post deployment by doing an anonymous contact and a logged-in contact and checking Zoho Desk
- [ ] Monitor if multi-line looks right


cl/simple_pages/forms.py
```diff
@@ -249,6 +262,18 @@ def clean(self) -> dict[str, Any] | None:
...
 
+        # Email: required for anonymous submitters; forbidden for logged-in
+        # submitters (where we use the account email instead).
+        email = cleaned_data.get("email", "")
+        if self.is_authenticated and email:
+            self.add_error(
+                "email",
+                "Don't provide an email address while logged in — we use "
+                "your account's email automatically.",
+            )
```
mlissner: With the template changes, would this surface anywhere? I think not? 

Instead of throwing this as a field error, perhaps it should be a form error?

cl/simple_pages/forms.py
```diff
@@ -336,91 +361,117 @@ def label_for(self, field_name: str, separator: str = ", ") -> str | None:
...
-            lines.append(
-                f"Background (other): {cd.get('partner_background_other', '')}"
-            )
-            lines.append(f"Current Work: {cd.get('partner_current_work')}")
-            team_label = self.label_for("partner_team_size")
-            lines.append(f"Team Size: {team_label if team_label else ''}")
-            lines.append(f"Founded Year: {cd.get('partner_founded_year', '')}")
-            funding_total_label = self.label_for("partner_funding_total")
+        if logged_in_info is not None:
+            confirmed = "yes" if logged_in_info["email_confirmed"] else "no"
```
mlissner:
```suggestion
            confirmed = "Yes" if logged_in_info["email_confirmed"] else "No"
```

cl/simple_pages/views.py
```diff
@@ -432,51 +438,40 @@ async def contact(
...
     else:
         # the form is loading for the first time
         issue_type = request.GET.get("issue_type")
         if issue_type and issue_type.lower() in ContactForm.VALID_ISSUE_TYPES:
             initial["issue_type"] = issue_type.lower()
-        user = await request.auser()  # type: ignore[attr-defined]
-        if isinstance(user, User):
-            initial["email"] = user.email
+        if is_authenticated:
+            # Email field is hidden for logged-in users; don't pre-fill it.
```
mlissner: this comment is weird in the absence of other code populating the email field.

mlissner (commented):
a few suggestions for claude

mlissner:
Addressed in 00be455ef:
- Email-when-logged-in error is now a form-level error (`add_error(None, …)`) so it actually surfaces.
- `Yes`/`No` capitalized.
- Removed the stale comment in the contact view.

Also re-narrowed the user type in the view so the lint error mypy was reporting goes away — turns out propagating a local `is_authenticated` bool doesn't narrow the underlying union for mypy, so I do the isinstance up front and assign `user` + `account_email` + `is_authenticated` per branch.

mlissner:
Note that this page will need tweaking once this is landed: 

https://wiki.free.law/c/courtlistener/support/how-to-confirm-a-users-account

(It refernces this issue by link)

cl/simple_pages/forms.py
```diff
@@ -249,6 +262,19 @@ def clean(self) -> dict[str, Any] | None:
...
+        email = cleaned_data.get("email", "")
+        if self.is_authenticated and email:
+            # The email field is hidden in the template for authenticated
+            # users, so raise this as a form-level error rather than a
+            # field error — otherwise it would have nowhere to render.
+            self.add_error(
+                None,
+                "Don't provide an email address while logged in — we use "
+                "your account's email automatically.",
+            )
```
ERosendo: Two questions about this logic:

- Is the “logged-in user submitting an email” case actually reachable? The template hides the email field for authenticated users via `{% if not user.is_authenticated %}`, so there’s normally no input to fill out. I’m trying to understand what scenario the `clean()` rejection in `forms.py` is protecting against. If this path isn’t realistically reachable from the UI, it might be simpler to remove the form-level validation and just let the view override with `email = account_email if is_authenticated else cd["email"]`.

- I noticed that `contact_form.html` only renders field-level errors since errors are displayed manually for each field. As a result, the user would see the top “There were errors with your submission” banner without any actual explanation. If we keep this validation, we should probably render `{{ form.non_field_errors }}` near the banner so the message is visible.

ERosendo (approved):
Code looks good to me. Just one small question.

cl/simple_pages/forms.py
mlissner: Good call. I made a few tweaks to address this. Thank you.

ERosendo (approved):

PR #7377, opened 2026-05-22
feat(donate): Adds LSO 1 membership level

### LINKED ISSUE #7374: Add LSO membership level
### Describe the issue.

Create a new membership level for Legal Services Orgs
In Neon on the front end this will be available only as a group membership by approval
On the back end the webhook from Neon will send the membership LSO Level 1 tied to individual members
Need to make sure that it doesn't get stuck with "Awaiting Payment" since there is no cost to this membership
The parent group membership will be tied to companies only - and does not have any associated benefits in CL, so does not need to be created in the back end.

Benefits are the same as current Tier 1:
- unlimited docket alerts
- unlimited real time opinion search alerts
- 15 requests in Pray & Pay
- 5 real-time and 10 daily, weekly, or monthly RECAP search alerts
- API Access at rates of 10 calls per minute/75 calls per hour/300 calls per day.

Tied to https://github.com/freelawproject/internal/issues/1007

### Related Issue

_No response_

### Screenshots

_No response_

Research

ERosendo:
## Fixes

This PR fixes #7374 

## Summary

This PR introduces the LSO 1 (Legal Services Organization Tier 1) membership level, an individual tier that includes the same benefits as Tier 1:

- Tier 1 API throttle rates (`10/min`, `75/hour`, `300/day`)
- RECAP alert quotas (5 real-time alerts and 10 daily/weekly/monthly alerts)
- Unlimited docket alerts
- 15 daily Pray & Pay requests

Unlimited docket alerts and Pray & Pay access are already covered by the existing `is_member` validation, so only throttle rates and alert quotas required explicit mappings.

**Key changes:**

- **Treat empty payment status as succeeded for free tiers**:  Generalizes the existing `EDU-tier` behavior that maps an empty `paymentStatus` to `SUCCEEDED` into `_map_payment_status_value`. This makes sure free tiers (`EDU`) and memberships granted manually in Neon without an attached payment record are enabled immediately instead of remaining stuck in "Awaiting payment processing".

- **Add LSO 1 membership level:** Updates the `NeonMembershipLevel` class to add a new level, plus a no-op migration to refresh the model field choices.

- **Applies Tier 1 throttle rates to LSO 1 members:** Maps `LSO_1` to `TIER_1_RATES` in `LEVEL_TO_RATES` so LSO 1 memberships receive Tier 1 API limits.

- **Sets Tier 1 RECAP alert quotas for LSO 1 members:** Adds `LSO_1` to both `RECAP_ALERT_QUOTAS` mappings so LSO 1 members receive Tier 1 alert quotas instead of silently falling back to the free-tier defaults.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


cl/donate/models.py
```diff
@@ -262,6 +263,7 @@ class NeonMembershipLevel:
         (TIER_2, "CL Membership - Tier 2"),
         (TIER_3, "CL Membership - Tier 3"),
         (TIER_4, "CL Membership - Tier 4"),
+        (LSO_1, "LSO 1"),
```
mlissner: I think this value is sometimes shown to users. Do we want to make it more friendly? Maybe:


```suggestion
        (LSO_1, "Legal Services Org. - Tier 1"),
```

mlissner (approved):
Just one thought! Thanks for jumping on this.

cl/donate/api_views.py
```diff
@@ -376,20 +381,6 @@ def _handle_membership_creation(self, webhook_data) -> None:
...
-            if not membership_data["paymentStatus"]:
-                payment_status = MembershipPaymentStatus.SUCCEEDED
-            else:
-                logger.error(
-                    "EDU membership %s for account %s has unexpected "
-                    "payment status: %s. Review manually.",
-                    membership_data["membershipId"],
-                    membership_data["accountId"],
-                    membership_data["paymentStatus"],
-                )
```
albertisfu, line 391: Is there any risk of receiving a paymentStatus value other than `SUCCEEDED` for EDU memberships that should we still review them manually?

I found this related Sentry issue: https://freelawproject.sentry.io/issues/7447069692/events/ec633f5a59f348f8ba461dcdc99ab020/

However, it seems weird because the received value was `succeeded`, so it should not have been raised.

I just want to make sure the values are matched properly and EDU memberships get the correct status all the time.

albertisfu (commented):
Thanks @ERosendo this looks good. I just have a question.

cl/donate/models.py
ERosendo: I just tried this and remembered that the description in our model should match the description in Neon. this membership name in Neon is `LSO 1`.

If we want to update the description here, we’ll also need to update it in Neon.

cl/donate/api_views.py
ERosendo: @albertisfu Oh sorry, I should’ve mentioned this before.

> is there any risk of receiving a paymentStatus value other than SUCCEEDED for EDU memberships that should we still review them manually?

I don’t think so, and the error report you found in Sentry actually confirms that. We added that logger.error to catch any EDU memberships that contained payment information, since the expected behavior was for them not to have any.

So far, the errors show that those memberships either have no payment information or receive the `SUCCEEDED` payment status.

I also removed the check because we recently ran into a similar issue, the one this check was originally intended to fix (missing payment data), for manually added memberships. With this new approach, `_map_payment_status_value` should handle both cases correctly: manually added memberships and memberships with missing payment data.

albertisfu: oh I see. In that case this is the correct behaviour!

albertisfu (approved):
Thanks @ERosendo this is ready to be merged!

PR #7380, opened 2026-05-23
feat(users): Add admin action to refresh API throttles from membership

ERosendo:
## Summary

This PR adds a **Refresh API throttles from active Neon membership** action to the User admin changelist. Replaces the one-off script that previously had to be edited and run from the shell whenever a member's API throttles needed to be resynced.

For each selected user, the action:
- Looks up their current Neon membership.
- If the membership is active (succeeded + not past `termination_date`), calls `apply_membership_throttles` to replace their `MEMBERSHIP`-source `APIThrottle` rows with the rates mapped for that level in `LEVEL_TO_RATES`.
- If the user has no membership / an inactive one / a level not in `LEVEL_TO_RATES`, the user is skipped and surfaced in a `messages.warning` so the admin knows nothing changed.

`MANUAL` throttle rows are never touched. `apply_membership_throttles` only filters by `Source.MEMBERSHIP`.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`
   
## How to use

1. Open the Django admin and go to **Users**.
2. Search for the user(s) whose throttles need a refresh.
3. Open the user's change page and confirm their Neon membership level matches what Neon shows. The action only uses what's in our DB, if the webhook didn't land, refreshing throttles won't fix the underlying mismatch.
4. Back on the user list, check the box next to the user(s).
5. From the **Action** dropdown at the top of the list, pick **Refresh API throttles from active Neon membership**.
6. Click **Run**.

admin_panel


cl/users/admin.py
```diff
@@ -99,6 +101,44 @@ class UserAdmin(admin.ModelAdmin, AdminTweaksMixin):
...
+        skipped: list[str] = []
+
+        for user in queryset.select_related("membership"):
+            try:
+                membership = user.membership
+            except NeonMembership.DoesNotExist:
+                membership = None
+
+            if not membership or not membership.is_active:
+                skipped.append(user.username)
```
albertisfu, line 127: In case the membership is not active and the user still has the throttles belonging to their membership, shouldn't those throttles be reset to the ones belonging to the free tier? Or do we have another process that resets throttles for expired memberships?

Maybe we can use `clear_membership_throttles`? But that deletes the throttles. I'm not sure whether the expected behavior is to delete them or to create free-tier throttles instead.

I think this is a bit complex because we should also take the user's previous state into account. For example, maybe this is an old user for whom the free tier doesn't apply, and instead the legacy tier should apply?

cl/users/admin.py
```diff
@@ -99,6 +101,44 @@ class UserAdmin(admin.ModelAdmin, AdminTweaksMixin):
...
+                continue
+
+            apply_membership_throttles(user, membership.level)
+            refreshed += 1
```
albertisfu: We have two potential cases where `apply_membership_throttles` has no effect, yet refreshed is still increased.

If the `SYNC_MEMBERSHIP_THROTTLES_SWITCH` is not active (which might not be problematic), or if rates is None (the membership tier is not in `LEVEL_TO_RATES`), then we're only logging an INFO message:

`"No throttle mapping for membership level=%s (user=%s); skipping."`

However, this log can easily be missed, since the admin updating throttles will still see the successful message:

`Refreshed API throttles for {refreshed} user(s).`

I'd suggest only logging the successful message for users whose throttles were actually updated.

Maybe we can tweak the return value of `apply_membership_throttles` to return a boolean indicating whether the update was completed or skipped. That way, we can log messages to the admin accordingly, so admins are aware when something went wrong.

cl/users/admin.py
albertisfu: The same applies if the Neon membership is deleted (not sure if that's possible), but this might be a different case from simply checking whether the membership is still active.

albertisfu (changes requested):
Thanks @ERosendo this looks good. I just have a couple of comments/questions.

cl/users/admin.py
ERosendo: > In case the membership is not active and the user still has the throttles belonging to their membership, shouldn't those throttles be reset to the ones belonging to the free tier? Or do we have another process that resets throttles for expired memberships?

In `get_all_throttle_overrides`, we query `membership_throttles` for users with active memberships in CL. If the membership is no longer active, those throttles won’t be included in the output of the method, and the user will automatically fall back to the free-tier throttles defined in our settings.

> Maybe we can use clear_membership_throttles? But that deletes the throttles. I'm not sure whether the expected behavior is to delete them or to create free-tier throttles instead.

Oh no, we’re not creating throttles for the free tier. The free-tier limits are handled through settings.

> I think this is a bit complex because we should also take the user's previous state into account. For example, maybe this is an old user for whom the free tier doesn't apply, and instead the legacy tier should apply?

This is a valid concern. I had the same question some time ago as well. However, it seems the migration logic we ran has worked properly for users so far.

Here’s a quick summary of how we treated existing users during the migration:

```
  - If the user has a membership - Grant according to their tier.
  - If the user does not have a membership:
    - If ≥ 1000 API requests so far, give them 5k/hour rate (our current default; they're grandfathered in for now b/c they may have a commercial agreement)
    - If < 1000 requests so far, move the user to new default for free tier
```

For users with memberships during the migration, they were assigned throttles based on their membership tier, not based on their historical API usage.

cl/users/admin.py
ERosendo: Good catch, thanks!

`apply_membership_throttles` now returns a boolean: `True` when `MEMBERSHIP` throttles are actually written, and `False` when it’s a no-op.

The admin action now relies on that return value instead of blindly incrementing `refreshed`. Users whose throttles couldn’t be updated are collected separately and shown in their own message. the success count now only reflects users whose throttles were actually changed 👍

cl/users/admin.py
albertisfu: Thanks for the detailed explanation. I see now why it's not required to manually reset throttles when a user's membership expires. So this looks correct.

albertisfu (approved):
Thanks for the changes, this looks ready to be merged now! :shipit:

mlissner:
@ERosendo are you connecting with Jenifer on this one so she has the info? Can one of you get it into the wiki somewhere too, please?

ERosendo:
> @ERosendo are you connecting with Jenifer on this one so she has the info?

Yes! I just shared the link to this PR since it includes a video and a step-by-step guide on how to use it.

PR #7386, opened 2026-05-25
fix(mv_latest_opinions): create migration with new definition

### LINKED ISSUE #7382: `scrapers_mv_latest_opinion` materialized view refresh timing out
### Describe the issue.

Another issue discovered by @mlissner is that the update-scraper-material-view command is also failing to complete successfully. While checking Sentry, I found the following unresolved error report:

- [OperationalError: canceling statement due to statement timeout](https://freelawproject.sentry.io/issues/7070002115/?project=5257254&query=is%3Aunresolved%20canceling%20statement%20due%20to%20statement%20timeout&referrer=issue-stream)

Based on the breadcrumbs in Sentry, the query appears to run for about 20 minutes before timing out.

### Related Issue

_No response_

### Screenshots

_No response_

Research

grossir:
Fixes #7382

- Replaces scrapers_mv_latest_opinion query to fix the 20-min statement_timeout failures on update-scraper-material-view cron (#7382, Sentry COURTLISTENER-BJP).
- Old query joined search_docket → search_opinioncluster → search_opinion then GROUP BY + HAVING; cost scaled with search_opinion (tens of millions of rows). The MV worked at create time (2024-11-25) and degraded with growth.
- New query: 2-step CTE — recent-activity check via search_opinioncluster.date_created index, then GROUP BY aggregate over clusters only for the stale courts. Drops search_opinion from the join entirely.

Performance
- EXPLAIN ANALYZE on prod-sized data: ~3.4 min (vs. >20-min timeout on the old query). ~6× headroom against the configured timeout.
- Scales with search_opinioncluster growth rather than search_opinion, which is the faster-growing table.

Semantic changes (intentional)
- Drops the AND in_use filter on search_court. Now reports any court with has_opinion_scraper=True. in_use is used elsewhere (coverage pages, API filter, PACER free-opinions gate) but is not load-bearing for scraper-health monitoring.
- Uses cluster.date_created as a proxy for opinion.date_created. In practice identical — scrapers create cluster + opinion in the same transaction.
- Staleness threshold kept at 7 days to match original HAVING semantics.

Operational notes
- MV is created WITH NO DATA; migration applies instantly. Admin page will error until the next refresh_scrapers_status_view cron tick refreshes it. Materialized view contents are not carried over logical replication (PG docs), so every DB running the migration will pick up data via its own cron.


quevon24 (commented):
The original 0004 populated the view at create time, but this uses WITH NO DATA. Since the admin page (MVLatestOpinionAdmin) and model query the MV directly with no error handling. I think it it will 500 with `materialized view ... has not been populated` between deploy and the first refresh_scrapers_status_view run. Can't we add something to populate it on migration ( just drop WITH NO DATA so it's created populated?) to avoid the errors? Or are we fine living with errors until the next refresh run re-populates it?

grossir:
@quevon24 refreshes are daily . So, using WITH NO DATA just sets the definition on the DB, until the next REFRESH runs... This makes the migration run faster, which will make things easier for the person running it?

The error on the admin view doesn't really matters since the only person using it is the one looking for failing scrapers (me, for now), not regular users

quevon24 (approved):
LGTM

Issue #7376, opened 2026-05-21
Tags UI is being affected by the new throttle rates.


albertisfu:
### Describe the issue.

Similarly to the issue that @nadahlberg found with the webhooks UI in https://github.com/freelawproject/courtlistener/pull/7340#issuecomment-4430932102

There is a similar issue with the Tags UI, which also uses the API internally to function. The new throttle rates can affect this frontend page quite easily after browsing a few tags.

<img width="1163" height="505" alt="Image" src="https://github.com/user-attachments/assets/6a70a888-1dec-4522-a915-c7052b4d4b17" />

We should remove or increase the throttle for the Tags API endpoint used by the UI.

The solution might be a bit more complex than for the webhooks endpoint since this endpoint is publicly available apart from its internal usage. Maybe we can do something like bypassing the throttle when the endpoint is called using Session Auth, while keeping the throttles in place when called using Token Auth.

### What type of issue is this?

To Do

mlissner:
> Maybe we can do something like bypassing the throttle when the endpoint is called using Session Auth, while keeping the throttles in place when called using Token Auth.

That feels like a good fix for Nathan's bug too, yeah? If so, seems worth doing in a few select places?

albertisfu:
> That feels like a good fix for Nathan's bug too, yeah? If so, seems worth doing in a few select places?

We'll for the webhooks bug that Nathan found, the fix I think is a bit simpler since the webhooks endpoint is not public. We can just grant access to it via Session auth using:

`authentication_classes = [ReplicaRoutingSessionAuthentication]`

And disable throttles completely:

`throttle_classes = []`

mlissner:
Alberto, as a quick fix, can/should we just remove throttles on this until we can do it properly? I doubt that'll cause many problems, but it should fix the website in the meantime?

albertisfu:
@mlissner Yeah, we can remove the throttles from this endpoint completely or Eduardo suggested that using a scope would work too. That should fix the problem, and then we can apply final solution for the public and internal usage of this endpoint in next sprint.

In the meantime, we'll apply this temporary fix.

ERosendo:
We’ve added a temporary fix in https://github.com/freelawproject/courtlistener/pull/7400. I’m moving this issue back to the F column for review in the next sprint.

PR #7400, opened 2026-05-27
fix(api): Adds higher dedicated throttle for tag endpoints

ERosendo:
## Fixes

This PR relates to #7376 but it doesn't fully resolve it. 

## Summary

The Tags UI relies on internal API calls for rendering, so it inherits the standard user throttle limits. These limits are low enough that the UI begins failing after a user browses only a few tags.

This PR introduces a dedicated tags throttle scope (1000/hour) and applies it to UserTagViewSet and DocketTagViewSet. By separating this traffic from the general user scope, tag-related requests no longer compete with a user's overall API budget, resolving the throttling issues in production.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/api/utils.py
+++ b/cl/api/utils.py
@@ -887,6 +887,12 @@ def get_recent_api_request_count(user: User, window_seconds: int) -> int:
     return sum(1 for ts in history if ts > cutoff)
 
 
+class TagRateThrottle(UserRateThrottle):
+    """Higher dedicated rate limit for the tag endpoints."""
+
+    scope = "tags"
+
+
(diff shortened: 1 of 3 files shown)


albertisfu (approved):

I'll be monitoring the tags endpoint after this is merged. Currently we're getting a lot of 429 from that endpoint.

PR #7402, opened 2026-05-28
feat(api): Makes throttle cache clear opt-in in apply_membership_throttles

ERosendo:
## Summary

This PR makes the `clear_tiered_cache()` call inside `apply_membership_throttles` opt-in via a new `clear_cache` flag (default `False`).

## Motivation

The admin `refresh_api_throttles` action was reported to be slow when run on multiple users.

While investigating the issue, it became clear that repeatedly calling `clear_tiered_cache()` was likely the most expensive part of the workflow. The action was invalidating the entire tiered cache once per selected user, which is unnecessarily expensive for the batch operation itself and also disruptive for concurrent requests reading from the cache during the loop.

Key changes:

- `apply_membership_throttles(user, level, clear_cache=False)` now only clears the cache when callers explicitly request it.

- The membership webhook flows in `cl/donate/api_views.py` no longer explicitly clear the cache after each update.

- The admin `refresh_api_throttles` action also keeps the default behavior and does not invalidate the cache while processing users in bulk.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/api/tests.py
+++ b/cl/api/tests.py
@@ -5293,11 +5293,20 @@ class MembershipThrottleSyncTest(TestCase):
         clear_membership_throttles(user)
         self.assertTrue(APIThrottle.objects.filter(user=user).exists())
 
-    def test_apply_clears_throttle_cache(self) -> None:
-        """apply_* invalidates the get_all_throttle_overrides cache."""
+    def test_apply_skips_cache_clear_by_default(self) -> None:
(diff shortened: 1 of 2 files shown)


albertisfu (approved):
Thanks @ERosendo this looks good.

PR #7409, opened 2026-05-31
feat(alerts): Replace docket stats with yesterday's alert count on help page

### LINKED ISSUE #7379: Remove stats from the alert templates
I just waited several minutes, including hitting refresh a bunch of times, for the help page for alerts to load: 

https://www.courtlistener.com/help/alerts/

It was bad enough that I checked the code, where I found this: 

```
<p>In the last 24 hours, {{ d_update_count|intcomma }} dockets and {{ de_update_count|intcomma }} docket entries were updated.
  </p>
```

Which the view handles via: 

```
cache_key = "alert-help-stats"
    data = await cache.aget(cache_key)
    if data is None:
        data = {
            "d_update_count": await Docket.objects.filter(
                date_modified__gte=now() - timedelta(days=1)
            ).acount(),
            "de_update_count": await RECAPDocument.objects.filter(
                date_modified__gte=now() - timedelta(days=1)
            ).acount(),
        }
        one_day = 60 * 60 * 24
        await cache.aset(cache_key, data, one_day)
```

Bleh. This is fun, but it degrades the user experience if you're the sucker that has to load the cache, and it's no good for the DB either. It'll be a pain when we move this to the wiki too.

The stats currently read:

```
In the last 24 hours, 87,847 dockets and 392,162 docket entries were updated. 
```

In isolation, these stats don't make a ton of sense either. There are, maybe 200k new filings per day in PACER. As a user, I don't know how we'd update 87k dockets or get nearly 400k docket entry updates. I'm sure the numbers are right, but, yeah, no good. 

A nice replacement might be how many emails we sent yesterday. I think we have that in a fast place that populates the homepage.

mlissner:
## Fixes
Fixes: #7379

## Summary
The alerts help page (`/help/alerts/`) was slow because it hit the database for stats. This tweaks the stats on the page to pull a different one from redis, so it's fast. 

This also adds that stat to the wiki API endpoint so it's there when I migrate these pages to the wiki, and it does some minor refactoring work to consolidate things.

## Detailed Changes if you Prefer It This Way
- **`get_redis_stat_sum`** gains a `start` offset so callers can read a window that skips today's still-filling bucket (`days=1, start=1` → yesterday). Default `start=0` keeps existing 10-day callers unchanged.
- The `alerts.sent` metric name now flows through the existing **`StatMetric.ALERTS_SENT`** constant instead of a magic string, so readers can't drift from the `tally_stat` writers.
- The legacy and v2 alert-help templates + view render `alerts_sent_count`.
- The same stat is exposed on the **`wiki-data`** endpoint (issue notes the help page will move to the wiki), riding its existing one-day cache. Its test is updated.

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/api/tests.py
+++ b/cl/api/tests.py
@@ -218,12 +218,14 @@ class BasicAPIPageTest(ESIndexTestCase, TestCase):
         expected_keys = {
             "court_count",
             "citation_count",
+            "alerts_sent_count",
             "citation_lookup",
             "financial_disclosures",
         }
         self.assertEqual(set(data.keys()), expected_keys)
         self.assertIsInstance(data["court_count"], int)
         self.assertIsInstance(data["citation_count"], int)
+        self.assertIsInstance(data["alerts_sent_count"], int)
         citation = data["citation_lookup"]
         self.assertIn("throttle_count", citation)
         self.assertIn("throttle_period", citation)
--- a/cl/api/views.py
+++ b/cl/api/views.py
@@ -24,8 +24,10 @@ from cl.search.documents import (
 )
 from cl.search.forms import SearchForm
 from cl.search.models import SEARCH_TYPES, Citation, Court, OpinionCluster
+from cl.search.utils import get_redis_stat_sum
 from cl.simple_pages.coverage_utils import build_chart_data
 from cl.simple_pages.views import get_coverage_data_fds
+from cl.stats.constants import StatMetric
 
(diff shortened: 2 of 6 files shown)


ERosendo (approved):
LGTM 🚀

Issue #7372, opened 2026-05-20
Big Redis alert_hits sets are consuming too much memory.


albertisfu:
### Describe the issue.

@ERosendo found that we have some large alert_hits sets consuming a few MB of memory each.

> Top 10 keys by cardinality:
>    MEMBERS  KEY
>    110,655  alert_hits:17640.r
>     86,497  alert_hits:17052.r
>     84,801  alert_hits:12914.r
>     84,799  alert_hits:12928.r
>     84,798  alert_hits:12956.r
>     84,798  alert_hits:12942.r
>     84,797  alert_hits:12970.r
>     84,794  alert_hits:12981.r
>     84,788  alert_hits:12995.r
>     84,172  alert_hits:13695.r

After investigating these alert sets and their content, I found that:

- No alert hits have been scheduled for those alert IDs.
- Those alerts do not have a `date_last_hit`, as they have never actually been hit.
- All of them are RT alerts.
- The reason is that those RT alerts belong to users without an active membership.

So what’s happening here is that those users used the Alerts API to create those alerts. As part of our current design, we decided not to enforce validations when creating alerts via the API. Specifically, we do not check membership status or estimate the alert rate before creating the alert.

For instance, user `207878` has 1366 alerts created, and user `187938` has 279.

Most of those alerts are RT and too broad.

I sent the list of the first 250 `alert_hits` sets with the largest cardinality via Slack, as it might contain sensitive data.

You’ll be able to see that most of them belong to non-members.

So for every document matching the alert, we add a new item to the Redis set, but no alert is actually sent because we do not send alert emails for non-members.

If those users had active webhooks, those would still be sent, but in these cases the users do not have active webhooks either.

So we are basically accumulating items in the sets without ever sending alerts.

We decided to send webhooks in RT for all alert rates, including non-members. That’s the reason we store a case or document ID in the alert set even if no email is sent. However, this is causing those Redis sets to grow too quickly.

We can start solving this issue with a few options:

- Add validations (membership benefits and alert rate estimation) when creating alerts via the API.
- Only add an item (case or document) to the Redis set if the alert was actually sent, either by email or webhook.
- We could probably deactivate all RT alerts created by users without memberships and clean up their related Redis sets.
- This is not very frequent, but when an alert is deleted, the Redis set remains. We should fix this by deleting the related Redis set as well.

@mlissner what do you think about enforcing memberships and other validations when creating alerts via the API? And about cleaning up existing RT alert sets that belong to non-members?

[...]

PR #7410, opened 2026-06-01
feat(alerts): Only record percolator hits when delivered or scheduled

albertisfu:
## Fixes
Related to: #7372

## Summary
This PR reduces the unbounded growth of the Redis `alert_hits` sets that issue #7372 flags as consuming too much memory.

Previously, `percolator_response_processing` always added a triggered document to the `alert_hits` Redis sets via `add_document_hit_to_alert_set`, regardless of whether the alert was actually delivered or scheduled. For overly broad Real-Time alerts owned by non-member users without an active webhook, the hit is neither sent via webhook nor scheduled for an email digest (non-members aren't eligible for RT email alerts), so recording it only made the Redis sets grow without ever being used.

Changes:
- `send_webhook_alert_hits` now returns whether at least one webhook was queued for delivery.
- The hit is recorded in the `alert_hits` sets only when `webhook_sent or schedule_alert` is true.
- Adds `RECAPAlertsPercolatorTest` coverage confirming a non-member RT alert **without** a webhook stores nothing in the Redis set (with a member control proving the docket actually matched the percolator query), while a non-member RT alert **with** a webhook does store it.

This is the first of several PRs under #7372; the others further reduce alert volume by enforcing per-membership alert quotas.

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`


ERosendo (approved):
LGTM :shipit:

PR #7412, opened 2026-06-01
feat(alerts): Reject overly broad alert queries in the Search Alerts API

albertisfu:
## Fixes
Related to: #7372

## Summary
Part of #7372. The frontend (`search.html`) estimates an alert query's frequency over the last 100 days and blocks creation when it averages more than `MAX_ALERT_RESULTS_PER_DAY` hits/day. The Search Alerts API had no such guard, so API clients could create alerts broad enough to flood inboxes and bloat the Redis `alert_hits` sets.

Changes:
- Extract `get_alert_estimation_count()` in `cl.alerts.utils`,  the request-independent core of `api.views.get_result_count`: it validates a query via `SearchForm`, selects the ES document for the search type, and runs `do_es_alert_estimation_query`, returning `(total_hits, case_only_hits)` or `None` when the query can't be validated.
- Refactor `get_result_count` to delegate to that helper, removing the duplicated per-type match logic. The `alert_frequency` endpoint behavior is unchanged.
- `SearchAlertSerializer.validate` runs the estimation whenever the query is being set or changed (create or query-changing update). If hits/day exceeds `MAX_ALERT_RESULTS_PER_DAY` it raises a **400** `ValidationError` whose detail includes the estimated 100-day hit count under the `estimated_hits` key plus a narrow-your-query message on the `detail` key. On success, `to_representation` surfaces the same `estimated_hits` key with the 100 day estimation.
- Adds `AlertAPITests` coverage for the success key, the over-limit rejection, and an end-to-end estimation against ES.

Successful response with estimated_hits:
Screenshot 2026-06-02 at 2 17 36 a m

Rejected overly broad query:
Screenshot 2026-06-02 at 2 17 17 a m

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

No special deploy steps required. This only changes the API serializer and the `alert_frequency` endpoint (web tier).

### DIFF
--- a/cl/alerts/api_serializers.py
+++ b/cl/alerts/api_serializers.py
@@ -1,5 +1,3 @@
-import math
-
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.http import QueryDict
--- a/cl/alerts/api_serializers.py
+++ b/cl/alerts/api_serializers.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.http import QueryDict
 from rest_framework import serializers
@@ -12,12 +13,21 @@ from cl.alerts.models import (
 from cl.alerts.utils import (
     AlertLimitViolation,
     check_alert_limits,
+    get_alert_estimation_count,
     is_match_all_query,
 )
 from cl.api.utils import DynamicFieldsMixin, HyperlinkedModelSerializerWithId
 from cl.search.models import SEARCH_TYPES
 
 FLP_MEMBERSHIP_URL = "https://free.law/membership/"
+
+# The number of days the alert frequency estimation averages over, matching
+# the value used by the alert_frequency endpoint in the frontend.
+ALERT_ESTIMATION_DAY_COUNT = 100
+
+# Key included in the API response with the estimated number of hits the alert
+# query would have produced over the last ALERT_ESTIMATION_DAY_COUNT days.
+ALERT_ESTIMATION_RESPONSE_KEY = "alert_frequency_estimation"
 
 
 class SearchAlertSerializer(
@@ -38,11 +48,15 @@ class SearchAlertSerializer(
 
     def validate(self, attrs):
         """Validate the query type, set a default for alert_type if not
-        provided, and enforce the user's alert creation quotas."""
+        provided, enforce the user's alert creation quotas, and reject alerts
+        whose query is estimated to be too broad."""
 
+        # The query provided in this request, if any. Used to decide when to
+        # re-run the frequency estimation (only when the query is being set).
+        request_query = attrs.get("query")
         # Get the query from the request or fall back to the instance, as done
         # during PATCH requests.
-        query = attrs.get("query") or getattr(self.instance, "query", "")
+        query = request_query or getattr(self.instance, "query", "")
         if query:
             match_all_query = is_match_all_query(query)
             if match_all_query:
@@ -56,6 +70,11 @@ class SearchAlertSerializer(
         # Enforce the membership and per-rate alert quotas, mirroring the
         # checks performed by CreateAlertForm.clean_rate in the frontend.
         self._validate_alert_quota(attrs)
+
+        # Estimate the alert frequency and reject overly broad queries. Only
+        # run when the query is being created or changed.
+        if request_query:
+            self._validate_alert_frequency(request_query)
         return attrs
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


cl/alerts/api_serializers.py
```diff
@@ -164,6 +185,51 @@ def _validate_alert_quota(self, attrs):
...
+        :param query: The alert's search query string.
+        :return: None
+        :raises ValidationError: If the query is estimated to be too broad.
+        """
+        qd = QueryDict(query.encode(), mutable=True)
+        estimation = get_alert_estimation_count(qd, ALERT_ESTIMATION_DAY_COUNT)
+        if estimation is None:
+            # The query couldn't be validated for estimation; skip the check.
+            return
+
```
ERosendo, line 235: Should we raise an exception here instead? It looks like `get_result_count` returns `{"error": "Invalid SearchForm"}` when the form is invalid.

cl/alerts/api_serializers.py
```diff
@@ -164,6 +185,51 @@ def _validate_alert_quota(self, attrs):
...
+        # Use the first value, as it is the broadest value present across all
+        # search types.
+        total_hits = estimation[0]
+        hits_per_day = math.floor(total_hits // ALERT_ESTIMATION_DAY_COUNT)
```
ERosendo: I think `math.floor()` might be redundant here since the `//` operator already performs floor division.

cl/alerts/api_serializers.py
```diff
@@ -164,6 +185,51 @@ def _validate_alert_quota(self, attrs):
...
+            raise serializers.ValidationError(
+                {
+                    ALERT_ESTIMATION_RESPONSE_KEY: total_hits,
+                    "detail": (
+                        f"This query averages about {hits_per_day} results per "
+                        "day, which is more than our system can support. Please "
+                        "narrow your query to have fewer results per day."
+                    ),
+                }
+            )
```
ERosendo, line 248: I think the shape of the error response feels a little off for a couple of reasons.

First, a consumer reading estimated_hits would now need to handle both `int` and `list[int]` depending on whether the request succeeded or failed.

Second, having detail as a key inside a field-errors dictionary reads a bit like an error on a field named detail, which feels slightly confusing.

I wonder if we could simplify this by returning a single error message instead. Something along these lines:

```python

from rest_framework.exceptions import APIException

class BroadAlertQueryError(APIException):
    status_code = 403
    default_detail = (
        "This query is estimated to produce more results than our system can support. "
        "Please narrow your query to have fewer results per day."
    )
    default_code = "broad_alert_query"
```

And then:

```python
        if hits_per_day > settings.MAX_ALERT_RESULTS_PER_DAY:
            message = (f"This query averages about {hits_per_day} results per "
                        "day, which is more than our system can support. Please "
                        "narrow your query to have fewer results per day.")
            raise BroadAlertQueryError(message)

```

The error response would then look like:

Image

Feel free to ignore this suggestion if you prefer the current approach, just wanted to share another option that might keep the response shape a bit simpler for API consumers.

ERosendo (changes requested):
The code looks good overall. I just left a couple of small comments to address before merging.

cl/alerts/api_serializers.py
albertisfu: Yeah, this is better and more reliable. In the frontend, if an alert query is invalid, the button to create the alert is not displayed, so users can't create alerts with invalid queries. However, through the API, users could still create those kinds of invalid alerts.

I've tweaked this so it now raises:

raise serializers.ValidationError(
    {
        "query": "This query is invalid and can't be used for an alert."
    }
)

cl/alerts/api_serializers.py
albertisfu: Thanks. I've removed `math.floor`

cl/alerts/api_serializers.py
albertisfu: Yeah, we wanted to include the `estimated_hits` key in both successful responses and responses where the estimated count exceeds the threshold.

That said, I agree it's a bit odd for `estimated_hits` to be an integer in successful responses and a list in error responses. I've tweaked your suggestion so that `BroadAlertQueryError` returns an integer for `estimated_hits` instead of a list. 

I'll open a proposal to update the wiki so we can document these new keys.

albertisfu:
Thanks @ERosendo I've addressed your suggestions.

I noticed that when running estimation queries through the validation endpoint, errors can occur due to Elasticsearch requests, as well as query syntax errors that are already handled in search, such as `UnbalancedQuotesQuery` and others.

So I've added logic to catch these exceptions and raise either an `ElasticServerError` or an `ElasticBadRequestError`, as appropriate. This ensures that such errors are handled properly if they occur when users try to create alerts.

Let me know what you think.

cl/alerts/api_serializers.py
ERosendo: Thanks! The new error response looks much better.

ERosendo (approved):
LGTM :shipit:

albertisfu:
Thanks @ERosendo !

@mlissner I've submitted a proposal for changes to the API Alerts documentation page in the wiki that covers the new `estimated_hits` field in the response, as well as the new validation errors introduced for overly broad queries and other syntax-related errors.

https://wiki.free.law/c/courtlistener/help/api/rest/v4/alerts/proposals/12/

PR #7411, opened 2026-06-01
feat(alerts): Enforce alert quotas in the Search Alerts API

albertisfu:
## Fixes
Related to: #7372

## Summary
Part of #7372. The frontend `CreateAlertForm.clean_rate` limits how many RECAP/DOCKETS alerts a user can create per rate based on their membership level, and restricts Real-Time alerts to members. The Search Alerts **API** had no equivalent checks, letting API clients bypass these limits and create the overly broad alerts that grow the Redis `alert_hits` sets.

This PR contains two commits:

**1. `feat(alerts): Enforce alert quotas in the Search Alerts API`**
- Ports the `clean_rate` business logic to `SearchAlertSerializer`:
  - RT alerts require RT eligibility (membership or the unlimited-alerts flag).
  - RECAP/DOCKETS alerts are capped per rate via `RECAP_ALERT_QUOTAS`, skipping users with the unlimited flag and excluding the alert being edited from the count.
- Over-limit / ineligible requests raise `PermissionDenied` (HTTP **403**) with a membership-appropriate message, distinct from the 400s used for malformed input.
- `AlertAPITests` gains profile-backed users + member fixtures and covers: non-member RT rejection, member RT success, the unlimited-flag bypass, free-user and per-tier RECAP quota enforcement, non-RECAP alerts not counting toward the quota, and edits excluding themselves from the count.

**2. `refactor(alerts): Extract shared alert limit business logic`**
- Extracts the decision logic shared by `CreateAlertForm.clean_rate` and `SearchAlertSerializer` into `cl.alerts.utils.check_alert_limits`, which returns an `AlertLimitResult` (an `AlertLimitViolation` enum value or `None`, plus the free-tier quota for messaging).
- Each caller only translates the result into its own presentation: the form raises HTML `ValidationError`s; the serializer raises plain-text `PermissionDenied`. Behavior is unchanged; the duplicated logic now lives in one place.

## Deployment

**This PR should:**
- [ ] `skip-deploy` (skips everything below)
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

No special deploy steps required. This only changes the API serializer and the alert form (web tier).

PR #7413, opened 2026-06-01
feat(alerts): Bypass alert quota for commercial-agreement users

albertisfu:
## Fixes
Related to: #7372

## Summary
Part of #7372. Adds a tier for users with a commercial agreement, allowing them to create as many alerts as they want, subject only to the `ALERTS` API throttle rate limits. So this bypasses the standard alert quotas for free and member users.

In summary this works like:
- **Members and Non-members** → bound by their free and membership quota set in https://github.com/freelawproject/courtlistener/pull/7411
- **Commercial-agreement users** (with an `APIThrottle` of type `ALERTS`) → unlimited alert count, capped only by their alert throttle rate.

Changes:
  - Add `AlertThrottle`(scope `"alerts"`, type `ALERTS`). The alerts scope has no default rate, so it never throttles regular users; only users with an `ALERTS` `APIThrottle` row are rate-limited, at their configured rate.
  - Add `has_throttle_override(user, throttle_type)`, reusing the cached `get_all_throttle_overrides` so MANUAL/MEMBERSHIP precedence and membership expiry are honored.
  - Override `SearchAlertViewSet.get_throttles()` to route throttling per user and request method, instead of stacking all throttles on every request:
    - Commercial-agreement users (those with an `ALERTS` `APIThrottle` row) have their alert **writes** (POST/PUT/PATCH) governed only by the `AlertThrottle` rate, bypassing the global per-user API throttle. Otherwise the global throttle (e.g. `5/min`) would bind first and the configured alert rate (e.g. `100/min`) would never take effect.
    - Everyone else, and commercial users' **reads** (GET/LIST/DELETE), fall back to the standard global `ExceptionalUserRateThrottle`. 
  - In `check_alert_limits`, bypass the membership/non-member quota when the user has an `ALERTS` throttle override. This applies to both the API serializer and the frontend form, which share that helper.
  - Adds `AlertAPITests` coverage for: the quota bypass; the `AlertThrottle` rate-limiting a commercial user's writes (429); commercial writes bypassing a tight global throttle; non-commercial users remaining bound by the global throttle; and commercial reads not consuming the write budget.

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

Includes a no-op DB migration (`api/0018_alter_apithrottle_throttle_type_and_more_noop`) that adds the new ALERTS `APIThrottle` type.


mlissner:
Cool! When this one merges, let's make sure to loop back with me, Eri, and Jessica on how to apply alert throttles to an account.

albertisfu:
> Cool! When this one merges, let's make sure to loop back with me, Eri, and Jessica on how to apply alert throttles to an account.

Sure! Once https://github.com/freelawproject/courtlistener/pull/7411 gets merged, we'll start enforcing membership quotas for all users creating alerts via the API (previously, quotas were only enforced by the frontend).

Would it be enough to coordinate merging that PR and then merge this one immediately afterward so we can create the new `AlertThrottle` for users with a commercial agreement?

Alternatively, if you think this could significantly impact users with commercial agreements who actively create alerts, we could modify these PRs to enable membership quota enforcement behind a waffle flag. That would give us more time to merge this PR and configure the new `AlertThrottle`.

btw one change I just committed is that the new `AlertThrottle` will only bypass the regular API throttles for alert creation and modification methods (POST, PUT, and PATCH). Other methods will continue to be governed by the standard API throttles.

This means that when an `AlertThrottle` exists, it takes precedence over the regular API throttles for creating and editing alerts, allowing users to perform those actions as quickly as their alert throttle permits. All other requests will still be subject to the regular API throttles.

Creating the new `AlertThrottle` for a user follows the same process as creating a regular API throttle. Just select the new `Alerts` throttle type and set the source to Manual.

Screenshot 2026-06-02 at 4 10 54 p m

mlissner:
> btw one change I just committed is that the new AlertThrottle will only bypass the regular API throttles for alert creation and modification methods (POST, PUT, and PATCH). Other methods will continue to be governed by the standard API throttles.

Perfect, thank you. Really good decision there. 

> Would it be enough to coordinate merging that PR and then merge this one immediately afterward so we can create the new AlertThrottle for users with a commercial agreement?

Yep, that'll be fine. Nobody has an alert agreement yet, so if things start breaking, I think that should be fine.

ERosendo (approved):
LGTM 👍

PR #7417, opened 2026-06-02
fix(corpus_importer): keep free-opinion scraper advancing past failures

### LINKED ISSUE #7316: scrape_pacer_free_opinions: SDNY stalled since 2025-11-26, peer districts lag 2-3 months — likely caused by slow WrtOpRpt.pl + bail-on-error per-court loop
> I noticed CL was missing recent SDNY opinions while running a citation-verification benchmark and spent some time with Claude digging into what was going on. The notes below are our joint observations and a hypothesis about the cause — flagging in case it's useful. The \"What I observed\" section is directly verified; \"Most likely explanation\" is unconfirmed.

**Severity:** Medium

### What I observed

**(1) SDNY (`court=nysd`) ingest hasn't advanced past 2025-11-26 in 5 months.**

Goldenberg v. NeoGenomics ([docket 66615404](https://www.courtlistener.com/docket/66615404/goldenberg-v-neogenomics-inc/)), `1:22-cv-10314`, entry #80 — Opinion & Order filed 2026-03-13:

```
GET /api/rest/v4/recap-documents/472092124/
{ "pacer_doc_id": "127039233429",
  "is_free_on_pacer": null,
  "is_available": false,
  "date_upload": null }
```

Direct PACER lookup of `pacer_doc_id 127039233429` shows price \$0.00 — the court has flagged it free, but CL hasn't picked that up. Sorting `type=o&court=nysd&stat_*=on` by `dateFiled desc` confirms the most recent SDNY opinion in CL is dated 2025-11-26.

**Positive control (nyed):** `rd 468902270` (entry filed 2026-02-16) has `is_free_on_pacer=true`, `is_available=true`, `date_upload=2026-05-01`. The importer is working for at least some peer courts.

**(2) NYSD's `WrtOpRpt.pl` is slow but functional.** A direct logged-in browser query of `https://ecf.nysd.uscourts.gov/cgi-bin/WrtOpRpt.pl` takes ~3 minutes for a one-month date range. Querying just April 2026 returns actual results — so PACER does have free SDNY opinions filed after 2025-11-26; the report just takes a long time to generate.

**(3) Peer districts (nyed, cand) respond quickly.** Same browser test on `ecf.nyed.uscourts.gov/cgi-bin/WrtOpRpt.pl` and `ecf.cand.uscourts.gov/cgi-bin/WrtOpRpt.pl` returns April 2026 results promptly. So the slowness isn't generic to PACER — it's specific to NYSD's instance.

**(4) Peer districts also lag, just less.** Most recent CL `date_filed` per court (via cluster API):

| Court | Most recent `date_filed` in CL | Lag from 2026-05-01 |
|---|---|---|
| nyed | 2026-02-16 | ~2.5 mo |
| cand | 2026-01-22 | ~3 mo |
| txnd | 2026-03-03 | ~2 mo |

Combined with (3), this rules out upstream curation lag as the cause of peer-court lag — the upstream is current; the lag is on CL's side.

### What the source shows

Reading `cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py`:

```python
for _start, _end in dates:
    exc = fetch_doc_report(pacer_court_id, _start, _end)
    if exc:
        court_failed = True
        break
    time.sleep(1)
```

`fetch_doc_report` catches `RequestException, ReadTimeoutError, IndexError, TypeError, PacerLoginException, ValueError` and returns `True`. When that happens the per-court catch-up loop bails for the rest of the run, `date_queried` does not advance for the court, and the next scheduled run starts from the same date. There is no retry-with-backoff and no alerting when a court hasn't advanced in N days.

A relevant note: juriscraper's `FreeOpinionReport.query()` carries a stale comment that says \"Iterate one day at a time. Any more and PACER chokes,\" but the default `day_span=7` and CL's caller passes `gap=7`. The original guidance may have been correct, especially for high-volume courts.

### Most likely explanation (unconfirmed)

For SDNY: the request timeout used by `juriscraper.PacerSession` for `WrtOpRpt.pl` queries is, I suspect, shorter than NYSD's ~3-minute response time. If true, every 7-day nysd chunk would raise `ReadTimeoutError`, hit the `break` path, and leave `date_queried` frozen at 2025-11-26.

For peer districts: the same `break-on-any-error` mechanism could plausibly explain a 2-3 month lag if the same courts have intermittent (rather than reliable) chunk failures. I have not confirmed this; it could equally be Celery queue throughput, scheduling cadence, or another factor.

I can't confirm either hypothesis without inspecting Sentry traces and `PACERFreeDocumentLog` history.

### Diagnostic to confirm

```python
from cl.scrapers.models import PACERFreeDocumentLog
for log in PACERFreeDocumentLog.objects.filter(court_id='nysd').order_by('-date_queried')[:10]:
    print(log.date_queried, log.status)
```

If the most recent non-`SCRAPE_FAILED` row is around 2025-11-26 and rows after that are all `SCRAPE_FAILED`, the stall pattern is confirmed. Sentry traces for `fetch_doc_report(pacer_court_id='nysd', ...)` will name the actual exception class — most likely `ReadTimeoutError`, but it could be `IndexError` (PACER 6.3 bug per the existing comments) or another. Running the same diagnostic on `nyed`/`cand`/`txnd` would clarify whether peer-court lag has the same shape.

### Fix candidates

These compound, not exclusive — and the right combination depends on what the diagnostic shows:

1. **Don't bail the whole court on a single chunk failure.** Track failed chunks for retry; continue to the next chunk so good chunks still advance the cursor.
2. **Bump the request timeout** for `WrtOpRpt.pl` queries — direct fix if `ReadTimeoutError` is the actual exception.
3. **Default `gap` from 7 to 1** for free-opinion queries (matches juriscraper's stale comment). Especially helps slow or high-volume courts.
4. **Alert when a court hasn't successfully advanced in N days** (e.g., 14). Catches silent stalls regardless of failure mode.

### Related

- [juriscraper#1268](https://github.com/freelawproject/juriscraper/issues/1268) — prior NYSD-specific PACER quirk (closed 2024-12-10). Not the same root cause; just historical evidence that NYSD's PACER frontend has unique characteristics.

### LINKED ISSUE #7381: Issues encountered while running the `scrape-free-opinions` daemon
### Describe the issue.

Recently, @mlissner mentioned in the infra channel that the scrape-free-opinions daemon was not completing successfully. While investigating in Sentry, I found a few unresolved errors that may be related:

- [OperationalError: consuming input failed: SSL connection has been closed unexpectedly](https://freelawproject.sentry.io/issues/6821974352/?project=5257254&query=is%3Aunresolved%20server_name%3A%EF%80%8DContains%EF%80%8Dscrape-free-opinions&referrer=issue-stream)
- [HTTPError: 500 Server Error: Internal Server Error for url: http://ecf.miwd.uscourts.gov/cgi-bin/WrtOpRpt.pl?None](https://freelawproject.sentry.io/issues/6595250218/?project=5257254&query=is%3Aunresolved%20server_name%3A%EF%80%8DContains%EF%80%8Dscrape-free-opinions&referrer=issue-stream)
- [IndexError: list index out of range](https://freelawproject.sentry.io/issues/6595246872/?project=5257254&query=is%3Aunresolved%20server_name%3A%EF%80%8DContains%EF%80%8Dscrape-free-opinions&referrer=issue-stream)

I think each of these probably deserves its own issue, but I wanted to open this as a high-level tracking issue so the team owning this area can help triage and decide how to split them up.

### Related Issue

I also noticed there’s another open issue related to the same daemon: https://github.com/freelawproject/courtlistener/issues/7316

### Screenshots

_No response_

Research

quevon24:
## Fixes

Fixes: #7316

## Summary

The free-opinion scraper had stalled for SDNY since 2025-11-26 (peer districts lagged 2-3 months). Two things combined to freeze a court:

1. The per-court loop bailed on the first failed date chunk (`court_failed = True; break`).
2. `get_last_complete_date` ignores `SCRAPE_FAILED` rows, so the cursor stays pinned to the last success and every run re-attempts the same failing range.

Root cause of the failures themselves: PACER's `WrtOpRpt.pl` for busy courts (NYSD especially) can take longer than the `webhook-sentry` proxy's timeout, so the proxy returns a truncated HTTP 200 and juriscraper raises `IndexError` (the path we label "PACER 6.3 bug"). It is not a `PacerSession` timeout.

This PR:

- **Stops bailing the whole court on a chunk failure.** The loop now continues, so good days advance even when one day fails.
- **Retries outstanding failed days.** A failed day is kept as history; on later runs it is re-queued until it succeeds (resolved via a DB anti-join), so an advancing cursor can't leave a permanent gap. Bounded by a look-back window.
- **Queries one day at a time by default** (`day_span=1`, with a `--day-span` override for low-volume courts) so most days fit under the proxy timeout. Also fixes a latent bug where the date range was built once and reused across courts.
- **Adds a `report-stalls` action** (also run at the end of `do-everything`) that alerts via Sentry when a court hasn't advanced in N days (`--stale-days`, default 14) and enumerates individual gap days for investigation.

The slowest-day case (a few NYSD days still exceed the proxy's 120s) is being addressed separately by raising the `webhook-sentry` `connectionLifetime` / `readTimeout` in infra config.

## Running the stall report manually

The `report-stalls` action can be run on its own, optionally scoped to specific courts:

```bash
docker exec cl-django python manage.py scrape_pacer_free_opinions \
  --action report-stalls --courts nysd --stale-days 14
```

It logs (to Sentry in prod) any stalled court plus its outstanding gap days, collapsed into ranges so each line is usable as `--date-start` / `--date-end`:

```
Free opinion scrape has 45 outstanding failed day(s) in 26 range(s) for nysd (some of these days may simply have no opinions; re-running them will mark them complete):
2018-04-02
2018-12-25 to 2018-12-27
2019-06-07 to 2019-06-08
...
```

Omit `--courts` to check every in-use PACER court.

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
+++ b/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
@@ -50,20 +50,26 @@ RECENT_REQUERY_DAYS = 5
 # this many days. Used by the report-stalls action.
 DEFAULT_STALE_DAYS = 14
 
+# Courts that don't provide a usable written opinions report, so we never scrape
+# or report on them.
+EXCLUDED_COURT_IDS = ["casb", "gub", "ilnb", "innb", "miwb", "ohsb", "prb"]
+
 
 def get_last_complete_date(
     court_id: str,
 ) -> datetime.date | None:
-    """Get the next start query date for a court.
+    """Get the date a court's next forward scrape should start from.
 
-    Check the DB for the last date for a court that was completed. Return the
-    day after that date + span days into the future as the range to query for
-    the requested court.
+    Look up the court's most recent non-failed log row. Return its
+    ``date_queried``, but never more recent than ``RECENT_REQUERY_DAYS`` ago,
+    so the last few days are always re-queried to catch late-posted opinions.
 
-    If the court is still in progress, return (None, None).
+    If the most recent row is still in progress, return ``None``.
 
     :param court_id: A PACER Court ID
-    :return: last date queried for the specified court or None if it is in progress
+    :return: the date to resume querying from, or None if the court's latest
+    scrape is still in progress
+    :rtype: datetime.date | None
     """
     court_id = map_pacer_to_cl_id(court_id)
     try:
@@ -216,7 +222,10 @@ def fetch_doc_report(
         if isinstance(exc, (RequestException | ReadTimeoutError)):
             reason = "network error."
         elif isinstance(exc, IndexError):
-            reason = "PACER 6.3 bug."
+            reason = (
+                "PACER 6.3 bug or incomplete/truncated response "
+                "(likely proxy timeout)."
+            )
         elif isinstance(exc, (TypeError | ValueError)):
             reason = "failing PACER website."
         elif isinstance(exc, PacerLoginException):
@@ -260,19 +269,18 @@ def get_and_save_free_document_reports(
     documents. Do not download those items, as that step is done later. For now
     just get the list.
 
-    Note that this uses synchronous celery chains. A previous version was more
-    complex and did not use synchronous chains. Unfortunately in Celery 4.2.0,
-    or more accurately in redis-py 3.x.x, doing it that way failed nearly every
-    time.
-
-    This is a simpler version, though a slower one, but it should get the job
-    done.
+    Note that the ``get_and_save_free_document_report`` Celery task is invoked
+    synchronously (in-process) here, one court and date at a time, rather than
+    being enqueued. A previous version dispatched the work asynchronously, but
+    in Celery 4.2.0 (more accurately in redis-py 3.x.x) that failed nearly every
+    time. This is simpler and slower, but reliable.
 
     :param courts: optionally a list of courts to scrape
     :param date_start: optionally a start date to query all the specified courts or all
     courts
     :param date_end: optionally an end date to query all the specified courts or all
     courts
+    :param day_span: how many days each PACER sub-query should cover
     """
     # Kill any *old* logs that report they're in progress. (They've failed.)
     three_hrs_ago = now() - datetime.timedelta(hours=3)
@@ -281,9 +289,7 @@ def get_and_save_free_document_reports(
         status=PACERFreeDocumentLog.SCRAPE_IN_PROGRESS,
     ).update(status=PACERFreeDocumentLog.SCRAPE_FAILED)
 
-    excluded_court_ids = ["casb", "gub", "ilnb", "innb", "miwb", "ohsb", "prb"]
-
-    base_filter = Q(in_use=True, end_date=None) & ~Q(pk__in=excluded_court_ids)
+    base_filter = Q(in_use=True, end_date=None) & ~Q(pk__in=EXCLUDED_COURT_IDS)
     if courts:
         base_filter &= Q(pk__in=courts)
 
@@ -527,8 +533,7 @@ def report_free_document_scrape_stalls(
     :returns: list of (court_id, latest_success_date) for stalled courts
     :rtype: list[tuple[str, datetime.date | None]]
     """
-    excluded_court_ids = ["casb", "gub", "ilnb", "innb", "miwb", "ohsb", "prb"]
-    base_filter = Q(in_use=True, end_date=None) & ~Q(pk__in=excluded_court_ids)
+    base_filter = Q(in_use=True, end_date=None) & ~Q(pk__in=EXCLUDED_COURT_IDS)
     if courts:
         base_filter &= Q(pk__in=courts)
 
--- a/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
+++ b/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
@@ -50,9 +50,6 @@ RECENT_REQUERY_DAYS = 5
 # this many days. Used by the report-stalls action.
 DEFAULT_STALE_DAYS = 14
 
-# Cap on how many gap dates to list in a single report-stalls log line.
-GAP_REPORT_LIMIT = 30
-
 
 def get_last_complete_date(
     court_id: str,
@@ -133,6 +130,28 @@ def get_outstanding_failed_dates(
         .distinct()
         .order_by("date_queried")
     )
+
+
+def collapse_date_ranges(
+    dates: list[datetime.date],
+) -> list[tuple[datetime.date, datetime.date]]:
+    """Collapse a sorted list of dates into consecutive-day ranges.
+
+    Consecutive calendar days are merged into a single (start, end) tuple so
+    they can be scraped in one --date-start/--date-end call.
+
+    :param dates: a sorted list of dates
+    :returns: list of inclusive (start, end) ranges
+    :rtype: list[tuple[datetime.date, datetime.date]]
+    """
+    ranges: list[tuple[datetime.date, datetime.date]] = []
+    one_day = datetime.timedelta(days=1)
+    for day in dates:
+        if ranges and day == ranges[-1][1] + one_day:
+            ranges[-1] = (ranges[-1][0], day)
+        else:
+            ranges.append((day, day))
+    return ranges
 
 
 def mark_court_in_progress(
@@ -552,18 +571,23 @@ def report_free_document_scrape_stalls(
         gaps = get_outstanding_failed_dates(court_id, before=gap_cutoff)
         if not gaps:
             continue
-        shown = gaps[:GAP_REPORT_LIMIT]
-        suffix = (
-            f" (+{len(gaps) - len(shown)} more)"
-            if len(gaps) > len(shown)
-            else ""
+        ranges = collapse_date_ranges(gaps)
+        # One line per range: a single day shows the date, a span shows
+        # "start to end". Each line is usable as --date-start / --date-end.
+        range_lines = "\n".join(
+            start.isoformat()
+            if start == end
+            else f"{start.isoformat()} to {end.isoformat()}"
+            for start, end in ranges
         )
         logger.error(
-            "Free opinion scrape has %s outstanding failed day(s) for %s: %s%s",
+            "Free opinion scrape has %s outstanding failed day(s) in %s "
+            "range(s) for %s (some of these days may simply have no opinions; "
+            "re-running them will mark them complete):\n%s",
             len(gaps),
+            len(ranges),
             court_id,
-            ", ".join(d.isoformat() for d in shown),
-            suffix,
+            range_lines,
             extra={"fingerprint": ["pacer-free-opinion-gaps", court_id]},
         )
 
--- a/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
+++ b/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
@@ -8,7 +8,7 @@ from collections.abc import Callable
 from typing import cast
 
 from celery.canvas import chain
-from django.db.models import F, Q, Window
+from django.db.models import F, Max, Q, Window
 from django.db.models.functions import RowNumber
 from django.utils.timezone import now
 from juriscraper.lib.date_utils import make_date_range_tuples
@@ -34,6 +34,24 @@ from cl.lib.types import OptionsType
 from cl.scrapers.models import PACERFreeDocumentLog, PACERFreeDocumentRow
 from cl.scrapers.tasks import extract_pdf_document
 from cl.search.models import Court, RECAPDocument
+
+# How far back to re-attempt days that previously failed and never succeeded.
+# Wide enough to drain a multi-month backlog, but bounded so a date PACER can
+# never render isn't retried forever (each retry can cost a full proxy
+# timeout). Days older than this need manual attention.
+OUTSTANDING_FAILED_LOOKBACK_DAYS = 365
+
+# The most recent days are always re-queried (to catch late-posted opinions),
+# so a failed day this recent is still being actively retried and shouldn't be
+# reported as a gap yet.
+RECENT_REQUERY_DAYS = 5
+
+# A court is considered stalled if its newest successful scrape is older than
+# this many days. Used by the report-stalls action.
+DEFAULT_STALE_DAYS = 14
+
+# Cap on how many gap dates to list in a single report-stalls log line.
+GAP_REPORT_LIMIT = 30
 
 
 def get_last_complete_date(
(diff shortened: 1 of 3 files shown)


mlissner:
This sounds nice. Thanks Kevin!

quevon24:
I inspected a dump of the `PACERFreeDocumentLog` table (data up to May 28th) to see how bad the backlog is. There are **2,224 outstanding gap ranges across 185 courts** (a "gap" being a date with a `SCRAPE_FAILED` log row and no `SCRAPE_SUCCESSFUL` row for the same court and date, consecutive days are collapsed into a single range).

Worth keeping in mind: some of these gaps are genuinely days with no opinions (PACER just returns an empty report), while others do have results that we never ingested. We can't tell which is which without actually querying them, but both cases are handled the same way below.

I've attached a CSV of all the gaps with columns `court_id, date_start, date_end` (single days have `date_start == date_end`).

[gap_ranges_2026-06-02.csv](https://github.com/user-attachments/files/28531730/gap_ranges_2026-06-02.csv)


To ingest them, run this through the Django shell (`docker exec -i cl-django python manage.py shell < ingest_gaps.py`), replacing `CSV_PATH` with wherever you put the CSV:

```python
import csv, datetime, time

from cl.corpus_importer.management.commands.scrape_pacer_free_opinions import (
    get_and_save_free_document_reports,
)

CSV_PATH = "/replace/with/your/path/gap_ranges.csv"

with open(CSV_PATH, newline="") as fh:
    rows = list(csv.DictReader(fh))

total = len(rows)
print(f"{total} ranges to ingest")
for i, row in enumerate(rows, 1):
    court = row["court_id"]
    start = datetime.date.fromisoformat(row["date_start"])
    end = datetime.date.fromisoformat(row["date_end"])
    print(f"[{i}/{total}] {court} {start} -> {end}")
    try:
        get_and_save_free_document_reports([court], start, end, day_span=1)
    except Exception as exc:
        print(f"  FAILED: {exc!r}")
    time.sleep(2)
print("done")
```

It calls the command's `get_and_save_free_document_reports()` directly (rather than `call_command`, which skips the argparse type coercion for kwargs) with `day_span=1`, so each range is queried one day at a time.

One nice side effect: even for the days that turn out to have no opinions, a successful query creates a new `PACERFreeDocumentLog` row with status `SCRAPE_SUCCESSFUL`. Since the gap detection ignores any date that has a success row, those empty days will stop showing up as gaps on the next `report-stalls` run, so we won't keep re-checking them.

grossir (approved):
LGTM!

grossir:
I ran the command on the PR description, seems like I overloaded Sentry. But we can gather those to fill any gaps. Will track that in a separate issue

```
./manage.py scrape_pacer_free_opinions --action report-stalls --stale-days 14
```

quevon24:
> I ran the command on the PR description, seems like I overloaded Sentry. But we can gather those to fill any gaps. Will track that in a separate issue
> 
> ```
> ./manage.py scrape_pacer_free_opinions --action report-stalls --stale-days 14
> ```

Thanks, I'll push a small change to group them into one issue

grossir:
Attaching logs @quevon24 . Sorry, I missed the first 700/2000 entries due to shell history limits :/
[pacer-free-gaps.txt](https://github.com/user-attachments/files/28575717/pacer-free-gaps.txt)

quevon24:
The gap CSV I attached earlier was built from PACERFreeDocumentLog (dates with a SCRAPE_FAILED row and no SCRAPE_SUCCESSFUL row for the same court+date). That only finds days the scraper tried and errored on. It cannot find "silent" gaps, where a scrape was recorded as SCRAPE_SUCCESSFUL but actually ingested zero opinions, because nothing in the log distinguishes "succeeded and captured everything" from "succeeded and returned nothing" (we don't store a result count).

That blind spot is exactly the original SDNY symptom. nysd's log shows SCRAPE_SUCCESSFUL for every day from 2025-11-27 through 2026-05-28 (each scraped three times by the daily/weekly/monthly cadence), yet the newest nysd opinion actually in CourtListener is dated 2025-11-26. So the log-based gap CSV listed none of those days, because by its own records they were fine.

The reliable signal is the ingested outcome, not the scrape log: for each court, the latest opinion dateFiled present in CL (queried via the search API, all precedential statuses on). A court whose newest opinion is months old while the log claims recent success has a silent gap. This catches what the log-based approach misses.

Based on that, the re-scrape list is built as: date_start = max(latest_opinion_dateFiled, 2024-08-05) (we already ran a full 1950 to 2024-08-05 sweep, so nothing earlier needs re-scraping), date_end = today. Then we drop two groups: courts whose latest opinion is <= 2025-11-12 (low-frequency courts where an old date is expected, not a gap), and courts within the daemon's own 5-day re-query window (get_last_complete_date already clamps to today - RECENT_REQUERY_DAYS, so those re-sweep automatically). That leaves 133 courts to re-scrape, with the longest real lags on the high-volume courts (nysd, caed, nvd, cand, cacd, ...).

Longer term, storing the per-scrape result count on PACERFreeDocumentLog would make these silent gaps detectable from the log directly, instead of having to cross-check against ingested opinions.

This is the file:

[rescrape_targeted_2026-06-04.csv](https://github.com/user-attachments/files/28579117/rescrape_targeted_2026-06-04.csv)

grossir:
@quevon24 I am running the script with the new file

Can you open a separate issue to track this 
> Longer term, storing the per-scrape result count on PACERFreeDocumentLog would make these silent gaps detectable from the log directly, instead of having to cross-check against ingested opinions.

quevon24:
> @quevon24 I am running the script with the new file
> 
> Can you open a separate issue to track this
> 
> > Longer term, storing the per-scrape result count on PACERFreeDocumentLog would make these silent gaps detectable from the log directly, instead of having to cross-check against ingested opinions.

Thank you, when it finishes can you upload the logs please

quevon24:
We need to run both csv again. The previous script I posted only got the document references but I missed the lines to download them.

For this: https://github.com/user-attachments/files/28531730/gap_ranges_2026-06-02.csv we can set `GET_REPORT = False` in the next script

And for this: https://github.com/user-attachments/files/28579117/rescrape_targeted_2026-06-04.csv we need to set `GET_REPORT = True`

```
import csv, datetime, time, logging

from cl.corpus_importer.management.commands.scrape_pacer_free_opinions import (
    get_and_save_free_document_reports, get_pdfs, ocr_available
)

CSV_PATH = "/replace/with/your/path/gap_ranges.csv"
QUEUE_NAME="celery"
LOG_PATH = f"/tmp/ingest_{datetime.datetime.now():%Y%m%d_%H%M%S}.log"
GET_REPORT = True

# Send our messages and the app/juriscraper loggers to both file and console.
file_handler = logging.FileHandler(LOG_PATH)
file_handler.setFormatter(
logging.Formatter("%(asctime)s %(levelname)s %(name)s: %(message)s")
)
logging.getLogger().addHandler(file_handler)
logging.getLogger().setLevel(logging.INFO)

logger = logging.getLogger("ingest")
logger.info("Writing logs to %s", LOG_PATH)


with open(CSV_PATH, newline="") as fh:
    rows = list(csv.DictReader(fh))

total = len(rows)
logger.info("%d ranges to ingest", total)
for i, row in enumerate(rows, 1):
    court = row["court_id"]
    start = datetime.date.fromisoformat(row["date_start"])
    end = datetime.date.fromisoformat(row["date_end"])
    logger.info("[%d/%d] %s %s -> %s", i, total, court, start, end)
    try:
        if GET_REPORT:
            logger.info("Running and compiling free document reports.")
            get_and_save_free_document_reports([court], start, end, day_span=1)
        logger.info("Getting PDFs from free document reports")
        get_pdfs([court], row["date_start"], row["date_end"], QUEUE_NAME)
        logger.info("Doing OCR and saving items.")
        ocr_available(QUEUE_NAME)

    except Exception as exc:
        logger.exception("  FAILED on %s %s -> %s", court, start, end)
    time.sleep(2)
logger.info("done")
```

quevon24:
I updated the script so it can be rerun safely and resume where it stopped, it will generate all the free document report for every range and then it will proceed to download the pdfs.


```python
import csv
import datetime
import logging
import time

from cl.corpus_importer.management.commands.scrape_pacer_free_opinions import (
  collapse_date_ranges,
  get_and_save_free_document_reports,
  get_pdfs,
  ocr_available,
)
from cl.scrapers.models import PACERFreeDocumentLog

CSV_PATH = "/replace/with/your/path/gap_ranges.csv"
QUEUE_NAME = "batch1"
LOG_PATH = f"/tmp/ingest_{datetime.datetime.now():%Y%m%d_%H%M%S}.log"

# Send our messages and the app/juriscraper loggers to both file and console.
file_handler = logging.FileHandler(LOG_PATH)
file_handler.setFormatter(
  logging.Formatter("%(asctime)s %(levelname)s %(name)s: %(message)s")
)
logging.getLogger().addHandler(file_handler)
logging.getLogger().setLevel(logging.INFO)

logger = logging.getLogger("ingest")
logger.info("Writing logs to %s", LOG_PATH)

# compile the free document reports for every range, one day at a time
with open(CSV_PATH, newline="") as fh:
  rows = list(csv.DictReader(fh))
total = len(rows)
logger.info("%d ranges to query", total)
one_day = datetime.timedelta(days=1)
for i, row in enumerate(rows, 1):
  court = row["court_id"]
  start = datetime.date.fromisoformat(row["date_start"])
  end = datetime.date.fromisoformat(row["date_end"])

  # a day that already has a SCRAPE_SUCCESSFUL log row was captured by a previous run, so only the remaining days are queried
  succeeded = set(
      PACERFreeDocumentLog.objects.filter(
          court_id=court,
          status=PACERFreeDocumentLog.SCRAPE_SUCCESSFUL,
          date_queried__range=(start, end),
      ).values_list("date_queried", flat=True)
  )
  pending = []
  day = start
  while day <= end:
      if day not in succeeded:
          pending.append(day)
      day += one_day
  if not pending:
      logger.info(
          "[%d/%d] %s %s -> %s already done, skipping",
          i, total, court, start, end,
      )
      continue

  for sub_start, sub_end in collapse_date_ranges(pending):
      logger.info(
          "[%d/%d] %s %s -> %s", i, total, court, sub_start, sub_end
      )
      try:
          get_and_save_free_document_reports(
              [court], sub_start, sub_end, day_span=1
          )
      except Exception:
          logger.exception(
              "FAILED on %s %s -> %s", court, sub_start, sub_end
          )
      time.sleep(2)

# get_pdfs call over everything accumulated in PACERFreeDocumentRow
logger.info("Getting PDFs from free document reports")
get_pdfs([], None, None, QUEUE_NAME)

logger.info("Doing OCR and saving items.")
ocr_available(QUEUE_NAME)

logger.info("done; logs at %s", LOG_PATH)

```

PR #7418, opened 2026-06-03
7372 Reclaim alert hits redis set

albertisfu:
## Fixes
  Related to: #7372

  ## Summary
  Search alerts store their triggered document IDs in per-alert Redis SETs keyed
  `alert_hits:{alert_id}.{document_type}` via `add_document_hit_to_alert_set`
  (document types `d`, `co`, `r`, `o`). When an Alert was deleted, these sets were
  left behind and leaked Redis memory indefinitely.

  This PR adds `remove_alert_hits_set(r, alert_id)` in `cl/alerts/utils.py`, which
  uses `scan_iter` over the pattern `alert_hits:{alert_id}.*` and deletes every
  matching set. A scan pattern is used so it stays correct if new document types are added later.

  It is wired into the existing `remove_alert_from_es_index` `post_delete`
  receiver.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

After this is merged we can clean up unused Redis sets manually.


cl/alerts/utils.py
```diff
@@ -509,6 +509,23 @@ def add_document_hit_to_alert_set(
...
+    and would otherwise leak Redis memory, so we scan and delete all of them.
+
+    :param r: Redis client instance.
+    :param alert_id: The alert identifier.
+    :return: None
+    """
+    match_pattern = f"{get_alerts_set_prefix()}:{alert_id}.*"
+    keys = list(r.scan_iter(match=match_pattern))
+    if keys:
+        r.delete(*keys)
```
ERosendo, line 545: Is it possible to implement this as a Celery task?

We recently noticed that `clear_tiered_cache` was adding some latency to the admin action that refreshes membership data, so we decided to skip that method when the admin action runs. I bring that up because this method follows a similar pattern (`scan_iter` followed by `delete`).

The admin action wasn't a big concern since it's only triggered by staff users, but this signal will run every time a user removes alerts from their profile. I don't think that happens very often, but it is exposed to more requests than the admin action.

Moving it to a background task wouldn't make the operation itself any more efficient, but it would at least prevent it from blocking the request cycle and potentially impacting the user experience.

ERosendo (changes requested):
Code looks good overall. I just left a small comment about a recent performance issue we ran into that might be worth considering before merging this PR. 👍

cl/alerts/utils.py
albertisfu: Thanks! Yeah, this is a good idea. It would also help in the event of a network issue communicating with Redis, since the task could then be retried. I've applied the change.

ERosendo (approved):
LGTM :shipit:

Issue #6878, opened 2026-02-05
Abstract Docket Merging Process


MorganBennetDev:
### Describe the issue.

Since we've decided that we're going to have unique `<State>DocketEntry` and `<State>Document` models for every state we ingest (see discussion in #6589), we need mergers for all of those new models. This will quickly get repetitive and waste time; Alberto and I are already doing very similar work in the Texas and SCOTUS mergers (#6761 and #6709). We should put some upfront effort into creating merger logic that can be easily adapted to each new model. Alberto, Brennan, and I have already had a brief chat about this issue, and I will be experimenting with approaches when it comes time to create mergers for Florida.

Tagging @Brennan-Chesley-FLP and @albertisfu in case they have any context to add.

### What type of issue is this?

Research

Brennan-Chesley-FLP:
What were some of the things that made this difficult/fraught for Scotus and Texas?

Brennan-Chesley-FLP:
# Virtual Aggregate-Root Mergers, simply

The very brief version:
- Identify [an aggregate(ish)-root](https://martinfowler.com/bliki/DDD_Aggregate.html)
- Shape the data we would like into a tree based on that root
- Hand that tree of data, along with a set of resolution choices to an engine that does react-style reconcilliation.
- Handle follow-ups, like file processing.

## A slightly longer description

After identifying the root, there's 3 steps, here's a small code example for each:

### 1. Pre-processing

Reshape one raw scrape record into the dict the Pydantic tree expects, and
slot in any **pre-resolved references** that are stable across the batch —
most commonly the `Court`. Think of this as a custom caching strategy for 
data we know will be consistent across multiple merges. 

```python
class NYCoAMerger(KentMerger[NYCoADocket]):
    aggregate_cls = NYCoADocket
    sql_path = "load.sql"

    def __init__(self, kent_db_path, *, using: str = "default") -> None:
        super().__init__(kent_db_path, using=using)
        self.court = Court.objects.using(using).get(pk="ny")

    def normalize(self, row: dict[str, Any]) -> dict[str, Any]:
        return {
            "court": self.court,
            "docket_number": row["docket_number"],
            "case_name": _collapse_whitespace(row.get("case_name")) or None,
            "date_argued": row.get("date_argued"),
            "date_filed": row.get("date_filed"),
            "nycoadocketentry_set": list(row.get("filings_entries") or [])
                + list(row.get("file_entries") or []),
        }
```

### 2. Pydantic tree definition

[...]

PR #7421, opened 2026-06-03
Florida docket merger and abstraction prototype

### LINKED ISSUE #7361: Create `Docket` merger for Florida
Implement docket merging logic for Florida scraper output. This issue is only for `Docket` and possibly `OriginatingCourtInformation` objects; other data will be handled in different issues.

Also should compose the first phase of experimentation for #6878

MorganBennetDev:
## Fixes

Closes #7361 
Progresses #6878 

## Summary

Adds prototype merger abstraction in `corpus_importer/state/merger.py`

Adds `Docket` and `OriginatingCourtInformation` mergers for Florida in `corpus_importer/state/florida` using prototype merger abstraction

- Add method for Florida DN core in `corpus_importer/state/florida/utils.py` and add it in the appropriate places to ensure correct DN core generation

Moves `MergeResult` from `cl/corpus_importer/tasks.py` to `cl/corpus_importer/state/utils.py`

I put the Florida merger in its own directory to keep things organized since. `tasks.py` is already quite large and adding more non-task merger methods to it is only going to make it harder to work in. I think going forward, we should limit `tasks.py` to only tasks and put related methods in other files/packages when possible.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [ ] `skip-cronjob-deploy`
    - [ ] `skip-daemon-deploy`


cl/corpus_importer/state/florida/mergers.py
```diff
@@ -0,0 +1,225 @@
...
+
+def _date_last_filing(docket_data: FloridaCase) -> date | None:
+    filing_dates = sorted(
+        e.date_filed for e in docket_data.entries if e.date_filed
+    )
+    return filing_dates[-1] if filing_dates else docket_data.date_filed
+
+
+def _originating_case_number(docket_data: FloridaCase) -> str:
+    return docket_data.originating_cases[0].case_number
```
albertisfu: Shouldn't this method also validate that there are `originating_cases` instances before accessing `docket_data.originating_cases[0]`?

```suggestion
     if len(docket_data.originating_cases) > 1:
         return None
    return docket_data.originating_cases[0].case_number
```

cl/corpus_importer/state/florida/utils.py
```diff
@@ -0,0 +1,80 @@
...
+            "Unable to find valid Florida DN for court %s in string %s",
+            court_id,
+            docket_number,
+        )
+        return docket_number
+
+    matches.sort()
+    if len(matches) > 1:
+        logger.warning(
+            "Found multiple Florida DNs for court %s in string %s. Using %s",
```
albertisfu, line 86: This log expect 3 different arguments but only `matches[0]` is passed.

cl/corpus_importer/tasks.py
```diff
@@ -4888,40 +4797,42 @@ def merge_texas_docket(
...
-                federal_dn_judge_initials_assigned=None,
-                federal_dn_judge_initials_referred=None,
                 docket_source=Docket.SCRAPER,
-                allow_create=False,
             )
-            if docket is not None:
+            if changed:
                 logger.info(
-                    "Disaggregating Texas appellate docket %s", docket_number
+                    f"Disaggregated Texas appellate docket {docket_number}"
```
albertisfu: also here:
```suggestion
                    "Disaggregated Texas appellate docket %s", docket_number
```

cl/recap/mergers.py
```diff
@@ -287,6 +296,58 @@ async def find_docket_object(
...
+        docket_number=docket_number,
+        federal_defendant_number=None,
+        federal_dn_judge_initials_assigned=None,
+        federal_dn_judge_initials_referred=None,
+        docket_source=docket_source,
+        allow_create=False,
+    )
+    if docket is not None:
+        logger.info(
+            f"Disaggregating docket {docket.pk} from {aggregate_court_id} to {court_id}"
```
albertisfu: In general, we should use parameterized logging (`%s`) for log messages.

cl/corpus_importer/state/florida/utils.py
```diff
@@ -0,0 +1,80 @@
+import logging
+import re
+
+from cl.lib.string_utils import normalize_dashes
+
+logger = logging.getLogger(__name__)
+
+# Non-comprehensive list of court IDs that can be handled by Florida make_docket_number_core
+FLORIDA_COURT_IDS: set[str] = {
+    "fl",
```
albertisfu: should be? 
```suggestion
    "fla",
```

Also, can we derive these IDs from `FLORIDA_COURT_ID_MAP` instead? To have IDs from a single source?

cl/corpus_importer/state/merger.py
```diff
@@ -0,0 +1,513 @@
...
+            scrape_value = getattr(scrape_obj, name)
+            db_value = getattr(db_obj, name)
+            merged_value = attr_merger.merge_values(scrape_value, db_value)
+            if merged_value != db_value:
+                setattr(db_obj, name, merged_value)
+                update = True
+
+        if update:
+            db_obj.save()
+            return MergeResult.updated(
```
albertisfu: For new dockets created via `find_and_disaggregate_docket_object`, this incorrectly marks the case as updated even when it's new.

The problem is that `find_and_disaggregate_docket_object` can return a Docket instance that has not yet been persisted to the database.

As a result, `db_obj` is not None, so the code enters the update path. In this scenario, we should be able to correctly distinguish between newly created cases and genuinely updated cases.

cl/recap/mergers.py
```diff
@@ -287,6 +296,58 @@ async def find_docket_object(
...
+
+    :param court_id: The primary court to look for the docket in.
+    :param aggregate_court_id: The backup court to look in and disaggregate from.
+    :param docket_number: The docket number to lookup.
+    :param docket_source: The source to set when creating a new docket.
+    :return: The docket found or created and a boolean indicating whether the
+        docket's court was changed."""
+    docket = await find_docket_object(
+        court_id=aggregate_court_id,
+        pacer_case_id=None,
```
albertisfu: If I not wrong I remember we agreed in https://github.com/freelawproject/courtlistener/issues/7361#issuecomment-4566459292 to use `case_uuid` as a parameter to merge Florida cases unambiguously. To do that, we'd also need to persist the `case_uuid` in the `pacer_case_id` field.

cl/corpus_importer/state/merger.py
```diff
@@ -0,0 +1,513 @@
...
+    information which won't change over the course of many merges (i.e., a parent `Docket` for `DocketEntry`s)."""
+
+    default: T | None = None
+    transform: Callable[[T], T] = lambda x: x
+
+    @override
+    def map(
+        self, i: Any, *args: Any, value: T | None = None, **kwargs: Any
+    ) -> T | None:
+        if value:
```
albertisfu: In cases where the value is 0, "", or False, this will return `self.default`, which might not be the same as the original value.

Would it be better to check:
```suggestion
        if value is not None:
```

?

albertisfu (commented):
Thanks @MorganBennetDev this looks good!

I just have a couple of comments/suggestions.

cl/recap/mergers.py
MorganBennetDev: Forgot about this, thank you for the reminder

cl/corpus_importer/state/florida/utils.py
```diff
@@ -0,0 +1,85 @@
...
+        logger.error(
+            "Cannot make Florida DN core from non-Florida court %s", court_id
+        )
+        return docket_number
+    docket_number = docket_number.strip().upper()
+    if not docket_number:
+        return ""
+    docket_number = normalize_dashes(docket_number)
+
+    if court_id == "fl":
```
albertisfu: should be `"fla"`? Can we reference it directly from `FLORIDA_COURT_ID_MAP`?

same in the method signature default 

`court_id: str = "fl"`

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -0,0 +1,226 @@
...
+    )
+    docket_number: str = AttributeMerger[FloridaCase, str](
+        InputField("docket_number"),
+        strategy=OverwriteExisting(),
+    )
+    docket_number_raw: str = AttributeMerger[FloridaCase, str](
+        InputField("docket_number"), strategy=OverwriteExisting()
+    )
+    docket_number_core: str = AttributeMerger[FloridaCase, str](
+        InputField("docket_number", transform=make_docket_number_core),
```
albertisfu: `make_docket_number_core` expects a `court_id` otherwise, it will use the default `fla`. Is this method receiving the corresponding court ID to generate the correct `docket_number_core`?

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -0,0 +1,226 @@
...
+                court_id=docket.court_id,
+                pacer_case_id=docket.pacer_case_id,
+                docket_number=docket.docket_number,
+                federal_defendant_number=None,
+                federal_dn_judge_initials_assigned=None,
+                federal_dn_judge_initials_referred=None,
+                docket_source=Docket.SCRAPER,
+                allow_create=False,
+            )
+        found, changed = async_to_sync(find_and_disaggregate_docket_object)(
```
albertisfu: This is performing a `save()` internally. Is this save operation also being performed in `_merge_object` after a case is disaggregated, resulting in a duplicate save?

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -0,0 +1,226 @@
...
+    )
+    appeal_from_id: str | None = AttributeMerger(
+        InputMap(_appeal_from_id), strategy=OverwriteExisting()
+    )
+    appeal_from_str: str | None = AttributeMerger(
+        InputMap(_appeal_from_str), strategy=OverwriteExisting()
+    )
+    # See https://github.com/freelawproject/courtlistener/issues/7361#issuecomment-4566459292
+    pacer_case_id: str = AttributeMerger(InputField("case_uuid"))
+    originating_court_information: OriginatingCourtInformation = RelatedMerger[
```
albertisfu: What’s the difference between this method for merging OCI data and `merge_oci`? I found that `merge_oci` is unused right now, as its result is not linked to the docket.

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -0,0 +1,226 @@
...
+def merge_docket(docket_data: FloridaCase) -> MergeResult:
+    """Merger for Florida docket data and originating court information.
+
+    :param docket_data: The scraped Florida docket information.
+    :return: The result of the attempted merge operation.
+    """
+    if docket_data.court_id not in FLORIDA_COURT_ID_MAP:
+        logger.error("Unknown court id: %s", docket_data.court_id)
+        return MergeResult.failed("Docket")
+    return FloridaDocketMerger.merge(docket_data)
```
albertisfu: What will be the approach to merge the full docket data: docket, parties, entries, documents, etc.? Would it be similar to `process_scotus_docket`, which is a Celery task that merges the full docket in the background? If that’s the case, should this method also be a task?

Merging the full case I think should be asynchronous in celery so we can run case merges in parallel to accelerate the ingestion of historical cases and also improve the performance of the ongoing scraper.

albertisfu (changes requested):
Thank you, @MorganBennetDev! I've added a couple of additional comments and suggestions.

I tested this branch by trying to adapt it to what we have for the Texas docket merger, and I struggled a bit to follow the initialization logic of `Merger` and its related helper methods. That made me wonder whether we might be overengineering things a bit. I'm curious if there's a way to simplify the merger abstraction.

My understanding is that `Merger` is currently generic over the model, right? Is that why it needs all the machinery around `__init_subclass__`? You also have classes like `RelatedMerger` and `Relationship` that are intended to keep the merged model agnostic, which is good and aligns with the goals of this effort.

However, I'm wondering whether we could reduce the complexity based on two observations: We already have typed input data such as `FloridaCase` and we already know which database model each merger will target.

Instead of having a single generic `Merger`, my proposal would be to create a small family of model-specific sub-classes based on `Merger`, such as `FloridaDocketMerger` (this PR), followed by `FloridaPartyMerger`, `FloridaDocketEntryMerger`, and `FloridaDocumentMerger` in future PRs. Each base class would simply define `model = Docket` (or the corresponding model) as a regular class attribute, eliminating the need for `__init_subclass__`. 

So these state-specific mergers would then become lightweight subclasses that provide field mappings and implement methods such as `find_existing()`. The base class could still iterate over a declarative field list and perform the "set if changed" logic, so we would retain that flexibility while removing the metaprogramming that seems to exist primarily to support full model agnosticism.

In summary, I'd like to explore whether we could remove:

* `__init_subclass__` model extraction and validation (replaced with an explicit `model` attribute)
* The `Any` based abstractions
* `InputMap`, `InputField`, `Parameter`, `Constant`, and `PassAll` (replaced with simple callables, e.g., `lambda c: c.case_name`)
* `Relationship`, `OneToOne`, and `Child`

I'd keep `MergeResult` exactly as it is, keep strategies as plain `(scrape, db) -> value` functions, and keep the declarative field list.

If we decide to simplify the approach some of the new suggestions in the code might not be required anymore.

What do you think? I may be missing something that becomes important for upcoming mergers such as entries, parties or documents, so please push back if there's a reason this additional abstraction it's not the best in the long run.

MorganBennetDev:
Pasting Alberto's message from Slack to keep context in one place:

<details>

Morning! regarding my comments in https://github.com/freelawproject/courtlistener/pull/7421/ I asked claude to make a proposal based on those comments and this is its output in case is helps to explain what I was suggesting and you can evaluate it better 

 ```python
# state/merger.py
  from abc import ABC
  from collections.abc import Sequence
  from typing import Any, ClassVar
  from django.db import models, transaction
  from cl.corpus_importer.state.utils import MergeResult
  from .fields import Field   # the @dataclass(frozen=True) from before
​
​
  class Merger[D, M: models.Model](ABC):
      """Merge scraped data ``D`` into DB model ``M``.
​
      One class per (state, model). The base owns find-or-create + set-if-changed +
      save + compose. Subclasses supply ``model`` + ``fields`` and optionally override
      ``find_existing`` / ``merge_related``.
​
      ``M`` is for static typing; ``model`` is the runtime handle. They're stated
      twice on purpose — the alternative is reflecting ``M`` out of the type args,
      which is exactly the machinery we're removing.
      """
      model: ClassVar[type[models.Model]]
      fields: ClassVar[Sequence[Field]] = ()
      atomic: ClassVar[bool] = False
      lock_parent: ClassVar[str | None] = None   # ctx kwarg to select_for_update before matching
​
      @classmethod
      def find_existing(cls, data: D, **ctx: Any) -> M | None:
          return ctx.get("existing")             # default: caller-supplied (one-to-one parents)
​
      @classmethod
      def merge_related(cls, obj: M, data: D, **ctx: Any) -> MergeResult:
          return MergeResult.unnecessary()       # override to compose relatives
​
      @classmethod
      def _apply(cls, obj: M, data: D) -> bool:
          changed = False
          for f in cls.fields:
              merged = f.strategy(f.get(data), getattr(obj, f.attr))
              if merged != getattr(obj, f.attr):
                  setattr(obj, f.attr, merged)
                  changed = True
          return changed
​
      @classmethod
      def merge_into(cls, obj: M | None, data: D, **ctx: Any) -> tuple[M, MergeResult]:
          """Merge into a specific instance; return (obj, result). Use when the caller
          must wire a FK back (parent-side one-to-one)."""
          obj = obj or cls.model()
          created = obj.pk is None
          changed = cls._apply(obj, data)
          result = MergeResult.unnecessary()
          if created or changed:
              obj.save()
              ctor = MergeResult.created if created else MergeResult.updated
              result = ctor(cls.model.__name__, obj.pk)
          return obj, result | cls.merge_related(obj, data, **ctx)
​
      @classmethod
      def merge(cls, data: D, **ctx: Any) -> MergeResult:
          def run() -> MergeResult:
              if cls.lock_parent and (p := ctx.get(cls.lock_parent)) is not None:
                  type(p).objects.select_for_update().get(pk=p.pk)   # universal, not entry-specific
              _, result = cls.merge_into(cls.find_existing(data, **ctx), data, **ctx)
              return result
          if cls.atomic:
              with transaction.atomic():
                  return run()
          return run()
```
​
  ```python
# state/florida/mergers.py
  from typing import ClassVar
  from cl.corpus_importer.state.merger import Merger, Field
  from cl.corpus_importer.state.strategies import overwrite, add_scraper_source
  from cl.search.models import Docket, OriginatingCourtInformation
  # ...find_docket_object, FLORIDA_COURT_ID_MAP, make_docket_number_core, helpers as before...
​
  # Shared docket bits live as DATA, not a base class. When a 2nd state repeats
  # these, that's the signal to lift a DocketMerger base — not before.
  DOCKET_BASE_FIELDS: tuple[Field, ...] = (
      Field("source", lambda _: None, add_scraper_source),   # policy field; getter unused
  )
​
​
  class FloridaOCIMerger(Merger[FloridaCase, OriginatingCourtInformation]):
      model = OriginatingCourtInformation
      fields = (
          Field("docket_number",     lambda d: _first_originating(d).case_number, overwrite),
          Field("docket_number_raw", lambda d: _first_originating(d).case_number, overwrite),
      )
​
​
  class FloridaDocketMerger(Merger[FloridaCase, Docket]):
      model = Docket
      atomic = True
      fields = (
          *DOCKET_BASE_FIELDS,
          Field("court_id",          lambda d: FLORIDA_COURT_ID_MAP[d.court_id], overwrite),
          Field("case_name",         lambda d: d.case_name,           overwrite),
          Field("docket_number",     lambda d: d.docket_number,       overwrite),
          Field("docket_number_core",
                lambda d: make_docket_number_core(
                    d.docket_number, court_id=FLORIDA_COURT_ID_MAP[d.court_id]),
                overwrite),
          Field("pacer_case_id",     lambda d: d.case_uuid,           overwrite),
          # ...case_name_full, case_name_short, date_filed, appeal_from_id, ...
      )
​
      @classmethod
      def find_existing(cls, data, **ctx):
          ...   # unchanged: supreme → find_docket_object; appellate → disaggregate
​
      @classmethod
      def merge_related(cls, docket, data, **ctx):
          if docket.court_id != "fla" or not data.originating_cases:
              return MergeResult.unnecessary()
          oci, result = FloridaOCIMerger.merge_into(docket.originating_court_information, data)
          docket.originating_court_information = oci          # the link, right here
          docket.save(update_fields=["originating_court_information"])
          return result
​
​
  def merge_florida_docket(data: FloridaCase) -> MergeResult:
      if data.court_id not in FLORIDA_COURT_ID_MAP:
          return MergeResult.failed("Docket")
      return FloridaDocketMerger.merge(data)
```
​
  And the follow-up-PR shape needs no new base layer — locking is base config, follow-ups are a base mechanism:
​
 ```python
 class FloridaDocketEntryMerger(Merger[FloridaDocketEntry, DocketEntry]):
      model = DocketEntry
      lock_parent = "docket"          # the base does select_for_update — no DocketEntryMerger needed
      fields = (...)
​
      @classmethod
      def find_existing(cls, data, *, docket, **ctx):
          return DocketEntry.objects.filter(
              docket=docket, date_filed=data.date, entry_type=data.type
          ).first()
```
</details>

File: cl/corpus_importer/state/florida/mergers.py (hunk 12 above)
Comment (MorganBennetDev) [reply cl/corpus_importer/state/florida/mergers.py]: I initially wrote these mergers in the same style as the Texas mergers and then converted them once I had an abstraction figured out. I converted `merge_oci` to a wrapper in that migration so I didn't have to update tests. Will get rid of it.

File: cl/corpus_importer/state/florida/mergers.py (hunk 13 above)
Comment (MorganBennetDev) [reply cl/corpus_importer/state/florida/mergers.py]: > Would it be similar to process_scotus_docket, which is a Celery task that merges the full docket in the background? If that’s the case, should this method also be a task?

Yes. The plan is to take the same approach as SCOTUS and Texas i.e. chained tasks for parsing and merging each individual docket along with all associated objects. The actual task would take the raw text as input and call this method after it's been parsed.

File: cl/corpus_importer/state/florida/mergers.py (hunk 11 above)
Comment (MorganBennetDev) [reply cl/corpus_importer/state/florida/mergers.py]: I think I'm confused. Shouldn't the `save` in `find_and_disaggregate_docket_object`/`find_docket_object` be disabled by setting `allow_create=False`?

**Comment by MorganBennetDev** [comment]
> My understanding is that Merger is currently generic over the model, right? Is that why it needs all the machinery around `__init_subclass__`? You also have classes like RelatedMerger and Relationship that are intended to keep the merged model agnostic, which is good and aligns with the goals of this effort.

Currently `__init_subclass__` handles two things:

The first is doing a bunch of shenanigans with introspection on `DBModel` since Python doesn't do monomorphization for generics so the type parameter is never mapped to a concrete type. I agree that this is overly complex and we should just require a `ClassVar[type[Model]]` be set on inheritors; the minor front-end convenience isn't worth the confusing (and probably brittle) backend.

The second is finding our merger mappings and caching them somewhere for easy access. One of my goals with this prototype was to give us a way to define mergers that was as simple as Pydantic models, which the `attr = <merge logic definition>` pattern is the core of. Regardless of the shape of the `<merge logic>` component, I think we should keep the attribute registration part of `__init_subclass__`. The logic isn't too complex and removing it would involve either sacrificing performance or moving to something like the `fields: ClassVar[Sequence[Field]]` pattern in Claude's proposal, which I feel is a bit less intuitive to work with (but that's a personal preference).

> Instead of having a single generic `Merger`, my proposal would be to create a small family of model-specific sub-classes based on `Merger`, such as `FloridaDocketMerger` (this PR), followed by `FloridaPartyMerger`, `FloridaDocketEntryMerger`, and `FloridaDocumentMerger` in future PRs. Each base class would simply define `model = Docket` (or the corresponding model) as a regular class attribute, eliminating the need for `__init_subclass__`.

This is roughly what I'm aiming for as well. My idea with merger inheritance was to use my work in Juriscraper on base classes for different kinds of parser output to define some base mergers that can handle the things that are the same for every merger (i.e. `case_name` from Juriscraper always goes to `case_name` in CourtListener) and limit state mergers to state-specific logic (i.e. different OCI merging or DN normalization).

> `InputMap`, `InputField`, `Parameter`, `Constant`, and `PassAll` (replaced with simple callables, e.g., `lambda c: c.case_name`)

Sounds good. I was a bit too aggressive on the abstraction front with those.

tl;dr: I agree on replacing `InputMap` and subclasses with pure functions and getting rid of the introspection to find `DBModel`, but I think we should keep `__init_subclass__`.

**Comment by Brennan-Chesley-FLP** [comment]
Overall I like the approach! 

A questions about how I would adapt this for the NYCoA, and one about future expansion.

Docket entries are just a file with a name about half the time. The names aren't guaranteed to be unique (there are at least 11 cases where one docket has multiples of the same file name). When there's no natural key, should we be synthesizing one? What happens if our synthetic natural keys aren't stable?

When we observe data contraction from a state (they remove a docket entry, or a party from a case or the like), how do you foresee expanding this to account for pruning those outdated records?

**Comment by MorganBennetDev** [comment]
> Docket entries are just a file with a name about half the time. The names aren't guaranteed to be unique (there are at least 11 cases where one docket has multiples of the same file name). When there's no natural key, should we be synthesizing one? What happens if our synthetic natural keys aren't stable?

In this case you would have to synthesize a key or override `get_existing`, yes. Texas (and SCOTUS IIRC) do this currently with `sequence_number`. I don't think we've had an issue with docket entries changing order and messing up that matching yet.

> When we observe data contraction from a state (they remove a docket entry, or a party from a case or the like), how do you foresee expanding this to account for pruning those outdated records?

This isn't currently handled in this prototype or the Texas merger, so I'm not sure. Probably worth thinking about.

File: cl/corpus_importer/state/florida/mergers.py (hunk 11 above)
Comment (albertisfu) [reply cl/corpus_importer/state/florida/mergers.py]: Yeah, the issue is in these lines within `find_and_disaggregate_docket_object`:

```python
if docket is not None:
    docket.court = await Court.objects.aget(pk=court_id)
    _ = await docket.asave()    # save 1 (court reassigned only)
    return docket, True
```

Then `_merge_tree` writes the merged attributes and calls `db_obj.save()`, which is the second save.

Could we keep `court_id` only in memory by just doing `docket.court_id = court_id` and avoid that first save?

So `allow_create=False` isn't helping here because the first save is happening outside `find_docket_object`.

cl/recap/mergers.py
```diff
@@ -287,6 +296,64 @@ async def find_docket_object(
...
+        allow_create=False,
+    )
+    if docket is not None:
+        logger.info(
+            "Disaggregating docket %s from %s to %s",
+            docket.pk,
+            aggregate_court_id,
+            court_id,
+        )
+        docket.court = await Court.objects.aget(pk=court_id)
```
albertisfu: Can we just do:
```suggestion
        docket.court_id = court_id
```

To avoid doing a `Court` lookup here?

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -0,0 +1,196 @@
...
+    )
+    case_name: str = AttributeMerger[FloridaCase, str](
+        lambda d: d.case_name, strategy=overwrite
+    )
+    case_name_full: str = AttributeMerger[FloridaCase, str](
+        lambda d: d.case_name_full,
+        strategy=overwrite,
+    )
+    case_name_short: str = AttributeMerger[FloridaCase, str](
+        lambda d: d.case_name, strategy=overwrite
```
albertisfu: I'm noticing that we're using `make_case_name_short` for RECAP dockets:

```
from juriscraper.lib.string_utils import CaseNameTweaker

cnt = CaseNameTweaker()
```

to pre-process case_name_short.

Maybe we can using it here?

```suggestion
        lambda d: cnt.make_case_name_short(d.case_name), strategy=overwrite
```

It looks like we're not populating case_name_short for SCOTUS and Texas, which is why we're not using this there. If we want, we can populate it later with a script.

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -0,0 +1,196 @@
...
+    )
+    appeal_from_id: str | None = AttributeMerger(
+        _appeal_from_id, strategy=overwrite
+    )
+    appeal_from_str: str | None = AttributeMerger(
+        _appeal_from_str, strategy=overwrite
+    )
+    # See https://github.com/freelawproject/courtlistener/issues/7361#issuecomment-4566459292
+    pacer_case_id: str = AttributeMerger(
+        lambda d: d.case_uuid, strategy=overwrite
```
albertisfu: It looks like we need to cast this to a `str`. Otherwise, the value of `pacer_case_id` is a UUID, and within `_merge_tree` it won't match the `db_object` value, which is a string. As a result, it'll always take the update path, causing an unnecessary `save()` whenever any fields have changed.

I suggest adding a test to cover this merge case to prevent a regression.

```suggestion
        lambda d: str(d.case_uuid), strategy=overwrite
```

albertisfu (changes requested):
Thanks for the changes @MorganBennetDev ! I just left some additional suggestions/changes please.

cl/corpus_importer/state/florida/mergers.py
MorganBennetDev: Oh, I see. I'll figure out a way to fix this.

cl/recap/mergers.py
MorganBennetDev: I've realized we don't even need this method with how things are set up. Will be removing and reverting Texas changes.

cl/corpus_importer/state/florida/mergers.py
MorganBennetDev: Since `CaseNameTweaker` is from `juriscraper`, should we just run this over there instead?

cl/corpus_importer/state/florida/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    :param court_id: The court ID to check.
+    :return: Whether the ID belongs to a Florida state court."""
+
+    return court_id in FLORIDA_COURT_IDS
+
+
+# <District>D<Year>-<Case number>
+FLORIDA_APPELLATE_DN_RE = re.compile(r"[1-6]D\d{4}-\d{4}")
+# SC<Year>-<Case number>
+FLORIDA_SUPREME_DN_RE = re.compile(r"SC\d{4}-\d{4}")
```
Brennan-Chesley-FLP: These regex diverge from the slightly more broad one [in juriscraper](https://github.com/freelawproject/juriscraper/blob/main/juriscraper/state/florida/common.py#L48)) . 5 digits seems likely to me for Florida, but I don't recall off the top of my head what the yearly volume is in that state.

cl/corpus_importer/state/merger.py
```diff
@@ -0,0 +1,574 @@
...
+        :param existing: Optional existing object to override the lookup in `get_existing`.
+        :param manager: The manager to use for lookups and creation of objects. Defaults to the model's default manager."""
+        self.guarantee_valid()
+        result = None
+        valid = self.validate(scrape)
+        if not valid:
+            logger.error(
+                f"Merger {self.__class__.__name__} received invalid input."
+            )
+            result = MergeResult.failed(self.model.__name__)
```
Brennan-Chesley-FLP, line 440: These get raised at line 483 as "already merged; cannot merge again" . I don't think it's very likely to happen in production, but if there's a clean way of returning the MergeResult.failed and keeping this clearer error reason attached, it would be nice.

cl/corpus_importer/state/merger.py
```diff
@@ -0,0 +1,574 @@
...
+
+    def after(self) -> None:
+        """Run extra processes after the merge operation completes or fails."""
+        ...
+
+    def query(self) -> QuerySet[M]:
+        """Constructs a queryset to find an existing object in the DB, using the natural key defined by `cls.key`.
+
+        :return: The queryset to find the object."""
+        return self.manager.filter(
```
Brennan-Chesley-FLP, line 474: Slightly in favor of raising an error here if self.key isn't overridden, but that's a matter of personal preference.

cl/corpus_importer/state/merger.py
```diff
@@ -0,0 +1,574 @@
...
+        if isinstance(merger_input, (str, bytes)):
+            return MergeResult.failed(self.merger.model.__name__)
+        result = MergeResult.unnecessary()
+        related_manager: RelatedManager[Any] = getattr(parent, self.name)
+        for child in cast(Iterable[Any], merger_input):
+            result |= self.merger(
+                child, manager=related_manager, params=params
+            ).merge()
+        #     children.append(cast(Model, child_obj))
+        # related_manager.add(*children)
```
Brennan-Chesley-FLP: Can we safely delete these?

cl/corpus_importer/state/merger.py
```diff
@@ -0,0 +1,574 @@
...
+                )
+        return errors
+
+    @abstractmethod
+    def merge(
+        self, parent: RM, scrape: ScrapeType, params: ParamType
+    ) -> MergeResult[Any]: ...
+
+
+class OneToOneMerger[ScrapeType, ParamType, ChildType, RM: Model](
```
Brennan-Chesley-FLP, line 190: If we need a reverse 1-1 relation in the future, do you think we should expand support here or make another class so it's explicit?

Brennan-Chesley-FLP (approved):
Looked over the changes since my last approval and I still like it. Nothing that looks like a blocker to me.

cl/corpus_importer/state/florida/utils.py
MorganBennetDev: Good catch. There are 35,837 cases with 5-digits after the dash in the full scrape.

cl/corpus_importer/state/merger.py
MorganBennetDev: I've updated it to just return the failed merge and log an error again. Regardless of what the caller does with the failed results, the error logs will surface in Sentry where (hopefully!) someone will fix them.

cl/corpus_importer/state/merger.py
MorganBennetDev: I'm relying on the merger just not being very useful if you don't override `key` or `query`, and review catching that.

cl/corpus_importer/state/merger.py
MorganBennetDev: I really hope we don't, but if we do need to merge one of those someday, we can expand support here. I'm trying to keep the terminology for mergers lined up with Django's model definitions to minimize cognitive overhead.

cl/corpus_importer/state/florida/utils.py
```diff
@@ -0,0 +1,91 @@
...
+    return court_id in FLORIDA_COURT_IDS
+
+
+# <District>D<Year>-<Case number>
+FLORIDA_APPELLATE_DN_RE = re.compile(r"[1-6]D\d{4}-\d{4,5}")
+# SC<Year>-<Case number>
+FLORIDA_SUPREME_DN_RE = re.compile(r"SC\d{4}-\d{4,5}")
+
+
+def make_docket_number_core(
```
albertisfu, line 43: Can we apply the same transformation we’re using for the Texas docket number core by stripping all non-alphanumeric characters and lowercasing it for consistency?

Also, adding tests for the different FL numbers would be helpful to prevent variations in the number format this method generates.

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -0,0 +1,183 @@
...
+    )
+    originating_court_information: OriginatingCourtInformation = (
+        OneToOneRelation(
+            FloridaOriginatingCourtInformationMerger,
+            _originating_case,
+        )
+    )
+
+    @override
+    def query(self) -> QuerySet[Docket]:
```
albertisfu, line 150: I was testing this and found that two cases with the same UUID but different docket numbers are not matching. We should prioritize the UUID (`pacer_case_id`) as the primary source of truth when matching existing cases, even if the docket number has changed (for example, due to a correction from the court).

We could use logic similar to `find_docket_object`, where `pacer_case_id` is always prioritized, either in combination with the docket number or on its own.

Why wasn't it possible to continue using `find_docket_object` here? It would be great to keep the matching logic in a single place if possible.

albertisfu (changes requested):
Thanks, @MorganBennetDev! This looks almost ready.

I just left a couple of additional comments. Also, could you please add all the new files introduced in this PR to the mypy checks in lint.yml?

It might be better to add the entire state directory:

`cl/corpus_importer/state/`

MorganBennetDev:
> Also, could you please add all the new files introduced in this PR to the mypy checks in lint.yml?

I think they're already covered by `corpus_importer` being in the package check. It's already caught a few errors in this PR.

cl/corpus_importer/state/florida/mergers.py
MorganBennetDev: > Why wasn't it possible to continue using `find_docket_object` here? It would be great to keep the matching logic in a single place if possible.

When I wrote this, I didn't want to update `find_docket_object` to be able to return a query set, but I should probably just do that. Maybe something like:

- Create `find_docket_object_query` which will house the `for kwargs in lookups` loop from `find_docket_object`, returning a queryset, and
- Update  `find_docket_object` to only do the DN core validation and checks, leaving the actual lookup logic in `find_docket_object_query`.

Then Florida and future `Merger` subclasses can use `find_docket_object_query` to keep the same logic as `find_docket_object` (minus the DN core conversion, which they should handle themselves).

What do you think?

MorganBennetDev:
@albertisfu Updated based on your feedback and implemented my proposed solution to your comment about `find_docket_object`. I also slightly simplified the query construction for `find_docket_object`, but the logic should be unchanged.

cl/recap/mergers.py
```diff
@@ -126,119 +130,107 @@ def confirm_docket_number_core_lookup_match(
     return docket
 
 
-async def find_docket_object(
+async def find_docket_object_query(
```
albertisfu, line 105: Is there any advantage to using a queryset in `FloridaDocketMerger`?

The downside I see with this refactor is that the original `find_docket_object` method returned the `Docket` object directly from the database:

```python
d = await ds.afirst()
...
if d:
    break  # Nailed it!
```

Having `find_docket_object_query` return a queryset could duplicate some database queries. The method still needs to retrieve the docket from the database because the instance is used in `confirm_docket_number_core_lookup_match`. However, with this refactor, that `Docket` instance is not reused. Instead, a queryset is returned, and the caller queries the database again to retrieve the same docket.

This may seem harmless because these queries are relatively fast. However, `find_docket_object` is one of the methods that hits our database most frequently, so we should be careful not to increase the database load unnecessarily during this refactor.

As an alternative, could `FloridaDocketMerger` accept a `Docket` instance instead of a queryset?

cl/corpus_importer/state/florida/tests.py
```diff
@@ -0,0 +1,363 @@
...
+)
+from cl.corpus_importer.state.florida.utils import make_docket_number_core
+from cl.search.factories import CourtFactory, DocketFactory
+from cl.search.models import Docket, OriginatingCourtInformation
+from cl.tests.cases import TestCase
+
+
+class FloridaUtilsTest(TestCase):
+    def test_docket_number_core(self) -> None:
+        """Can we correctly normalize Texas docket numbers?"""
```
albertisfu: Docstring should be updated

cl/corpus_importer/state/florida/tests.py
```diff
@@ -0,0 +1,363 @@
...
+        self.assertEqual(make_docket_number_core("SC1983-2014"), "sc19832014")
+        self.assertEqual(
+            make_docket_number_core("SC1983-2014", court_id="fla"),
+            "sc19832014",
+        )
+        self.assertEqual(
+            make_docket_number_core("3D2001-20145", court_id="fla"),
+            "3D2001-20145",
```
albertisfu: Is this an invalid docket number for `fla`? Is that why it is not being transformed?

If it is invalid, or in any other case where the docket number is not a valid Florida docket number, the `make_docket_number_core` method should return an empty string, similar to what we do for PACER, SCOTUS, and Texas.

albertisfu (changes requested):
Thanks for the changes @MorganBennetDev I just have a couple of extra comments, please.

cl/recap/mergers.py
MorganBennetDev: > Is there any advantage to using a queryset in `FloridaDocketMerger`?

Not in this PR, but in the party merging PR, I found `QuerySet`s made m2m relations with `through` models much easier to work with. If we return an object instead of a queryset, the mergers for `through` models will have to manually implement the matching and merging of source and target FKs, where it's handled automatically right now.

Additionally, my hope with returning a `QuerySet` was to leave us open to optimizing merger DB access with `prefetch_related` in the future.

> Having `find_docket_object_query` return a queryset could duplicate some database queries. The method still needs to retrieve the docket from the database because the instance is used in `confirm_docket_number_core_lookup_match`. However, with this refactor, that `Docket` instance is not reused. Instead, a queryset is returned, and the caller queries the database again to retrieve the same docket.

I misunderstood how Django's query caching works so I thought this wouldn't create an additional query, but `get` seems to bypass that cache. However, `first` does use the cache if the `QuerySet` is ordered. So if we add `.order_by("pk")` to `ds` on line 217 and change the return on 344 to `dqs.afirst()`, we can get rid of the redundant queries without changing behavior.

cl/corpus_importer/state/florida/tests.py
MorganBennetDev: > Is this an invalid docket number for `fla`? Is that why it is not being transformed?

Yes, since this is a district court DN and the FLSC is being passed as the `court_id`.

cl/recap/mergers.py
albertisfu: I ran some tests on the refactored `find_docket_object()` method to confirm that no extra queries are being made compared to the original implementation. I found that the code paths that use the confirm method are still executing one additional query.

Here's the code I used for this test:

```
from asgiref.sync import async_to_sync
from django.db import connection
from django.test.utils import CaptureQueriesContext

from cl.lib.model_helpers import (
    make_docket_number_core,
    make_scotus_docket_number_core,
)
from cl.recap.mergers import find_docket_object
from cl.search.factories import CourtFactory, DocketFactory
from cl.tests.cases import TestCase

DN = "2:20-cv-01234"


class FindDocketObjectQueryCountTest(TestCase):
    @classmethod
    def setUpTestData(cls):
        cls.court = CourtFactory(id="canb", jurisdiction="FB")
        cls.scotus = CourtFactory(id="scotus", jurisdiction="F")

    def _run(self, label, expect_new=False, expect_pk=None, **kwargs):
        kwargs.setdefault("federal_defendant_number", None)
        kwargs.setdefault("federal_dn_judge_initials_assigned", None)
        kwargs.setdefault("federal_dn_judge_initials_referred", None)
        with CaptureQueriesContext(connection) as ctx:
            d = async_to_sync(find_docket_object)(**kwargs)
        n = len(ctx.captured_queries)
        print(f"\n### {label}: {n} queries")
        for q in ctx.captured_queries:
            print("    ", q["sql"][:160])
        if expect_new:
            self.assertIsNotNone(d, label)
            self.assertIsNone(d.pk, f"{label}: expected a new unsaved docket")
        if expect_pk is not None:
            self.assertIsNotNone(d, label)
            self.assertEqual(d.pk, expect_pk, f"{label}: wrong docket matched")
        return n

    def test_query_counts(self):
        core = make_docket_number_core(DN)
        results = {}

        # S1: lookup (pacer_case_id + core), count==1, confirm=False
        d1 = DocketFactory(
            court=self.court,
            source=Docket.RECAP,
            pacer_case_id="11111",
            docket_number=DN,
            docket_number_raw=DN,
            docket_number_core=core,
        )
        results["S1 pcid+core, count=1, no confirm"] = self._run(
            "S1 pcid+core, count=1, no confirm",
            expect_pk=d1.pk,
            court_id=self.court.pk,
            pacer_case_id="11111",
            docket_number=DN,
        )
        d1.delete()

        # S2: reaches lookup (pcid=None + core) because pcid doesn't match,
        # count==1, confirm=True (federal)
        d2 = DocketFactory(
            court=self.court,
            source=Docket.DEFAULT,
            pacer_case_id=None,
            docket_number=DN,
            docket_number_raw=DN,
            docket_number_core=core,
        )
        results["S2 null-pcid+core, count=1, confirm"] = self._run(
            "S2 null-pcid+core, count=1, confirm",
            expect_pk=d2.pk,
            court_id=self.court.pk,
            pacer_case_id="22222",
            docket_number=DN,
        )
        d2.delete()

        # S3: reaches lookup (pcid only) after two misses, count==1,
        # confirm=False
        d3 = DocketFactory(
            court=self.court,
            source=Docket.RECAP,
            pacer_case_id="33333",
            docket_number="9:99-cv-09999",
            docket_number_raw="9:99-cv-09999",
            docket_number_core=make_docket_number_core("9:99-cv-09999"),
        )
        results["S3 pcid only, count=1, no confirm"] = self._run(
            "S3 pcid only, count=1, no confirm",
            expect_pk=d3.pk,
            court_id=self.court.pk,
            pacer_case_id="33333",
            docket_number=DN,
        )
        d3.delete()

        # S4: no pacer_case_id, lookup (pcid=None + core), count==1,
        # confirm=True
        d4 = DocketFactory(
            court=self.court,
            source=Docket.DEFAULT,
            pacer_case_id=None,
            docket_number=DN,
            docket_number_raw=DN,
            docket_number_core=core,
        )
        results["S4 no-pcid, null-pcid+core, count=1, confirm"] = self._run(
            "S4 no-pcid, null-pcid+core, count=1, confirm",
            expect_pk=d4.pk,
            court_id=self.court.pk,
            pacer_case_id=None,
            docket_number=DN,
        )
        d4.delete()

        # S5: count>1, refined to exactly 1 by DN components
        d5a = DocketFactory(
            court=self.court,
            source=Docket.DEFAULT,
            pacer_case_id=None,
            docket_number=DN,
            docket_number_raw=DN,
            docket_number_core=core,
            federal_dn_judge_initials_assigned="AB",
        )
        d5b = DocketFactory(
            court=self.court,
            source=Docket.DEFAULT,
            pacer_case_id=None,
            docket_number=DN,
            docket_number_raw=DN,
            docket_number_core=core,
            federal_dn_judge_initials_assigned="CD",
        )
        results["S5 count>1, dn-refine to 1"] = self._run(
            "S5 count>1, dn-refine to 1",
            expect_pk=d5a.pk,
            court_id=self.court.pk,
            pacer_case_id=None,
            docket_number=DN,
            federal_dn_judge_initials_assigned="AB",
        )
        d5a.delete()
        d5b.delete()

        # S6: count>1, no DN components, fallback to oldest, confirm=True
        d6a = DocketFactory(
            court=self.court,
            source=Docket.DEFAULT,
            pacer_case_id=None,
            docket_number=DN,
            docket_number_raw=DN,
            docket_number_core=core,
        )
        d6b = DocketFactory(
            court=self.court,
            source=Docket.DEFAULT,
            pacer_case_id=None,
            docket_number=DN,
            docket_number_raw=DN,
            docket_number_core=core,
        )
        results["S6 count>1, fallback oldest, confirm"] = self._run(
            "S6 count>1, fallback oldest, confirm",
            expect_pk=d6a.pk,
            court_id=self.court.pk,
            pacer_case_id=None,
            docket_number=DN,
        )
        d6a.delete()
        d6b.delete()

        # S7: SCOTUS (confirmation skipped), count>1, fallback to oldest
        sdn = "22-123"
        score = make_scotus_docket_number_core(sdn)
        d7a = DocketFactory(
            court=self.scotus,
            source=Docket.DEFAULT,
            pacer_case_id=None,
            docket_number=sdn,
            docket_number_raw=sdn,
            docket_number_core=score,
        )
        d7b = DocketFactory(
            court=self.scotus,
            source=Docket.DEFAULT,
            pacer_case_id=None,
            docket_number=sdn,
            docket_number_raw=sdn,
            docket_number_core=score,
        )
        results["S7 scotus count>1, fallback, no confirm"] = self._run(
            "S7 scotus count>1, fallback, no confirm",
            expect_pk=d7a.pk,
            court_id=self.scotus.pk,
            pacer_case_id=None,
            docket_number=sdn,
        )
        d7a.delete()
        d7b.delete()

        # S8: nothing matches at all -> new unsaved docket
        results["S8 no match, new docket"] = self._run(
            "S8 no match, new docket",
            expect_new=True,
            court_id=self.court.pk,
            pacer_case_id="55555",
            docket_number="1:11-cv-11111",
        )

        # S9: core matches but confirmation FAILS (cv vs mc) -> continue ->
        # new docket
        other_dn = "2:20-mc-01234"
        self.assertEqual(
            core,
            make_docket_number_core(other_dn),
            "S9 precondition: cores must collide",
        )
        d9 = DocketFactory(
            court=self.court,
            source=Docket.DEFAULT,
            pacer_case_id=None,
            docket_number=other_dn,
            docket_number_raw=other_dn,
            docket_number_core=core,
        )
        results["S9 confirm fails -> new docket"] = self._run(
            "S9 confirm fails -> new docket",
            expect_new=True,
            court_id=self.court.pk,
            pacer_case_id=None,
            docket_number=DN,
        )
        d9.delete()

        print("\n" + "=" * 60)
        print("SUMMARY (queries per scenario)")
        for k, v in results.items():
            print(f"  {v}  {k}")
        print("=" * 60)
```

You can run it first on main and then on your branch so you can compare the number of queries in each. Ideally, both versions should execute the same number of queries across all code paths.

If returning a queryset is helpful for other mergers, like the one you mentioned for parties, then that makes sense. However, if it's not possible to avoid the extra queries, maybe we could make an exception only for the Florida docket merger and have it accept a Docket instance instead of a queryset.

One option would be to add a bridge method like `_get_existing()` to `Merger`, which by default would simply call `self.query().get()`. The `merge()` method would then use `_get_existing()`.

The Florida docket merger could override `_get_existing()` and use the Docket instance returned by `find_docket_object()` instead.

MorganBennetDev:
@albertisfu Updated to get rid of the extra query, and managed to also eliminate one compared to pre-refactor from the S5 case in your comment.

I moved some of the confirmation logic from Python to the query as well, but that can be reverted if necessary.

cl/recap/mergers.py
```diff
@@ -82,48 +86,169 @@
...
+    for confirm, query in lookups:
+        q = Q(court_id=court_id)
+        if confirm:
+            q &= confirm_query
+        ds = (
+            Docket.objects.filter(q & query)
+            .order_by("date_created")
+            .using(using)
+        )
+        count = await ds.values("pk")[:2].acount()
```
albertisfu, line 222: Could you please add a comment here to explain why we are using `[:2] `to improve the query performance?

albertisfu (changes requested):
Thanks @MorganBennetDev! The refactor looks great. I noticed that the new `find_docket_object` method is not only performing the same number of queries, but is also saving one query in some cases. Along with the performance improvement in the count queries, this is looking great.

I just made one suggestion to document the improved count query, please.

MorganBennetDev:
@albertisfu Added a comment explaining the slice.

albertisfu (approved):
Thanks, @MorganBennetDev, for all the work here. This is a big step toward making mergers easier to implement for future state content. The `find_docket_object` query should also be more performant now!

mlissner:
Wooo! Congrats all!

PR #7438, opened 2026-06-10
fix(users): Fix bar-membership wipe and purge user data/PII on account deletion

### LINKED ISSUE #7427: SearchQuery objects not deleted along with users
When we delete users, the function `delete_user_assets` is run. It's supposed to delete anything related to the user, but it doesn't currently delete the user's `SearchQuery` objects. After it's run, the account is converted to a stub account.

We should:
 - Fix that so SearchQuery objects get deleted properly.
 - Go through any stub accounts and make sure they don't have any SearchQuery objects associated with them

mlissner:
## Fixes

A funny bug is that every now and then somebody writes us a note saying that their state isn't listed in the Bar Memberships list. About a month ago when it happened, I laughed it off, and fixed it carefully, once and for all. 

Well, guess what? It came back today, so I had claude dig into it. We had a bug that was deleting rows from the table instead of clearing the M2M relationship. 

This code fixes that by switching from:

```
    profile.barmembership.all().delete()
```

To:

```
    profile.barmembership.all().clear()
```

🤦🏻 

While I was in here, I had claude look for other things that weren't getting deleted, and it turned up:

 - Event tables
 - Visualizations (which we soft deleted prior to this)
 - Search queries (which would get nuked by our privacy policy eventually)
 - Emails (which would get nuked by our privacy policy eventually)

So this fixes that too, and adds a long test.

Fixes: #7427

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

We also need to run this to clean up past records:

```
from django.contrib.auth.models import User

from cl.search.models import SearchQuery
from cl.users.models import (
    EmailSent,
    UserProfile,
    UserProfileBarMembershipEvent,
    UserProfileEvent,
    UserProxyEvent,
)
from cl.visualizations.models import SCOTUSMap

# Deleted accounts only. Donation stubs stay is_active=True, so this skips
# them (we must not touch active donors' data).
deleted_users = User.objects.filter(
    profile__stub_account=True, is_active=False
)
deleted_profiles = UserProfile.objects.filter(
    stub_account=True, user__is_active=False
)
user_ids = deleted_users.values("pk")
profile_ids = deleted_profiles.values("pk")

# Each entry: label, queryset of rows that will be deleted.
targets = [
    ("SearchQuery", SearchQuery.objects.filter(user__in=deleted_users)),
    ("EmailSent", EmailSent.objects.filter(user__in=deleted_users)),
    ("SCOTUSMap", SCOTUSMap.objects.filter(user__in=deleted_users)),
    ("UserProfileEvent", UserProfileEvent.objects.filter(user_id__in=user_ids)),
    ("UserProxyEvent", UserProxyEvent.objects.filter(pgh_obj_id__in=user_ids)),
    (
        "UserProfileBarMembershipEvent",
        UserProfileBarMembershipEvent.objects.filter(
            userprofile_id__in=profile_ids
        ),
    ),
]

# Dry run: show the scope and the per-model row counts before touching anything.
print(f"\nDeleted stub accounts in scope: {deleted_users.count()}")
print("Rows that will be deleted:")
for label, qs in targets:
    print(f"  {label}: {qs.count()}")

answer = input("\nProceed with deletion? Type 'yes' to continue: ")
if answer.strip().lower() != "yes":
    print("Aborted. Nothing was deleted.")
else:
    for label, qs in targets:
        print(f"{label}:", qs.delete())
    print("Done.")
```


cl/users/utils.py
```diff
@@ -77,7 +82,18 @@ def convert_to_stub_account(user: User) -> User:
     profile.zip_code = None
     profile.save()
 
-    profile.barmembership.all().delete()
+    profile.barmembership.clear()
```
mlissner, line 85: This is the fix for bar memberships. The rest is just making the deletion stuff better since I was here.

cl/users/tests.py
```diff
@@ -688,6 +698,115 @@ async def test_nuke_user_tag_and_docket_tag_history_objects(self) -> None:
...
+            ).exists()
+        )
+        self.assertTrue(
+            UserProfileBarMembershipEvent.objects.filter(
+                userprofile_id=bystander.pk
+            ).exists()
+        )
+        self.assertTrue(
+            MonthlyDonation.objects.get(donor=bystander.user).enabled
+        )
```
ERosendo, line 808: Could we add a couple of assertions that verify the BarMembership data directly? Something like:

```python
self.assertTrue(BarMembership.objects.filter(pk=bar.pk).exists())
self.assertTrue(bystander.barmembership.filter(pk=bar.pk).exists())
```

This would help the test fail more explicitly if the shared membership row ever gets deleted again, making it easier to catch a regression if that behavior is accidentally reintroduced.

ERosendo (approved):
Looks good to me! I just left a small suggestion around the test assertions to make the regression a little more obvious if it ever comes back. 👍

mlissner:
I added the tweak to the test. 

Did you also review the code in my original comment? If so, want to merge and run it?

ERosendo:
@mlissner I ran the cleanup script and here's the summary:

```
Deleted stub accounts in scope: 303

Proceed with deletion? Type 'yes' to continue: yes 

SearchQuery: (55799, {'search.SearchQuery': 55799})
EmailSent: (13818, {'users.EmailSent': 13818})
SCOTUSMap: (173, {'visualizations.SCOTUSMap_clusters': 165, 'visualizations.JSONVersion': 4, 'visualizations.SCOTUSMap': 4})
UserProfileEvent: (1085, {'users.UserProfileEvent': 1085})
UserProxyEvent: (1805, {'users.UserProxyEvent': 1805})
UserProfileBarMembershipEvent: (16, {'users.UserProfileBarMembershipEvent': 16})
Done.
```

mlissner:
Wonderful, thank you. That cleaned up a lot more than I thought it would!

PR #7441, opened 2026-06-11
feat(api): Add parentheticals v4 endpoint

### LINKED ISSUE #4082: Create a parentheticals API
We put these in the bulk data, but we don't yet have an API for them. Nobody has asked for this yet, but we need a placeholder for people to upvote and comment on. 

This issue can serve that purpose.

isabellaaguilar:
## Summary

Adds a new `/api/rest/v4/parentheticals/` endpoint that exposes parenthetical data programmatically.

Fixes #4082

## What it does

- Lists parentheticals describing a given opinion (`?described_opinion=<id>`)
- Lists parentheticals authored by a given opinion (`?describing_opinion=<id>`)
- Embeds group information inline when already computed
- Filters by score (`?score__gte=<n>`)
- Includes case names for both opinions to avoid follow-up requests

## Design decisions

See the design document posted in #4082 for full rationale, including:
- Why a flat endpoint (not nested)
- Why the API never computes groups on demand
- Why cursor pagination by id (not score)

## AI disclosure

I used an AI assistant to explore the codebase, draft the design document, and scaffold the implementation. All design decisions and trade-off analysis are my own.

## Testing

- 10 tests covering filtering, ordering, lazy computation, group embedding, and N+1 prevention
- All pass locally (`manage.py test cl.api.tests_parentheticals`)
- `ruff check` and `ruff format` clean

Screenshot 2026-06-11 at 10 45 35 AM


isabellaaguilar:
# Design: A Parentheticals API for CourtListener (v4)

*Issue #4082*

> **AI disclosure (per the challenge rules and FLP's AI policies):** I used an AI
> assistant to help explore the codebase, draft this document, and sanity-check the
> approach. All design decisions, trade-off analysis, and the final reasoning are my
> own, and I reviewed every claim against the actual source.

## Goal

Expose parentheticals through the v4 REST API so a client can, for a given opinion:

1. List the parentheticals that **describe** it (how other courts summarized this case).
2. List the parentheticals it **authored** about other cases (its table of authorities).
3. Access grouping information where it exists.

The existing `/api/rest/v4/opinions-cited/` endpoint is the closest precedent and my
reference point.

## What the data looks like (and why it matters)

From `cl/search/models.py`:

- `Parenthetical`: `describing_opinion` (FK → Opinion, `authored_parentheticals`),
  `described_opinion` (FK → Opinion, `parentheticals`), `group` (FK → ParentheticalGroup,
  **nullable, `SET_NULL`**), `text`, and `score` (**`db_index=True`**, 0–1).
- `ParentheticalGroup`: `opinion` (the described opinion), `representative` (a Parenthetical),
  `score`, `size`. It has a DB index on `score`.

Two facts drive the whole design:

1. **The cheap filters are already indexed.** Both FKs are indexed (Django default) and
   `score` is explicitly indexed. Filtering by described/describing opinion or by score
   range is cheap. Free-text search over `text` is not, so I will not expose it here.
2. **Groups are normally computed at ingestion; lazy computation is a fallback.** When an
   opinion is processed, `find_citations_and_parentheticals_for_opinion_by_pks`
   (`cl/citations/tasks.py`) creates the parentheticals *and* their groups via
   `create_parenthetical_groups`. So in the common case groups already exist. The web UI's
   `get_or_create_parenthetical_groups` (`cl/citations/parenthetical_utils.py`) is a
   **fallback**: the first time someone views a cluster's "Summaries" tab, if any
   parenthetical lacks a group it **deletes and recomputes all groups for that cluster** and
   re-indexes them in Elasticsearch via signals — an expensive write triggered by what looks
   like a read.

   The decisive evidence sits in the code itself: batch jobs deliberately turn group
   creation **off**. The `find_citations` command exposes `--disable-parenthetical-groups`,
   and the team's own comment explains it is used to avoid overloading the database and
   Elasticsearch during bulk work. The maintainers have already concluded that computing
   groups at scale is too expensive to do inline.

## Options considered

I compared three shapes.

### A. A new top-level endpoint: `/api/rest/v4/parentheticals/`

A `ParentheticalViewSet` filterable by `described_opinion`, `describing_opinion`, and
`score`, mirroring `OpinionsCitedViewSet`. Group data is embedded per row via the `group` FK.

- **Pros:** One endpoint serves all three required capabilities (describe / authored /
  grouping) by changing a filter. Consistent with `opinions-cited`. Reuses the existing
  cursor pagination, hyperlinked serializers, filters, and permission classes. Both query
  directions hit indexed columns.
- **Cons:** Group data is denormalized onto each row rather than presented as a first-class
  collection. Clients who want "the groups for this case" must read group fields off
  parenthetical rows (or we add a light grouped view later).

### B. Nest under the existing `opinions` / `clusters` endpoints

E.g. `/opinions/{id}/parentheticals/` and `/clusters/{id}/summaries/`.

- **Pros:** Reads naturally for the "give me this opinion's parentheticals" use case.
  URL encodes the relationship.
- **Cons:** Needs two nested routes to cover both directions (described vs. describing),
  which duplicates wiring. Harder to filter across the table (e.g. "all parentheticals with
  score > 0.8"). Diverges from the flat, filter-driven convention the other v4 relationship
  endpoints follow. The cluster-level "summaries" route is also where the lazy-computation
  temptation is strongest.

### C. Two separate endpoints: `/parentheticals/` and `/parenthetical-groups/`

- **Pros:** Cleanest representation of groups as a first-class resource with its own
  filters and pagination. Good if grouping becomes a primary access pattern.
- **Cons:** More surface area to build, document, and secure for a feature nobody has
  requested yet (the issue is a placeholder). A client wanting rows + their group context
  makes two requests or we duplicate fields anyway.

## Chosen design

**Option A: a single `/api/rest/v4/parentheticals/` endpoint, with group data embedded,
and no lazy computation in the request path.** Option C is the natural next step if demand
for group-centric access appears; A delivers all three required capabilities now with the
least new surface and the most reuse.

### Endpoint behavior

- `GET /parentheticals/?described_opinion=<id>` → parentheticals describing that opinion.
- `GET /parentheticals/?describing_opinion=<id>` → parentheticals it authored (table of
  authorities).
- `GET /parentheticals/?score__gte=<n>` (and friends) → score filtering, combinable with
  the above.
- Cursor pagination keyed on `id` (descending). Relevance is expressed through the `score`
  filter rather than ordering: CourtListener's cursor paginator only supports int/date keys
  (`determine_cursor_position_type`) and needs a unique, monotonic column, so paging by the
  non-unique `score` float would be both unsupported and unstable. This matches every other
  DB endpoint.

Implementation mirrors `OpinionsCitedViewSet`:

```python
class ParentheticalViewSet(
    LoggingMixin, NoFilterCacheListMixin, DeferredFieldsMixin, viewsets.ModelViewSet
):
    serializer_class = ParentheticalSerializer
    filterset_class = ParentheticalFilter
    permission_classes = [DjangoModelPermissions, V3APIPermission]
    ordering = "-id"
    cursor_ordering_fields = ["id"]
    queryset = Parenthetical.objects.select_related(
        "describing_opinion__cluster",
        "described_opinion__cluster",
        "group",
    ).order_by("-id")
```

### The lazy-computation decision

This is the crux, and the right answer depends on what guarantee the API wants to make. A
`GET` should stay a cheap, side-effect-free read, so the real question is *who* is allowed to
pay for an expensive recomputation, and *when*. I weigh three options.

- **(i) Compute inline on `GET`** — rejected. It turns reads into writes, breaks `GET`
  idempotency, and lets an unauthenticated crawler walking opinion IDs trigger cluster-wide
  clustering plus Elasticsearch re-indexing across millions of rows. This is the
  unbounded-cost scenario FLP operates against. Decisively, the maintainers already reached
  this conclusion: batch jobs set `--disable-parenthetical-groups` precisely to avoid that
  load. An API is uncontrolled bulk access by nature; it would be incoherent for it to do
  inline exactly what the batch pipeline refuses to do.

The remaining two are both defensible; the choice turns on the desired guarantee:

- **(ii) Serve only what exists** — my default. Raw `Parenthetical` rows are always returned
  (they are cheap and, after ingestion, already present). Embedded `group` fields are
  populated only when groups were already computed — by ingestion in the common case, or by
  the UI fallback. When a cluster has no groups yet, group fields are null/empty. The
  response honestly reflects stored state and the endpoint can never be coerced into doing
  work. This favors abuse resistance and predictable latency — the properties the challenge
  emphasizes ("huge scale", "prevent expensive or unbounded queries", "consequences at API
  traffic volumes"). The trade-off I accept out loud: group coverage in the API mirrors what
  ingestion produced and what the UI has warmed; a never-processed-or-viewed cluster returns
  parentheticals without groups.

- **(iii) Explicit asynchronous trigger (HTTP 202 + Celery)** — the right choice if a stronger
  guarantee is required: that an API client can obtain groups without depending on the web UI.
  Reads stay cheap as in (ii). Computation moves behind a separate, authenticated,
  rate-limited write action — e.g. `POST /clusters/{id}/compute-parenthetical-groups/`. The
  handler does **not** compute inline; it enqueues a Celery task (the same machinery
  ingestion already uses) and returns `202 Accepted`, meaning "request accepted, processing
  asynchronously." The client polls the normal `GET` later and finds the groups populated.
  This cleanly separates the cheap read from the expensive write, stays safe under load
  (work is queued, not run in the request), and gives clients a first-class way to ask for
  computation. Its costs: more surface area (the action, Celery wiring, permissions, and
  throttling on the trigger to prevent a different abuse vector), and eventual consistency
  instead of an immediate result.

**Decision for this PR:** implement (ii) now — it is safe, predictable, and fits the
"API only" scope and time box. I document (iii) as the natural evolution if API-driven group
completeness becomes a requirement; for very high-citation clusters, an offline pre-warming
job is a third complement. The point is not that (ii) is universally correct, but that the
read path must never carry write load — and between (ii) and (iii) the deciding factor is
whether the API must guarantee group availability independently of the UI.

### Response shape (avoid one follow-up request per row)

Each row includes `text`, `score`, hyperlinks to both `describing_opinion` and
`described_opinion` (declared as explicit `HyperlinkedRelatedField`s — the `opinions-cited`
serializer notes this avoids a query that pulls back *all* opinions), and enough display
context (case name / citation from the related cluster) so a client can render a summary
without a second call. Group context (`group` id, `score`, `size`) is embedded inline. The
`select_related` above keeps this at a constant number of queries; I will lock that in with
`assertNumQueries`.

### Filters, ordering, pagination, abuse resistance

- **Filters:** `ParentheticalFilter(NoEmptyFilterSet)` with `RelatedFilter`s on
  `described_opinion` and `describing_opinion`, plus numeric lookups on `score`. All hit
  indexed columns. `NoEmptyFilterSet` (already used by `opinions-cited`) blocks unfiltered
  full-table scans.
- **Ordering:** by `score` and `id`, both indexed — stable and cheap.
- **Pagination:** cursor pagination (the v4 convention), which avoids deep-offset cost on a
  table with millions of rows.
- **Permissions:** reuse `DjangoModelPermissions` + `V3APIPermission`, same as the
  surrounding endpoints, inheriting the existing throttling/auth posture.

## What I'm not doing (and why)

- No free-text search on `text` (unindexed; expensive at scale; out of scope).
- No on-demand group computation (see above).
- No nested routes or a second groups endpoint yet (Option C remains the clean future step
  if group-centric demand materializes).

## Testing plan

Using existing factories (`ParentheticalFactory`, `ParentheticalGroupFactory`,
`ParentheticalWithParentsFactory`) with `setUpTestData()`:

1. List by `described_opinion` — ordering and shape.
2. List by `describing_opinion` — the authored / table-of-authorities direction.
3. Score filtering.
4. Lazy computation: a cluster with no groups returns rows with empty group fields and
   triggers **no** writes / no clustering.
5. Pagination and `NoEmptyFilterSet` behavior.
6. `assertNumQueries` to prove there is no N+1 across describing/described/group.

PR #7442, opened 2026-06-12
fix(users): Create Zoho Contacts instead of Leads on API milestones

mlissner:
## Fixes

Part of: https://github.com/freelawproject/internal/issues/1021

## Summary

This PR changes the `create_or_update_zoho_account` Celery task so that when a user hits an API v4 usage milestone and has no existing record in Zoho, we create a **Contact** instead of a **Lead**. We're phasing out the Leads module in Zoho altogether.

After I've migrated all Leads over to Contacts, we'll want another PR to remove Lead-related stuff altogether.

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

No special deployment steps are required beyond a standard celery deploy.

### DIFF
--- a/cl/users/tasks.py
+++ b/cl/users/tasks.py
@@ -138,7 +138,7 @@ def create_or_update_zoho_account(
     This task:
     - Builds a Zoho payload from the User model
     - Checks if the user exists as a Contact or Lead in Zoho
-    - Updates the existing record or creates a new Lead if none exist
+    - Updates the existing record or creates a new Contact if none exist
 
     :param user_id: The primary key of the user to sync with Zoho
     :param milestone: A milestone value to store in the 'API_calls' field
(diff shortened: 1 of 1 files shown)


ERosendo (approved):
LGTM :shipit:

ERosendo:
@mlissner I initially set this up for auto-merge, but I disabled it after realizing that the cadences @eriodiah created for **Leads** won't apply to new records in the **Contacts** module. I think we should first create equivalent cadences for Contacts so that new records continue to receive the appropriate email sequences when applicable.

eriodiah:
@ERosendo, Adding this to my to-do. When do you need me to create the cadence for Accounts and Contacts by?

ERosendo:
Hey @eriodiah , that's probably a great question for @mlissner. I'm not sure what the timeline is for migrating from Leads to Contacts.

Also, just to clarify, I think we only need cadences for Contacts (I may have mixed up Accounts and Contacts in my original comment, Sorry! 🙂 )

I took a quick look at the `Contacts` module, and it seems to already include the key fields we use in Leads (the CourtListener ID, the API call tracking field and field for Zoho forms responses). Based on that, my impression is that the migration should be fairly smooth.

eriodiah:
@ERosendo, ok — I'll check in with Mike. But I'm planning to hopefully get the cadence for Contacts done by the end of this week.

mlissner:
OK, Eri's got her cadences in place, so this is set for auto-merge. Thanks all!

PR #7446, opened 2026-06-12
feat(scrapers): store per-scrape result count on PACERFreeDocumentLog

### LINKED ISSUE #7424: Store per-scrape result count on PACERFreeDocumentLog to detect silent gaps
## Background

Follow-up from #7417.

The free-opinion scraper records a `PACERFreeDocumentLog` row as `SCRAPE_SUCCESSFUL` whether the Written Opinions report returned 100 results or 0. The row stores `status` but not how many results the query produced. As a result, a "silent" gap (a scrape that succeeded but ingested nothing, e.g. due to a truncated proxy response or opinions the court flagged free only later) is indistinguishable in the log from a genuinely complete scrape.

This is the original SDNY symptom: nysd's log showed `SCRAPE_SUCCESSFUL` for every day from 2025-11-27 through 2026-05-28, while the newest nysd opinion actually in CL was 2025-11-26. The only way we could detect it was to cross-check the latest ingested opinion `dateFiled` per court against the search API, which is indirect and slow.

## Proposal

Persist the result count on each `PACERFreeDocumentLog` row. `get_and_save_free_document_report` already computes `rows_to_create`; we just throw it away after logging. Add a nullable integer field (e.g. `document_count`) and populate it on completion.

With that in place, silent gaps become detectable directly from the log:
- a `SCRAPE_SUCCESSFUL` day with `document_count = 0` surrounded by non-zero neighbors, or
- a re-scrape that returns more than a prior run for the same date.

## Notes

This is a model change, so it needs a migration following the project's migration guide.

quevon24:
Fixes: #7424

The free-opinion scraper logged `SCRAPE_SUCCESSFUL` whether the Written Opinions report returned 100 results or 0, so a "silent" gap (a scrape that succeeded but ingested nothing) was indistinguishable from a complete scrape. The counts were already computed during the scrape; we just threw them away after logging them.

Following @grossir's review, this stores three counts per scrape rather than one. They narrow from what PACER claimed to what we can actually ingest, so a shortfall can be attributed to a stage instead of merely noticed.

This PR:

- Adds three nullable count fields to `PACERFreeDocumentLog` and populates them when a scrape completes. `reported_document_count` is PACER's own "Total number of opinions reported", summed over every page queried; `parsed_document_count` is the rows we parsed out of the report; `saved_document_count` is the parsed rows we saved for ingestion. Null means unknown (the scrape failed, or the row predates these fields); 0 means the scrape succeeded with no results. Rows skipped for a missing docket number (#4547) still land in `parsed_document_count`, so they stay visible instead of vanishing.
- Makes silent gaps detectable directly from the log: a successful day with `saved_document_count = 0` surrounded by non-zero neighbors, a `reported > parsed` gap pointing at a stale parser, or a re-scrape that returns more than a prior run for the same date. A `reported > parsed` gap also emits a `logger.error` naming the court, since that one usually means the docket-number / case-name / date repair logic in `FreeOpinionRow` is out of date for that court.
- Takes the reported count from `FreeOpinionReport.reported_opinion_count`, added in freelawproject/juriscraper#2055 and released in juriscraper 3.0.34, and moves the floor in `pyproject.toml` from `>=3.0.0` to `>=3.0.34` (the lockfile already resolved to 3.0.34). That property is worth having on its own: the `cand_2` fixture is a real case where PACER reports 52 opinions and only 12 rows parse.
- Shows the three counts in the `PACERFreeDocumentLog` admin list view.
- Fixes an inconsistent return type in `get_and_save_free_document_report`: the retries-exhausted failure paths returned a bare status int despite the `tuple[int, int]` annotation, so unpacking it in `fetch_doc_report` raised a `TypeError` that was only handled by accident and mislabeled as "failing PACER website". Both paths now return `(SCRAPE_FAILED, None)` — `None` rather than zeros, because "we saw nothing" and "we never looked" are different facts — and `fetch_doc_report` handles a returned failure status explicitly. The task returns a `FreeOpinionScrapeCounts` NamedTuple instead of a bare count.

Two things worth knowing while reviewing. None of the three fields is named `document_count`: a single `document_count` sitting beside two `*_document_count` siblings read badly, and since the migration hadn't merged it was regenerated in place rather than stacked with a second one. And `saved_document_count` will change meaning if #7600 lands, since deduplication makes it a function of DB state at scrape time rather than a pure property of the scrape — the reported and parsed counts are unaffected.

The migration is three nullable `ADD COLUMN`s with no default, so it's metadata-only and instantaneous. The `CHECK (>= 0)` constraints from `PositiveIntegerField` are free here too: the columns are brand new, so every row is NULL and there is nothing to scan. The model has no event tables or triggers, so the replica `.sql` and `_customers.sql` files are identical.


cl/scrapers/models.py
```diff
@@ -59,6 +59,14 @@ class PACERFreeDocumentLog(models.Model):
     status = models.SmallIntegerField(
         help_text="The status of the scrape.", choices=SCRAPE_STATUSES
     )
+    document_count = models.IntegerField(
```
grossir: should this be a PositiveIntegerField? Since it's 0 or more

cl/corpus_importer/tasks.py
```diff
@@ -490,7 +491,7 @@ def get_and_save_free_document_report(
 
         if self.request.retries == self.max_retries:
             logger.error(f"{msg} at %s (%s to %s).", court_id, start, end)  # noqa: G004
-            return PACERFreeDocumentLog.SCRAPE_FAILED
+            return PACERFreeDocumentLog.SCRAPE_FAILED, 0
```
grossir: If this `0` will be used to populate the new `document_count` field, it would be better to leave it as None?

cl/corpus_importer/tasks.py
```diff
@@ -500,7 +501,7 @@ def get_and_save_free_document_report(
         # IndexError: When the page isn't downloaded properly.
         # HTTPError: raise_for_status in parse hit bad status.
         if self.request.retries == self.max_retries:
-            return PACERFreeDocumentLog.SCRAPE_FAILED
+            return PACERFreeDocumentLog.SCRAPE_FAILED, 0
```
grossir: same reasoning for this one, putting a 0 when we actually couldn't get the report itself seems incorrect. Even if we have the SCRAPE_FAILED status in place

cl/corpus_importer/tasks.py
grossir, line 565: is there any way to compare the "valid" results returned by the parser against the actual rows existing in the HTML? This to catch cases when the "repairing" of docket number, case name or date fails. Maybe it fails because the parser is not up to date

I think we could modify juriscraper slightly and save the total count as metadata. See the "Total number of opinions reported"

https://github.com/freelawproject/juriscraper/blob/3730f3c09553d09e49917a52bacd2ce23aa70aeb/juriscraper/pacer/free_documents.py#L151-L177


We should analyze if this should be just a `logger.error` or upgraded to a model field (like "number_of_opinions_reported") so we can save it on the DB 


Repairs:
https://github.com/freelawproject/juriscraper/blob/3730f3c09553d09e49917a52bacd2ce23aa70aeb/juriscraper/pacer/free_documents.py#L337-L339

https://github.com/freelawproject/juriscraper/blob/3730f3c09553d09e49917a52bacd2ce23aa70aeb/juriscraper/pacer/free_documents.py#L319-L326

https://github.com/freelawproject/juriscraper/blob/3730f3c09553d09e49917a52bacd2ce23aa70aeb/juriscraper/pacer/free_documents.py#L358-L359

grossir, line 573: These skips go together with what I mentioned above

We have 3 counts for each scrape
- number of opinions reported according to PACER (see the comment above)
- number of opinions we actually parsed 
- number of documents we will actually ingest (right now, we skipped missing docket numbers)

Maybe we should keep track of the 3 counts? Counts are cheap on storage, this may help us debug any silent failures

grossir: even if a document has no docket number, and is skipped, we should still count it. It may mean we need to improve our parsers.

Also, from https://github.com/freelawproject/courtlistener/pull/7600/ ; the actual count of the "PACERFreeDocumentLog" will now be even smaller when excluding duplicates. 

But the duplicate cleaning is not an attribute itself of this object that represents a scrape, but a computation from the state of the DB at that moment. 

In other words, we should also have a count field that reflects the actual value related to the scrape, which is a property of the scrape. We could also have a count that reflects the number of expected documents to be ingested from this scrape, which is `len(document_rows_to_create)`.

grossir (commented):
I think we could introduce 1 or 2 more count fields to have a better sense of the data returned by the scrape. This should help us debug further. What do you think?

cl/corpus_importer/tasks.py
quevon24: Good call. I've split the PACER-reported count off into Juriscraper, since that's where it's parsed: issue freelawproject/juriscraper#2054 and PR freelawproject/juriscraper#2055 add a `reported_opinion_count` property exposing PACER's "Total number of opinions reported" (summed across pages). It's a real problem in practice — the `cand_2` fixture reports 52 opinions but only 12 rows parse.

Once that's merged and released, I'll extend this PR to store the reported count on `PACERFreeDocumentLog` and compare it against the parsed and to-ingest counts, so silent gaps become queryable directly from the log. Leaving that DB field out of this PR until the Juriscraper release lands.

PR #7447, opened 2026-06-13
refactor (components): Unify <c-filter-drawer> with <c-dialog>

### LINKED ISSUE #7318: refactor(components): Unify <c-filter-drawer> with <c-dialog>
### Context

[#7227](https://github.com/freelawproject/courtlistener/pull/7227) introduced `<c-filter-drawer>` for the docket entries page. It uses the same Alpine UI dialog primitives (`x-data="dialog"`, `x-dialog`, `:overlay`, `:panel`, `:title`) as `<c-dialog>` but without reusing the component itself, so both components end up duplicating the same plumbing.

### Goal

Make `<c-filter-drawer>` a thin wrapper around `<c-dialog>` — the drawer owns its layout (header / body / footer) and visual identity (funnel trigger, slide-in animation), but all dialog mechanics (open/close, escape, overlay, breakpoint-close, ARIA wiring) live in `<c-dialog>`.

### Required work

- Extend `<c-dialog>` with transition support so consumers can opt into slide / fade animations.
- Refactor `<c-filter-drawer>` to call `<c-dialog>` and project content via slots.

### Notes

- Consumer-facing API of `<c-filter-drawer>` (component name, slots, props) stays nearly identical in the experiment, so the docket entries page implementation in [#7068](https://github.com/freelawproject/courtlistener/issues/7068) is not affected.

mexomagno:
## Fixes
Closes https://github.com/freelawproject/courtlistener/issues/7318 

## Summary
This PR reimplements `filter_drawer` component as a `dialog` wrapper instead of an independent component depending on dialog composable to implement behavior that the `dialog` component already has. It makes the effort of preserving the existing `filter_drawer` contract.

* Adds `variant` prop to `dialog` to provide the "drawer" variant
* Adds `disable_transitions` prop to `dialog` to opt-out of transitions
* Refactors the `filter_drawer` component to use c-dialog
* Documents these changes on the storybook (`v2_components.html`)
* Adds unit tests to validate the new `dialog` props

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

To deploy just update the frontend components. Nothing else is affected.


cl/assets/templates/cotton/filter_drawer.html
```diff
@@ -1,94 +1,51 @@
...
-  x-data="dialog"
-  x-id="dialogIdGroup"
-  x-on:open-filter-drawer="open"
+<c-dialog
+  variant="drawer"
   class="md:hidden {{ class }}"
-  {% if has_errors %}data-has-errors{% endif %}
-  {{ attrs }}
+  x-on:open-filter-drawer="open"
+  :attrs="{% if has_errors %}{'data-has-errors': True}{% else %}{}{% endif %}"
```
elisa-a-v, line 12: I was actually surprised to see this worked, because `attrs` isn't a free prop name: [Cotton fills it with whatever undeclared attributes the caller passed](https://django-cotton.com/docs/components#attrs), and `<c-dialog>` renders it onto its root div. That's how `x-on:open-filter-drawer` on the line above gets there, so setting `:attrs` explicitly only works because Cotton merges them, which is non-obvious and isn't behavior I'd want to depend on.

I'm assuming the corner you were working around is that you can't put a conditional attribute on a `<c-dialog>` tag the way you could on the plain `<div>` this used to be, but I think there's a cleaner fix. Pass the prop like this:

```suggestion
  data-has-errors="{{ has_errors|yesno:'true,false' }}"
```

Then in `docket_filter.js` we can do:

```diff
// If the filter form was submitted with errors, pop the mobile  
// drawer open so the user can see the validation messages inside.
- const drawer = this.$el.querySelector("[data-has-errors]");
+ const drawer = this.$el.querySelector('[data-has-errors="true"]');
```

That way we keep the filter-specific prop out of the dialog component, and we don't touch Cotton's `attrs`, thus avoiding potentially weird edge cases.

cl/simple_pages/templates/v2_components.html
```diff
@@ -1046,6 +1081,14 @@ <h4 class="mt-3">Demo</h4>
         <c-slot name="label"><code>panel_class</code></c-slot>
         <c-slot name="description">CSS classes added to the panel itself.</c-slot>
       </c-library.item>
+      <c-library.item optional only>
+        <c-slot name="label"><code>variant</code></c-slot>
+        <c-slot name="description">Visual style. Options: <code>drawer</code>. When omitted, the default centered dialog is displayed.</c-slot>
+      </c-library.item>
+      <c-library.item optional only>
+        <c-slot name="label"><code>disable_transitions</code></c-slot>
+        <c-slot name="description">Opt-out of dialog transitions on open and close. Only affects the <code>drawer</code> variant</c-slot>
```
elisa-a-v, line 1090: "Only affects the `drawer` variant" isn't right. The overlay's `{% if not disable_transitions %}` guard sits outside the variant check, so the flag strips the overlay fade on the default centered dialog too, and in fact `test_disable_transitions_removes_overlay_transitions` renders the default variant and asserts exactly that.

Maybe we could do:

```suggestion
        <c-slot name="description">Opt out of dialog transitions on open and close. Disables the overlay fade on every variant, and the panel slide on the <code>drawer</code> variant.</c-slot>
```

cl/assets/templates/cotton/dialog/index.html
```diff
@@ -31,17 +43,38 @@
...
+      x-transition:enter="motion-safe:transition motion-safe:ease-out motion-safe:duration-200 motion-safe:transform"
+      x-transition:enter-start="motion-safe:translate-x-full"
+      x-transition:enter-end="motion-safe:translate-x-0"
+      x-transition:leave="motion-safe:transition motion-safe:ease-in motion-safe:duration-150 motion-safe:transform"
+      x-transition:leave-start="motion-safe:translate-x-0"
+      x-transition:leave-end="motion-safe:translate-x-full"
+        {% endif %}
+      {% else %}
+      class="w-full flex flex-col bg-white shadow-md z-50 {{ panel_class }}"
+      {% endif %}
```
elisa-a-v: Is it intentional that a centered dialog gets no panel transition at all? The `{% else %}` branch sets a class and nothing else, so `disable_transitions` only has the overlay to strip there.

cl/assets/templates/cotton/dialog/index.html
```diff
@@ -31,17 +43,38 @@
...
     >
-      {{ panel }}
-    </div>
+    {{ panel }}
+  </div>
```
elisa-a-v, line 78: Indentation slipped here:

```suggestion
      {{ panel }}
    </div>
```

cl/simple_pages/templates/v2_components.html
```diff
@@ -1046,6 +1081,14 @@ <h4 class="mt-3">Demo</h4>
...
+      </c-library.item>
     </c-library.list>
 
     <c-library.list title="Slots" only>
```
elisa-a-v, line 1094: This one is optional and not something you introduced: the `trigger_button`  slot has been undocumented since it was added, how about we add something like this:

```suggestion
    <c-library.list title="Slots" only>
      <c-library.item optional only>
        <c-slot name="label"><code>trigger_button</code></c-slot>
        <c-slot name="description">Replaces the default trigger button entirely. Use when the trigger needs its own styling or content, as <code>c-filter-drawer</code> does.</c-slot>
      </c-library.item>
```

cl/simple_pages/templates/v2_components.html
```diff
@@ -1410,7 +1486,7 @@ <h4 class="mt-3">Demo</h4>
   {# FILTER DRAWER #}
   <section class="max-w-full w-full border-t-2 border-greyscale-200" x-intersect.margin.-100px="show" id="filter-drawer">
     <h2 class="mt-6 mb-3">Filter drawer</h2>
-    <p>A mobile slide-in panel for filter controls. Only visible below the <code>md</code> breakpoint. Reuses the dialog composable for open/close, escape key, and auto-close on breakpoint change. Auto-opens when <code>data-has-errors</code> is present.</p>
+    <p>A mobile slide-in panel for filter controls. Only visible below the <code>md</code> breakpoint. Reuses the <code>c-dialog</code> component which provides open/close, escape key, and auto-close on breakpoint change. Auto-opens when <code>data-has-errors</code> is present.</p>
```
elisa-a-v, line 1489: All of this predates your PR, but since you already tweaked it, I think we can make it even better.

This paragraph describes `data-has-errors`, which is the attribute the component sets internally rather than anything a caller passes. From the outside the API is the `has_errors` prop, so this is probably more useful:

```suggestion
    <p>A mobile slide-in panel for filter controls. Only visible below the <code>md</code> breakpoint. Reuses the <code>c-dialog</code> component which provides open/close, escape key, and auto-close on breakpoint change. Auto-opens when <code>has_errors</code> is truthy.</p>
```

Separately and optionally: `has_errors` isn't in this section's Props list either. Same as the `trigger_button` note: it's a nice-to-have more than a blocker.

cl/assets/templates/cotton/dialog/index.html
elisa-a-v, line 1: Not yours originally: these section markers (`<!-- Trigger -->`, and the same for Dialog, Overlay, and Panel below) are HTML comments, so they ship to users in page source. Django comments `{# ... #}` don't. Would you mind fixing them while you're on this file? 🙏🏽

elisa-a-v (changes requested):
Nice, thank you! Refactor works and there are three things I liked that the issue didn't ask for:
- `variant="drawer"` matches how `<c-button>` already handles variants and makes it easier for consumers to use the component without needing to know the exact classes required to make it work
- the `motion-safe:` prefixes give the transitions reduced-motion support the old drawer never had
- the `DialogComponentTest` coverage, good call adding tests for this

I do have some comments that should be addressed before we merge this, left them all inline.

Side note: you can use [GitHub's keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to automatically link the issue to the PR so the issue will be closed once the PR gets merged. In this case, you can do "Closes #7318" instead of "Implements #7318" and that'll do it 👌🏽

PR #7450, opened 2026-06-15
7353 term locations search results 20260615

### LINKED ISSUE #7353: Return term locations with search API
The MCP is using keyword search, then pulling entire documents into its context. Like a human, it would be better if it could map the search results to the correct location within the results and then pull just that content from the DB (or into the context window). 

It seems like Elastic has a few options for this:
 - Include the locations in the search results?
 - Provide another API that takes a document and a query and returns locations
 - Something else? 

What we *don't* want to do is reverse engineer the Elastic query engine. That way lies madness.

Can we begin in an upcoming sprint by analyzing this and figuring out the design and then in a subsequent one we can implement it?

Cc @nadahlberg and @albertisfu.

nikolly:
## Fixes
Fixes: #7353

[Technical Challenge_ Term Locations in Search Results.pdf](https://github.com/user-attachments/files/28967129/Technical.Challenge_.Term.Locations.in.Search.Results.pdf)

## Summary
This PR introduces a new custom @action detail endpoint on the RECAP document viewset (/api/rest/v4/recap-documents/{id}/matches/) to allow LLM agents to inspect specific context regions within a long document.

What it fixes:
Previously, agents discovering a relevant document through global search had to pull the entire plain_text payload into their context window just to find the occurrences of their query. This was highly inefficient, consuming amounts of context memory and causing massive network payloads on long court filings.

The Approach:
To solve this problem, this PR implements an optimized text retrieval mechanism using Elasticsearch's Unified Highlighter:
- It isolates the text search strictly to the document ID (id=pk).
- It disables the full document body transport by enforcing .source(False).
- It configures the highlighter with fragment_size=1500 and order="score" to return high-quality, semantically intact paragraph chunks ordered by internal relevance.

By pushing this work to Elasticsearch, we keep the API extremely lightweight.

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

No special steps, database migrations, or Elasticsearch index rebuilds are required. The current mappings fully support the Unified Highlighter.

### DIFF
--- a/cl/search/api_views.py
+++ b/cl/search/api_views.py
@@ -3,7 +3,14 @@ from http import HTTPStatus
 from django.db.models import Prefetch
 from django.http.response import Http404
 from django.urls import reverse
+from elasticsearch.exceptions import (
+    ApiError,
+    ConnectionError,
+    RequestError,
+    TransportError,
+)
 from rest_framework import pagination, permissions, response, viewsets
+from rest_framework.decorators import action
 from rest_framework.exceptions import NotFound
 from rest_framework.pagination import PageNumberPagination
 from rest_framework.permissions import (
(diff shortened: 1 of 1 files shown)


albertisfu:
Thanks @nikolly I'm closing this one since it's only a prototype, and as we discussed during its review, it would require a different approach to extract and expose term locations.

PR #7493, opened 2026-06-24
Florida scraper command

### LINKED ISSUE #7057: Scrape historial Florida court cases
### Describe the issue.

After https://github.com/freelawproject/courtlistener/issues/6831 is complete, the work in this issue is to build a Django command to scrape historical Florida court cases.

The scraper should store the cases in S3, and in a follow-up issue we will import that content.

### Related Issue

_No response_

### Screenshots

_No response_

To Do

MorganBennetDev:
## Fixes

Closes #7057 

## Summary

Add a `back_scrape_fl_dockets` command to scrape Florida dockets in a given date range, optionally archiving responses and parsed data to S3.

Additionally moves some shared functionality between the existing `back_scrape_dockets` command and the new command to `management/utils.py`

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`


MorganBennetDev:
Small note: This will need to be updated slightly to handle the new retry configuration when https://github.com/freelawproject/juriscraper/pull/2014/ is merged. Also, Juriscraper will need to be updated before this is merged; I set uv to point at the git repository locally for testing.

MorganBennetDev:
PR has been updated to handle new retry configuration. Should be 100% ready to go.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -0,0 +1,307 @@
...
+    ScraperCheckpointTracker,
+    StateBackScrapeCommand,
+    _make_case_number_key,
+    _parse_date,
+)
+
+S3_BASE = "responses/dockets/florida"
+
+
+def make_s3_key(request: ScheduledRequest) -> str:
```
albertisfu: I ran a test scrape and found it a bit difficult to use the responses stored in S3 for debugging purposes and to understand what's going on.

What do you think about adding the endpoint path as a prefix to the hash?

For example:

`/courts/cms/cases/`

This would create directories, allowing us to keep all the `case` responses in one place, while storing `court` responses under `/courts/`, and so on.

I think this would make debugging much easier.

You can see an example of this in:

`dev-com-courtlistener-private-storage/responses/dockets/florida/`

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -0,0 +1,307 @@
...
+
+async def _backfill(
+    date_ranges: dict[FloridaCourtID, tuple[date, date]],
+    full_scrape: bool,
+    checkpoint_trackers: dict[FloridaCourtID, ScraperCheckpointTracker],
+    scraper: FloridaScraper,
+    save_queue: queue.Queue[tuple[str, bytes]],
+):
+    logger.info("Starting Florida backfill...")
+    archive_loop = asyncio.to_thread(_archive_loop, save_queue)
```
albertisfu: It seems that this only creates the `archive_loop`, but it never schedules it, so the task never starts.

I found this issue while running a test scrape. It got stuck when trying to store the files in S3.

Claude suggested the following refactor, and it worked:

```
archive_task = asyncio.create_task(
        asyncio.to_thread(_archive_loop, save_queue)
    )

    try:
        for court_id, (start_date, end_date) in date_ranges.items():
            i = 0
            async for case in scraper.backfill(
                start_date,
                end_date,
                court_ids=[court_id],
                full_scrape=full_scrape,
            ):
                key = f"{S3_BASE}/parsed/{_make_case_number_key(case.docket_number)}.json"
                content = case.model_dump_json(ensure_ascii=True).encode(
                    "utf-8"
                )
                save_queue.put((key, content))
                i += 1
                if i % 5 == 0:
                    logger.info(
                        "Updating checkpoint for %s to %s",
                        court_id,
                        case.date_filed,
                    )
                    checkpoint_trackers[court_id].set(case.date_filed)

        logger.info(
            "Florida backfill complete. Waiting for archiving to complete..."
        )
        save_queue.join()
    finally:
        # shutdown() unblocks _archive_loop's blocking get() so the thread
        # exits cleanly, even if the scrape raised mid-way.
        save_queue.shutdown()
        await archive_task
```

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -0,0 +1,307 @@
...
+        if not self.save:
+            return
+
+        try:
+            response = await request.response
+        except Exception:
+            return
+
+        loop = asyncio.get_running_loop()
+        _ = loop.run_in_executor(None, self.save_to_s3, request, response)
```
albertisfu: Got this error while finalizing the scraper:

```
026-0962.json to dev-com-courtlistener-private-storage"
INFO 2026-06-25 15:17:26,883 (/opt/courtlistener/cl/scrapers/management/commands/back_scrape_fl_dockets.py _backfill): "Florida backfill and archive complete!"
INFO 2026-06-25 15:17:26,943 (/opt/courtlistener/cl/scrapers/management/commands/back_scrape_fl_dockets.py save_to_s3): "Adding responses/dockets/florida/v1/courts/cms/docketentrydocumentsaccess/882a98efa06733ad3159d5e99da30f18001d6d39 to archive queue"
Future exception was never retrieved
future: <Future finished exception=ShutDown()>
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/opt/venv/lib/python3.13/site-packages/sentry_sdk/integrations/threading.py", line 176, in wrapped_fn
    return fn(*args, **kwargs)
  File "/opt/courtlistener/cl/scrapers/management/commands/back_scrape_fl_dockets.py", line 100, in save_to_s3
    self.save_queue.put((key, response.content))
    ~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/queue.py", line 152, in put
    raise ShutDown
queue.ShutDown
Sentry is attempting to send 2 pending events
Waiting up to 2 seconds
Press Ctrl-C to quit
```

It looks like this should be awaited to prevent the request from being considered complete before it actually finishes.

Also, should we catch exceptions here so that if storing the file in S3 fails, the scraper can continue running?

I don't remember exactly, but I think boto3 retries requests by default.
Maybe that's enough? If we start seeing this error frequently, we could consider adding custom retry logic for S3Cache requests.

```suggestion
        try:
            await loop.run_in_executor(None, self.save_to_s3, request, response)
        except Exception:
            logger.exception("Failed to archive response for %s", request.url)
```

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -0,0 +1,307 @@
...
+        if auto_resume:
+            logger.info("Auto resume enabled. Getting checkpoints...")
+            checkpoints = {
+                cid: self.checkpoint_trackers[cid].get() for cid in court_ids
+            }
+        else:
+            checkpoints = {}
+
+        date_ranges = {
+            cid: (checkpoints.get(cid) or start, end) for cid in court_ids
```
albertisfu, line 322: It looks like the auto-resume process here can terminate the scrape too early and miss cases.

In `FloridaScraper.enumerate_cases`, it uses a LIFO stack to keep track of date ranges after splitting them when a range contains more than 10,000 cases, correct?

A simple example:

start: 2010 -> end: 2026 (15,000 cases)

stack = {(2010,2026)}

stack.pop()
Fetch: start: 2010 -> end: 2026 (15,000 cases)

It splits the range in half:

stack.append(2010,2018)  # 9,000 cases
stack.append(2018,2026)  # 6,000 cases

stack = {(2010,2018), (2018,2026)}

stack.pop()
Fetch: start: 2018 -> end: 2026 (6,000 cases)

The scraper gets interrupted while processing a case from 2019.

The checkpoint is now 2019.

When it resumes:

start: 2019 -> end: 2026

At that point, it will skip everything older than 2019, including the unprocessed 2010 -> 2018 range that is still on the stack.

Should we fix this in Juriscraper? One option would be to process the oldest date ranges first?

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -0,0 +1,307 @@
...
+):
+    logger.info("Starting Florida backfill...")
+    archive_loop = asyncio.to_thread(_archive_loop, save_queue)
+
+    for court_id, (start_date, end_date) in date_ranges.items():
+        i = 0
+        async for case in scraper.backfill(
+            start_date, end_date, court_ids=[court_id], full_scrape=full_scrape
+        ):
+            key = f"{S3_BASE}/parsed/{_make_case_number_key(case.docket_number)}.json"
```
albertisfu: Are case numbers unique across all courts? I'm concerned they may be repeated across different courts, which could cause cases to be overwritten.

Maybe including the `court_id` in the key would help?

albertisfu (commented):
Thanks @MorganBennetDev, this looks good!

I just found a couple of issues that should be fixed, along with a few suggestions.

I think one of them might be worth handling in Juriscraper.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
MorganBennetDev: Yes, we should fix this in Juriscraper. Using LIFO was a mistake on my part; I meant to keep things in chronological order. Will add to the Juriscraper PR.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
MorganBennetDev: Let's do that. Better safe than sorry. In the test scrape, all courts had a unique prefix on the docket numbers, but we're not guaranteed that will be true in the past.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
MorganBennetDev: I misread the documentation of `to_thread` and assumed that it also started the thread. Fixing.

MorganBennetDev:
@albertisfu Addressed your changes and updated the Juriscraper PR to keep case output in the correct order.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -115,14 +131,21 @@ def _archive_loop(responses: queue.Queue[tuple[str, bytes]]) -> None:
         try:
             key, content = responses.get()
         except queue.ShutDown:
+            logger.exception("Archive queue forcibly shut down.")
```
albertisfu: In a regular scrape that appears to finish successfully, this is being logged as an error.

Is this expected, or could it be logged less loudly?

 ```
 INFO 2026-06-26 13:51:55,644 (/opt/courtlistener/cl/scrapers/management/commands/back_scrape_fl_dockets.py _archive_loop): "Archived  
  responses/dockets/florida/parsed/flsc/SC2026-0971.json to dev-com-courtlistener-private-storage"                                      
  INFO 2026-06-26 13:51:55,644 (/opt/courtlistener/cl/scrapers/management/commands/back_scrape_fl_dockets.py _backfill): "Florida       
  backfill completed successfully!"                                                                                                     
  INFO 2026-06-26 13:51:55,644 (/opt/courtlistener/cl/scrapers/management/commands/back_scrape_fl_dockets.py _backfill): "Waiting for   
  Florida archive loop to finish..."                                                                                                    
  ERROR 2026-06-26 13:51:55,645 (/opt/courtlistener/cl/scrapers/management/commands/back_scrape_fl_dockets.py _archive_loop): "Archive  
  queue forcibly shut down."                                                                                                            
    File "/opt/courtlistener/cl/scrapers/management/commands/back_scrape_fl_dockets.py", line 132, in _archive_loop                     
      key, content = responses.get()                                                                                                    
                     ~~~~~~~~~~~~~^^                                                                                                    
    File "/usr/local/lib/python3.13/queue.py", line 201, in get                                                                         
      raise ShutDown                                                                                                                    
  queue.ShutDown                                                                                                                        
  INFO 2026-06-26 13:51:55,675 (/opt/courtlistener/cl/scrapers/management/commands/back_scrape_fl_dockets.py _backfill): "Florida       
  archive loop complete." 
```

File: cl/scrapers/management/commands/back_scrape_fl_dockets.py [hunk 7]
```diff
@@ -0,0 +1,344 @@
...
+    :ivar save: Whether to archive responses
+    :ivar load: Whether to intercept requests and load responses from S3 if available"""
+
+    base: Path
+    save: bool
+    load: bool
+    save_queue: queue.Queue[tuple[str, bytes]]
+
+    def make_s3_key(self, request: ScheduledRequest) -> str:
+        url_path = Path(request.url.path.lower())
```
albertisfu: This is creating an absolute path for `/courts/...` and throwing away `responses/dockets/florida`.

As a result, all raw responses end up being stored at the bucket root under `/courts/` instead of `responses/dockets/florida/courts/`.

It looks like the solution is just:


```suggestion
        url_path = request.url.path.lower().strip("/")
```

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -0,0 +1,344 @@
...
+            [f"{k}={v}" for k, v in sorted(request.url.params.items())]
+        )
+        method = request.method.upper()
+        request_hash = hashlib.sha1(
+            f"{method}|{url_path}|{url_params}".encode()
+        ).hexdigest()
+        return str(self.base / url_path / request_hash)
+
+    def load_from_s3(self, request: ScheduledRequest) -> Response | None:
+        logger.info("Attempting to get %s from S3", request.url.path.lower())
```
albertisfu: Could we omit this log to reduce the noise?

It looks like we'll always log a follow-up message indicating whether the file was found in S3 or not, so this one seems a bit redundant.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -0,0 +1,344 @@
...
+    def save_to_s3(
+        self, request: ScheduledRequest, response: Response
+    ) -> None:
+        key = self.make_s3_key(request)
+        try:
+            storage = S3GlacierInstantRetrievalStorage()
+            # Don't try to cache responses that we pulled from the cache
+            if self.load and storage.exists(key):
+                return
+            logger.info("Queueing archive of response for %s", key)
```
albertisfu: Maybe this log can also be omitted?

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -0,0 +1,344 @@
...
+    )
+
+    while True:
+        try:
+            key, content = responses.get()
+        except queue.ShutDown:
+            logger.exception("Archive queue forcibly shut down.")
+            break
+
+        logger.info("Archiving %s to %s", key, storage.bucket_name)
```
albertisfu: And this one too?

This way, we can have a single log entry for the same item being stored, which is the one below:

`logger.info("Archived %s to %s", key, storage.bucket_name)`

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -0,0 +1,344 @@
...
+            storage = S3GlacierInstantRetrievalStorage()
+            if not storage.exists(key):
+                logger.info("Did not find %s in S3", request.url.path.lower())
+                return None
+
+            logger.info("Loading %s from S3", request.url.path.lower())
+
+            with storage.open(key, "rb") as f:
+                return Response(200, content=f.read())
+        except:
```
albertisfu: To be safe, so that no `BaseException` subclasses are caught here (such as `KeyboardInterrupt`, `SystemExit`, etc.).

```suggestion
        except Exception:
```

We should apply this change to the other places where a bare `except:` is used as well.

albertisfu (changes requested):
Thanks @MorganBennetDev a few more changes before we can merge it please.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
MorganBennetDev: That is not correct. Fixing

MorganBennetDev:
@albertisfu Updated based on your feedback. Did not know there was a distinction between `except Exception` and `except`. Learned something new today.

albertisfu (approved):
Thanks @MorganBennetDev this looks great and ready to go 🚀

PR #7495, opened 2026-06-24
Fix(recap): Add composite GFK index on PacerHtmlFiles to fix delete timeouts

### LINKED ISSUE #7494: Docket / DocketEntry queries/deletes time out due to missing index on recap_pacerhtmlfiles
### Describe the issue.

I needed to perform a `DocketEntry` cleanup related to a couple of support tickets for cases affected by issue https://github.com/freelawproject/courtlistener/issues/5247.

However, all of my attempts timed out.

After investigating the issue, I found that the problem is related to the `PacerHtmlFiles` table.

When removing a `DocketEntry` using either the Django admin or the ORM, all related instances configured with `on_delete=CASCADE` are also collected and deleted.

I analyzed the queries executed while collecting those related instances:

```text
[    1]      29.4ms  SELECT "search_docketentry_tags"."id", "search_docketentry_tags"."docketentry_id", "search_docketentry_tags"."tag_id", "search_docketentry"."id", "sea
[    2]      33.4ms  SELECT "search_recapdocument"."id", "search_recapdocument"."date_created", "search_recapdocument"."date_modified", "search_recapdocument"."sha1", "sea
[    3]       2.1ms  SELECT "favorites_note"."id", "favorites_note"."date_created", "favorites_note"."date_modified", "favorites_note"."user_id", "favorites_note"."cluster
[    4]       2.0ms  SELECT "favorites_prayer"."id", "favorites_prayer"."date_created", "favorites_prayer"."user_id", "favorites_prayer"."recap_document_id", "favorites_pr
[    5]      12.0ms  SELECT "favorites_prayeravailability"."id", "favorites_prayeravailability"."recap_document_id", "favorites_prayeravailability"."last_checked", "search
[    6]      22.2ms  SELECT "recap_emailprocessingqueue_recap_documents"."id", "recap_emailprocessingqueue_recap_documents"."emailprocessingqueue_id", "recap_emailprocessi
[    7]     117.3ms  SELECT "search_recapdocument_tags"."id", "search_recapdocument_tags"."recapdocument_id", "search_recapdocument_tags"."tag_id", "search_recapdocument".
[    8]       3.2ms  SELECT "search_opinionscitedbyrecapdocument"."id", "search_opinionscitedbyrecapdocument"."citing_document_id", "search_opinionscitedbyrecapdocument"."
[    9]       3.1ms  SELECT "django_content_type"."id", "django_content_type"."app_label", "django_content_type"."model" FROM "django_content_type" WHERE ("django_content_
[   10]   10012.0ms  SELECT "recap_pacerhtmlfiles"."id", "recap_pacerhtmlfiles"."date_created", "recap_pacerhtmlfiles"."date_modified", "recap_pacerhtmlfiles"."content_type  <<<< SLOW

FAILED after 10.2s — 10 queries: OperationalError('canceling statement due to statement timeout')
```

The query against `recap_pacerhtmlfiles` is the one that timed out.

The underlying issue is that `PacerHtmlFiles` uses a `GenericForeignKey`. While `content_type` is indexed by default, `object_id` is not.

As a result, when Django attempts to retrieve the related `PacerHtmlFiles` rows using the `(content_type_id, object_id)` pair, PostgreSQL cannot leverage an index. Instead, it must scan millions of rows matching the selected `content_type`, causing the query to timeout.

To resolve this, we can add a composite index on `(content_type_id, object_id)`.

### Related Issue

_No response_

### Screenshots

_No response_

To Do

albertisfu:
## Fixes
Fixes: #7494

## Summary
Querying or Deleting  a `Docket` or `DocketEntry` could hang or time out in production.

The delete cascade follows the `html_documents` `GenericRelation` to
`PacerHtmlFiles` and queries `recap_pacerhtmlfiles` by its GenericForeignKey
pair `(content_type_id, object_id)`. The table only had an index
on `content_type_id` alone, so Postgres scanned millions of rows for the
dominant content types and the statement was cancelled by `statement_timeout`.

The cascade query `WHERE content_type_id = X AND object_id IN (...)` used the
`content_type_id` index, returned millions of candidate rows, then filtered
`object_id`.

**Fix:** add the composite `(content_type, object_id)` index Django recommends
for [GenericForeignKeys](https://docs.djangoproject.com/en/6.0/ref/contrib/contenttypes/#generic-relations). It matches the exact access pattern (both columns are always supplied).
 The migration uses `AddIndexConcurrently` + `atomic = False` so the build does not lock the large
table.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`


ERosendo (approved):
LGTM 👍

Issue #6213, opened 2025-08-25
[Parent] Incorporate LLMs into Document Ingestion & Processing


flooie:
### Describe the issue.

We should begin experimenting with—and gradually incorporating—large language models (LLMs) into parts of our document ingestion and processing workflows.

This issue is intended to serve as a parent issue for several smaller, focused tasks, so we can safely explore and validate this process.

--------

Our current scraped and OCR’d plain_text is often ugly and hard to read.
Preliminary testing suggests that LLMs can:

1.	Improve readability of OCR’d/scraped PDFs (replace Doctor)
	•	Reflow bad/ugly text, remove page lines and bates stamps.
	•	Handle footnotes, spacing, and flow better.
2.	Provide more reliable case name extraction and classification (RECAP → Opinions)
	•	Address mismatches in criminal RECAP dockets where captions differ from docket case names.
3.	Parsing of opinions
	•	Extract separate opinions from combined opinions
4.	Improve ingestion pipeline (Juriscraper/Doctor)
	•	Identify file types.
	•	Identify failed scrapes (redirects, .bin/.js files, 404 pages).
5.	Opinion tagging and data extraction
	•	Settle on a schema for the scanning project.

### What type of issue is this?

To Do

mlissner:
Talked to a client just now that was hesitating about using our replica service because of the quality of the text: https://github.com/freelawproject/crm/issues/734

flooie:
Not for long

rachlllg:
One concern that was raised to me, and I think is very valid, is that, because LLM could hallucinate and inject texts into an opinion (or any other legal documents), we should make it abundantly clear to the end user that the texts are being extracted using an LLM (this is not a concern with using OCR or traditional ML techniques because these techniques do not have a risk of including language that may otherwise not be there).

Also, the fact that the texts are being extracted using an LLM could undermine the integrity of the legal document, again, not a concern from using OCR or traditional ML techniques. In another word, if I were doing legal research and using our CourtListener data, the level of trust I place in the text dramatically decreases if I know the data was extracted using an LLM.

I again must voice the possibility of using non-LLM techniques to perform this task, I think many encoder-based OCR models have the capabilities without these risks.

mlissner:
Thanks Rachel. I think we should check this really carefully. Maybe there's a way to use OCR+LLM to get the best of both worlds or to just use the up-and-coming OCR engines that are showing up everywhere. 

[...]

PR #7497, opened 2026-06-25
feat(ai): add Anthropic batch provider

artell0:
## Fixes

Part of #6213 (Incorporate LLMs into Document Ingestion & Processing).

## Summary
Adds `AnthropicBatchWrapper` (`cl/ai/llm_providers/anthropic.py`), an
Anthropic **Message Batches** implementation that mirrors the public surface
of `GoogleGenAIBatchWrapper` so the batch management commands can target
either provider. `LLMProvider.ANTHROPIC` was already declared and
`LLMRequest.api_model_name`'s help text already references a Claude model —
this fills in the provider itself.

Following the pattern noted in `check_gemini_batch_status.py`
("If you add other providers (OpenAI, Anthropic), create separate
commands"), I kept this as an independent wrapper alongside `google.py`.

Two differences from the Gemini path simplify the implementation:
- `messages.batches.create` accepts the request list directly — there is no
  intermediate JSONL file to upload.
- Prompt caching is implicit prefix caching applied via a `cache_control`
  breakpoint on the system block — there is no cache object to create or
  look up, so `get_or_create_cache` has no analog here.

**Choices I made (happy to change if you'd prefer otherwise):**
- **File inputs:** inline base64 `document` blocks (matches the Gemini
  wrapper's `input_file_path` / `input_text` task shape). Inline documents
  are capped at 32 MB / 600 pages per request; if your scans routinely
  exceed that, I'll switch to the Files API instead.
- **No commands yet:** this PR is the provider + tests only, so the shape
  can be reviewed before I add `send_anthropic_file_batches` /
  `check_anthropic_batch_status`. The S3/task-staging helpers in the Gemini
  command are provider-agnostic — I can factor them into a shared module
  when wiring the Anthropic command, if that's the direction you want.
- **`max_tokens`** defaults to 8192 per task (overridable on
  `execute_batch`); no default model is hard-coded in the wrapper.

**Tests:** `SimpleTestCase` with the Anthropic client mocked (no live calls,
no DB) — model validation, request preparation (inline base64 document +
text, skip-missing-key), system `cache_control` attachment, status
retrieval, unfinished-batch guard, and result parsing (succeeded + errored).
Mirrors the existing `ValidateGeminiModelTest` / `GoogleGenAIBatchWrapperTest`
style.

The Anthropic API key is referenced by env var only and routed through the
same secret mechanism as the Gemini key; nothing is committed.

## Deployment

**This PR should:**
    
    
    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [ ] `skip-cronjob-deploy`
    - [ ] `skip-daemon-deploy`

### DIFF
--- a/cl/ai/llm_providers/anthropic.py
+++ b/cl/ai/llm_providers/anthropic.py
@@ -1,0 +1,353 @@
+import base64
+import json
+import logging
+import mimetypes
+from datetime import date
+from typing import Any, TypedDict
+
+import anthropic
+from anthropic.types.message_create_params import (
+    MessageCreateParamsNonStreaming,
+)
+from anthropic.types.messages import MessageBatch
+from anthropic.types.messages.batch_create_params import Request
+
+logger = logging.getLogger(__name__)
+
+# Default ceiling on tokens generated per task. Extraction outputs can be
+# long; callers may override via ``execute_batch``.
+DEFAULT_MAX_TOKENS = 8192
+
+# Processing status returned by the Message Batches API once a batch has
+# finished (whether or not every request succeeded). Individual request
+# outcomes are inspected separately in ``process_results``.
+BATCH_ENDED_STATUS = "ended"
+
+# Supported Anthropic models for batch operations.
+# Maps model name to its retirement date (None if no date announced).
+# See https://docs.claude.com/en/docs/about-claude/model-deprecations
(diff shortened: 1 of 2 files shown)


mlissner:
Hi, I'm not sure this is something we need? Is there more (human) context for this?

PR #7501, opened 2026-06-25
Florida party merger

### LINKED ISSUE #7364: Florida case party merging
Create a merger for parties in Florida dockets.

MorganBennetDev:
## Fixes

Closes #7364 

## Summary

Creates merger for Florida parties and attorneys by porting existing logic from `add_parties_and_attorneys` to new merger architecture.

Moves logic common to merging all inheritors of `juriscraper.state.docket` members to a shared `cl.corpus_importer.state.common` module.

Updates Florida case party and representative factories for testing.

Adds tests for Florida party merging functionality.

Adds `RelatedParams` and `ThroughParameters` wrapper types to base merger module to allow passing parent information (for `Role` model merging) and `through` table field names and values (for all relations with a `through` table) respectively.

Implement `ManyToManyMerger` with support for models with custom `through` relations.

## Deployment

    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [ ] `skip-cronjob-deploy`
    - [ ] `skip-daemon-deploy`


cl/corpus_importer/state/merger.py
```diff
@@ -266,33 +355,173 @@ def validate(self, field: Field | ForeignObjectRel) -> list[Exception]:
...
+            result |= child_merger.merge()
+            if child_merger.out is None:
+                continue
+            related_objects.append((child, child_merger.out))
+
+        if self.strategy is ManyStrategy.REPLACE:
+            # Delete everything we didn't update or create along with associated objects
+            _ = related_manager.exclude(
+                pk__in=[r.pk for _, r in related_objects]
+            ).delete()
```
Brennan-Chesley-FLP: If we have the same lawyer representing multiple clients on a case, and they get replaced by one of the clients but not both, this might remove them entirely from the DB.

cl/corpus_importer/state/merger.py
```diff
@@ -266,33 +355,173 @@ def validate(self, field: Field | ForeignObjectRel) -> list[Exception]:
...
+                    target_name=related_manager.target_field_name,  # type: ignore[attr-defined]
+                ),
+                existing=None,
+            )
+            result |= through_merger.merge()
+            if through_merger.out is None:
+                continue
+            through_objects.append(through_merger.out)
+
+        if self.strategy is ManyStrategy.REPLACE:
```
Brennan-Chesley-FLP: Should this be `through_strategy`?

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -1,27 +1,65 @@
...
+    role: int = Attribute(_florida_representative_role)
+
+
+class FloridaPartyMerger(PartyMerger[FloridaParty, RelatedParams[None]]):
+    attorneys: list[Attorney] = ManyToManyRelation(
+        AttorneyMerger, FloridaRoleMerger, _party_representatives
+    )
+
+
 def add_scraper_source(scrape: int | None, db: int | None) -> int:
```
Brennan-Chesley-FLP: I think we can get rid of this one in favor of the common one now.

cl/corpus_importer/state/merger.py
Brennan-Chesley-FLP, line 558: Type swap?

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -1,27 +1,65 @@
...
+from cl.corpus_importer.state.common.docket import (
+    DocketMerger,
+    _docket_parties,
+)
+from cl.corpus_importer.state.common.party import (
+    AttorneyMerger,
+    PartyMerger,
+    PartyTypeMerger,
+    RoleMerger,
+    _party_representatives,
```
Brennan-Chesley-FLP: Are this and `_docket_parties` intended for general consumption?

cl/corpus_importer/state/common/docket.py
```diff
@@ -0,0 +1,72 @@
...
+    )
+    case_name: str = Attribute(
+        lambda d, params: d.case_name, strategy=overwrite
+    )
+    case_name_full: str = Attribute(
+        lambda d, params: d.case_name_full,
+        strategy=overwrite,
+    )
+    case_name_short: str = Attribute(
+        lambda d, params: d.case_name, strategy=overwrite
```
Brennan-Chesley-FLP: Oh man, the naming of names.

cl/corpus_importer/state/merger.py
MorganBennetDev: Yes. Thanks!

cl/corpus_importer/state/florida/mergers.py
MorganBennetDev: Not really. This is just because I couldn't figure out a great way to override the `through` merger, but I've got an idea now that you've brought it up.

cl/corpus_importer/state/merger.py
MorganBennetDev: Good catch. Thanks

Brennan-Chesley-FLP (approved):
Nice.

cl/corpus_importer/state/florida/mergers.py
MorganBennetDev: Updated so you shouldn't have to import private methods anymore (and it's less verbose for subclasses to boot).

cl/corpus_importer/state/merger.py
```diff
@@ -266,33 +355,185 @@ def validate(self, field: Field | ForeignObjectRel) -> list[Exception]:
...
+            if child_merger.out is None:
+                continue
+            related_objects.append((child, child_merger.out))
+
+        if self.strategy is ManyStrategy.REPLACE:
+            # Delete everything we didn't update or create along with associated objects
+            to_keep = {r.pk for _, r in related_objects} | {
+                m.existing.pk for m in child_mergers if m.existing
+            }
+            _ = related_manager.exclude(pk__in=to_keep).delete()
```
albertisfu, line 485: Just wanted to confirm that this is the desired behavior.

It looks like `related_manager` can be either `party.attorneys` or `docket.parties` here, right?

So in this case, we'll remove both `Parties` and `Attorneys`?

Is this logic intended to replicate the `disassociate_extraneous_entities` logic? There we have:

> * Do not delete parties or attorneys; just allow them to be orphaned. Later, we can decide what to do with them, but keeping them around at least lets us recover them if needed.

Maybe it's correct to remove them here, but do you see any scenarios where deleting `Parties` or `Attorneys` could be risky, or where keeping them around might be useful in the future?

As I'm seeing in line 509:

> Only prune through objects belonging to this parent; other
> parents' relationships are out of scope for this merge.

So wanted to confirm this is not redundant as removing parent objects would remove the related instances.

cl/corpus_importer/state/common/party.py
```diff
@@ -0,0 +1,103 @@
...
+        docket = cast(Docket, self.params.parent)
+        return self.manager.filter(
+            name=self.transformed["name"], party_types__docket=docket
+        ).distinct()
+
+
+def _party_type_name(
+    party: ScrapeParty[ScrapeRepresentative], params: Any
+) -> str:
+    return party.party_type.value
```
albertisfu: Should this be in Title Case, like in RECAP? I'm seeing these values come in lowercase in Florida.

cl/corpus_importer/state/merger.py
```diff
@@ -472,27 +732,26 @@ def query(self) -> QuerySet[M]:
...
         """Merge scraped data into the DB."""
         if self.result is not None:
-            logger.error(
-                f"Merger {self.__class__.__name__} result already set; cannot run merge."
-            )
             return self.result
         if self.existing is None:
             try:
-                self.existing = self.query().get()
+                self.existing = (
```
albertisfu: In `add_parties_and_attorneys`, it catches an `IntegrityError` under a race condition and then tries to retrieve the `Party` again. If we encounter the same race condition here, do we need to handle it as well? And do it conditionally only for Parties.

But I'm not sure how that `IntegrityError` can occur there, since `Party` doesn't have a unique constraint. The unique constraint is on `PartyType` 🤔

cl/corpus_importer/state/merger.py
```diff
@@ -472,27 +732,26 @@ def query(self) -> QuerySet[M]:
...
+                    self.query().filter(**self._through_params).get()
+                )
             except self.model.MultipleObjectsReturned:  # type: ignore[attr-defined]
```
albertisfu: `add_parties_and_attorneys` also handles cases where the `Party` count is >= 2 by selecting the earliest `Party`.

I assume that scenario can be handled here as well, but do we also need to select the earliest one?

Maybe we could add a `resolve_multiple` method and call it here before raising the error:

```python
self.existing = self.resolve_multiple(qs)
```

This method could return `None` by default and be implemented in the Florida party merger to select the earliest instance.

cl/corpus_importer/state/common/docket.py
```diff
@@ -0,0 +1,77 @@
...
+    Attribute,
+    ManyToManyRelation,
+    Merger,
+    overwrite,
+)
+from cl.people_db.models import Party
+from cl.search.models import Docket
+
+
+def add_scraper_source(scrape: int | None, db: int | None) -> int:
```
albertisfu, line 23: I remember you introduced a similar method like this in the Docket merger PR. Can you reuse it that one?

albertisfu (changes requested):

A couple of suggestions in the code and some questions here:

* I remember we talked about adding a `pro_se_flag` to the model so we can identify self-represented parties. Is that something we can do here?
* If we add the `pro_se_flag`, we still need to decide whether we want to create an `Attorney` object for a self-represented party. I remember we were going to look at how other states handle this (Texas, California, New York, etc.).
* I also remember you mentioned we could store the `party_uuid` in `extra_info`.

cl/corpus_importer/state/merger.py
MorganBennetDev: I did some research into that issue since I wanted to cover the same edge cases as `add_parties_and_attorneys`, and this may be inaccurate for CL (please correct me if so), but the posts I was able to find seemed to agree that this specific issue is too rare to care about too much. If you'd rather stay on the safe side like the existing logic, I can add a retry here.

cl/corpus_importer/state/merger.py
MorganBennetDev: > It looks like `related_manager` can be either `party.attorneys` or `docket.parties` here, right?
> 
> So in this case, we'll remove both `Parties` and `Attorneys`?

Correct. The `REPLACE` strategy is intended to remove all objects that weren't in the scraped data.

> Maybe it's correct to remove them here, but do you see any scenarios where deleting `Parties` or `Attorneys` could be risky, or where keeping them around might be useful in the future?

Keeping them could be useful if we want to keep a record of changes to parties and representation in a case. However, that's more of a question for the specific merger implementation where we can make the choice between `REPLACE` and `APPEND` strategies.

> So wanted to confirm this is not redundant as removing parent objects would remove the related instances.

This is not redundant because there _might_ be a scenario where we don't need to delete a `Party` but we do need to delete a `Role` (e.g. an `Attorney` is deleted but the `Party` isn't).

cl/corpus_importer/state/common/docket.py
MorganBennetDev: This is the same method, but it's been moved to the common module since it will be useful for future mergers as well.

MorganBennetDev:
> I remember we talked about adding a `pro_se_flag` to the model so we can identify self-represented parties. Is that something we can do here?

Yes. Should I generate the migrations and update the merger?

> If we add the `pro_se_flag`, we still need to decide whether we want to create an `Attorney` object for a self-represented party. I remember we were going to look at how other states handle this (Texas, California, New York, etc.).

This is something I need to do, but forgot to while handling other tasks. Will get on that.

> I also remember you mentioned we could store the `party_uuid` in `extra_info`.

Adding this. Thanks for the reminder.

cl/corpus_importer/state/merger.py
MorganBennetDev: > Maybe we could add a `resolve_multiple` method and call it here before raising the error:
> 
> ```python
> self.existing = self.resolve_multiple(qs)
> ```
> 
> This method could return `None` by default and be implemented in the Florida party merger to select the earliest instance.

This is a good idea. If we do this, we should also use the optimized counting strategy from `find_docket_object` since `get` by default fetches a full page of results, which is very much not what we want here. This won't cost anything since if I remember, `get` doesn't use the cache.

As an alternative, we could use `list(qs)` and then run `resolve_multiple(qs)` if the length is greater than one, in case `resolve_multiple` needs to look at results, but that seems unlikely.

albertisfu:
Thanks @MorganBennetDev !

> Yes. Should I generate the migrations and update the merger?

Could you put that migration in the Florida models PR, then stack this PR on top of that one and apply the required changes? We'll need to apply those migrations before merging that PR, but that will make things easier for the infra team.

> This is something I need to do, but forgot to while handling other tasks. Will get on that.

Yeah, please. With that information, and maybe by asking in slack, we can make a decision on whether we should create an `Attorney` instance for self-represented parties or leave it blank.

MorganBennetDev:
> Could you put that migration in the Florida models PR, then stack this PR on top of that one and apply the required changes?

Done

> With that information, and maybe by asking in slack, we can make a decision on whether we should create an `Attorney` instance for self-represented parties or leave it blank.

Asked in Slack. Will update here with response.

MorganBennetDev:
> > With that information, and maybe by asking in slack, we can make a decision on whether we should create an Attorney instance for self-represented parties or leave it blank.
>
> Asked in Slack. Will update here with response.

The conclusion of the discussion in Slack was that we should add a small int field to `PartyType` with values `PRO_SE_YES`, `PRO_SE_NO`, and `PRO_SE_UNKNOWN` (default). This leaves us open to adding more options for pro se info in the future and makes it so we don't have to backfill the entire database with the correct values. This migration has already been applied in the parent PR #7552.

albertisfu (commented):
Thanks, @MorganBennetDe.  Just one thing, could you please add an assertion to one of the existing test cases to confirm that the `pro_se` field is populated correctly based on the boolean flag?

albertisfu (approved):
Thanks @MorganBennetDev this is ready to be merged!

PR #7506, opened 2026-06-27
fix(scrapers): encode opinion content to bytes before S3 upload

### LINKED ISSUE #7504: Scraper opinion uploads fail with S3 500 InternalError for non-ASCII text content
## Summary

`make_objects` in `cl_scrape_opinions.py` wraps opinion content in `ContentFile(content)` where `content` is a `str`. For text content containing non-ASCII characters, the resulting S3 PUT declares the **character count** as `X-Amz-Decoded-Content-Length` while the body serializes to a larger number of **UTF-8 bytes**. S3 receives more payload bytes than the request declares, the `aws-chunked` CRC32 trailer no longer aligns, and S3 returns `500 InternalError` on every attempt. botocore exhausts its 5 retries and the exception propagates uncaught, aborting the scrape and surfacing in Sentry.

Sentry: issue **7575975495** / `COURTLISTENER-DHH` — `ClientError: An error occurred (InternalError) when calling the PutObject operation (reached max retries: 4)`, originating at `cl_scrape_opinions.py:188` `opinion.local_path.save(...)`.

## Impact

- Affected opinions are **never stored** (the whole court's `ingest_a_case` aborts).
- Recurring Sentry noise (25+ occurrences), currently only from `alaska` and `alaskactapp`.
- Deterministic per-document: any opinion whose content is a `str` with non-ASCII bytes fails 100% of the time.

## Root cause

```python
# cl/scrapers/management/commands/cl_scrape_opinions.py  (make_objects, ~L184)
cf = ContentFile(content)                              # content is a str
...
opinion.local_path.save(file_name, cf, save=False)     # L188 -> raises
```

`ContentFile(<str>)` reports `.size = len(content)` (a **character** count). The real write path — `IncrementingAWSMediaStorage` → django-storages `_save` → `obj.upload_fileobj(..., Config=self.transfer_config)` → `client.put_object` — uses that size as `X-Amz-Decoded-Content-Length` for the streaming `aws-chunked` body, but transmits the content as UTF-8. When `len(str) != len(str.encode("utf-8"))` (i.e. any non-ASCII), the declared length is short, S3's chunked decoder misreads the payload/trailer boundary, and returns `500 InternalError`.

## Why only `alaska` / `alaskactapp`

Two conditions must both hold to trigger it:

1. `site.download_content()` returns a **`str`** (text/HTML responses, not PDF bytes), **and**
2. the content contains **non-ASCII** characters (so char count != UTF-8 byte count).

Westlaw-sourced opinions (alaska, alaskactapp) deliver HTML with Unicode typography — curly quotes `'` `"`, em-dashes `—`, `§`. PDF courts return `bytes` (size is correct → fine); plain-ASCII HTML courts have char count == byte count (→ fine). So this is **not** an Alaska-scraper bug — it's a generic CL write-path bug that Alaska happens to trigger reliably. Any current/future HTML court with non-ASCII content hits the identical path.

## Minimal reproduction

Self-contained — no scraper or network needed. Run in `docker exec -it cl-django python manage.py shell`. The only thing that matters is a `str` whose character count differs from its UTF-8 byte count:

```python
from datetime import date
from django.core.files.base import ContentFile
from cl.search.models import Opinion

# em-dash (U+2014) is 3 UTF-8 bytes, so char count != byte count
content = "<html>" + "—" * 1000 + "</html>"
assert len(content) != len(content.encode("utf-8"))   # 1013 chars vs 3013 bytes

op = Opinion(sha1="0" * 40, download_url="https://example.com/repro")
op.file_with_date = date(2026, 6, 24)   # make_upload_path needs this

# FAILS: ContentFile(str) -> X-Amz-Decoded-Content-Length = 1013 (chars);
#        body is 3013 UTF-8 bytes -> S3 500 InternalError x5 -> ClientError
op.local_path.save("zz_repro_nonascii.html", ContentFile(content), save=False)

# PASSES: encode first -> declared length = 3013 = actual bytes
op.local_path.save("zz_repro_nonascii.html", ContentFile(content.encode("utf-8")), save=False)
```

The first `save()` raises `ClientError(... InternalError ... reached max retries: 4)`; the second succeeds. (Clean up the `zz_repro_nonascii*.html` objects from the bucket afterward.)

To see the mismatch on the wire, enable `boto3.set_stream_logger("botocore", logging.DEBUG)` first and compare `X-Amz-Decoded-Content-Length` between the two calls. A raw `client.put_object(Body=<str>)` does **not** reproduce it (that path counts encoded bytes), so use the `ContentFile(str)` + `.save()` path shown above.

## Fix

Encode to bytes before wrapping, in `make_objects` (central, so it covers every court). `force_bytes` is already imported in the module:

```python
cf = ContentFile(force_bytes(content))   # was: ContentFile(content)
```

This makes `.size` the true UTF-8 byte count; declared and actual lengths match; the PUT succeeds.

## Regression test

Add a test that pushes a non-ASCII `str` through the opinion save path and asserts the stored object's size equals the UTF-8 byte count (fails today, passes after the fix), using content containing `—`, `'`, `§`. Guards every court, not just Alaska.

## Defense-in-depth (optional, separate)

Independent of this fix, transient S3 `500/503`s are real. Consider bumping `AWS_S3_CLIENT_CONFIG` retries to `standard`/`adaptive` mode and/or wrapping the `make_objects` save in a small retry-on-`ClientError`, so an unrelated transient blip doesn't abort an entire court's scrape. Not required to close this issue.

grossir:
## Fixes

Fixes: #7504

## Summary

Scraper opinion uploads were failing with S3 `500 InternalError` on `PutObject` for non-ASCII text opinions (currently `alaska` / `alaskactapp`).

`make_objects` wrapped content with `ContentFile(content)` where `content` is a `str`. `ContentFile(str)` reports `.size` as the **character** count, which django-storages sends as `x-amz-decoded-content-length` on the `aws-chunked` streaming PUT — but the body serializes to **UTF-8 bytes**. For non-ASCII text (Westlaw HTML: curly quotes, em-dashes, `§`) the declared length is short of the real byte count, so S3 rejects the upload, botocore exhausts its 5 retries, and the exception propagates uncaught, aborting the court's scrape (Sentry `COURTLISTENER-DHH` / 7575975495).

Fix: wrap with `ContentFile(force_bytes(content))` so the reported size is the true UTF-8 byte count.

`force_bytes` is type-aware and side-effect-free across scrapers:
- **bytes content** (PDFs, and every `cleanup_content` override that returns bytes — including `okla`'s ISO-8859-1) → returned unchanged (no-op).
- **str content** (alaska/alaskactapp) → encoded UTF-8, which matches the dedup hash already computed as `sha1(force_bytes(content))` in `get_opinions_content`, so stored bytes now equal the hashed bytes (no dedup change).

A full breakdown of every `cleanup_content` return type and the safety argument is in the issue: https://github.com/freelawproject/courtlistener/issues/7504#issuecomment-4814402438

Adds `MakeObjectsContentEncodingTest`, which pushes non-ASCII `str` content through `make_objects` and asserts the stored object equals `content.encode("utf-8")` (fails before the fix, passes after).

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [ ] `skip-cronjob-deploy`
    - [ ] `skip-daemon-deploy`

No special deploy steps required.

### DIFF
--- a/cl/scrapers/management/commands/cl_scrape_opinions.py
+++ b/cl/scrapers/management/commands/cl_scrape_opinions.py
@@ -181,7 +181,13 @@ def make_objects(
             ordering_key=opinion_metadata.get("ordering_key"),
         )
 
-        cf = ContentFile(content)
+        # Encode str content to bytes so the file's reported size is the UTF-8
+        # byte count, not the character count. django-storages sends that size
+        # as x-amz-decoded-content-length on the aws-chunked PUT; for non-ASCII
+        # text (e.g. Westlaw HTML with curly quotes/em-dashes) a character count
+        # is short of the real byte count, and S3 rejects the upload with
+        # 500 InternalError. See #7504.
+        cf = ContentFile(force_bytes(content))
         extension = get_extension(content)
         file_name = trunc(item["case_names"].lower(), 75) + extension
         opinion.file_with_date = cluster.date_filed
--- a/cl/scrapers/tests.py
+++ b/cl/scrapers/tests.py
@@ -15,6 +15,7 @@ from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
 from django.core.files.base import ContentFile
 from django.test import SimpleTestCase
+from django.utils.encoding import force_bytes
 from django.utils.timezone import now
 from juriscraper.AbstractSite import logger
 from juriscraper.lib.exceptions import UnexpectedContentTypeError
@@ -35,6 +36,7 @@ from cl.citations.models import UnmatchedCitation
 from cl.corpus_importer.tasks import (
     merge_scotus_docket as real_merge_scotus_docket,
 )
+from cl.lib.crypto import sha1
 from cl.lib.exceptions import ScrapeFailed
 from cl.lib.juriscraper_utils import get_module_by_court_id
 from cl.lib.microservice_utils import microservice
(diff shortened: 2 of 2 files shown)


quevon24 (approved):
The fix seems good, and it's good to have a test for it

PR #7508, opened 2026-06-27
feat(api): double API throttles for membership promotion

ERosendo:
## Fixes

This PR fixes https://github.com/freelawproject/internal/issues/1050

## Summary

This PR adds a time-boxed x2 boost to API rate limits, controlled by a new `double_api_throttles` waffle switch. This lets us enable the promotion and disable it when it ends without requiring a redeploy.

**Who gets doubled**

| Group | Doubled? |
|---|---|
| Paid memberships (TIER_1–4, GROUP_T*, HEYCOUNSEL_T*) | ✅ |
| Non-members (default user rate) | ✅ |
| LSO (Included in [9772253](https://github.com/freelawproject/courtlistener/pull/7508/commits/9772253c7ec526495504942e678bf2a2614c5886)) | ✅ |
| EDU members | ❌ |
| MANUAL / commercial overrides | ❌ |

The promotion is implemented as shared eligibility logic that's reused throughout the codebase:

- **Throttle enforcement**: `ExceptionalUserRateThrottle.allow_request` doubles the resolved rates for eligible users.
- **No retroactive penalties**: promo traffic is stored in a separate `_promo2x` throttle cache window (via get_cache_key). When the promotion ends, users return to the normal limits with a clean slate instead of having a window full of requests counted against the lower limits.
- **Consistent reporting**: the API usage page and `get_recent_api_request_count` use the same eligibility check, so the
displayed limits always match the enforced limits.

Lifetime and daily usage stats are untouched, they're tracked on a separate Redis path (`_log_request`), independent of the throttle history.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


cl/api/utils.py
```diff
@@ -796,6 +801,63 @@ def get_all_throttle_overrides(
...
+            | Q(termination_date__date__gte=today)
+        )
+        .values_list("user__username", flat=True)
+    )
+    return manual_users | edu_members
+
+
+def promo_doubling_applies(user: User) -> bool:
+    """Return whether the x2 API promotion applies to this user."""
+    if not switch_is_active(DOUBLE_API_THROTTLES_SWITCH):
```
albertisfu: This will hit the Redis cache on each request. Could we cache the `switch_is_active` status in memory to avoid hitting Redis too often? Maybe with a 10-minute cache?

It’s fine if it takes up to 10 minutes for the increase to be enabled or disabled after the switch is activated.

cl/api/utils.py
```diff
@@ -944,6 +1018,8 @@ def allow_request(self, request, view):
 
         overrides = get_all_throttle_overrides(self.throttle_type)
         rates = overrides.get(request.user.username) or self.default_rates
+        if promo_doubling_applies(request.user):
```
albertisfu: In `get_cache_key`, we’re doing:

`if self.scope == "user" and promo_doubling_applies(request.user):`

But here, we only check:

`if promo_doubling_applies(request.user):`

To be consistent, could we use the same condition?

`if self.scope == "user" and promo_doubling_applies(request.user):`

Maybe we can create a helper method for it.

albertisfu (changes requested):
Thanks @ERosendo this looks good!

I just a couple of suggestions before we merge it.

cl/api/utils.py
ERosendo: Good catch! I extracted a `_promo_applies(request)` helper on `ExceptionalUserRateThrottle` that encapsulates `self.scope == "user"` and `promo_doubling_applies(request.user)`, then used it in both `get_cache_key()` and `allow_request()`. That way the eligibility check is defined in one place and stays consistent across both methods.

albertisfu (approved):
Thanks @ERosendo this looks ready to go!

PR #7514, opened 2026-06-29
Offload FL back-scrape S3 archiving to a Celery task

MorganBennetDev:
Description generated by Claude:

> The Florida back-scraper archived each parsed docket and raw response to S3 from a single background thread doing blocking, serial PUTs (~166ms each). Log analysis showed this pacing the entire scrape: ~96% of wall-clock was serial S3 writes while the court API sat idle.
> 
> Move the S3 write into a new `save_response_to_s3` Celery task so archiving runs concurrently across workers, with retry/backoff on transient S3 errors. The command now dispatches tasks paced by CeleryThrottle (run in an executor so the async scrape loop isn't blocked) instead of an in-process queue and thread. Adds --queue and --throttle-min-items options.

## Fixes

Fixes Florida scraper S3 archive bottleneck.

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/scrapers/management/commands/back_scrape_fl_dockets.py
+++ b/cl/scrapers/management/commands/back_scrape_fl_dockets.py
@@ -2,10 +2,12 @@
 
 import asyncio
 import hashlib
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from datetime import date
+from functools import lru_cache
 from pathlib import Path
 
+import botocore.exceptions
 from asgiref.sync import async_to_sync
 from django.core.management import CommandParser
 from httpx import Response
@@ -20,6 +22,7 @@ from juriscraper.state.RequestManager import (
 
 from cl.lib.celery_utils import CeleryThrottle
 from cl.lib.command_utils import logger
+from cl.lib.decorators import retry
 from cl.lib.storage import S3GlacierInstantRetrievalStorage
 from cl.scrapers.management.utils import (
     ScraperCheckpointTracker,
@@ -38,12 +41,35 @@ class S3Cache(RequestHandler):
 
     :ivar base: The prefix where cached values are stored
     :ivar save: Whether to archive responses
-    :ivar load: Whether to intercept requests and load responses from S3 if available"""
+    :ivar load: Whether to intercept requests and load responses from S3 if available
+    :ivar throttle: The celery throttle to use when adding responses to the queue
+    :ivar queue_name: The celery queue to dispatch archive tasks to
+    :ivar storage: The S3 storage backend to use for archiving and retrieval"""
 
     base: Path
     save: bool
     load: bool
+    throttle: CeleryThrottle
     queue_name: str
+    storage: S3GlacierInstantRetrievalStorage = field(
+        default_factory=S3GlacierInstantRetrievalStorage
+    )
+
+    @retry(
+        (
+            botocore.exceptions.HTTPClientError,
+            botocore.exceptions.ConnectionError,
+        ),
+        delay=1,
+        backoff=2,
+    )
+    @lru_cache(512)
+    def s3_key_exists(self, key: str) -> bool:
+        """Cached check for whether a key exists in S3 with retries. Allows us to avoid an extra `storage.exists` call
+        when `save` and `load` are both enabled."""
+        if self.storage.exists(key):
+            return True
+        return False
 
     def make_s3_key(self, request: ScheduledRequest) -> str:
         url_path = request.url.path.lower().strip("/")
@@ -60,14 +86,13 @@ class S3Cache(RequestHandler):
         key = self.make_s3_key(request)
 
         try:
-            storage = S3GlacierInstantRetrievalStorage()
-            if not storage.exists(key):
+            if not self.s3_key_exists(key):
                 logger.info("Did not find %s in S3", request.url.path.lower())
                 return None
 
             logger.info("Loading %s from S3", request.url.path.lower())
 
-            with storage.open(key, "rb") as f:
+            with self.storage.open(key, "rb") as f:
                 return Response(200, content=f.read())
         except Exception:
             logger.exception(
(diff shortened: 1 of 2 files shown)


cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -87,8 +85,9 @@ def save_to_s3(
             # Don't try to cache responses that we pulled from the cache
             if self.load and storage.exists(key):
                 return
-            # We want this to block so our scrape doesn't get too far ahead of the archive
-            self.save_queue.put((key, response.content))
+            save_response_to_s3.si(key, response.content).set(
```
albertisfu, line 113: This task that stores the responses isn't throttled, right?

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -87,8 +85,9 @@ def save_to_s3(
             # Don't try to cache responses that we pulled from the cache
             if self.load and storage.exists(key):
```
albertisfu: Would it be possible to avoid this additional `exists` request, considering that `load_from_s3` has already performed one, so we already know whether the file exists in S3? Maybe we could keep a dictionary that stores the existence status for each key, and then remove the key from the cache once the request flow ends.

albertisfu (changes requested):
Thanks @MorganBennetDev this looks good. Just a couple of suggestions, please.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
MorganBennetDev: It is not. Should I add a throttle (maybe 10/s which is 4x our current request rate limit)?

albertisfu: I was thinking of using the same `throttle.maybe_wait` backpressure method, considering we're going to use the same queue to store both raw and parsed responses. Or do you think we should use a different throttle rate for each of them?

cl/scrapers/management/commands/back_scrape_fl_dockets.py
MorganBennetDev: That sounds like a good idea, but I think it would be easier to make a `_exist_in_s3(key: str)` function and decorate it with `functools.lru_cache` so we don't have to handle the eviction logic ourselves.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
MorganBennetDev: Same throttle should be fine. Was tired and misunderstood your comment as asking about decorating the celery task itself with a throttle.

MorganBennetDev:
@albertisfu Updated. Should be good to go once tests pass. Added some type hints as well.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -40,12 +41,35 @@ class S3Cache(RequestHandler):
...
+    @retry(
+        (
+            botocore.exceptions.HTTPClientError,
+            botocore.exceptions.ConnectionError,
+        ),
+        delay=1,
+        backoff=2,
+    )
+    @lru_cache(512)
+    def s3_key_exists(self, key: str) -> bool:
```
albertisfu, line 67: I ran this version locally, and the scraper completely freezes the first time this method is reached.

It looks like the problem is the use of `lru_cache`.

`@lru_cache` needs to build a cache key from all the arguments, and for a method the first argument is `self`. So `lru_cache` tries to hash `(self, key)`, which calls `hash(self)`. But `S3Cache` is an unhashable dataclass, so this breaks the request loop.

I’m not sure if `lru_cache` is still the best tool here. My initial idea was to use a custom instance-level dictionary.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -83,19 +106,20 @@ def save_to_s3(
...
         key = self.make_s3_key(request)
         try:
-            storage = S3GlacierInstantRetrievalStorage()
             # Don't try to cache responses that we pulled from the cache
-            if self.load and storage.exists(key):
+            if self.load and self.s3_key_exists(key):
                 return
-            # We want this to block so our scrape doesn't get too far ahead of the archive
-            self.save_queue.put((key, response.content))
+            self.throttle.maybe_wait()
```
albertisfu, line 112: Why isn’t it required to run `maybe_wait` inside `run_in_executor` here? (As you did in `_backfill`)

```python
await loop.run_in_executor(None, throttle.maybe_wait)
```

Also, could there be any concurrency issues when calling `throttle.maybe_wait` for both storing parsed responses and raw files?

albertisfu (changes requested):
Thanks @MorganBennetDev a couple of comments, please.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
MorganBennetDev: This method is synchronous. The `run_in_executor` call is in `listen` on line 144.

MorganBennetDev: > Also, could there be any concurrency issues when calling `throttle.maybe_wait` for both storing parsed responses and raw files?

Yes, and now that I'm thinking about it I think we should just let `throttle.maybe_wait` block instead of running it in a separate thread and waiting. If `full_scrape=False`, I think it'd be fine since we'd do at most 1 new response archive per iteration of the backfill loop (results), but for the full scrape each iteration can cause an arbitrary number of requests (1 per docket entry attachment) which could get out of hand.

albertisfu: Yeah! That was my concern too. I agree with letting `throttle.maybe_wait` block.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
MorganBennetDev: I added

```python
def __hash__(self):
    return hash(id(self))
```

which seems to work. According to the docs for `id`:

> This is an integer which is guaranteed to be unique and constant for this object during its lifetime.

I can also just set up an instance-level LRU cache if you think we should do that.

MorganBennetDev:
@albertisfu Took care of the method cache and the potential race condition. The throttle is blocking now, but that shouldn't be an issue if we have enough workers.

albertisfu (approved):
Thanks @MorganBennetDev this is working properly now. Merging...

PR #7515, opened 2026-06-30
fix(scrape_pacer_free_opinions): materialize get_pdfs rows to avoid DB cursor death

### LINKED ISSUE #7507: get_pdfs server-side cursor dies mid-iteration — scrape-free-opinions CronJob failing in production every day
Follow-up from #7417 (see also #7445, #7489, #7490).

This promotes the "connection-death observation (not yet ticketed)" noted at the bottom of #7489 into its own issue, because it is no longer a one-off manual sweep — it is **failing the scheduled production CronJob every single day**.

## Summary

The `scrape-free-opinions` CronJob (`court-listener` namespace, image `freelawproject/courtlistener:735655f-prod`) has failed on **5 consecutive days**. Every pod exits `Error` during the `get-pdfs` step:

```
scrape-free-opinions-29701920-n84vv   0/1   Error   4d15h
scrape-free-opinions-29703360-jpcn9   0/1   Error   3d15h
scrape-free-opinions-29704800-nqrd2   0/1   Error   2d15h
scrape-free-opinions-29706240-vl9ck   0/1   Error   39h
scrape-free-opinions-29707680-s9w85   0/1   Error   15h
```

Because the CronJob command is a single `&&`-chained sequence (`get-report-results` ×3 → `get-pdfs` ×3 → `ocr-available`), the crash in `get-pdfs` **aborts the rest of the run**: no PDFs are downloaded and OCR never runs. Free-opinion PDF ingestion has effectively been down for ~4+ days.

## Error

Consistently crashes **right after dispatching 2000 tasks**:

```
INFO Sent 2000/888397 tasks to celery for downloading so far.
  ...
  File "/opt/courtlistener/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py", line 439, in get_pdfs
    for row in rows.iterator():
  ...
django.db.utils.OperationalError: consuming input failed: SSL connection has been closed unexpectedly
```

## Root cause

`get_pdfs()` iterates the result set with `rows.iterator()` (`scrape_pacer_free_opinions.py:439`), which opens a **server-side cursor**. Django's `.iterator()` defaults to a chunk_size of **2000** — exactly where the crash lands:

1. The cursor fetches the first 2000 rows.
2. For each row the loop dispatches a Celery chain, but it also blocks on `throttle.maybe_wait()` (waiting for the `free_pacer_docs` queue to drain) and `time.sleep(1)` per court cycle. With an ~888k-row backlog the loop spends most of its time **idle, issuing no DB activity**.
3. By the time it has worked through the first chunk and calls `fetchmany()` for the next, the idle connection backing the cursor has been reset (idle/LB/NAT timeout, `idle_in_transaction_session_timeout`, or a brief failover). A server-side cursor **cannot be resumed by reconnecting**, so psycopg raises `SSL connection has been closed unexpectedly`.

In short: **a server-side cursor is held open for hours across a long, mostly-idle dispatch loop, and the DB connection does not survive the gap between chunk fetches.** It fails at the 2000-row boundary on every run.

The huge `count` in the log (888,397) is itself a symptom of the backlog inflation tracked in #7489 (zombie `error_msg=""` rows that never drain) and #7490 (~5.25x duplicate rows from non-idempotent ingestion). The larger that eligible set, the longer the cursor stays open and the more certain this crash becomes.

## Proposed fix

Don't hold a server-side cursor open across the slow dispatch loop. The query only selects `pk` and `court_id`, so materialize the ordering up front and iterate a plain Python list, freeing the connection between enqueues:

```python
# Materialize before the slow dispatch loop so no server-side cursor is
# held open across Celery throttling / sleeps.
rows_list = list(rows.values_list("pk", "court_id"))
count = len(rows_list)
...
for pk, court_id in rows_list:
    throttle.maybe_wait()
    ...
```

For 2 ints × ~888k rows this is only tens of MB. (If constant memory is ever required, the alternative is keyset pagination over `(row_number, court_id)` with a fresh `LIMIT n` query per batch — more code, not needed at this scale.)

This is the same fix already proposed in the #7489 footnote; this issue just makes it a standalone, independently-shippable bug since it's an active production outage rather than a backlog-quality concern.

## Relationship to other issues

- **#7489** (`error_msg=""` zombie-row leak) — first recorded this exact crash as an unticketed observation; that leak inflates the eligible-row count feeding the long cursor.
- **#7490** (non-idempotent report ingestion, ~5.25x row bloat) — other major inflator of the backlog.
- **#7445** (speed up backfills) — its recommended workaround is the single whole-table `get_pdfs` call that triggers this crash; #7489 already notes this crash as "a separate failure mode of #7445's recommended single-`get_pdfs` workaround." A worker-side per-court rate limit (#7445) would also let the throttle stop keeping the queue near-empty, indirectly shortening idle gaps.
- **#7424** (per-scrape `document_count` on `PACERFreeDocumentLog`) — independent (silent-gap detection).

## How to confirm

```bash
kubectl logs <newest scrape-free-opinions pod> -n court-listener --tail=200
# -> "Sent 2000/<N> tasks..." immediately followed by the psycopg OperationalError at get_pdfs line 439
```

grossir:
## Fixes
Fixes #7507

## Summary

The scrape-free-opinions CronJob has been failing daily: get_pdfs() iterated the result set with .iterator(), opening a server-side cursor held open across the slow, mostly-idle dispatch loop (Celery throttling + per-cycle sleeps). The backing DB connection was reset between chunk fetches, so psycopg raised "SSL connection has been closed unexpectedly" at the 2000-row boundary on every run, aborting the rest of the &&-chained command (PDF download + OCR).

- Replace .only("pk", "court_id") + .iterator() with a materialized .values_list("pk", "court_id"), freeing the DB connection before the loop
- Iterate the in-memory (pk, court_id) tuples instead of model instances
- values_list keeps the per-row footprint minimal (~127 MiB at current backlog) vs ~3.6x larger for hydrated .only() instances


quevon24 (approved):
The fix looks good and the change makes sense

PR #7536, opened 2026-07-02
perf: Optimize citation pipeline ORM usage

KRRT7:
- preload opinion clusters in citation batch processing
- fetch parenthetical clusters in one bulk query instead of one query per cluster
- use existing cluster foreign key IDs when updating citation counts
- isolate the chunked citation HTML test from Elasticsearch-backed resolution

### DIFF
--- a/cl/citations/tasks.py
+++ b/cl/citations/tasks.py
@@ -379,10 +379,10 @@ def store_opinion_citations_and_update_parentheticals(
             # Update parenthetical groups for clusters that we have added
             # parentheticals for from this opinion
             logger.debug("Create parenthetical groups: %s", opinion.pk)
-            clusters = OpinionCluster.objects.in_bulk(
-                clusters_to_update_par_groups_for
+            clusters = OpinionCluster.objects.filter(
+                pk__in=clusters_to_update_par_groups_for
             )
-            for cluster in clusters.values():
+            for cluster in clusters:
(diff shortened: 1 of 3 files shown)


mlissner:
@claude review

cl/citations/tasks.py
```diff
@@ -379,10 +379,11 @@ def store_opinion_citations_and_update_parentheticals(
...
             logger.debug("Create parenthetical groups: %s", opinion.pk)
-            for cluster_id in clusters_to_update_par_groups_for:
-                create_parenthetical_groups(
-                    OpinionCluster.objects.get(pk=cluster_id)
-                )
+            clusters = OpinionCluster.objects.in_bulk(
+                clusters_to_update_par_groups_for
+            )
+            for cluster in clusters.values():
+                create_parenthetical_groups(cluster)
```
grossir: This optimization works. 

However, seems like `clusters = OpinionCluster.objects.filter(pk__in=clusters_to_update_par_groups_for)` would be more appropiate? The dictionary built by `in_bulk` is discarded right away

grossir (approved):
@KRRT7 thanks for the contribution. These are nice optimizations

Just a minor comment you can address, I will probably merge this later anyway

cl/citations/tasks.py
KRRT7: you're right

PR #7537, opened 2026-07-02
perf: Defer search alert quota context

KRRT7:
- move search alert quota context construction behind a helper
- defer the alert quota aggregate until search.html is actually rendered with alert UI
- avoid doing that alert-count aggregate on homepage GETs and successful alert POST redirects


mlissner:
@claude review

mlissner (approved):

PR #7552, opened 2026-07-06
Create `FloridaDocketEntry` and `FloridaDocument` models

### LINKED ISSUE #7362: Create `FloridaDocketEntry` and `FloridaDocument` models
As decided in #6589, we should create new docket entry and document models for each state we scrape. This issue is to create them for Florida.

MorganBennetDev:
## Fixes

Closes #7362 

## Summary

Creates `FloridaDocketEntry` and `FloridaDocument` models per discussion in #7362.

Adds new models to admin panel.

Adds test factories for new models.

## Deployment

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`


cl/lib/model_helpers.py
```diff
@@ -421,6 +422,13 @@ def make_pdf_path(instance, filename, thumbs=False):
         root = Path(f"us/state/tx/{court_id}")
         file_name = f"gov.tx.{court_id}.{slug}{ext}"
         return str(root / file_name)
+    elif isinstance(obj, FloridaDocument):
```
albertisfu: This should be `instance` right?
```suggestion
    elif isinstance(instance, FloridaDocument):
```

cl/search/state/florida/models.py
```diff
@@ -0,0 +1,111 @@
...
+    )
+    date_filed = models.DateTimeField(
+        null=True,
+        blank=True,
+    )
+    date_submitted = models.DateTimeField(
+        null=True,
+        blank=True,
+    )
+    entry_type = models.SmallIntegerField(choices=DocketEntryType.CHOICES)
```
albertisfu: Should we provide a default option here?

cl/search/state/florida/models.py
```diff
@@ -0,0 +1,111 @@
...
+    entry_type_raw = models.TextField(blank=True)
+    entry_name = models.TextField(blank=True)
+    description = models.TextField(null=True, blank=True)
+    submitted_by = models.ForeignKey(
+        "people_db.Person",
```
albertisfu: Who is the person represented here in the Florida content? Is it a judge or a party?

In the spreadsheet, I see it's proposed to link this to `people_db.Party`. Is this party always present among the docket's parties? Otherwise, it might not be possible to merge it?

cl/search/state/florida/models.py
```diff
@@ -0,0 +1,111 @@
...
+    docket_entry_uuid = models.UUIDField()
+
+    class Meta:
+        app_label = "search"
+        ordering = ["-date_filed"]
+        indexes = [models.Index(fields=["docket_entry_uuid"])]
+        verbose_name_plural = "Florida Docket Entries"
+        constraints = [
+            models.UniqueConstraint(
+                fields=["docket", "docket_entry_uuid"],
```
albertisfu: What if we invert the fields here to `docket_entry_uuid, docket`? That way, the unique constraint's leading column could be leveraged for queries that only filter by `docket_entry_uuid`, allowing us to remove the standalone index:

`indexes = [models.Index(fields=["docket_entry_uuid"])]`

cl/search/state/florida/models.py
```diff
@@ -0,0 +1,111 @@
...
+    processing_error = models.SmallIntegerField(
+        choices=ProcessingError.CHOICES,
+        null=True,
+        blank=True,
+    )
+
+    class Meta:
+        app_label = "search"
+        ordering = ["link_uuid"]
+        indexes = [models.Index(fields=["link_uuid"])]
```
albertisfu: Can we add an index on `filepath_local` column too? similar to what we have in Texas and RECAP.

cl/search/state/florida/models.py
```diff
@@ -0,0 +1,111 @@
...
+    :ivar url: Download URL for attachment. Derived from uuid and link_uuid. Stored for safety.
+    :ivar processing_error: The processing error for the document, if any.
+    """
+
+    docket_entry = models.ForeignKey(
+        FloridaDocketEntry,
+        on_delete=models.CASCADE,
+        related_name="documents",
+    )
+    content_type = models.CharField(max_length=63, null=True, blank=True)
```
albertisfu: Is there a reason for using null on this field?

Django generally [discourages](https://docs.djangoproject.com/en/6.0/ref/models/fields/#field-options) using `null=True` for `CharField` and `TextField`.

The same applies to other fields, such as `description` here and the document `description`.

albertisfu (changes requested):
Thanks @MorganBennetDev this looks good I just have a few comments and suggestions, please.

cl/lib/model_helpers.py
MorganBennetDev: Yes. I think I fixed this in a different PR but forgot to fix it here.

cl/search/state/florida/models.py
MorganBennetDev: Ah I made a mistake with the model the FK points to. Fixing.

> In the spreadsheet, I see it's proposed to link this to `people_db.Party`. Is this party always present among the docket's parties?

In the trial scrape (https://github.com/freelawproject/courtlistener/issues/7362#issuecomment-4714169136), this was true in all but one case. This docket was consolidated with a different one, but kept tracking docket entries, which included parties that were in the consolidated case but not the original. Maybe we can add a `submitted_by_name` column to handle these cases (since `submitted_by` is already nullable)?

cl/search/state/florida/models.py
```diff
@@ -0,0 +1,115 @@
...
+    class Meta:
+        app_label = "search"
+        ordering = ["link_uuid"]
+        indexes = [
+            models.Index(fields=["link_uuid"]),
+            models.Index(fields=["filepath_local"]),
+        ]
+        constraints = [
+            models.UniqueConstraint(
+                fields=["docket_entry", "link_uuid"],
```
albertisfu: Similarly if we invert the columns in the UniqueConstraint  to: `"link_uuid", "docket_entry"`

We'll be able to remove the standalone index: `fields=["link_uuid"]`

cl/search/state/florida/models.py
```diff
@@ -0,0 +1,115 @@
...
+        FloridaDocketEntry,
+        on_delete=models.CASCADE,
+        related_name="documents",
+    )
+    content_type = models.CharField(max_length=63, blank=True)
+    document_name = models.TextField(blank=True)
+    document_type = models.TextField(blank=True)
+    description = models.TextField(blank=True)
+    link_uuid = models.UUIDField()
+    url = models.URLField(max_length=250)
```
albertisfu, line 110: Just wanted to confirm that this max_length will be enough to store document URLs. I know the URLs use UUIDs, and their length shouldn't vary much between documents.

cl/search/state/florida/models.py
```diff
@@ -0,0 +1,115 @@
...
+    entry_type_raw = models.TextField(blank=True)
+    entry_name = models.TextField(blank=True)
+    description = models.TextField(blank=True)
+    submitted_by = models.ForeignKey(
+        "people_db.Party",
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+    )
+    status = models.TextField(blank=True)
```
albertisfu: Will this be populated from the `entry_status` field? IF so it looks like this is a short value. Do you know what different values it can have? Could this be a `SmallIntegerField` with choices? Or, if we're not sure and the values are always short, maybe a regular `CharField` with a small `max_length` would be sufficient?

cl/search/state/florida/models.py
albertisfu: Got it! Yeah, a `submitted_by_name` field makes sense here for these cases.

albertisfu (changes requested):
Thanks, @MorganBennetDev! This looks good. I just have a few additional comments, and I also agree with adding the `submitted_by_name` field.

cl/search/state/florida/models.py
MorganBennetDev: The URL for attachments has two UUIDs in the parameters, which is 72 characters total. The URL and parameter names (with separators) account for 123 characters. The value of the `page` (max 3 digits based on our strategy) and `caseHeader.courtID` parameters (1 digit in practice). This leaves us 51 characters of headroom, which I think should be sufficient.

cl/search/state/florida/models.py
MorganBennetDev: Yes. I think a `CharField` makes more sense (there are a lot of values). Will update this once I find where I put the list of values.

MorganBennetDev: The values from the trial scrape are

```json
[
  "***STRICKEN***",
  "***VACATED***",
  "Docketed"
]
```

Updating `status` to be a `SmallInt` based on this.

cl/search/state/florida/models.py
```diff
@@ -54,7 +66,7 @@ class FloridaDocketEntry(AbstractDateTimeModel, CSVExportMixin):
         blank=True,
     )
     submitted_by_name = models.TextField(blank=True)
-    status = models.CharField(max_length=50, blank=True)
+    status = models.SmallIntegerField(choices=STATUS_CHOICES)
```
albertisfu: Should we set `STATUS_UNKNOWN` as the default value here?

albertisfu (commented):
Thanks @MorganBennetDev just a small change please.

albertisfu (approved):
Thanks, @MorganBennetDev! This looks great now.

We'll just hold off on merging it until the infrastructure team is ready to apply the migrations.

@ERosendo, could you please help us apply the migrations in this PR when you have a chance? Here's the related infrastructure issue:

https://github.com/freelawproject/infrastructure/issues/784

cl/search/state/florida/models.py
```diff
@@ -0,0 +1,130 @@
...
+@document_model
+class FloridaDocument(AbstractDateTimeModel, AbstractPDF):
+    """
+    Represents an attachment to a Florida docket entry.
+
+    :ivar docket_entry: The Docket entry this document is associated with.
+    :ivar content_type: The MIME type indicated by Florida ACIS
+    :ivar document_name: The name of the document in Florida ACIS
+    :ivar document_type: The type of the document in Florida ACIS
+    :ivar description: The description of the document in Florida ACIS
```
albertisfu: Is this field `description` outdated from the docstrings?

MorganBennetDev: I'm not sure I understand the question. The `description` field is still present on this model.

albertisfu: You're right. I only looked at the docstring.

Which value from Juriscraper will this be populated from?

I just reviewed https://github.com/freelawproject/courtlistener/pull/7554 and didn't see this description in the merger, which is why I thought this field was missing.

MorganBennetDev: > Which value from Juriscraper will this be populated from?

It will be populated from the `entry_description` field. #7554 is a bit out of date since I was waiting for reviews/changes on the models to be finished before I updated it.

albertisfu: I saw that `entry_description` is already merged in `FloridaDocketEntryMerger.description`.

Wouldn't it be redundant to also merge it here in the document?

MorganBennetDev: Sorry, I was looking at the wrong model in Juriscraper. Yes, this seems like it can be removed. It's present in my `FloridaDocument` model proposal, but I think that might have been a typo that I then copied into this PR. Removing.

PR #7554, opened 2026-07-06
Mergers for `FloridaDocketEntry` and `FloridaDocument` models

### LINKED ISSUE #7363: Merger for `FloridaDocument` and `FloridaDocketEntry`
Create merger methods for `FloridaDocument` and `FloridaDocketEntry` models.

After #7361 and #7362

MorganBennetDev:
## Fixes

Closes #7363

## Summary

Updates `Merger` to allow abstract bases (i.e. base classes that don't yet have a model defined).

Moves field validation from `MergerSpec.__set_name__` to `Merger.__init_subclass__` to ensure fields are always validated on every model.

Creates merger base classes for Juriscraper `DocketEntry` and `Document` data types.

Creates mergers for `FloridaDocketEntry` and `FloridaDocument` models.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`


cl/lib/model_helpers.py
```diff
@@ -421,6 +422,13 @@ def make_pdf_path(instance, filename, thumbs=False):
         root = Path(f"us/state/tx/{court_id}")
         file_name = f"gov.tx.{court_id}.{slug}{ext}"
         return str(root / file_name)
+    elif isinstance(obj, FloridaDocument):
```
Brennan-Chesley-FLP: I think we want `instance` not `obj` here.

cl/corpus_importer/state/merger.py
```diff
@@ -336,7 +339,10 @@ def merge(
             case ManyStrategy.APPEND:
                 related_manager.add(*related_objects)
             case ManyStrategy.REPLACE:
-                related_manager.set(related_objects)
+                _ = related_manager.exclude(
+                    pk__in=[r.pk for r in related_objects]
```
Brennan-Chesley-FLP: I think the semantics of this one are slightly different than the M2M merger below, where we keep children if their merger failed. Should we unify the semantics or have different strategy names?

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -55,6 +73,74 @@ class FloridaPartyMerger(PartyMerger[FloridaParty, RelatedParams[None]]):
...
+    )
+    docket_entry_uuid: UUID = Attribute(
+        lambda e, params: e.docket_entry_uuid, strategy=overwrite
+    )
+    documents: list[FloridaDocument] = AttachmentRelation(
+        FloridaDocumentMerger
+    )
+
+
+def add_scraper_source(scrape: int | None, db: int | None) -> int:
```
Brennan-Chesley-FLP: I think we deleted this one as duplicate in one of the pre stacked with this (looking forward to GH supporting real stacked prs).

Brennan-Chesley-FLP:
One thing that I think is going to be slightly edge-casey for the merger framework is California and Washington state spreading some of their info across multiple sites, so briefs would come from one scrape and docket info from another. I think there's a relatively clear path forward within this framework for that though. 
One opportunity to delete some code and a couple of small things. It's lookin good.

cl/corpus_importer/state/florida/mergers.py
MorganBennetDev: Stacking PRs is very good except for when I have to merge branches all the way down the chain. Then it's evil.

cl/corpus_importer/state/merger.py
MorganBennetDev: Unified for now

Brennan-Chesley-FLP (approved):

Brennan-Chesley-FLP:
Ausgeseichnet!

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -75,6 +92,66 @@ def resolve_query(self, qs: QuerySet[Party]) -> tuple[bool, Party | None]:
...
+        lambda e, params: e.entry_type_raw, strategy=overwrite
+    )
+    entry_name: str = Attribute(
+        lambda e, params: e.entry_name, strategy=overwrite
+    )
+    description: str = Attribute(
+        lambda e, params: e.entry_description, strategy=overwrite
+    )
+    status: str = Attribute(
+        lambda e, params: e.entry_status, strategy=overwrite
```
albertisfu: This should be updated now that status is a `SmallIntegerField` right?

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -75,6 +92,66 @@ def resolve_query(self, qs: QuerySet[Party]) -> tuple[bool, Party | None]:
...
+    )
+    docket_entry_uuid: UUID = Attribute(
+        lambda e, params: e.docket_entry_uuid, strategy=overwrite
+    )
+    documents: list[FloridaDocument] = AttachmentRelation(
+        FloridaDocumentMerger
+    )
+
```
albertisfu, line 210: Also it looks that we're missing to merge `submitted_by_name` and `submitted_by` fields

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -75,6 +92,66 @@ def resolve_query(self, qs: QuerySet[Party]) -> tuple[bool, Party | None]:
...
+    DocumentMerger[ScrapeFloridaDocument, ParamType, FloridaDocument]
+):
+    model: ClassVar[type[Model]] = FloridaDocument
+    key: ClassVar[Iterable[str]] = ["link_uuid"]
+
+    document_name: str = Attribute(
+        lambda doc, params: doc.document_name, strategy=overwrite
+    )
+    document_type: str = Attribute(_document_type, strategy=overwrite)
+    content_type: str | None = Attribute(lambda doc, params: doc.content_type)
```
albertisfu: Should we also use a pattern like `_document_type` here? As `content_type` from juriscraper can be `str` or  `None`

cl/corpus_importer/state/merger.py
```diff
@@ -337,7 +344,15 @@ def merge(
...
             case ManyStrategy.APPEND:
                 related_manager.add(*related_objects)
-            case ManyStrategy.REPLACE | ManyStrategy.DISASSOCIATE:
+            case ManyStrategy.REPLACE:
+                # Keep existing children whose merge failed, matching the
+                # many-to-many REPLACE semantics below.
+                to_keep = {r.pk for r in related_objects} | {
+                    m.existing.pk for m in child_mergers if m.existing
+                }
+                _ = related_manager.exclude(pk__in=to_keep).delete()
```
albertisfu: Is it possible that `m.existing` is not populated in case of a failure? 

For instance if the merger looks it up in the database, it finds two or more matching rows and can't decide which one to update (`resolve_query()`), or if the validation (`merge()`) of a bad entry fails so it gives up and `m.existing` is never actually set in when a merge fails.

So in that escenario the existing rows for that entry are not on the `keep` list, and the `REPLACE` cleanup deletes them. Which is not the expected behaviour here?

albertisfu (changes requested):
Thanks @MorganBennetDev this looks good! I just have a few suggestions/comments.

cl/corpus_importer/state/merger.py
MorganBennetDev: Updated it so `REPLACE` only deletes extraneous rows if all the child mergers succeeded.

cl/corpus_importer/state/merger.py
```diff
@@ -345,12 +345,9 @@ def merge(
...
             case ManyStrategy.REPLACE:
-                # Keep existing children whose merge failed, matching the
-                # many-to-many REPLACE semantics below.
-                to_keep = {r.pk for r in related_objects} | {
-                    m.existing.pk for m in child_mergers if m.existing
-                }
-                _ = related_manager.exclude(pk__in=to_keep).delete()
+                if all(m.out is not None for m in child_mergers):
+                    to_keep = {r.pk for r in related_objects}
+                    _ = related_manager.exclude(pk__in=to_keep).delete()
```
albertisfu, line 345: I think there’s still one gap, the through-model pruning is triggered only on `through_mergers`, but those mergers are created only for child mergers that succeed.

If a child merger fails, its through merger is never created. The pruning step can still run and delete the through-model row linking the parent to the existing row that the failed child would have matched.

For example, in `test_related_mergers_m2m_replace_keeps_rows_on_ambiguous_child`, the ambiguous child fails, so `through_mergers` is empty. Since `all([])` evaluates to `True`, pruning runs with an empty keep set and deletes `stale_party`’s `PartyType` row. The `Party` itself survives, but it is no longer linked to the docket. The test still passes because it only verifies that the `Party` exists.

It may also be worth updating the test to assert that the `PartyType` row survives.

albertisfu (changes requested):
@MorganBennetDev I just have one more question, please.

MorganBennetDev:
@albertisfu Addressed your comments and got some decent reductions to query count. I've added a utility that prints out the merger query count when tests are run, but I can remove that if you don't think it belongs in the final merged code.

I think the most dubious optimization is `OR`ing all the queries together in `init_many`; I was thinking we could just do `manager.all()` since that already filters relations, but I wasn't sure. It seems like a straight improvement to me since we're going to be fetching almost all associated objects every merge anyway and that way the DB doesn't have to evaluate (many) filters.

I've included the numbers on the improvements below.

<details>

### FloridaDocumentMergerTest

| Test                                              | Before | After | %Change |
| ------------------------------------------------- | ------ | ----- | ------- |
| test_merge_creates_documents                      | 20     | 17    | -15.0%  |
| test_merge_document_without_content_type_is_blank | 20     | 17    | -15.0%  |
| test_merge_document_without_type_is_blank         | 20     | 17    | -15.0%  |
| test_merge_keeps_documents_missing_from_scrape    | 35     | 29    | -17.1%  |
| test_remerge_documents_is_idempotent              | 35     | 29    | -17.1%  |
| total                                             | 130    | 109   | -16.2%  |

### FloridaDocketEntryMergerTest

| Test                                                  | Before | After | %Change |
| ----------------------------------------------------- | ------ | ----- | ------- |
| test_entry_merger_standalone                          | 2      | 2     | 0.0%    |
| test_merge_creates_all_docket_entries                 | 21     | 17    | -19.0%  |
| test_merge_creates_docket_entries                     | 17     | 15    | -11.8%  |
| test_merge_keeps_entries_missing_from_scrape          | 30     | 27    | -10.0%  |
| test_merge_submitted_by_links_docket_party            | 13     | 11    | -15.4%  |
| test_merge_submitted_by_unknown_party_keeps_name_only | 18     | 16    | -11.1%  |
| test_merge_unrecognized_entry_status_is_unknown       | 17     | 15    | -11.8%  |
| test_remerge_entries_is_idempotent                    | 29     | 25    | -13.8%  |
| test_remerge_keeps_resolved_submitted_by_party        | 21     | 18    | -14.3%  |
| test_remerge_updates_entry_fields                     | 29     | 26    | -10.3%  |
| total                                                 | 197    | 172   | -12.7%  |

### FloridaMergerTest

| Test                                                     | Before | After | %Change |
| -------------------------------------------------------- | ------ | ----- | ------- |
| test_merge_creates_new_oci                               | 16     | 14    | -12.5%  |
| test_merge_docket_appellate_creates_new                  | 13     | 13    | 0.0%    |
| test_merge_docket_appellate_disaggregates_existing       | 19     | 18    | -5.3%   |
| test_merge_docket_existing_supreme_court_is_update       | 15     | 14    | -6.7%   |
| test_merge_docket_no_entries_falls_back_to_date_filed    | 15     | 14    | -6.7%   |
| test_merge_docket_supreme_court_creates_new              | 16     | 15    | -6.3%   |
| test_merge_docket_unknown_court_fails                    | 0      | 0     | N/A     |
| test_merge_docket_uses_latest_entry_for_date_last_filing | 22     | 18    | -18.3%  |
| test_merge_multiple_originating_cases_uses_first         | 16     | 14    | -12.5%  |
| test_merge_no_originating_cases_skips_oci                | 14     | 13    | -7.1%   |
| test_merge_skips_non_sc_oci                              | 16     | 15    | -6.3%   |
| test_merge_updates_existing_oci                          | 15     | 14    | -6.7%   |
| test_uuid_matches                                        | 18     | 15    | -16.7%  |
| total                                                    | 195    | 177   | -9.2%   |

### FloridaPartyMergerTest

| Test                                                 | Before | After | %Change |
| ---------------------------------------------------- | ------ | ----- | ------- |
| test_merge_creates_all_parties                       | 16     | 13    | -18.8%  |
| test_merge_creates_party_with_type                   | 12     | 10    | -16.7%  |
| test_merge_does_not_modify_unrelated_parties         | 12     | 10    | -16.7%  |
| test_merge_empty_parties_preserves_existing          | 22     | 18    | -18.2%  |
| test_merge_non_primary_representative_role_unknown   | 17     | 15    | -11.8%  |
| test_merge_party_with_multiple_representatives       | 21     | 18    | -14.3%  |
| test_merge_preserves_unrelated_party_types_and_roles | 17     | 15    | -11.8%  |
| test_merge_primary_representative_is_lead_attorney   | 17     | 15    | -11.8%  |
| test_merge_sets_attorney_name                        | 17     | 15    | -11.8%  |
| test_party_type_change_renames_in_place              | 21     | 17    | -19.0%  |
| test_remerge_is_idempotent                           | 29     | 25    | -13.8%  |
| total                                                | 201    | 171   | -14.9%  |

</details>

cl/corpus_importer/state/tests.py
```diff
@@ -26,7 +27,89 @@
 from cl.tests.cases import TestCase
 
 
-class BaseMergerTest(TestCase):
+class QueryCountTestCase(TestCase):
```
albertisfu: This utility seems very helpful, especially for all the merger tests.

Just one suggestion, instead of printing a report with the number of queries, could we tweak it so it asserts the expected number of queries for each test?

I think we could add a variable either in each test or in setUpTestData to lock down the expected number of queries for each test method. The utility could then compare the actual number of queries against the expected value and fail if it exceeds the limit.

albertisfu (changes requested):
Thanks @MorganBennetDev this looks good. Just a small suggestion regarding `QueryCountTestCase` and potential orphan OCIs.

cl/corpus_importer/state/merger.py
```diff
@@ -245,20 +251,13 @@ def merge(
...
-        if db_obj is None:
-            model = self.merger.model
-            model_name = model.__name__
-            if model_name in result.creates:
-                db_obj_pk = next(iter(result.creates[model_name]))
-                field = parent._meta.get_field(self.name)
-                attname = getattr(field, "attname", f"{self.name}_id")
-                setattr(parent, attname, db_obj_pk)
-                parent.save(update_fields=[attname])
-        return result
```
albertisfu, line 261: It looks like the removal of this block introduces a bug where a new OCI is created but never linked to an existing docket, leaving it orphaned.

With this new code, when the case and the OCI are new the fk relation `docket.originating_court_information_id` is properly set in `create_new`.

However the problem arises when the parent already exists and the OCI is new (e.g: a re-scrape that adds the OCI data). In `update_existing` it calls `o2o_spec.merge(self.existing, ...),` which creates the OCI but never sets the fk back on the docket (this old block did that here with `setattr`). So the OCI is created but never linked, and since the fk is never persisted, every re-scrape would create another orphan.

`test_merge_creates_new_oci` doesn't catch it because it only asserts the OCI was created, not that it's linked to the docket.

MorganBennetDev:
@albertisfu Updated to assert query count instead of printing and fixed the regression you found.

albertisfu (approved):
Thanks, @MorganBennetDev, for all your work here! It's great to have the query count available as a `merger_test` decorator. It'll be very helpful for future merger work.

This is ready to be merged!

PR #7559, opened 2026-07-08
feat(florida): Targeted backscrapes

### LINKED ISSUE #7553: Update Florida backscrape command for targeted rescrapes
In #7513, we found some cases that failed with the current parser. These issues were fixed in https://github.com/freelawproject/juriscraper/pull/2033. We now need a way to rerun the scraper on specific cases to remap `UNASSIGNED` court IDs and reparse the pages we failed to.

What we need:

- A flag to have the scraper command skip cases we already have parsed entries for.
- A way to rerun the scraper for a specific case, set of cases, or page of results.

MorganBennetDev:
## Fixes

Fixes #7553

## Summary

Adds `--skip-parsed` and `--scrape-uuids` options to the Florida backscrape command to allow targeted rescraping of failed cases.

The `--skip-parsed` option will modify the existing `_backfill` loop to first check if a parsed file exists in S3, and if so, skip fetching full and parsing full case data. This is intended to help with broad rescrapes for cases that failed to parse in a given date range.

The `--scrape-uuids` option takes the path to a CSV file of Florida court IDs and case UUIDs. It then fetches full case data for each of these cases and saves them to S3.

Once the full Florida scrape is complete and we have any necessary fixes in a new Juriscraper version, we will run the command using these options to fill in any gaps caused by parsing or network failures.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/scrapers/management/commands/back_scrape_fl_dockets.py
+++ b/cl/scrapers/management/commands/back_scrape_fl_dockets.py
@@ -6,12 +6,14 @@ from dataclasses import dataclass, field
 from datetime import date
 from functools import lru_cache
 from pathlib import Path
+from typing import ClassVar
+from uuid import UUID
 
 import botocore.exceptions
 from asgiref.sync import async_to_sync
 from django.core.management import CommandParser
 from httpx import Response
-from juriscraper.state.florida import FloridaScraper
+from juriscraper.state.florida import FloridaCase, FloridaScraper
 from juriscraper.state.florida.courts import FloridaCourtID
 from juriscraper.state.RequestManager import (
     ExponentialBackoff,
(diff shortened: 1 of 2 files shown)


cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -157,26 +159,92 @@ def __hash__(self) -> int:
...
+    logger.info("Starting targeted backfill of %d cases...", len(cases))
+    for i, (court_id, case_uuid) in enumerate(cases):
+        logger.info("Fetching case %s for %s", case_uuid, court_id)
+        case = await _get_full_case(court_id.value, str(case_uuid), scraper)
+        if case is None:
+            continue
+
+        content = case.model_dump_json(ensure_ascii=True).encode("utf-8")
+
+        key = f"{S3_BASE}/parsed/{court_id.value}/{_make_case_number_key(case.docket_number)}.json"
```
albertisfu: This line is used twice in this file.

I suggest creating a helper method that receives `court_id` and `docket_number` so the key can be computed consistently in both parts of the code.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -157,26 +159,92 @@ def __hash__(self) -> int:
...
+
+        throttle.maybe_wait()
+        save_response_to_s3.si(key, content).set(
+            queue=queue_name
+        ).apply_async()
+
+        if i % 100 == 0:
```
albertisfu: Could you please reduce this number to 10? Since there might not be too many missing cases, the total number of documents to scrape would also be small. If we log only every 100 entries and the scraper breaks, we might not be able to determine the last successfully scraped case and resume from there, since `_backfill_targeted` does not support automatic resumption.

albertisfu (changes requested):
Thanks @MorganBennetDev just minor changes please.

albertisfu (approved):
Thanks! This is ready to be merged :shipit:

PR #7566, opened 2026-07-10
feat(florida): Basic poller daemon

### LINKED ISSUE #7433: Poll the Florida docketentrydocuments endpoint to keep cases up to date.
### Describe the issue.

In https://github.com/freelawproject/courtlistener/issues/6831 we decided to poll Florida `docketentrydocuments` endpoint as a strategy to keep existing cases up to date.

For this sprint, we should only draft a plan for this scraper, including the polling rate, rate limits, and how it will interact with other scrapers (such as the ongoing new-cases scraper) to stay within the rate limits. We should also define the storage and extraction pipeline, along with any other important details needed to develop it.

### Related Issue

_No response_

### Screenshots

_No response_

To Do

MorganBennetDev:
## Fixes

#7433

## Summary

Implements a polling daemon to keep Florida data up to date using the `docketentrydocuments` and case search endpoints. Submitting the implementation for review now without the merger implementation since that doesn't affect the functionality of this (yay pipelining!).

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [ ] `skip-daemon-deploy`


cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def send_merge_task(
+        self,
+        court_id: FloridaCourtID,
+        case: FloridaCase,
+        throttle: CeleryThrottle,
+        queue_name: str,
+    ):
+        # TODO: Spawn merge task here when that's ready
+        logger.error("Florida merge task not implemented")
```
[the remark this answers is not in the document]
MorganBennetDev: This is intentional. This should not be merged or run in production until this method is implemented.

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,255 @@
...
+                    ),
+                    "caseHeader.filedDateFrom": start.strftime(
+                        "%Y-%m-%dT%H:%M:%S.000Z"
+                    ),
+                    "sort": "caseHeader.filedDate,desc",
+                },
+            ):
+                match page:
+                    case PaginationFailed() as e:
+                        logger.error("Failed to get page: %r", e)
```
Brennan-Chesley-FLP, line 271: Confirming the plan for this is `Sentry alert -> Targeted backscrape`?

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,255 @@
...
+                    continue
+                seen.add(update)
+                court_id = external_id_map[update.court_external_id]
+                try:
+                    maybe_case = await scraper.fetch_case_data(
+                        str(update.case_uuid),
+                        court_id.value,
+                    )
+                except Exception as e:
+                    logger.error(
```
Brennan-Chesley-FLP: If we use `logger.exception` here (and elsewhere where we have an exception) I believe that Sentry will get more reliable traceback data for us to peruse.

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,255 @@
...
+        polling_delay: int,
+        case_backfill_days: int,
+        queue_name: str,
+    ):
+        scraper_courts = await scraper.courts
+        external_id_map = {
+            metadata.court.external_identifier: court
+            for court, metadata in scraper_courts.items()
+        }
+        start = datetime.now(UTC) - timedelta(days=case_backfill_days)
```
Brennan-Chesley-FLP: There's a case to be made (and Alberto might make it) that we would benefit from storing the last successful poll date in Redis and resuming from that instead of `now-bfd`. I think we're probably operating with enough rps headroom that just doing targeted backfills (so long as we're doing doc update search there as well) in the event of downtime is likely simpler. I'm considering the case where k8s bounces the poller over a 3day weekend here.

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,255 @@
...
+        for court_id in courts:
+            parser = FloridaDocumentPollParser(court_id=court_id.value)
+            async for page in scraper._enumerate_pages(
+                DOCKET_ENDPOINT,
+                parser,
+                {
+                    "caseHeader.courtID": str(
+                        scraper_courts[court_id].court.resource_id
+                    ),
+                    "caseHeader.filedDateFrom": start.strftime(
```
Brennan-Chesley-FLP, line 263: I think we should befriend someone on one of the 6 CTrack courts and see if they can give us the docs to answer questions like "is backdating filing dates possible".

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,255 @@
...
+                        "%Y-%m-%dT%H:%M:%S.000Z"
```
Brennan-Chesley-FLP, line 264: Glad they're only doing date clipping on the UI. Have you seen times of day when updates are clustered?

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,255 @@
...
+        logger.info(
+            "Checking docketentrydocuments endpoint for new entries since %s",
+            start,
+        )
+        for court_id in courts:
+            async for page in scraper._enumerate_pages(
+                DE_DOC_ENDPOINT,
+                parser,
+                {
+                    "sort": "docketEntryHeader.filedDate,desc",
```
Brennan-Chesley-FLP: I think there might be some small advantage in having our twin searches sort by date asc so if there is a failure, our backfill window is a little easier to reason about.

Brennan-Chesley-FLP:
The refactor is nice. I have a couple of operational questions that might suggest some small changes, but other than those, once the merger code is rolled in I'll give this the stamp.

cl/scrapers/management/commands/fl_poller.py
MorganBennetDev: Correct

cl/scrapers/management/commands/fl_poller.py
MorganBennetDev: I didn't look into this with my trial scrape, but we can certainly check that once the full scrape is complete!

cl/scrapers/management/commands/fl_poller.py
MorganBennetDev: I think this would be easy enough to solve by just running once with `--case-backfill-days=4`. The only scenario where that would fail is if the documents or dockets endpoint ended up with over 10,000 new entries in a single court over a 3-day weekend, which seems unlikely. I'll leave it up to Alberto's review to make the final decision on whether having Redis as a safety is worth it.

cl/scrapers/management/commands/fl_poller.py
MorganBennetDev: I hope to whatever's out there that it isn't.

cl/scrapers/management/commands/fl_poller.py
MorganBennetDev: Done

cl/scrapers/management/commands/fl_poller.py
MorganBennetDev: Done

MorganBennetDev:
@Brennan-Chesley-FLP Import task has been integrated. Review of that is happening one PR down the stack.

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def send_merge_task(
+        self,
+        case: FloridaCase,
+        throttle: CeleryThrottle,
+        queue_name: str,
+        download_attachments: bool,
+    ):
+        throttle.maybe_wait()
+        fl_ingest_docket_task.si(case, download_attachments).set(
```
albertisfu: This task expects `download_result` which is a 3-tuple and here we're just passing the case.

It expects the serialized case, the bucket and the  s3 key.

cl/scrapers/management/utils.py
```diff
@@ -103,5 +123,249 @@ def add_arguments(self, parser: CommandParser):
...
+    "fladistctapp2": FloridaCourtID.SECOND_COA,
+    "fladistctapp3": FloridaCourtID.THIRD_COA,
+    "fladistctapp4": FloridaCourtID.FOURTH_COA,
+    "fladistctapp5": FloridaCourtID.FIFTH_COA,
+    "fladistctapp6": FloridaCourtID.SIXTH_COA,
+}
+
+
+class FLScrapeCommand(BaseCommand):
+    def add_arguments(self, parser: CommandParser):
```
albertisfu, line 168: Some of the arguments defined here are already defined in `back_scrape_fl_dockets.Command` so this is breaking the `back_scrape_fl_dockets` on initialization:

`argparse.ArgumentError: argument --archive-responses: conflicting option string: --archive-responses`

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -380,38 +226,29 @@ def handle(
...
-            rps=rps,
-            retry=ExponentialBackoff(
-                max_retries=max_retries,
-                backoff=backoff,
-                backoff_growth=backoff_growth,
-            ),
-            handlers=[cache],
+        logger.info("Setting up Florida back-scrape...")
+        start = _parse_date(backscrape_start)
+        end = _parse_date(backscrape_end)
```
albertisfu: These are breaking the `scrape_uuids` mode. As there are no dates to parse when running a targeted scrape.

cl/scrapers/management/commands/fl_poller.py
albertisfu: I agree with Brennan that it’ll be more reliable to persist the poller checkpoint in Redis using ScraperCheckpointTracker on every cycle, so in case of a multi-day failure, it can recover automatically. Otherwise, it’ll require human intervention to restart the command.

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+class Command(FLScrapeCommand, StatePollCommand):
+    help = "Continuously polls Florida ACIS for new cases and backfills when new entries are detected."
+
+    def add_arguments(self, parser: CommandParser):
+        super().add_arguments(parser)
+
+        parser.add_argument(
+            "--download-attachments",
```
albertisfu: In case we want to run the poller without downloading attachments, it’s currently not possible to disable attachment downloads because this defaults to True, and evaluating any non-empty string, such as `bool("False")`, will always return True.


What about downloading attachments by default and only disabling it if the `--no-download-attachments` flag is passed?

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -37,130 +26,20 @@
...
-    "fla": FloridaCourtID.SUPREME_COURT,
-    "fladistctapp1": FloridaCourtID.FIRST_COA,
-    "fladistctapp2": FloridaCourtID.SECOND_COA,
-    "fladistctapp3": FloridaCourtID.THIRD_COA,
-    "fladistctapp4": FloridaCourtID.FOURTH_COA,
-    "fladistctapp5": FloridaCourtID.FIFTH_COA,
-    "fladistctapp6": FloridaCourtID.SIXTH_COA,
-}
+def _make_case_key(court_id: FloridaCourtID, docket_number: str) -> str:
+    return f"{S3_BASE}/parsed/{court_id.value}/{_make_case_number_key(docket_number)}.html"
```
albertisfu: Shouldn't be .json?

```suggestion
    return f"{S3_BASE}/parsed/{court_id.value}/{_make_case_number_key(docket_number)}.json"
```

albertisfu (changes requested):
Thanks, @MorganBennetDev. This looks good overall. I only found a few issues and have some suggestions.

I didn't run a real poll because the command isn't working right now. So, I suggest doing a real test to confirm that the full workflow behaves as expected. I also suggest adding a few tests to cover the basic functionality of the poller, like the ones for SCOTUS in `test_scotus_daemon.py`.

Finally, we need to add the poller to `docker-entrypoint.sh`.

cl/scrapers/management/commands/fl_poller.py
MorganBennetDev: This got messed up in a merge. Fixing.

MorganBennetDev:
@albertisfu Fixed the issues and added some tests.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
```diff
@@ -327,12 +178,6 @@ def add_arguments(self, parser: CommandParser):
...
             help="If set the scraper will use the cache to avoid re-downloading files. Useful when running a second pass with the --full-scrape option.",
         )
-        parser.add_argument(
-            "--archive-responses",
-            action="store_true",
-            default=False,
-            help="If set the scraper will archive responses to S3.",
-        )
         parser.add_argument(
             "--queue",
```
albertisfu: This argument is still duplicated here (it's already in FLScrapeCommand) and it's breaking this command.

cl/scrapers/management/commands/back_scrape_fl_dockets.py
albertisfu: Also this one duplicated.

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,305 @@
...
+    def send_merge_task(
+        self,
+        case: FloridaCase,
+        throttle: CeleryThrottle,
+        key: str,
+        queue_name: str,
+        download_attachments: bool,
+    ):
+        throttle.maybe_wait()
+        self.checkpoint_tracker.set(case.date_filed)
```
albertisfu: What if the scraped document is related to an old case? If we set the checkpoint based on `case.date_filed`, we could end up storing an old date that is very far from the current date. What if, instead, we set it to the day the scraper runs, using local time?

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,305 @@
...
+            type=bool,
+            default=False,
+            help="Set to download attachments for newly discovered dockets.",
+        )
+        parser.add_argument(
+            "--auto-resume",
+            type=bool,
+            default=False,
+            help="Set to resume polling from the last saved checkpoint.",
+        )
```
albertisfu, line 76: These two arguments are bool type so they expect a value to work. 

We should remove the type and only add:
`action="store_true",`

So passing the flag is enough.

albertisfu (changes requested):
Thanks for the changes @MorganBennetDev they look good. I just have a couple of extra comments.

Also, I wanted to confirm how it would work when the poller downloads attachments.

It looks like `AbstractStateDocument.download` will always download the file, even if it's already present. Can we confirm that we only download the document if there is no document already or if the document UUID has changed?

MorganBennetDev:
@albertisfu Fixed the backfill argument issue, added tests to make sure we only download updated/new documents, moved the checkpoint to the end of the loop, and fixed a typing issue in the TAMES poller.

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,307 @@
...
+                court_id = external_id_map[update.court_external_id]
+                try:
+                    maybe_case = await scraper.fetch_case_data(
+                        str(update.case_uuid),
+                        court_id.value,
+                    )
+                except Exception:
+                    logger.exception(
+                        "Failed to fetch case data for %s",
+                        update.case_uuid,
```
albertisfu: Can you please also add the `court_id` to this log so that, in case of a failure, it's easy to identify which case failed and use a targeted backscrape to fetch those cases?

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,307 @@
...
+                    )
+                    continue
+                if isinstance(maybe_case, Exception):
+                    logger.error(
+                        "Failed to fetch case data for %s: %r",
+                        update.case_uuid,
+                        maybe_case,
```
albertisfu: Also here `court_id` please

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,307 @@
...
+                    )
+                    continue
+                case, errors = maybe_case
+                if errors:
+                    logger.error(
+                        "Failed to fetch case data for %s: %r",
+                        update.case_uuid,
+                        errors,
```
albertisfu: And here too `court_id`

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,307 @@
...
+                        scraper_courts[court_id].court.resource_id
+                    ),
+                    "docketEntryHeader.docketEntryFiledDateFrom": start.strftime(
+                        "%Y-%m-%dT%H:%M:%S.000Z"
+                    ),
+                },
+            ):
+                match page:
+                    case PaginationFailed() as e:
+                        logger.error("Failed to get page: %r", e)
```
albertisfu: `court_id` also here please.

docker/django/docker-entrypoint.sh
```diff
@@ -54,6 +54,9 @@ case "$1" in
 'tames-scraper-daemon')
     exec ./manage.py tames_poller
     ;;
+'fl-scraper-daemon')
+    exec ./manage.py fl_poller
```
albertisfu: we need to pass the auto-resume flag here so it can automatically recover in a failure

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,307 @@
...
+            # Add a little overlap between segments to make extra sure we don't miss anything
+            now = datetime.now(UTC) - timedelta(minutes=1)
+            logger.info(
+                "Looking for new and updated cases from %s to %s",
+                last_polled,
+                now,
+            )
+            seen = set()
+            async for update in self.gather_all(
+                scraper, scraper_courts, courts, last_polled
```
albertisfu: Would it be possible for cases or documents to be published later than their `date_filed` date?

For example, could they appear in search a few days later, causing us to miss those cases or documents?

To be cautious, I imagine we could do this:

* Use `last_polled - case_backfill_days`, so on every cycle we also look back a few days to ensure we don't miss anything that was published later.
* Tweak `gather_updated` and `gather_new` to only scrape cases that are not in the DB, to prevent hitting the endpoint when it's not necessary (see suggestions in those methods)

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,307 @@
...
+                    ),
+                    "sort": "caseHeader.filedDate,asc",
+                },
+            ):
+                match page:
+                    case PaginationFailed() as e:
+                        logger.error("Failed to get page: %r", e)
+                    case FloridaPaginatedResults(results=results):
+                        for result in results:
+                            yield result
```
albertisfu: We can rely on the DB to only return cases that haven't been merged.

```suggestion
                        known = await sync_to_async(set)(
                            Docket.objects.filter(
                                court_id__in=FLORIDA_CL_COURT_IDS,
                                pacer_case_id__in=[str(u.case_uuid) for u in results],
                            ).values_list("pacer_case_id", flat=True)
                        )
                        for result in results:
                            if str(result.case_uuid) not in known:
                                yield result
```

cl/scrapers/management/commands/fl_poller.py
```diff
@@ -0,0 +1,307 @@
...
+                    "docketEntryHeader.docketEntryFiledDateFrom": start.strftime(
+                        "%Y-%m-%dT%H:%M:%S.000Z"
+                    ),
+                },
+            ):
+                match page:
+                    case PaginationFailed() as e:
+                        logger.error("Failed to get page: %r", e)
+                    case FloridaPaginatedResults(results=results):
+                        for result in results:
```
albertisfu: We can do something similar as suggested in `gather_new` but in this case targeting documents:

```
known = await sync_to_async(set)(
      FloridaDocument.objects.filter(
          link_uuid__in=[r.document_link_uuid for r in results]
      ).values_list("link_uuid", flat=True)
  )
```

Good news is that we can rely on the `unique_link_uuid_per_docket_entry` constraint for this query as the first column is `link_uuid` so the planner can relay on it to make queries faster.

cl/corpus_importer/state/florida/tests.py
```diff
@@ -1419,6 +1420,66 @@ def test_ingest_skips_downloads_when_disabled(
...
+        old_document.save()
+        download_mock.reset_mock()
+
+        case.entries[0].attachments[0].document_link_uuid = uuid4()
+        result = fl_ingest_docket_task((case, "bucket", "key"))
+
+        self.assertTrue(result.success)
+        new_document = FloridaDocument.objects.exclude(
+            pk=old_document.pk
+        ).get()
```
albertisfu: Is this the expected behavior? If the `link_uuid` changes on a document, is it considered a completely new document?

I'm thinking about the difference between a new document being added to an entry and a PDF being replaced in one of the existing documents.

I don't think there's a reliable way to differentiate between the two cases. I looked for potentially problematic examples like this one:

https://acis.flcourts.gov/portal/court/8a73be7f-864b-42a9-b8ad-1ec0a3e890b4/case/a9af1990-bd86-4755-99c1-e12a924b8c52

See the first entry:

`10/26/2023 | Notice | Notice of Appearance`

Both documents there have the same name, but they're different documents.

I was wondering what we can do to prevent duplicate documents if the second case is possible (an existing document being replaced).

albertisfu (changes requested):
Thanks @MorganBennetDev I have a few more suggestions, please.

Also one question. Is this command compatible with `--archive-responses` so we can store the raw responses too? Also I'm not sure if that would be worthwhile. For instance how hard it'll be if later we found a parsing issue and we want to re-ingest those cases. Would it be easy to do it from those raw responses?

Issue #7077, opened 2026-03-04
Allow many-to-many link between dockets and opinions


rachlllg:
### Describe the issue.

In principle, court `cases` start from the lowest court, as it goes through the appeal process, `dockets` get created  by the courts and the courts issue `opinions` to express its decisions. Both `dockets` and `opinions` are children of a `case`.  
- One `case` can have many `dockets`.
- One `docket` can have many `opinions`.
- One `opinion` can be related to many `dockets`.

So `dockets` and `opinions` really should have a many-to-many relationship, which is not the case currently. See the comments in #7034 for more details. 

By enabling the many-to-many relationship, we can link `dockets` and `opinions` to one another, and link the appellate chain as each `case` goes through the appeal process.

### What type of issue is this?

Research

grossir:
I think this may be related to opinions on "consolidated" cases, when an opinion is issued that applies to more than 1 docket number.
We see it in
- PACER / RECAP: https://github.com/freelawproject/courtlistener/issues/6179#issuecomment-3211135773
- juriscraper (see below)

<details>
<summary>expand for examples from the courts we scrape</summary>
  Separators Found                                                                                                                                                           
                                                                                                                                                                             
  Comma-separated                                                                                                                                                            
                                                                                                                                                                             
  - asbca — up to 21 numbers: "59419, 59420, ..., 61319"                                                                                                                     
  - conn — "SC20763, SC20764, SC20765"                                                                                                                                       
  - connappct — "AC44836, AC45267"                                                                                                                                           
  - ohioctapp_{3,4,5,6,7,8,9,11} — many instances across districts
  - kyctapp — "2022-CA-1454, 1456, 1457"                                                                                                                                     
  - njtaxct_p / njtaxct_u — "006396-2016, 07619-2016, 004282-2017, ..."                                                                                                      
  - nd — "20240010, 20240009"                                                                                                                                                
  - ca8 — "24-2330, 24-2526"                                                                                                                                                 
  - wisctapp — "2023AP000078, 2023AP001339"                                                                                                                                  
  - ca2 oral_args — "25-148, 25-237"                                                                                                                                         
                                                                                                                                                                             
  Semicolon-separated                                                                                                                                                        
                                                                                                                                                                             
  - ohioctapp_{1,2,5,7,8,12} — e.g. "CA2021-06-017; CA2021-06-018"
  - masslandct — "MISC 15-000302; MISC 15-000545"                                                                                                                            
[...]

PR #7567, opened 2026-07-10
Florida `CaseTransfer` merger

### LINKED ISSUE #7365: Florida `CaseTransfer` merging
Create logic to merge `CaseTransfer` data for Florida dockets. Should be relatively straightforward due to the shape of Florida scraper output.

MorganBennetDev:
## Fixes

#7365

## Summary

Adds base classes for general merging of `CaseTransfer`s and implements specific instances for Florida.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`


cl/corpus_importer/state/common/case_transfer.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def _destination_docket_number(scrape: Any, params: RelatedParams[Any]) -> str:
+    return cast(Docket, params.parent).docket_number or ""
+
+
+def _transfer_date(scrape: Any, params: RelatedParams[Any]) -> date | None:
+    # The standard docket format doesn't include transfer dates, so treat the
+    # destination docket's filing date as the date the transfer occurred.
+    return cast(Docket, params.parent).date_filed
```
Brennan-Chesley-FLP, line 107: Creating [a link here](https://github.com/freelawproject/courtlistener/issues/7077) for future discussion.

cl/corpus_importer/state/common/case_transfer.py
```diff
@@ -0,0 +1,179 @@
...
+        Prefer a partial transfer missing its destination docket FK so the FK
+        can be filled in (see `update_existing`); otherwise match a transfer
+        already attached to the parent docket.
+
+        :return: The queryset to find the transfer."""
+        candidates = cast(
+            QuerySet[CaseTransfer],
+            CaseTransfer.objects.filter(
+                **{name: self.transformed[name] for name in self.key}
+            ),
```
Brennan-Chesley-FLP, line 167: Flagging this as a potential indexing optimization target.

cl/corpus_importer/state/common/case_transfer.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+        fillable = candidates.filter(destination_docket__isnull=True)
+        if fillable.exists():
+            return fillable
+        return candidates.filter(destination_docket=self.params.parent)
```
Brennan-Chesley-FLP, line 172: If we find we need some performance gains here, this seems like one possibility.

cl/corpus_importer/state/common/case_transfer.py
```diff
@@ -0,0 +1,179 @@
...
+    define the ``origin_court_id`` attribute to translate the transfer's
+    Juriscraper court ID to a CourtListener one."""
+
+    model: ClassVar[type[Model]] = CaseTransfer
+    key: ClassVar[Iterable[str]] = [
+        "origin_court_id",
+        "origin_docket_number",
+        "destination_court_id",
+        "destination_docket_number",
+        "transfer_date",
```
Brennan-Chesley-FLP: It seems quite unlikely to me, but noting this seems the most likely to get changed under us by the court and lead to duplicates. Have you seen cases where this is necessary to disambiguate between transfers?

cl/corpus_importer/state/florida/mergers.py
```diff
@@ -200,6 +212,45 @@ def _originating_case(
...
+    for transfer in transfers:
+        court_id = FLORIDA_TRANSFER_COURT_ID_MAP.get(transfer.court_id)
+        if court_id is None:
+            logger.info(
+                "Skipping CaseTransfer for Florida docket %s: no matching court for Juriscraper court %s",
+                docket_data.docket_number,
+                transfer.court_id,
+            )
+            continue
+        if not Court.objects.filter(pk=court_id).exists():
```
Brennan-Chesley-FLP, line 311: Flagging this as a potential site for caching if we find this to be slow in prod.

Brennan-Chesley-FLP (approved):
Solid addition. One question about the CaseTransfer key that isn't a blocker.

cl/corpus_importer/state/common/case_transfer.py
MorganBennetDev: This is just here to match the pattern from Texas for now. It's entirely possible that date of appeal/transfer is not a reliable thing to match on, but I haven't investigated this and we don't have any logging to check.

albertisfu: I agree with Brennan, and it seems we can remove `transfer_date` from the lookup. Right now, we’re taking the `transfer_date` from the case’s `date_filed`, correct?

It’s possible for the case’s filing date to be corrected while the `CaseTransfer` remains the same. Therefore, using it in the lookup could lead to duplicate records. Instead maybe we can use `strategy=overwrite` for `transfer_date` ?

The other fields, `origin_court_id`, `origin_docket_number`, `destination_court_id`, and `destination_docket_number`, should already be enough to match the correct `CaseTransfer`. It seems unlikely that a case could have more than one transfer involving the same origin and destination cases?

If we agree on this maybe we can also tweak it in Texas?

cl/corpus_importer/state/common/case_transfer.py
```diff
@@ -0,0 +1,179 @@
...
+    number and transfer type from the scraped transfer; subclasses must
+    define the ``origin_court_id`` attribute to translate the transfer's
+    Juriscraper court ID to a CourtListener one."""
+
+    model: ClassVar[type[Model]] = CaseTransfer
+    key: ClassVar[Iterable[str]] = [
+        "origin_court_id",
+        "origin_docket_number",
+        "destination_court_id",
+        "destination_docket_number",
```
albertisfu, line 130: One thing I'm realizing right now is that we don't have indexes on `origin_docket_number` and `destination_docket_number` which is probably fine, but because of that this query can become slow once `CaseTransfer` has millions of records.

Would it be possible to instead filter by either `destination_docket_id` or `origin_docket_id`, since those foreign keys already have a built-in index? Since we know the direction of the `CaseTransfer` being merged, maybe we can do something like this:

```
CaseTransfer.objects.filter(
    destination_docket_id=current_docket.id,
    origin_court_id=transfer.origin_court_id,
    origin_docket_number=transfer.origin_docket_number,
)
```

Note that I'm not including `destination_court_id`, as it's implicit in `current_docket.court_id` given that it was just merged.

albertisfu (changes requested):
Thanks @MorganBennetDev this looks great just a few suggestions in code and some questions here:

- Just confirming: does the `transfers` data come from the same `originating_cases` source, or can it be different in some courts?

The difference when merging `originating_cases` is that we only select the first entry, whereas with `transfers` we can merge all of the entries using `CaseTransfer` instances, correct?

- Another question is if we are we planning to use `CaseTransfer._fill_null_docket_side` to fill in null dockets once we merge them? If so, it looks like we'll need to review `_fill_null_docket_side` to make it compatible with Florida,  for instance, the line that says:

```
if court.jurisdiction == Court.STATE_APPELLATE:
                docket_number = normalize_texas_appellate_docket_number(
                    docket_number
                )
```

MorganBennetDev:
> Just confirming: does the transfers data come from the same `originating_cases` source, or can it be different in some courts?

I have not encountered any Florida courts/cases where it does not.

> The difference when merging `originating_cases` is that we only select the first entry, whereas with transfers we can merge all of the entries using `CaseTransfer` instances, correct?

Correct

> Another question is if we are we planning to use `CaseTransfer._fill_null_docket_side` to fill in null dockets once we merge them? If so, it looks like we'll need to review `_fill_null_docket_side` to make it compatible with Florida

Yes. Should be fixed now

albertisfu (changes requested):
Thanks, @MorganBennetDev! Just a couple of things, please:

Why was this test removed?

`test_merge_fills_existing_partial_transfer`

Can't it be preserved with the change to using `destination_docket_id`?

Also, there's a merge conflict that needs to be resolved.

MorganBennetDev:
> Why was this test removed?

I was a bit tired when I addressed the feedback and my fix broke the test, but I thought it was because the test wasn't necessary anymore. I've had Claude add it back.

cl/corpus_importer/state/common/case_transfer.py
MorganBennetDev: The `query` method does this filtering to see if there's an existing transfer it can work with. The presence of `key` and `query` is a bit confusing, but I'm not sure how best to handle that.

albertisfu (approved):
Thanks, @MorganBennetDev! This is ready to be merged. I’ll hold off until the base PR is merged.

albertisfu:
@MorganBennetDev could you please review the mypy complaints that raised after I updated this PR? https://github.com/freelawproject/courtlistener/actions/runs/31045122045/job/92438633169?pr=7567

Also tests needs to be updated.

MorganBennetDev:
@albertisfu Fixed the mypy issues and added merger query count assertions for case transfers.

PR #7568, opened 2026-07-10
feat(citations): propagate queue argument to recap document citation extraction for batch runs

### LINKED ISSUE #7565: Update `find_citations_and_parantheticals_for_recap_documents` for batch runs
While running this [script](https://github.com/freelawproject/courtlistener/pull/7417#issuecomment-4695775298) we found that the citation extraction process was filling the default celery queue, delaying regular jobs.

We should find a way to propagate the queue value so we can control where costy / slow tasks such as these are run. This will also be useful for the upcoming citation extraction re-run

grossir:
Solves #7565

- Read citation_queue off the RECAPDocument instance in handle_recap_doc_change, defaulting to celery, so signal-triggered citation tasks are queue-controllable.
- Thread a citation_queue kwarg through the extraction tasks and set it on the doc before save.
- Pass the batch queue from ocr_available and the get_pdfs chain (get_and_process_free_pdf).
- Update signal tests for the queue kwarg and add an override case.

### DIFF
--- a/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
+++ b/cl/corpus_importer/management/commands/scrape_pacer_free_opinions.py
@@ -484,7 +484,7 @@ def get_pdfs(
                 court_id,
                 cnt,
             ).set(queue=q),
-            get_and_process_free_pdf.s(pk, court_id).set(queue=q),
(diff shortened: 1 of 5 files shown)


quevon24 (approved):
LGTM. This will help to use a different queue than the default in the other processes para la extraccion.

PR #7573, opened 2026-07-13
feat(alerts): Adds progress logging to scheduled alert sends

ERosendo:
## Fixes

Relates to https://github.com/freelawproject/infrastructure/issues/750 but it doesn't fully fix it.

## Summary

The `cl_send_scheduled_alerts` command has been running out of memory, but it produced almost no logs to debug with it , only logged two totals at the very end of a run. When the process was killed mid-run, there was nothing to point at which user or which phase caused the growth.

This PR adds progress logging throughout `query_and_send_alerts_by_rate` and `delete_old_scheduled_alerts`:

- A startup line with the alert rate and the total number of users to process.
- A per-user line with the count of scheduled hits being loaded( a single user with a large backlog of scheduled hits, each storing a full document JSON in `document_content`, is the most likely source of the spike)
- A per-user line with the number of alerts and total documents queued into each Celery task, right before `send_search_alert_emails.delay`, so the size of the payload being serialized is visible.
- Markers before each of the two cleanup deletes, to separate the send phase from the deletion phase during debugging.

This is logging only — no behavior changes. The actual memory reductions (capping queued hits to `SCHEDULED_ALERT_HITS_LIMIT` and batching the bulk deletes) are intended as a follow-up so these diagnostics can be observed in production first.

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/alerts/management/commands/cl_send_scheduled_alerts.py
+++ b/cl/alerts/management/commands/cl_send_scheduled_alerts.py
@@ -115,13 +115,14 @@ def query_and_send_alerts_by_rate(rate: str) -> None:
     alerts_sent_count = 0
     now_time = datetime.now()
     # Get unique alert users with scheduled alert hits
-    user_ids = (
+    user_ids = list(
         ScheduledAlertHit.objects.filter(
             alert__rate=rate, hit_status=SCHEDULED_ALERT_HIT_STATUS.SCHEDULED
         )
         .values_list("user", flat=True)
         .distinct()
     )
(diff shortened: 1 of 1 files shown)


albertisfu (approved):
Thanks @ERosendo this looks good. Merging...

PR #7577, opened 2026-07-14
perf(alerts): batch scheduled-alert cleanup deletes to bound memory

ERosendo:
## Fixes

Fixes https://github.com/freelawproject/infrastructure/issues/750

## Summary

Follow-up PR to https://github.com/freelawproject/courtlistener/pull/7573. 

The additional logging confirmed the root cause: in production, the pod is terminated immediately after logging `Deleting SENT hits older than ...` and never reaches the `Removed N Scheduled Alert Hits` message, so the memory blows up inside `delete_old_scheduled_alerts`, not in the per-user send loop.

The command cleaned up the entire 90+ day backlog of `ScheduledAlertHit` rows in two single `.delete()` calls. Django loads every matching row (plus cascade bookkeeping) into an in-memory collector before issuing the DELETE, so on a large backlog that one call exhausts memory.

This PR adds `delete_scheduled_hits_in_batches()`, which deletes at most `DELETE_BATCH_SIZE` (5,000) primary keys per pass and re-queries until the filter is empty. Peak memory is now bounded by the batch size instead of the total number of rows to delete. The returned total count and the overall cleanup behavior are unchanged.

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [ ] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/alerts/management/commands/cl_send_scheduled_alerts.py
+++ b/cl/alerts/management/commands/cl_send_scheduled_alerts.py
@@ -4,6 +4,7 @@ from typing import Any
 
 import pytz
 from django.conf import settings
+from django.db.models import QuerySet
 from django.utils.timezone import get_default_timezone, make_aware
 
 from cl.alerts.models import (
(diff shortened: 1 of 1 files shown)


albertisfu (approved):
Thanks, @ERosendo! This is working as expected and should help reduce memory usage when deleting old hits.

PR #7587, opened 2026-07-17
Reduce opinion authority queries with database aggregation

KRRT7:
## Summary

- Aggregate citation depth across opinion clusters in PostgreSQL instead of issuing one query per authority.
- Traverse opinion citation relationships directly in the `authorities` queryset.
- Prefetch citations and select docket/court display data in constant query count.
- Preserve the existing asynchronous API.
- Add regression coverage for aggregation, ordering, self-citations, unrelated citations, async behavior, and query count.

## Performance

Benchmarked against `origin/main` using identical PostgreSQL rows and warm-cache A/B/B/A runs.

| Citation graph | Queries | Main median | This PR median | Speedup |
|---|---:|---:|---:|---:|
| 20 authorities / 40 edges | 25 → 2 | 41.4 ms | 24.5 ms | 1.69× |
| 100 authorities / 500 edges | 108 → 2 | 173.4 ms | 102.0 ms | 1.70× |
| 500 authorities / 5,000 edges | 513 → 2 | 967.8 ms | 473.4 ms | 2.04× |

Query count is now constant instead of growing with the number of authorities. At 500 authorities, this eliminates 99.6% of SQL statements.

Full ordered-result fingerprints matched between `main` and this PR at every benchmark size.


mlissner:
@claude review

cl/search/models.py
```diff
@@ -2766,31 +2732,22 @@ async def ahas_private_authority(self):
...
-            authorities_base.prefetch_related("citations")
+    @property
+    def authorities_with_data(self) -> QuerySet["OpinionCluster"]:
+        """Return authorities with their total citation depth and display data."""
+        return (
+            self.authorities.prefetch_related("citations")
             .select_related("docket__court")
-            .order_by("-citation_count", "-date_filed")
+            .order_by("-citation_depth", "-citation_count", "-date_filed")
         )
```
ERosendo, line 2824: Nice cleanup on the aggregation! 

Since we're already touching this method, I think this could also be a nice opportunity to defer some fields that can contain large blocks of text we never render. That way the query doesn't have to fetch those TOASTed values.

I took a quick look on the dev database using one large cases. For that page, the query pulls about 829 KB of text across those columns, while the template only renders around ~60 KB. Most of the extra data comes from `headmatter`, `attorneys`, and `summary` (about 95% of it), and one `headmatter` value alone was ~60 KB.

Since the table of authorities only displays the caption, date, status, court, and citations, I think we could safely defer the following fields:

```python
return (
    self.authorities.defer(
        "arguments",
        "attorneys",
        "correction",
        "cross_reference",
        "disposition",
        "headmatter",
        "headnotes",
        "history",
        "judges",
        "nature_of_suit",
        "other_dates",
        "posture",
        "procedural_history",
        "summary",
        "syllabus",
    )
    .prefetch_related("citations")
    .select_related("docket__court")
    .order_by("-citation_depth", "-citation_count", "-date_filed")
)
```

I chose `defer()` over `only()` because `only()` would require enumerating the `docket__court` fields as well, and caption reaches through to `docket.docket_number` and `docket.court.citation_string`. It feels pretty easy to accidentally miss one of those and end up reintroducing an N+1 query.

One small thing that might pair well with this: since deferred fields trigger a query if they're accessed later, it could be worth extending the `assertNumQueries(0)` section of the new test (`test_authorities_with_data_uses_constant_queries`) to also access `authority.caption` and `authority.get_absolute_url()`. `caption` is probably the most useful one to exercise since it traverses both docket and citations.

ERosendo (commented):
@KRRT7 Great refactor, it looks good! 👍 

I just left one comment about a small optimization around deferring some fields we don't render.

Let me know what you think. I'm happy to see it addressed in this PR, or in a follow-up PR if you prefer. Either works for me!

KRRT7:
somehow missed the review until your merge, will work on it. in this PR.

ERosendo (approved):

PR #7592, opened 2026-07-19
Fix SES bounce type classification

KRRT7:
## Summary

Fixes SES bounce classification so only Transient and Undetermined bounce types enter soft-bounce handling.

The previous condition was always true for every non-Permanent value because the string literal Undetermined is truthy.

## Validation

- Python compilation passed.
- git diff --check passed.


mlissner:
@claude review

PR #7593, opened 2026-07-19
Log SES event storage failures without failing webhooks

KRRT7:
## Summary

Log failures while storing sampled SES bounce and complaint payloads in S3, while keeping the webhook response successful.

The SES webhook intentionally acknowledges notifications with HTTP 200. Re-raising storage errors would cause SNS to redeliver the event and could repeat bounce or complaint side effects. The storage is auxiliary, sampled data, so failures are now observable through logging without failing event processing.

## Validation

- Python compilation passed.
- git diff --check passed.
- Added a regression test asserting storage failures are logged without escaping the handler.
- Django test execution is blocked locally because the configured cl-postgres host is unavailable.


mlissner:
@claude review

albertisfu (approved):
Looks good to me.
Thank you!

PR #7599, opened 2026-07-20
fix(contact): display issue-type hints and add DMCA category

### LINKED ISSUE #7305: Add DMCA complaint as a category on contact form.
These should come to me via Zoho Desk, so we need the new category there, and we need Zoho Desk to do the correct routing for it. 

The category can be "DMCA Complaint" and it can just have a simple box below that for folks to input their complaints. Keep it simple.

I'm assigning to Elisa for the contact form since she did it last time. Once she's done, I can see if I can get Zoho Desk working properly.

### LINKED ISSUE #7459: Page hint on contacts page does not display
See page:
https://www.courtlistener.com/contact/?issue_type=api

The contact page includes a page hint box that is supposed to display when a category is selected. The culprit is that it's rendered using `x-html` which isn't allowed by our CSP.

https://github.com/freelawproject/courtlistener/blob/253563e485669b83da479e94bbb8fc57061f0297/cl/simple_pages/templates/contact_form.html#L64

Maybe instead we can just render all of them and guard which ones show with an `x-show`.

elisa-a-v:
## Fixes
Fixes #7459
Closes #7305

## Summary
Two contact form changes:

- **#7459 — hints not displaying:** the per-issue-type hint text lived in the `contactForm` Alpine component's `hint` getter and was injected with `x-html`, which failed to render. The hint markup now lives in a Django include (`includes/contact_form_hint.html`) whose divs toggle on a `showHint` getter; the terms and MCP-help URLs render directly with `{% url %}` instead of round-tripping through Alpine via `json_script`. Because `contact_form.html` is a legacy (`base.html`) template, this uses a plain include rather than a Cotton component to keep the legacy/new stack boundary clean.
- **#7305 — DMCA category:** adds a "DMCA Complaint" option to the issue-type dropdown. Currently has no hint but it's easy to add one, might be worth pointing to the [TOS](http://localhost:8000/terms/) copyright section?

## Deployment
**This PR should:**
- [ ] `skip-deploy`
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

## Screenshots

<details>
<summary>DMCA Complaint category</summary>

image

</details>

<details>
<summary>Demo</summary>

[Screencast from 20-07-26 15:46:56.webm](https://github.com/user-attachments/assets/3dabd85d-54ec-4405-a0bb-99f2bd7c206c)

</details>


ERosendo (approved):
Thanks @elisa-a-v! The PR looks good.

I also created the new category in Zoho Desk, and the ticket category should map correctly when a new one is created.

mlissner:
Just a heads up that I added dmca_complaint to the wiki page about the contact form: https://wiki.free.law/c/courtlistener/dev-guide/working-with-the-contact-form

PR #7617, opened 2026-07-22
fix(corpus_importer): Prevent duplicated docket entries for ca8 and cadc

### LINKED ISSUE #7079: Add court-specific check for download confirmation page to get document numbers
### Describe the issue.

In courts like CADC that don't use regular document numbers in docket entries, we currently have inconsistent document number assignment depending on the source:

  - Extension uploads: Use `pacer_doc_id` as the document number (from the HTML docket).
  - recap.email: Extract the document number from the PDF header, which in some courts (e.g., CADC) returns a number that is not the `pacer_doc_id`.

  This mismatch causes duplicated docket entries when entries from both sources are merged into the same docket.

  Previously, we had court-specific code that checked the download confirmation page first for courts where the PDF header number doesn't match the `pacer_doc_id`. That code was removed to simplify the logic and avoid hardcoding court IDs.

**Proposed Fix:**

- Re-add the court-specific condition in `get_document_number_for_appellate` so that for courts known to show incorrect document numbers in the PDF, we check the download confirmation page first to retrieve the document number, falling back to the PDF if needed.
- Also, we should add logic to trigger an alert when these kinds of courts start using regular numbers, so we can clean up the entries.

### Related Issue

_No response_

### Screenshots

_No response_

Research

nikollycardoso:
document_number from the PDF header (recap.email) didn't match the value from the docket HTML (extension uploads) for ca8 and cadc, so process_recap_email treated the same entry as two documents. Check the confirmation page first for these courts instead, falling back to the PDF. Also alerts if a court starts returning a regular number, signaling it no longer needs this handling.

Fixes: #7079

## Fixes
Fixes: #7079

## Summary
For ca8 and cadc, document_number extracted from the PDF header (recap.email uploads) didn't match the value extracted from the docket HTML (extension uploads), causing the merge logic to create duplicated docket entries. This checks the download confirmation page first for these two courts, falling back to PDF extraction. Also adds a logger.error alert if a court starts returning a regular (short) document number on the confirmation page, signaling it may have switched to normal docket numbering.

## Deployment

**This PR should:**
    - [x] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`


mlissner:
And we're off!

cl/recap/tests/test_recap_email.py
```diff
@@ -3646,6 +3646,72 @@ async def test_nda_document_not_available_get_from_confirmation_page(
         self.assertEqual(recap_document_first.document_number, "148")
         self.assertEqual(recap_document_first.docket_entry.entry_number, 148)
 
+    @mock.patch(
+        "cl.recap.tasks.download_pdf_by_magic_number",
+        return_value=(None, "Document not available from magic link."),
+    )
+    @mock.patch(
+        "cl.corpus_importer.tasks.get_document_number_from_confirmation_page",
+        side_effect=lambda z, x: "0012345678",
```
albertisfu: In PACER, the first digit of a pacer_doc_id is always 0 or 1. So in this case, it should be 1:

`0011345678`

This doesn't affect the test, but it's better to keep it consistent.

cl/recap/tests/test_recap_email.py
```diff
@@ -3646,6 +3646,72 @@ async def test_nda_document_not_available_get_from_confirmation_page(
...
+        mock_cookies,
+        mock_cookies_cache,
+        mock_download_pdf_by_magic_number,
+        mock_get_document_number_from_confirmation_page,
+    ):
+        """This test verifies that a document number with two leading zeros
+        returned by the PACER download confirmation page for ca8/cadc is
+        normalized (leading zeros stripped) before being stored.
+        """
+
```
albertisfu, line 3671: This test should confirm that we're not duplicating docket entries or documents when merging an entry from RECAP Email that was previously created by a different source.

The idea is to use a `DocketFactory`, `DocketEntryFactory`, and `RECAPDocumentFactory` where the `DocketEntryFactory.entry_number` and `RECAPDocumentFactory.document_number` match the case and document being merged by the `RECAPEmailNotificationDataFactory`.

You can then verify that no additional `Docket`, `DocketEntry`, or `RECAPDocument` instances are created, and that the existing ones are simply updated. For example, set a `DocketEntryFactory.description` when creating the initial object, then use a different description in `RECAPEmailDocketEntryDataFactory` to confirm that the existing entry is updated.

cl/corpus_importer/tasks.py
```diff
@@ -2380,11 +2380,40 @@ def get_document_number_for_appellate(
 
     pdf_bytes = None
     document_number = ""
-    # Try to get the document number for appellate documents from the PDF first
-    if pq.filepath_local:
+
+    # Appellate courts that don't use regular docket entry numbering. Their PDF
```
albertisfu: A suggestion:

```suggestion
    # Appellate courts "ca8", "cadc" don't use regular docket entry numbering. Their PDF
```

We should also mention that the reason for fetching the `document_number` using `get_document_number_from_confirmation_page`, instead of using the `pacer_doc_id` directly (which is the number displayed on that page), is that it allows us to detect when a court starts using regular document numbers. When that happens, the confirmation page will display the regular document number instead of the `pacer_doc_id`. Otherwise, we wouldn't be aware that the court had changed its behavior.

albertisfu (changes requested):
Thanks @nikollycardoso this looks good.

I just have a couple of suggestions regarding documentation and testing.

albertisfu:
Thank you, @nikollycardoso! This looks great and is ready to be merged.

We just need you to sign the pending CLA agreement, please. It's listed in the pending checks.

albertisfu (approved):

PR #7624, opened 2026-07-23
fix(search): Throttle bulk document sealing to protect archive.org

### LINKED ISSUE #6613: Admin: Add button to seal document
### What happened?

We're really good at adding buttons to our admin site now. Let's add one for sealing documents, assuming it's not too hard. 

The links look like this: 

https://www.courtlistener.com/admin/search/recapdocument/?q=460314903

And then you can use the dropdown action to seal it. 

Is this feasible/doable with your latest code, @elisa-a-v?

### Expected behavior

A new button to make sealing easier.

### Actual behavior

Currently, I just hack the URL to accomplish this. I start at:

https://www.courtlistener.com/admin/search/recapdocument/460314903/change/

And I then modify it to: 

https://www.courtlistener.com/admin/search/recapdocument/?q=460314903

It's a bit error prone and a bit annoying. A button would be better! 

### Steps to reproduce

Try to seal an item using the admin panel. There's no button! 

### Screenshots

_No response_

Low

elisa-a-v:
Fixes #6613

The admin sealing flows added in #7100 (bulk seal action on RECAPDocument, and docket-entry-wide sealing including attachments) deleted files from archive.org in a tight loop with no delay.

This adds a 1-second sleep between documents in `seal_documents`, matching the pacing Mike used in his CLI workaround. The sleep only runs between documents — sealing a single document adds no delay.

The docket-entry seal confirmation page now shows how many documents are about to be sealed, and warns above ~100, where the throttled request may hit the 3-minute server timeout.

The existing `test_seal_documents_action` test now mocks `time.sleep` and asserts it fires exactly once when sealing two documents.


elisa-a-v:
Side note for reviewers: while adding the sleep mock I noticed the pre-existing mock parameter names in this test are swapped — decorators apply bottom-up, so `mock_delete_from_ia` actually receives the `invalidate_cloudfront` mock and vice versa. Harmless since neither is used by name. Want me to fix the naming in this PR, or leave it?

mlissner:
Feel free to do the rename if you want, but I don't think it's important. 

I think CloudFront is configured to a 60s timeout, so maybe do 50 docs as the threshold instead? I'd also tweak the wording to make it less specific so that it doesn't have to be updated in the future.

elisa-a-v:
Threshold is now 50, and I took the numbers out of both the warning text and the template comment so neither goes stale when the throttle or the timeout changes.

Fixing the swapped mock names was trivial so I just did it too, might save someone a headache if either mock is ever actually used.

mlissner (approved):

mlissner:
LGTM. I merged, but then realized we forgot the skip labels. I added them, but it's probably already too late. Ah well. Note for next time!

PR #7625, opened 2026-07-23
NYCoA models

Brennan-Chesley-FLP:
## Summary

This introduces models for the New York Court of Appeals data, and refactors a couple of things to reduce code duplication. 
- Common exact search functionality for admin
- Internalizes polymorphism in make_pdf_path

If we want to take out the refactors from this, I'm happy to. They are in the last two commits.

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


Brennan-Chesley-FLP:
The conflicts currently (6a19cbd) identified are for the intermediate commit prior to the refactor. Once we decide on keeping the refactor or not I'll fix the conflict accordingly.

cl/lasc/models.py
```diff
@@ -69,6 +73,16 @@ class LASCPDF(AbstractPDF, AbstractDateTimeModel):
...
 
+    def get_pdf_path(self, filename: str, thumbs: bool = False) -> str:
+        slug = slugify(trunc(filename, 40))
+        root = f"/us/state/ca/lasc/{self.docket_number}"
+        if thumbs:
+            root += "-thumbnails"
+        file_name = (
+            f"gov.ca.lasc.{self.docket_number}.{self.document_id}.{slug}.pdf"
+        )
+        return os.path.join(root, file_name)
```
MorganBennetDev, line 84: Since we're already messing with existing code could we also update everything to use `pathlib.Path` objects over `os.path` calls since this is the project preference?

cl/lib/tests.py
MorganBennetDev, line 1: If these tests are necessary, they should be handled by inheritors.

cl/lib/models.py
```diff
@@ -120,6 +120,12 @@ class AbstractPDF(models.Model):
     class Meta:
         abstract = True
 
+    def get_pdf_path(self, filename: str, thumbs: bool = False) -> str:
+        """Return the storage path for this document, or for its
+        thumbnail when ``thumbs`` is True. Used by the ``upload_to``
+        callbacks on ``filepath_local`` and ``thumbnail``."""
+        raise NotImplementedError
```
MorganBennetDev:
```suggestion python
    @abstractmethod
    def get_pdf_path(self, filename: str, thumbs: bool = False) -> str:
        """Return the storage path for this document, or for its
        thumbnail when ``thumbs`` is True. Used by the ``upload_to``
        callbacks on ``filepath_local`` and ``thumbnail``."""
        ...
```

cl/search/state/new_york/models.py
```diff
@@ -0,0 +1,157 @@
...
+        help_text="The date this case was decided, if it has been decided.",
+        blank=True,
+        null=True,
+        auto_now=False,
+    )
+    opinion_by = models.TextField(
+        help_text="The author of the opinion, for decided cases.",
+        null=True,
+        blank=True,
+    )
```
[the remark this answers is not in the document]
MorganBennetDev: This should be addressed (along with the others). See: https://github.com/freelawproject/courtlistener/pull/7552#discussion_r3642098906

cl/search/state/new_york/models.py
```diff
@@ -0,0 +1,157 @@
...
+    """New York Court of Appeals-specific metadata associated with a docket.
+
+    These fields capture information that only applies to NYCoA cases, so we
+    keep them in a separate model instead of adding them directly to the Docket
+    namespace.
+    """
+
+    docket = models.OneToOneField(
+        "search.Docket",
+        help_text="The docket this NYCoA metadata applies to.",
```
MorganBennetDev, line 32: Is there a reason we're not using `@document_model` for docket metadata specifically?

cl/search/state/new_york/models.py
MorganBennetDev, line 57: Would it be possible to make this an optional FK to `people_db.Person` since (presumably) the author of the opinion is always going to be a judge, who we will have or want to have in the database?

cl/search/state/new_york/models.py
```diff
@@ -0,0 +1,157 @@
...
+        on_delete=models.CASCADE,
+    )
+    page = models.IntegerField(
+        help_text=(
+            "The Court-PASS page this entry was scraped from. Entries are "
+            "numbered independently per page, so this is part of the natural "
+            "key."
+        ),
+        choices=EntryPage.choices,
+    )
```
MorganBennetDev, line 102: `help_text` doesn't need to be set here unless it differs from the docstring.

Would it be possible to combine this field with `filing_type` somehow? This setup feels a bit like we're letting NYCoA's front-end design seep into our backend.

cl/lib/model_helpers.py
```diff
@@ -382,63 +383,20 @@ def base_recap_path(instance, filename, base_dir):
...
-    from cl.lasc.models import LASCPDF
-    from cl.search.models import (
-        ClaimHistory,
-        FloridaDocument,
-        RECAPDocument,
-        ScotusDocketMetadata,
-        SCOTUSDocument,
-        TexasDocument,
-    )
+class SupportsPdfPath(Protocol):
```
MorganBennetDev, line 385: I'm wondering if we could just have `ScotusDocketMetadata` inherit from `AbstractPDF` to get rid of the need for a `Protocol`.

cl/search/state/new_york/factories.py
```diff
@@ -0,0 +1,49 @@
...
+
+    class Meta:
+        model = NYCoADocketEntry
+
+
+class NYCoADocumentFactory(DjangoModelFactory):
+    docket_entry = SubFactory(NYCoADocketEntryFactory)
+    file_name = Faker("file_name", extension="pdf")
+    document_number = Sequence(lambda n: n + 1)
+    available = True
```
MorganBennetDev: This should be randomized with a low chance of `False`

MorganBennetDev (changes requested):
Looks good! Some comments regarding the refactor and the design of NYCoA models (didn't see them discussed anywhere). I've been eyeing that nasty `if` chain in `make_pdf_path` for a bit now so it's good someone got around to it.

cl/lib/tests.py
Brennan-Chesley-FLP: They were helpful for the refactor, but I don't know that they need to stick around. Axed em.

cl/search/state/new_york/models.py
Brennan-Chesley-FLP: Removed this field entirely! This slipped in here from a little unreliable pdf parsing experiment that I've thought better of. This field has been replaced with a (sometimes available) lower court citation.

cl/search/state/new_york/models.py
Brennan-Chesley-FLP: Unfortunately, we are kind of constrained by how nycourts.gov has decided to display their information on the front end. There are docket entries that appear on two different pages, and since those sequences can get added to independently, to have stable natural keys, we're kinda stuck with identifying the page somehow. The alternative here is to have a blank and a null filing_type and just use nulls on one page and blanks on the other, though that seems a little more fraught to me.

cl/lib/model_helpers.py
Brennan-Chesley-FLP: It would be a little bit tidier, though it would require a migration and a bunch of backfill for the concrete columns that are slightly less than Abstract. We can certainly do it if we'd like to.

MorganBennetDev:
@Brennan-Chesley-FLP Could you please clarify what site these models are meant to support? I'm trying to go through your work and I'm having a bit of trouble following. There seems to be a proposal for models https://github.com/freelawproject/juriscraper/issues/1838 (?), which seems like it was completed, but other than that I can't find anything written down. Is this intended to support results from Court-PASS _and_ NYSCEF or just one?

cl/search/state/new_york/models.py
Brennan-Chesley-FLP: There was not!

Brennan-Chesley-FLP:
This is just for [Court-PASS](https://courtpass.nycourts.gov). NYSCEF will be a different scraper, which should give us a bit more info and more reliable files, but dev is currently parked pending some communication. That juriscraper PR is mostly done, but it's pending jkent and any changes we make to it.

cl/search/state/new_york/models.py
```diff
@@ -0,0 +1,152 @@
...
+        related_name="nycoa_metadata",
+        on_delete=models.CASCADE,
+    )
+    issue = models.TextField(
+        help_text=(
+            "The issue categories the Court of Appeals associated with this "
+            "case."
+        ),
+        blank=True,
+    )
```
MorganBennetDev, line 42: Does this appear to have a fixed set of values in the scraped data or does it just seem like there's a text box on the clerk's end?

cl/search/state/new_york/models.py
```diff
@@ -0,0 +1,152 @@
...
+
+    docket = models.ForeignKey(
+        "search.Docket",
+        related_name="nycoa_docket_entries",
+        on_delete=models.CASCADE,
+    )
+    page = models.IntegerField(
+        choices=EntryPage.choices,
+    )
+    filing_type = models.TextField(blank=True)
```
MorganBennetDev, line 103: Would it be possible to make this a `SmallInteger` with choices based on [Court-PASS' conventions](https://www.nycourts.gov/ctapps/techspecs.htm)? I assume there will be some weird situations with people not formatting things correctly, but we can capture those with a `filing_type_raw` field. Would only cost a few extra bytes and would give end users nicer data to work with.

cl/search/state/new_york/models.py
```diff
@@ -0,0 +1,152 @@
...
+    party = models.TextField(blank=True)
```
MorganBennetDev, line 104: Similarly here, can we have `party` be a nullable FK to `people_db.Party` and then have a `party_name` that's a `TextField`?

cl/search/state/new_york/models.py
```diff
@@ -0,0 +1,152 @@
...
+    :ivar available: Whether the file is available (False for sealed or
+    not-available files).
+    """
+
+    docket_entry = models.ForeignKey(
+        NYCoADocketEntry,
+        related_name="documents",
+        on_delete=models.CASCADE,
+    )
+    file_name = models.TextField(blank=True)
```
MorganBennetDev, line 134: Can we store the URL to the file on NY's server instead of/in addition to the name?

cl/search/state/new_york/models.py
```diff
@@ -0,0 +1,152 @@
...
+    document_number = models.IntegerField()
```
MorganBennetDev, line 135: I trust that they exist, but just to make sure I understand everything, can you also send a link to a NY docket where a single docket entry has multiple attachments?

cl/search/state/new_york/models.py
MorganBennetDev: Can you send a link to an example of this? I'm having trouble finding one just randomly searching dockets.

Also, is the `page` value useful to end users or should we just drop this field and construct `sequence_number` from `page` concatenated with whatever `sequence_number` is currently planned to be in the merger?

MorganBennetDev (changes requested):
Thanks for the changes @Brennan-Chesley-FLP. Thought of quite a few questions.

cl/search/state/new_york/models.py
```diff
@@ -0,0 +1,152 @@
...
+        choices=EntryPage.choices,
+    )
+    filing_type = models.TextField(blank=True)
+    party = models.TextField(blank=True)
+    date_received = models.DateField(null=True, blank=True, auto_now=False)
+    sequence_number = models.CharField(max_length=16)
+
+    class Meta:
+        app_label = "search"
+        ordering = ["page", "sequence_number"]
```
albertisfu, line 110: I assume this is the default order that will be used to sort the docket entries when they're rendered on the docket page. Does this mean we'll show all the `DOCUMENTS_PAGE` entries first, followed by the `FILINGS_PAGE` entries?

Would it make more sense to use `["sequence_number", "page"]` instead? I'm assuming sequence_number reflects chronological order. 

What's the ordering used in the court?

cl/search/state/new_york/models.py
```diff
@@ -0,0 +1,152 @@
...
+        verbose_name_plural = "NYCoA Docket Entries"
+        unique_together = [["docket", "page", "sequence_number"]]
```
albertisfu, line 112: We can change the order of the columns in this unique constraint to leverage the index it creates.

This can also match the `ordering` we decide to use, for example:

`[['sequence_number', 'page', 'docket']]`

That way, ordering can be performed more efficiently without requiring an additional index.

cl/search/state/new_york/models.py
```diff
@@ -0,0 +1,152 @@
...
+    file_name = models.TextField(blank=True)
+    document_number = models.IntegerField()
+    available = models.BooleanField(default=True)
+
+    class Meta:
+        app_label = "search"
+        indexes = [
+            models.Index(fields=["filepath_local"]),
+        ]
+        unique_together = [["docket_entry", "document_number"]]
```
albertisfu, line 143: If we invert this constraint to `["document_number", "docket_entry"]`, we'll be able to leverage the index created by the constraint to speed up `document_number` only queries.

cl/search/state/new_york/models.py
```diff
@@ -0,0 +1,152 @@
...
+    )
+    summary = models.TextField(
+        help_text="The detailed issue descriptions for this case.",
+        blank=True,
+    )
+    decision_date = models.DateField(
+        help_text="The date this case was decided, if it has been decided.",
+        blank=True,
+        null=True,
+        auto_now=False,
```
albertisfu, line 51: Is this redundant as a DateField by default is `auto_now=False`

albertisfu (changes requested):
Thanks @Brennan-Chesley-FLP, this looks good.

Just a few additional suggestions, along with the ones Morgan mentioned, please.

Brennan-Chesley-FLP:
I'm closing this PR in favor of the general discussions on coding (in the data sense) and further exploration of the merger framework on my part to see if I can remove some of the fields that are objectionable here.

PR #7632, opened 2026-07-24
Fix increment event throttle and skip bots

ERosendo:
## Summary

`EventCounterViewSet` (the `increment-event` endpoint) did not define its own `throttle_classes`, so it inherited the default API throttles. Since the view sets `authentication_classes = []`, DRF never authenticates these requests, causing all traffic, including browser requests, to be throttled by IP under the shared `user` scope. That scope is also used for regular API traffic and was recently tightened to 5/min, 50/hour, and 125/day, making it too restrictive for this endpoint.

Key changes
- Adds a dedicated events throttle scope so this internal endpoint no longer shares the same rate limit bucket as regular API traffic.
- Sets the events rate limit to `60/hour`, based on an analysis of one day of production request logs.
- Reuses the existing `cl.lib.bot_detector.is_bot()` helper to skip incrementing the counter for recognized bots, returning a normal 202 Accepted response without updating the database.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`


albertisfu (approved):
Thanks @ERosendo looks good to me.

PR #7635, opened 2026-07-25
fix(search): Server-render corpus search tabs (#7035)

### LINKED ISSUE #7035: Server-render corpus search tabs to prevent flash before JS loads
## Problem

On the homepage, the corpus search component's tabs (Case Law, RECAP Archive, Oral Arguments, Judges) are rendered client-side with Alpine's `x-for`. This causes a visible flash on page load where no tabs are shown until JavaScript finishes executing.

## Proposed fix

Switch from `x-for` to Django's `{% for %}` template tag so the tabs are included in the server-rendered HTML. The active tab state and interactivity can still be handled by Alpine, but the initial render should come from the server.

stephenstanwood:
## Fixes
Fixes: #7035

## Summary
The homepage search tabs were built with Alpine x-for, so they weren't in the HTML until JS ran and you got a flash. This switches the tabs to Django `{% for %}` so the labels show up right away. Alpine still handles which tab is selected. I also added a test that checks the labels are in the page HTML.

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

<details closed>
<details open>

I don't have local screenshots yet. Happy to add them if helpful.

</details>
<details open>

Same. Tabs are desktop-only in this UI (`hidden md:flex`), so mobile still uses the scope menu.

</details>
</details>


mlissner:
Hi! Thanks for submitting this. It's important to look at your work to be sure it's correct. Can you please submit screenshots as well?

stephenstanwood:
Screenshots here! Also verified with JS off: the four tab labels are in the initial HTML (no Alpine flash).

cl-pr-7635-desktop-recap-selected

cl-pr-7635-desktop-tabs-closeup

cl-pr-7635-desktop-tabs

cl-pr-7635-mobile-scope-menu

elisa-a-v (approved):
Hi! Thank you for this PR, and for your patience with the review. The regression test is a nice touch, good call.

I tested locally and it works correctly, so I'm merging, but two follow-ups came out of reading it:

- The menu variant of the scope selector is still built client-side with `<template x-for>`. It's used by the site header as well as the mobile controls, so that flash is still there on every page with the new header. #7827
- The scope list now lives in both `constants.py` and `corpus_search.js`, which is what the sync comment is working around. We have a `json_script` pattern for this elsewhere that would let the JS array go away entirely. #7828

Both are written up if you want them, but you've given us plenty of your time already, so genuinely no obligation either way. Thanks again for this!

PR #7653, opened 2026-07-29
Remove throttles for DCR and oauth metadata endpoints

nadahlberg:
## Fixes

The built-in Django OAuth Toolkit endpoints don't inherit our default throttles for anonymous users, but our hand-written DCR endpoint (`/o/register`) and oauth metadata (`/well-known/oauth-authorization-server`) do. These endpoints should remain publicly accessible and allow lots traffic from a small number of IPs (e.g. an LLM provider registering clients for all of their users connecting to the MCP).

Making more than 5 requests to these endpoints currently raises 429s. Note: The reason this hasn't appeared to be a larger problem is because there is a related issue with how our anonymous user rate limits are being applied (I'll open that separately).

I realize there might be a concern about not gating an endpoint which can write rows to the oauth application table. But we can in a separate PR implement a cleanup job that drops stale applications that haven't gone through the approval flow (which requires a logged in user and browser interaction).

I think this is a pretty high priority fix, so would prefer not to wait on the cleanup PR.

## Summary

Sets `throttle_classes=[]` for both endpoints.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [x] `skip-celery-deploy`
    
    
    - [x] `skip-cronjob-deploy`
    
    
    - [x] `skip-daemon-deploy`

### DIFF
--- a/cl/oauth/views.py
+++ b/cl/oauth/views.py
@@ -75,6 +75,7 @@ class DynamicClientRegistrationView(APIView):
 
     authentication_classes: list[Any] = []
     permission_classes: list[Any] = []
+    throttle_classes: list[Any] = []
 
     def handle_exception(self, exc: Exception) -> Response:
         if isinstance(exc, Ratelimited):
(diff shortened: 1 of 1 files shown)


mlissner:
This looks good to me, I think, but I do think a quick review from @albertisfu or @ERosendo would be helpful. It's only two lines in this PR.

albertisfu (approved):
Looks good to me :shipit:

PR #7664, opened 2026-07-31
Feat/issue 6085 active filter count

### LINKED ISSUE #6085: Introduce active filter count for user feedback in mobile
### Describe the issue.

As discussed [here](https://github.com/freelawproject/courtlistener/issues/5206#issuecomment-3133155703), if a user:

1. Goes to the homepage
2. Opens the advanced filters dialog
3. Enters some values
4. Then closes the dialog

The entered values are still in those inputs, so they will be sent along with the query if the user performs a search with the dialog closed. We need to inform the user that this will happen, so we should show the number of active filters on the "Filters" button when the dialog is closed:

<details>
<summary>Homepage</summary>

Image

</details>

<details>
<summary>Header</summary>

Image

Image

</details>

### Related Issue

_No response_

### Screenshots

_No response_

To Do

joanasouza:
## Fixes

This fixes #6085 originated from #5206 bug
## Summary

This PR adds a user feedback on homepage and search dialog for active filters values in advanced search on mobile. 

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

1. To deploy...

<details closed>
<details open>

</details>
<details open>
Screenshot 2026-07-31 at 14 31 46
Screenshot 2026-07-31 at 14 31 52

</details>


mlissner:
@claude review

elisa-a-v (changes requested):
Thank you @joanasouza for this PR! As discussed over email, we'll take it from here, so no further action from you is needed; @mexomagno will land the remaining fixes.

Now here's my review of the PR as it stands. First, the parts worth calling out as genuinely well done:

- The accessibility structure is exactly right: the count bound into the button's `aria-label`, the visible badge `aria-hidden` so the number isn't announced twice, and the unasked-for `aria-expanded` on the in-dialog button.
- All logic lives in the external Alpine component behind named getters, so the templates carry no inline expressions, just as our CSP settings require.
- Reusing the existing `data-ignore-input` convention instead of inventing a parallel one.

Claude's review already covers most of what I found:

- [The disabled-scope finding](https://github.com/freelawproject/courtlistener/pull/7664#discussion_r3694033274) is correct, including the compounding init gap: a child input's `.disabled` property doesn't reflect an ancestor `<fieldset disabled>`, so the code as it stands has a real bug where filters in inactive scopes are still counted as active, and inactive scopes are always visually hidden but their fieldsets are only marked disabled on scope change. I think the fixes suggested by Claude are the right call.
- [The date-selector timing finding](https://github.com/freelawproject/courtlistener/pull/7664#discussion_r3694033276) is also correct, though minor because it's display-only and self-correcting. The `$nextTick` deferral it suggests is fine.
- [The live-region finding](https://github.com/freelawproject/courtlistener/pull/7664#discussion_r3694033270) I only partly agree with. `aria-hidden` on the badge is the right call, since the button aria label includes the count already. I quickly tested the `aria-live="polite"` fix and never got VoiceOver to announce the count changes, so I think the live update announcement for SR is outside the scope of this issue; I'll open a separate issue for it.

One thing the automated pass didn't flag:

- The badge `<span>` is duplicated verbatim between `mobile/dialog.html` and `v2_homepage.html`. The buttons diverge in a few attributes but I think we could extract it and create a new component for the filters button that is flexible enough for the callsite to provide the different attributes.

mexomagno:
@claude review

cl/assets/static-global/js/alpine/components/corpus_search.js
```diff
@@ -173,21 +204,29 @@ document.addEventListener('alpine:init', () => {
...
     init() {
-      this.$watch('selectedScope', (newVal) => this.updateFieldsets(newVal.label));
+      this.$watch('selectedScope', (newVal) => {
+        this.updateFieldsets(newVal.label);
+        this.updateActiveFilterCount();
+      });
       this.onBreakpointChange(() => {
         this.advancedFiltersExpandedDesktop = false;
       });
+      this.updateActiveFilterCount();
```
elisa-a-v, line 229: As suggested by Claude, we need to call `updateFieldsets` on first load to mark the inactive scopes disabled.

This is a pre-existing issue that is not currently producing any bugs, but once we get pre-filled filters via query params, we could end up with some populated fields in inactive scopes which would be counted as active and, indeed, sent along with the search if no scope change ever happens because the fieldsets are only marked disabled by `updateFieldsets`.

```suggestion
      this.updateFieldsets();
      this.updateActiveFilterCount();
```

cl/assets/static-global/js/alpine/components/corpus_search.js
```diff
@@ -173,21 +204,29 @@ document.addEventListener('alpine:init', () => {
...
-        const shouldIgnore = el.dataset?.ignoreInput === 'true';
-        if (isEmpty || shouldIgnore) {
-          el.setAttribute('disabled', 'disabled');
-        }
+      this.formInputs.filter((el) => !willBeSubmitted(el)).forEach((el) => el.setAttribute('disabled', 'disabled'));
+    },
+
+    /**
+     * Recount the filters the next submit would apply, so the count can be shown
+     * before submitting. Runs on the next tick to wait x-bind:disabled and
```
elisa-a-v, line 212:
```suggestion
     * before submitting. Runs on the next tick to wait for x-bind:disabled and
```

elisa-a-v (changes requested):
Thanks, we're almost there! There's just one requested change that hasn't been addressed yet, and one tiny typo, then we can merge 👍🏽

cl/assets/static-global/js/alpine/components/corpus_search.js
mexomagno: Done

cl/assets/static-global/js/alpine/components/corpus_search.js
mexomagno: Oops. Done

PR #7723, opened 2026-08-03
feat(donate): Grant membership benefits while payment is pending

ERosendo:
## Summary

Members lose access the moment their payment is mid-flight. `NeonMembership` only counted as active when `payment_status == SUCCEEDED`, so a member whose charge is still processing (`PENDING` / "awaiting payment processing") was treated as a non-member everywhere access is gated: API throttle overrides, UserProfile.is_member, the CRM/Zoho sync, and the x2 promo.

This PR treats `PENDING` as active alongside `SUCCEEDED`, so members keep their benefits while a charge settles. Only an outright `FAILED/declined` payment revokes access.

Key changes: 

- Adds `NeonMembership.ACTIVE_PAYMENT_STATUSES = {SUCCEEDED, PENDING}` as the single source of truth and pointed NeonMembership.is_active at it.

-  Routed the two queryset-level that independently hardcoded `payment_status=SUCCEEDED` (`get_all_throttle_overrides` and `get_promo_excluded_usernames` in `cl/api/utils.py`) through the same constant, so the SQL and Python definitions of "active" can't drift apart.

- Everything else (is_member, search views, alerts, admin) already flows through the is_active property and inherits the change automatically.

## Deployment

**This PR should:**

    - [ ] `skip-web-deploy`
    
    
    - [ ] `skip-celery-deploy`
    
    
    - [ ] `skip-cronjob-deploy`
    
    
    - [ ] `skip-daemon-deploy`

(Since `NeonMembership.is_active` is a shared model property and its behavior changed, deploying all tiers together helps keep the web app and workers in sync)


mlissner (approved):
LGTM.

mlissner:
Just need the proper skip labels, @ERosendo, then I think we're good to merge.

ERosendo:
@mlissner I left a comment mentioning that I think changes like this to the models should be rolled out across all our deployments so the app and worker stay in sync. 😄 that's the reason i didnt add skip labels

PR #7368, opened 2026-05-20
docs(coverage): Add SCOTUS and Texas docket coverage pages

albertisfu:
## Fixes
Fixes: #7356

## Summary
Adds two new coverage pages and updates the main coverage index to advertise the SCOTUS and Texas docket data we now replicate.

**New pages**

- `/help/coverage/scotus/` — Supreme Court docket coverage. 
- `/help/coverage/texas/` — Texas state court coverage. 

**Main coverage page**

Adds a new "Supreme Court Dockets" section and a "State Courts" subsection (currently containing Texas) to both `help/coverage.html` and `v2_help/coverage.html`. The subsection is structured so additional states can be added later without restructuring.

**Tests**

- Registered both new URLs in `SimplePagesTest.test_simple_pages` (legacy load test).
- Registered both v2 templates in `V2PagesRegisterTest.V2_PAGES` so the redesign flag is exercised too.

**See Screenshots below for content details.**

## Deployment

**This PR should:**

- [x] `skip-deploy` (skips everything below)
    - [ ] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

<details closed>
<summary><h2>Screenshots</h2></summary>
<details open>
<summary><h4>Current design</h4></summary>

Coverage Page:
Screenshot 2026-05-20 at 10 48 31 a m

SCOTUS:
Screenshot 2026-05-20 at 10 48 44 a m

Texas:

Screenshot 2026-05-20 at 10 49 29 a m

Screenshot 2026-05-20 at 10 49 39 a m

</details>
<details open>
<summary><h4>New Design</h4></summary>
Coverage Page:
Screenshot 2026-05-20 at 11 23 39 a m

SCOTUS: 
Screenshot 2026-05-20 at 10 53 15 a m
Screenshot 2026-05-20 at 10 53 38 a m

Texas:
Screenshot 2026-05-20 at 10 52 22 a m
Screenshot 2026-05-20 at 10 52 31 a m
Screenshot 2026-05-20 at 10 52 42 a m

</details>
</details>

PR #7727, opened 2026-08-04
7584 scotus docket rendering 20260803

### LINKED ISSUE #7584: Support full SCOTUS case rendering on the docket page
### Describe the issue.

This is part of [https://github.com/freelawproject/courtlistener/issues/6102](https://github.com/freelawproject/courtlistener/issues/6102).

We already have the SCOTUS docket content in the database; however, it's not being exposed to users in the frontend.

Here's how a docket scraped from SCOTUS currently looks:

Image

As you can see, there are no docket entries or documents, and the metadata is incomplete.

The reason why docket entries and their documents are not shown is that we're using different models to store SCOTUS content than the regular RECAP ones:

* `DocketEntry` → `SCOTUSDocketEntry`
* `RECAPDocument` → `SCOTUSDocument`

So we should add support to render these SCOTUS models so they can be displayed like regular dockets:

Image

This is a PACER docket, so for SCOTUS there should be a few differences:

* No **Export CSV** button, as it relies on Elasticsearch. We'll add it once SCOTUS data is indexed in ES. For now, we can simply hide the button for SCOTUS.
* No **🙏 Pay and Pray** button, as this is a PACER-only feature.
* No **Buy on PACER** button.
* We should still support date and document filters, ASC/DESC ordering, and pagination.
* Update the coverage notice for SCOTUS to something like: `"Items on this page are sourced from ..."`.

Also, on the docket page we should consider:

* Rendering the additional data already stored in the `ScotusDocketMetadata` model.
* Rendering any additional `Docket` fields that are relevant for SCOTUS, such as **Date of Last Known Filing**.
* Hiding the **Get Alerts** button, as we currently don't support docket alerts for SCOTUS.
* Updating the **View on PACER** button to **View in SCOTUS**.
* Verifying that the **Parties** and **Attorneys** sections render correctly, or adjusting them if needed for SCOTUS.

All of these customizations should happen conditionally only for dockets with `court_id="scotus"`.

### Relevant files

Some relevant files to consider when exploring what's needed:

* `cl/opinion_page/views.py` (`view_docket`)

  For the refactor required in this view to fetch `SCOTUSDocketEntry`, `SCOTUSDocument`, and `ScotusDocketMetadata`, in addition to checking `court_id="scotus"`, we should consider an approach that makes it easier to introduce new docket models in the future. This will make it straightforward to add support for state content, where we'll have similar models such as `TexasDocketEntry` and `TexasDocument`.

  In general, we should try to make this view model-agnostic so that forms, pagination, document fetching, and related logic can be easily adapted to future state-specific models.

### Frontend files

* `cl/opinion_page/templates/docket.html`
* `cl/opinion_page/templates/docket_tabs.html`
* `cl/opinion_page/templates/de_list.html`

### Related Issue

_No response_

### Screenshots

_No response_

To Do

nikollycardoso:
## Fixes
Fixes: #7584

## Summary
SCOTUS dockets have been ingested, but the docket detail page only ever queried DocketEntry/RECAPDocument, so opening a SCOTUS docket rendered an empty entries list with PACER-only UI (Buy on PACER, Get Alerts, Pay-and-Pray) that doesn't apply to the Supreme Court.

Approach: added cl/opinion_page/docket_entry_sources.py, a small per-court config (RECAP_SOURCE/SCOTUS_SOURCE) describing how to fetch, sort, and display entries for a given docket. view_docket and the templates branch on this config instead of hardcoding court_id/model checks, so a future state-specific source can plug in by adding one more instance. Also added Docket.scotus_docket_url (mirrors the existing pacer_docket_url property) for linking out to the public case page on supremecourt.gov, and a new SCOTUS Docket Metadata section (capital case, linked-with, questions presented) in both template stacks.

Gotchas / compromises:
- CSV export: per the discussion on this issue, the Export CSV button stays visible for SCOTUS dockets, but SCOTUSDocketEntry/SCOTUSDocument don't implement real CSV export yet. Clicking it return a handled 501 (which triggers the existing "There was a problem. Try again later." message) instead of crashing. This is scoped so a genuine future bug in the RECAP CSV path still raises normally and reaches Sentry.

## Deployment

**This PR should:**
    - [ ] `skip-web-deploy`
    - [x] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`

1. To deploy...

Screenshot_2026-08-04_15-01-08

Screenshot_2026-08-04_15-02-57

Screenshot_2026-08-04_15-28-45

Screenshot_2026-08-04_15-29-11


mlissner:
Just a heads up that @elisa-a-v mentioned she was going to be working on the v2 docket page in today's sprint meeting. I think the timing should be fine, but we should be careful.

cl/opinion_page/docket_entry_sources.py
```diff
@@ -0,0 +1,79 @@
...
+    order_by_desc=("-sequence_number",),
+    has_pay_and_pray=False,
+    has_docket_alerts=False,
+)
+
+_SOURCES_BY_COURT_ID: dict[str, DocketEntrySource] = {"scotus": SCOTUS_SOURCE}
+
+
+def get_docket_entry_source(docket: Docket) -> DocketEntrySource:
+    return _SOURCES_BY_COURT_ID.get(docket.court_id, RECAP_SOURCE)
```
elisa-a-v: Maybe this could be a method in the class definition of the `Docket` model, that way instead of importing the method and passing the docket as an argument, we could just call it from the Docket object itself, like so: `docket.get_entry_source()`

cl/opinion_page/templates/docket_tabs.html
```diff
@@ -438,6 +438,47 @@ <h3 class="v-offset-above-3">Originating Court Information</h3>
       </p>
     {% endif %}
     {% endwith %}{# No more og_info variable #}
+
+    {% if is_scotus %}
+    {% with scotus_meta=docket.scotus_metadata %}
+    {% if scotus_metadata %}
```
albertisfu: This condition comes from `scotus_metadata` in `view_docket`, which is only available in the main tab. It is not available in other tabs, such as Parties, Authorities, etc., so this content is hidden on those tabs.

One suggestion is to move the `scotus_metadata` context assignment to `core_docket_data` so that this context is available across all the docket tabs.

cl/opinion_page/templates/docket_tabs.html
```diff
@@ -438,6 +438,47 @@ <h3 class="v-offset-above-3">Originating Court Information</h3>
       </p>
     {% endif %}
     {% endwith %}{# No more og_info variable #}
+
+    {% if is_scotus %}
+    {% with scotus_meta=docket.scotus_metadata %}
```
albertisfu: This might not be required. You can use the data from `scotus_metadata`.

What if you create a `for` loop in this template that iterates over `scotus_metadata` and renders what's required, such as the label, url, value, etc.? This way, you might be able to remove all the manual assignments below.

This might also be closer to how the values are rendered in V2.

cl/opinion_page/utils.py
```diff
@@ -316,6 +318,56 @@ def build_bankruptcy_metadata(
...
+        items.append({"label": "Capital Case", "value": "Yes"})
+
+    if scotus_metadata.date_discretionary_court_decision:
+        items.append(
+            {
+                "label": "Date of Discretionary Court Decision",
+                "value": (
+                    scotus_metadata.date_discretionary_court_decision.strftime(
+                        "%b. %d, %Y"
+                    )
```
albertisfu: This might be more reliable.

```suggestion
                "value": date_format(
                    scotus_metadata.date_discretionary_court_decision,
                    "N j, Y",
                ),
```

cl/opinion_page/utils.py
```diff
@@ -316,6 +318,56 @@ def build_bankruptcy_metadata(
...
+        items.append(
+            {"label": "Linked With", "value": scotus_metadata.linked_with}
+        )
+
+    if http_url(scotus_metadata.questions_presented_url):
+        items.append(
+            {
+                "label": "Questions Presented",
+                "value": "View",
+                "url": scotus_metadata.questions_presented_url,
```
albertisfu: The priority here should be to link to the file from our local copy in `questions_presented_file` and fall back to `questions_presented_url` if the file is not present.

cl/opinion_page/views.py
```diff
@@ -429,6 +453,7 @@ def _get_related(
             "docket_entries": paginated_entries,
             "sort_order_asc": sort_order_asc,
             "form": form,
+            "hide_docket_alerts": not source.has_docket_alerts,
```
albertisfu: We can remove hide_docket_alerts, as it seems to be used only in V2. The button should also be visible, just like in the current template.

cl/opinion_page/templates/includes/de_filter.html
```diff
@@ -110,6 +110,10 @@
   </div>
 
 <p style="margin-top: 10px; margin-bottom: 0;">
-  <i class="fa fa-info-circle"></i> Items on this page are sourced from <a href="https://www.pacer.gov">PACER</a> and <a href="https://free.law/recap/">RECAP</a> and may not be up to date. <a href="{% url 'coverage_recap' %}">Learn more about our coverage</a>.
+  {% if is_scotus %}
+    <i class="fa fa-info-circle"></i> Items on this page are sourced from the Supreme Court of the United States and may not be up to date.
```
albertisfu: Can you also mention the SCOTUS coverage page here? It will be available at `/help/coverage/scotus/`.

It's not available yet, but it will be once we launch the SCOTUS content: https://github.com/freelawproject/courtlistener/pull/7368

cl/opinion_page/utils.py
```diff
@@ -316,6 +318,56 @@ def build_bankruptcy_metadata(
     return items
 
 
+def build_scotus_metadata(
```
albertisfu: We expect to have more methods like this for other states as well. Could you please move this method to the new file you added, `docket_entry_sources`, and probably rename that file to `docket_sources_utils`?

albertisfu (changes requested):
Thanks, @nikollycardoso! This looks good.

I left some additional suggestions. 

I agree with Elisa's suggestion, so please apply it too.

Also, regarding the V2 template, could you apply the same changes you made to the legacy template?

For example, display the docket alerts button, remove the SCOTUS metadata title, and show the dropdown button for SCOTUS documents.

Image

cl/opinion_page/templates/includes/de_filter.html
mlissner: You can link it to the location Alberto references, but I opened a bunch of tickets just now to start moving coverage stuff over to the wiki, so it might happen sort of soon. When I do, I'll just redirect /help/coverage/scotus to the new wiki location, so it's just a matter of whether we have that one extra redirect.

cl/opinion_page/templates/includes/de_filter.html
```diff
@@ -111,7 +111,11 @@
 
 <p style="margin-top: 10px; margin-bottom: 0;">
   {% if is_scotus %}
-    <i class="fa fa-info-circle"></i> Items on this page are sourced from the Supreme Court of the United States and may not be up to date.
+    {% comment %}
+    Swap this hardcoded path for {% url 'coverage_scotus' %} once the
+    SCOTUS coverage page is ready.
+    {% endcomment %}
+    <i class="fa fa-info-circle"></i> Items on this page are sourced from the Supreme Court of the United States and may not be up to date. <a href="/help/coverage/scotus/">Learn more about our coverage</a>.
```
albertisfu: We should change this URL to point directly to the wiki page, since we ended up not having a SCOTUS coverage page in CL.

The URL is:
[https://wiki.free.law/c/courtlistener/help/data-coverage/surpeme-court-cases-and-filings](https://wiki.free.law/c/courtlistener/help/data-coverage/surpeme-court-cases-and-filings)

You can add it as a variable, similar to the other wiki pages, for example: `WIKI_COVERAGE_RECAP_URL`.

cl/opinion_page/utils.py
```diff
@@ -514,6 +509,10 @@ async def core_docket_data(
 
     has_alert = await user_has_alert(await request.auser(), docket)  # type: ignore[arg-type]
 
+    scotus_metadata = await sync_to_async(getattr)(
+        docket, "scotus_metadata", None
+    )
```
albertisfu: This is triggering a reverse query to check whether the one-to-one relation with `scotus_metadata` exists, even if the `Docket` is not from SCOTUS. So, this is an extra query that runs on every docket page load.

We should get rid of this extra query for dockets that are not from SCOTUS.

You can create a method like this:

```
async def get_scotus_metadata_items(docket: Docket) -> list[MetadataItem]:
      if docket.court_id != "scotus":
          return []
      metadata = await ScotusDocketMetadata.objects.filter(docket=docket).afirst()
      return build_scotus_metadata(metadata)
```

cl/opinion_page/utils.py
```diff
@@ -523,6 +522,8 @@ async def core_docket_data(
             "has_alert": has_alert,
             "timezone": COURT_TIMEZONES.get(docket.court_id, "US/Eastern"),
             "private": docket.blocked,
+            "is_scotus": docket.court_id == "scotus",
+            "scotus_metadata": build_scotus_metadata(scotus_metadata),
```
albertisfu: And use it here:
```suggestion
            "scotus_metadata": await get_scotus_metadata_items(docket),
```

cl/opinion_page/utils.py
albertisfu: This is triggering a reverse query to check whether the one-to-one relation with `scotus_metadata` exists, even if the `Docket` is not from SCOTUS. So, this is an extra query that runs on every docket page load.

We should get rid of this extra query for dockets that are not from SCOTUS.

You can create a method like this:

```
async def get_scotus_metadata_items(docket: Docket) -> list[MetadataItem]:
      if docket.court_id != "scotus":
          return []
      metadata = await ScotusDocketMetadata.objects.filter(docket=docket).afirst()
      return build_scotus_metadata(metadata)
```

cl/opinion_page/utils.py
albertisfu: And use it here:
```suggestion
            "scotus_metadata": await get_scotus_metadata_items(docket),
```

cl/opinion_page/templates/includes/de_filter.html
```diff
@@ -110,6 +110,14 @@
   </div>
 
 <p style="margin-top: 10px; margin-bottom: 0;">
-  <i class="fa fa-info-circle"></i> Items on this page are sourced from <a href="https://www.pacer.gov">PACER</a> and <a href="https://free.law/recap/">RECAP</a> and may not be up to date. <a href="{% url 'coverage_recap' %}">Learn more about our coverage</a>.
+  {% if is_scotus %}
+    {% comment %}
+    Swap this hardcoded path for {% url 'coverage_scotus' %} once the
+    SCOTUS coverage page is ready.
+    {% endcomment %}
+    <i class="fa fa-info-circle"></i> Items on this page are sourced from the Supreme Court of the United States and may not be up to date. <a href="/help/coverage/scotus/">Learn more about our coverage</a>.
```
albertisfu: We should change this URL to point directly to the wiki page, since we ended up not having a SCOTUS coverage page in CL.

The URL is:
[https://wiki.free.law/c/courtlistener/help/data-coverage/surpeme-court-cases-and-filings](https://wiki.free.law/c/courtlistener/help/data-coverage/surpeme-court-cases-and-filings)

You can add it as a variable, similar to the other wiki pages, for example: `WIKI_COVERAGE_RECAP_URL`.

albertisfu (changes requested):
Thank you @nikollycardoso changes look good!

I just have a couple of extra suggestions. Also can you please introduce a tweak in the Docket admin page so the `View Docket Entries` button switches conditionally to display `SCOTUSDocketEntry` (for SCOTUS dockets instead of `DocketEntry`)  

Image


There is also a merge conflict in `de_filter.html` that should be solved after you update this PR with main.

cl/opinion_page/templates/includes/pacer_links_dropdown.html
```diff
@@ -1,4 +1,13 @@
-{% if docket.pacer_docket_url %}
```
albertisfu, line 1: Is this file `pacer_links_dropdown.html` still used somewhere or it can be removed?

cl/settings/misc.py
```diff
@@ -68,7 +68,9 @@
 WIKI_COVERAGE_OA_URL = f"{WIKI_COVERAGE_BASE_URL}/oral-argument-recordings"
 WIKI_COVERAGE_OPINIONS_URL = f"{WIKI_COVERAGE_BASE_URL}/case-law"
 WIKI_COVERAGE_RECAP_URL = f"{WIKI_COVERAGE_BASE_URL}/federal-cases-and-filings"
-
+WIKI_COVERAGE_SCOTUS_URL = (
+    f"{WIKI_COVERAGE_BASE_URL}/surpeme-court-cases-and-filings"
```
albertisfu: There was a typo here. Fixed the wiki page slug and can be fixed here too.

```suggestion
    f"{WIKI_COVERAGE_BASE_URL}/supreme-court-cases-and-filings"
```

albertisfu (changes requested):
Thanks @nikollycardoso !

A couple of extra comments please.

And here some other details I found related to the refactor.

The View on SCOTUS button is missed in mobile templates below 768px:

Image
Image

You can see that in RECAP the small white button is visible and redirects to PACER 


- Also the Originating Court Information section is missed now.

This one:
Image

It's missed after the refactor

cl/opinion_page/templates/includes/pacer_links_dropdown.html
nikollycardoso: I confirmed it's not being used anymore, it was replaced by docket_source_button/{recap,scotus}.html but never deleted after that. Removed in 15598f3b0.

cl/opinion_page/templates/cotton/docket_source_attribution/scotus.html
```diff
@@ -0,0 +1 @@
+Items on this page are sourced from the Supreme Court of the United States and may not be up to date. <a class="underline" href="{{ WIKI_COVERAGE_SCOTUS_URL }}">Learn more about our coverage</a>.
```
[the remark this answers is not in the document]
elisa-a-v: /fp `WIKI_COVERAGE_SCOTUS_URL` is a hardcoded settings constant, not user input.

cl/opinion_page/templates/cotton/docket_source_attribution/recap.html
```diff
@@ -0,0 +1 @@
+Items on this page are sourced from <a class="links" href="https://www.pacer.gov">PACER</a> and <a class="links" href="https://free.law/recap/">RECAP</a> and may not be up to date. <a class="links" href="{{ WIKI_COVERAGE_RECAP_URL }}">Learn more about our coverage</a>.
```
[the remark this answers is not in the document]
elisa-a-v: /fp `WIKI_COVERAGE_RECAP_URL` is a hardcoded settings constant, not user input.

cl/opinion_page/templates/cotton/docket_entry_rows.html
```diff
@@ -15,174 +16,136 @@
...
-                      <p class="hidden md:block flex-1 text-sm text-greyscale-900 min-w-48">{{ rd.description }}</p>
+        {% if entry.documents %}
+          <ul role="list" class="list-none p-0 m-0" aria-label="Documents for this entry">
+            {% for document in entry.documents %}
+              <li class="mt-3 md:ml-28">
+                <div class="flex flex-wrap items-start gap-2">
+                  {# Type badge — fixed width to align with date column #}
+                  <div class="w-32 shrink-0">
+                    {% if document.detail_url %}
+                      <a href="{{ document.detail_url }}" rel="nofollow" class="links text-sm font-medium">
```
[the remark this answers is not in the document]
elisa-a-v: /fp Both variables this template renders into an href are covered by a contract on `DocketEntrySource`: `document_detail_url` returns a CourtListener path we build ourselves, and `document_external_url` must get its scheme from a hardcoded prefix or an `http_url` filter. The contract is documented in the class docstring and enforced by `test_document_detail_urls_are_internal`, both in 58d82e98b.

albertisfu:
@nikollycardoso I've applied a few tweaks after my final review, could you please review them?

[6dec9b4](https://github.com/freelawproject/courtlistener/pull/7727/commits/6dec9b45cb02b7c467ac5ef9c7ab88ab88e9cb4d)

In this commit, I moved the SCOTUS metadata to the top, next to where the regular docket metadata is displayed. I made this change after realizing that the SCOTUS metadata appeared immediately after the Originating Court Information, which made it look like the SCOTUS data was part of that section which is not.

Screenshot 2026-08-28 at 10 16 20 a m


[1b4fbd7](https://github.com/freelawproject/courtlistener/pull/7727/commits/1b4fbd77990f695235dd96bb087506ddb6a543d0)

I found that, on SCOTUS dockets, the buttons at the top were not being displayed in other tabs (like Parties, etc.) because `pacer_docket_url` now returns False for SCOTUS.

Screenshot 2026-08-28 at 10 30 34 a m

So I added a small tweak to use docket_source_url for this condition instead, since it now returns True for SCOTUS dockets as well.

[32788ad](https://github.com/freelawproject/courtlistener/pull/7727/commits/32788ad3abf51604cf1a969405390768ce27a0a8)

This CTA, which is shown when there are no entries for a PACER docket, was removed, so I restored it.
Screenshot 2026-08-28 at 10 34 14 a m


[fb1f029](https://github.com/freelawproject/courtlistener/pull/7727/commits/fb1f029803cbcf64950c7fca151e99ae187f0735)

On the `new_docket_alert` page, the Get Alerts button used the `pacer_docket_url` condition to determine whether it should be displayed. That condition was removed from the template, but now that we have `docket_source_url`, it makes sense to use it instead.

For now, the button is shown only for RECAP and SCOTUS dockets.

Screenshot 2026-08-28 at 10 37 32 a m


[2d4c5b1](https://github.com/freelawproject/courtlistener/pull/7727/commits/2d4c5b147445d71674fe60b48fd173241b18c29a)

Found that in the tags page we have a button that renders the Docket original source link.

It was missing for SCOTUS cases:
Screenshot 2026-08-28 at 10 00 19 a m

So I applied a fix so the View on SCOTUS is displayed.

[ce54c1e](https://github.com/freelawproject/courtlistener/pull/7727/commits/ce54c1e6b014a14832a0c64dc0e5919c16002725)

There was a tooltip in the OCI docket number that was removed during the refactor. 

Screenshot 2026-08-28 at 10 42 09 a m

So I restored it.

[e6fe7b8](https://github.com/freelawproject/courtlistener/pull/7727/commits/e6fe7b8ce9c1a6d03b8a4ec38d67da60be2bc38e)

I noticed that the OCI and Bankruptcy metadata had swapped order:

Screenshot 2026-08-28 at 10 15 13 a m

So I restored the original order. Bankruptcy metadata should be shown before the OCI metadata.

nikollycardoso (commented):
Went through all 7 commits, @albertisfu, thanks for picking these up. This looks great: tests pass locally and it's working for me manually too.

albertisfu:
Thank you Nikolly! I'll go ahead and merge it then.

I've created the `scotus_docket_page` flag and set it up to staff only for now.

albertisfu (approved):

PR #7738, opened 2026-08-05
fix(scotus): Prevent duplicate docket entries when an entry gains attachments

### LINKED ISSUE #7728: SCOTUS merger can create duplicate docket entries
In the process of working on #6956, I discovered that docket 25-1405 has duplicate docket entries.

I checked and the current outputs of JSON (poller) and HTML (email) parsing are the same.

My theory is that an attachment was added to the docket entry between the poller hitting 25-1405 and the email update, which is not a possibility we had considered. This would explain why the older duplicate has a blank `entry_number` field.

If my theory is correct, the poller hit 25-1405 before it had any attachments so `SCOTUSDocketReport` set `document_number` to `None`. Then when we got the email update, there _were_ attachments so `SCOTUSDocketReportHTML` set `document_number` to something, which caused `merge_scotus_docket_entry` to take the `if entry_number` branch, where the lookup for a matching entry number failed and it created a duplicate docket entry.

MorganBennetDev:
## Fixes

Fixes #7728

## Summary

If the `entry_number` lookup for a SCOTUS docket entry fails, fall back to the non-entry number lookups before creating a new docket entry.

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [ ] `skip-cronjob-deploy`
    - [ ] `skip-daemon-deploy`


cl/corpus_importer/tasks.py
```diff
@@ -3595,23 +3597,21 @@ def merge_scotus_docket_entry(
                 )
                 return False, None, []
 
-        else:
+        if de is None:
             normalize_long_description(input_docket_entry)
```
albertisfu: One suggestion is to constrain the lookup for the queries below so that they only match unnumbered entries when there is an entry number.

Otherwise, it could match an entry that belongs to a different (same description and date) `entry_number` and overwrite it.

```suggestion
            unnumbered_lookup = (
                {"entry_number__isnull": True} if entry_number else {}
            )
```

albertisfu: Also we should move `normalize_long_description` to the top before line `description = input_docket_entry["description"]` so the value used for lookups matches the one saved on the entry.

cl/corpus_importer/tasks.py
albertisfu: This text is outdated now, as the fallback can also match entries with numbers.

albertisfu (changes requested):
Thanks @MorganBennetDev this looks good. I just have a few suggestions, please.

cl/corpus_importer/tasks.py
```diff
@@ -3595,48 +3598,55 @@ def merge_scotus_docket_entry(
...
                         sequence_number=sequence_number,
+                        **unnumbered_lookup,
                     )
-                    de_created = False
                 except SCOTUSDocketEntry.DoesNotExist:
                     de = SCOTUSDocketEntry(
                         docket=docket,
                         date_filed=date_filed,
                         sequence_number=sequence_number,
+                        **unnumbered_lookup,
```
albertisfu: We can omit this one, right? Since this is not a queryset.

albertisfu (changes requested):
Thanks @MorganBennetDev just a small change please.

Also a merge conflict needs to be solved.

albertisfu (approved):
Thanks this looks good. Merging.

PR #7739, opened 2026-08-05
feat(api): Give the Fetch API its own throttle, independent of membership

### LINKED ISSUE #7503: Make fetch API not apply to API limits
I'm not sure if the [Fetch API](https://wiki.free.law/c/courtlistener/help/api/rest/v4/recap#pacer-fetch) uses the same throttles as the other APIs, but it probably shouldn't. Imagine that you're a CL member and you want to start buying PACER documents in some automated way. We should encourage that by allowing it as much as somebody wants (within reason), not block it behind a membership.

mlissner:
## Fixes
Fixes: #7503

**Stacked on `7114-add-api-usage-endpoint`** — merge that PR first.

## Summary

The Fetch API used the default throttle, which meant that since our membership changes, people couldn't use it very quickly. That's a bummer. If people want to use our system to buy lots of PACER stuff, we should allow that. 

This does two things:
 - Introduces a 30/m rate for the fetch API
 - Makes it work with the new API Usage API

Fetch requests are queued in celery and throttled by the rate that celery can process them, so I think it should be fine to allow these higher limits.

## Deployment

1. Deployment labels are checked. 
2. There is a no-op migration, alas.
3. I have a tweak to the wiki page ready to go.


cl/recap/views.py
```diff
@@ -122,6 +123,9 @@ class PacerFetchRequestViewSet(LoggingMixin, ModelViewSet):
     serializer_class = PacerFetchQueueSerializer
     filterset_class = PacerFetchQueueFilter
     permission_classes = (IsAuthenticatedOrReadOnly, V3APIPermission)
+    # Dedicated, more generous rate than the global per-user API throttle,
+    # applied regardless of membership status. See #7503.
+    throttle_classes = (FetchRateThrottle,)
```
albertisfu: The fetch endpoint is a public endpoint. Before adding the custom `FetchRateThrottle`, it was throttled by the default `AnonRateThrottle`, which is configured to allow 100 requests per day.

After introducing `FetchRateThrottle`, the `AnonRateThrottle` no longer applies.

This makes the endpoint more susceptible to abuse, as anonymous users can now make up to 30 requests per minute which is 43,200 requests per day.

The solution seems simple we just need to add `AnonRateThrottle` before `FetchRateThrottle`. This won't affect authenticated users, as they will continue to be throttled by `FetchRateThrottle`, while anonymous users will still be subject to the default daily limit.

```suggestion
    throttle_classes = (AnonRateThrottle, FetchRateThrottle)
```

albertisfu (changes requested):
This looks good! I just have one suggestion to help prevent abuse of this endpoint by anonymous users.

cl/recap/views.py
mlissner: Reverted my change in b6897749f — going with your `AnonRateThrottle` suggestion instead.

mlissner:
I went ahead and merged this without waiting for tests since GH CI is broken today.

PR #7749, opened 2026-08-06
Morgan/florida backfill

MorganBennetDev:
## Fixes

Implements Florida docket ingestion Celery task and Florida docket backfill management command.

## Summary

- Adds `import_fl_dockets` command to backfill Florida dockets from S3
- Adds `fl_corpus_download_task` to merge a parsed FL docket stored in S3
- Create `AbstractStateDocument` model to capture shared fields (`url` and `processing_error`) and functionality (download file)
- Updates Texas document download logic to use `AbstractStateDocket` utilities
- Update Texas tests
- Add `pre_update` method to `Merger` to allow resetting fields when necessary

## Deployment

**This PR should:**

    - [x] `skip-web-deploy`
    - [ ] `skip-celery-deploy`
    - [x] `skip-cronjob-deploy`
    - [x] `skip-daemon-deploy`


cl/corpus_importer/management/commands/import_fl_dockets.py
```diff
@@ -0,0 +1,35 @@
...
+    def download_task() -> app.Task:
+        return fl_corpus_download_task
+
+    @staticmethod
+    def merge_task() -> app.Task:
+        return fl_ingest_docket_task
+
+    def handle(self, *args, **options):
+        super().handle(*args, **options)
+        CaseTransfer.fill_null_dockets()
```
albertisfu: It might be better to make this execute only if a `--fill-case-transfer` argument is passed. That way, we can trigger it manually once the ingestion has finished. Otherwise, there might still be cases in the queue that haven't been merged by the time this runs.

cl/corpus_importer/state/common/docket_entry.py
```diff
@@ -44,11 +48,42 @@ def _document_url(document: ScrapeDocument, params: Any) -> str:
...
+        if (existing := self.existing) is None:
+            return updated
+        if existing.processing_error == ProcessingError.BAD_URL:
+            existing.processing_error = None
+            updated.append("processing_error")
+        if existing.filepath_local:
+            existing.filepath_local.delete(save=False)
+            updated.append("filepath_local")
+        existing.filepath_local = ""
+        existing.ocr_status = None
```
albertisfu, line 85: This `pre_update` hook runs on every document field update, even if the document URL hasn't changed, right?

And this cleans up the existing file and resets the `ocr_status`. Is that intended?

The right approach would be to clean up the file and reset the `ocr_status` only if the document URL has changed from the original source. If the document hasn't changed, we should preserve the original file because we don't want to duplicate download work for files we already have.

Also, there might be cases where we don't run the merger with `download_attachments=True`, so we could end up missing those documents.

cl/search/state/shared.py
```diff
@@ -47,3 +60,142 @@ class ProcessingError:
...
+                return None
+
+            filename = f"{document.make_filename()}{extension}"
+            downloaded_file = File(tmp)
+            document.filepath_local.save(filename, downloaded_file, save=False)
+            document.file_size = downloaded_file.size
+            document.sha1 = sha1_hash
+
+            if extension == ".pdf":
+                response = async_to_sync(doc_page_count_service)(document)
```
albertisfu: Florida already gives us the `page_count` of a document. Can we omit this `doc_page_count_service` call if we already have the page count from the source?

cl/corpus_importer/tasks.py
```diff
@@ -5135,3 +5027,97 @@ def texas_corpus_download_task(
...
+
+    :param task: The Celery task.
+
+    :param case_bytes: The bytes of the parsed JSON retrieved from S3.
+    :param download_attachments: Whether to download docket entry attachments.
+
+    :return: The result of the merge operation.
+    """
+    try:
+        case = FloridaCase.model_validate_json(
```
albertisfu: Can this use the new Juriscraper `deserialize` method?

cl/corpus_importer/tasks.py
```diff
@@ -5135,3 +5027,97 @@ def texas_corpus_download_task(
...
+            case_bytes, by_name=True, context={"deserialize": True}
+        )
+    except Exception:
+        logger.exception("Failed to deserialize Florida case")
```
albertisfu: In case there is an error during serialization, it'll be difficult to identify which cases we need to re-ingest.

Maybe we can also make `fl_corpus_download_task` return the `key`, receive it here in this task, and include it in this log in case of an error?

albertisfu (changes requested):
Thanks @MorganBennetDev this looks good!

I just have a few comments, please.

cl/corpus_importer/tasks.py
MorganBennetDev: Yes! Was waiting on the dependency update

cl/corpus_importer/tasks.py
```diff
@@ -5135,3 +5027,110 @@ def texas_corpus_download_task(
...
+    document = FloridaDocument.download(fl_document_pk)
+    return document.pk if document else None
+
+
+@app.task(bind=True, max_retries=5, ignore_result=True)
+def fl_ingest_docket_task(
+    task: Task,
+    case_bytes: bytes,
+    bucket: str,
+    key: str,
```
albertisfu: This task is failing because these three arguments are expected to be a tuple, as that's what `fl_corpus_download_task` returns.

`case_bytes, bucket, key = download_result`

cl/corpus_importer/tasks.py
albertisfu: I've updated juriscraper in main. So this should be ready to be updated.

albertisfu (changes requested):
Thanks, @MorganBennetDev! Just one small fix, please. Also, Juriscraper should be up to date on `main` now.

albertisfu (approved):

